Issue #1, opened 2017-08-08
PW parser cannot parse forces and stresses from QE 6.1


borellim:
Since QE 6.1, the output text preceding the forces and stresses has changed slightly, which prevents a correct parsing.

The changes were introduced in the following commits:
https://github.com/QEF/q-e/commit/521808b02d603688c0c6840eeabff50ba3427fee
https://github.com/QEF/q-e/commit/d11c68dc5ba46e68a4ae7c7d7b56778b63dee39f
https://github.com/QEF/q-e/commit/bc39402ebbcb23204c67e6b1aa6d77174870fc0f

PR #2, opened 2017-08-08
PW parser: correctly parse forces and stresses from QE 6.1

borellim:
This should fix issue #1.

These are the strictest changes possible that appear to work with my version of QE 6.1. The change can be made simpler (and a bit more general) by parsing only for `Forces acting on atoms` and `Computing stress` respectively. Given that the capitalization of `Cartesian` has changed in between the commits referenced from #1, I would be in favor of this.

I'll add a further commit to implement this if requested.

### DIFF
--- a/aiida_quantumespresso/parsers/raw_parser_pw.py
+++ b/aiida_quantumespresso/parsers/raw_parser_pw.py
@@ -1642,7 +1642,7 @@ def parse_pw_text_output(data, xml_data={}, structure_data={}, input_dict={}):
                     parsed_data['warnings'].append('Error while parsing Fermi energy from the output file.')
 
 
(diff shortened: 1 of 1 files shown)


sphuber:
Thanks Marco. I am wondering if it would be a good idea to simplify the conditionals. Indeed it would maybe reduce the chance of breaking the plugin if they slightly change the syntax again in the future, however, it might also then match parts of the output that it is not supposed to. Do you have older, say v5.* versions installed in your aiida? If so, could you maybe simplify the conditionals and run a test calculation with an older qe version, just to see if it breaks parsing. It is not the most thorough testing of course, but we don't have that set up anyway and if that test passes, I think we can merge the simplified conditionals. What do you think?

sphuber:
This was fixed in PR #5

PR #12, opened 2017-09-29
Load error handlers for PwBaseWorkChain from entry points

sphuber:
Since external plugins might want to add the functionality of the
PwBaseWorkChain, such as add additional error handlers, we need to
turn these into entry points so that new plugins can be registered.
We define custom pluginloader functions for this plugin and define
the new entry points:

    aiida_quantumespresso.workflows.error_handlers

External plugins can add new entry points here that adds another
list of additional error handlers

### DIFF
--- a/aiida_quantumespresso/common/pluginloader.py
+++ b/aiida_quantumespresso/common/pluginloader.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+from aiida.common.exceptions import MissingPluginError
+
+try:
+    from reentry import manager as epm
+except ImportError:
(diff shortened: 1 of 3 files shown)


sphuber:
Couple of questions regarding this solution:

* What do you think of the naming of the entry points. I kept the top level category generic for all workflows.
* I turned the error handlers into regular functions (used to be methods of the workchain class itself) and as a result now pass the workchain as an argument as opposed to `self`
* The entry point loads the list of error handler functions. It is defined in the same file as the workchain and the list is defined all the way at the bottom of the file. Bit hard to find maybe. Possibly better location for it?

PR #19, opened 2017-10-25
Release v1.0.0

sphuber:


sphuber:
The `.gitignore` is indeed a generic one for python projects and I needed in any case to add the `build` and `dist` folders for the `pypi` package builds. The spaces in the `setup.json` I also noticed but since that file is formatted by the `pre-commit` hook and is just what the `json.dumps` formatter spits out, it has nothing to do with my editor.

PR #26, opened 2017-11-02
Merge master into develop

sphuber:
Release v1.0.1


sphuber:
`Won't accept, you are polluting my commit log!` Something like that? :)

giovannipizzi:
Some people would go with that!

Issue #28, opened 2017-11-02
Incorrect method of retrieving last PwCalculation from PwBaseWorkChain in the PwBandsWorkChain


sphuber:
In the `results` step of the `PwBandsWorkChain` I need to retrieve the final `PwCalculation` of the `PwBaseWorkChain` to obtain the final `BandsData` and do so by calling `get_outputs` and specifying the link type to be `LinkType.CALL`, which returns a list and then getting the first element. However, if the base workchain launched multiple calculations, the retrieved calculation may not be the last and successful one.

sphuber:
Fixed in PR #29

PR #29, opened 2017-11-02
Add output_band as optional output to PwBaseWorkChain

sphuber:
This also prevents the PwBandsWorkChain of having to go through
the PwCalculation of the PwBaseWorkChaihn to try and obtain the
output_band BandsData node.


sphuber:
Fixes #28

PR #34, opened 2017-11-08
Parse bands by default

### LINKED ISSUE #30: Add parsed occupation numbers by default to PwCalculation outputs
The occupation numbers after a `PwCalculation` are often required for the calculation analysis and preparation of calculation restarts and should be parsed by default into a `ArrayData` node. To do this in the correct way, this requires a change in the `CalcInfo` data structure of `aiida_core` so this [issue](https://github.com/aiidateam/aiida_core/issues/886) has to be solved first, before this can be implemented in `aiida-quantumespresso`.

sphuber:
Fixes #30 

Note that this fix relies on changes in `aiida_core` that are only in develop. Therefore, to make the tests run, I had to explicitly run `pip install` in the `travis.yml` with the hash of the commit in `aiida_core`. I am not sure what the best procedure should be here. Once the next version of `aiida_core` is released we should revert the change and simply update the requirement in the `setup.py`.

### DIFF
--- a/.travis-data/test_pw_with_daemon.py
+++ b/.travis-data/test_pw_with_daemon.py
@@ -132,7 +132,7 @@ while time.time() - start_time < timeout_secs:
     print "Output of 'verdi calculation list':"
     try:
         print subprocess.check_output(
-            ["verdi", "calculation", "list"], 
+            ["verdi", "calculation", "list", "-a"], 
             stderr=subprocess.STDOUT,
         )
     except subprocess.CalledProcessError as e:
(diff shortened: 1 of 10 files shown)


giovannipizzi:
I'm not sure what's the best idea. However basically everybody using 'develop' here has to use 'develop' in aiida_core... maybe these things should be developed in a different branch?

sphuber:
So any changes to `aiida_quantumespresso` that can run of the latest `aiida_core` release go into `develop` but any changes that require a newer version of `aiida_core` should go into a temporary branch, let's call it `future-release`? Then the new version of `aiida_core` is released, we merge `future-release` into `develop` and create a new release of `aiida_quantumespresso`. Only thing is that if we want to be able to merge multiple features into this new `future-release` branch, we have to disable tests, because they are going to be broken. Or are we going to allow to change the `aiida_core` dependency in the `future-release` branch?

sphuber:
OK, I will close this PR as the branch contains the commit that messes with the `aiida_core` version. I will remove the commit and open the new `future-release` branch

PR #43, opened 2017-11-16
Reinstate the parser tests

### LINKED ISSUE #14: Add back parser tests
Test files are available, but not implemented yet

sphuber:
Fixes #14


giovannipizzi:
I got scared by the 886 files changed :-)

sphuber:
I was aiming for 666 for the triple threat

giovannipizzi:
```
OK (skipped=4)
Final summary of the run of tests:
* Tests skipped: 4
  Reasons for skipping:
  - Unable to import ase
* Tests run:     20
* Tests failed:  0
* Tests errored: 0
```
Should we run the tests with ASE installed?

sphuber:
We should install it through pip then, right? But do we really want to make that a required dependency? Or do you want to make it an optional dependency and in the travis setup install that optional?

sphuber:
OK so after various additional commits everything works, except for the error that crops up, which is a problem in `aiida_core`. The function `_collect_tags` in `aiida.tools.dbexporters.tcod` has some weird logic to get the plugin and so it throws. The following code fixes it, but I am not sure if it does so for the right reasons:

```
if calc is not None:

        code = calc.get_code()

        if code is not None:
            category = 'tools.dbexporters.tcod_plugins'
            plugin_name = code.get_input_plugin_name()

            try:
                plugin = get_plugin(category, plugin_name)
            except (MissingPluginError, LoadingPluginFailed):
                pass
            else:
                if calc._plugin_type_string.endswith(plugin._plugin_type_string + '.'):
                    translated_tags = translate_calculation_specific_values(calc, plugin)
                    tags.update(translated_tags)
```

The original problem was that it was getting all the plugins for the entrypoint `aiida.tools.dbexporters.tcod`, which could be more than one, but then raising if it found more than one. My attempt to determine which entrypoint to use is by getting the plugin name of the `calc` through its `code` but I think in these tests that always return `None`, so the whole logic is skipped.
Not sure what the original idea was supposed to be

giovannipizzi:
Thanks! I think having a optional dependency that we use in the tests is ok.
For the logic, I will check with Andrius who originally wrote it.

merkys:
Hi. It seems that the logic got inverted in aiidateam/aiida_core@da9767bd29be2ce719d6325e5f09d44eb8385c1d. Beforehand the code did the following:

1. Get all plugins in ``aiida.tools.dbexporters.tcod_plugins``
2. Find all plugins from Step 1. that match the given calculation ``calc``
3. Emit an ``MultipleObjectsError`` if more than one plugin matches the ``calc``
4. Otherwise use the single found plugin that matches ``calc``.

Now the code seems to be doing this:

1. Get all plugins in ``aiida.tools.dbexporters.tcod_plugins``
2. Emit an ``MultipleObjectsError`` if more than one is found in step 1.
3. Check whether the single found plugin matches ``calc`` and use it.

Could you please bring back the former logic using the novel ``all_plugins(category)`` function?

sphuber:
Thanks Andrius! That is my bad. I will put in a fix in `aiida_core` which should then fix this problem here.

merkys:
No prob @sphuber, thanks for a prompt fix! I should have documented that piece of code better.

PR #68, opened 2017-12-15
Fix small bug in the PwRelaxWorkChain

sphuber:
The new optional input final_scf when set to true, will trigger
a final PwBaseWorkChain after the structure has successfully relaxed
that performs an scf calculation. Since it is an scf run, the
workchain will not have a structure as output. Therefore, in that
case of a final scf we attach the structure that was its input. In
all other cases, we attach the structure that was the output of the
final completed PwBaseWorkChain in relax mode.


muhrin:
All good.

PR #102, opened 2018-01-31
PwBandsWorkChain: make 'relax' inputs and relaxation optional

### LINKED ISSUE #96: PwBandsWorkchain requires 'relax' inputs even if 'skip_relax' is True
Currently this is using the `input_group` construct which will be deprecated, but it should be fixed in the new system

sphuber:
Fixes #96 

The relaxation of the input structure was made optional by defining
a 'skip_relax' optional boolean input. However, the 'relax' input
group was always required. We remove the 'skip_relax' input and just
determine whether the input structure needs to be relaxed based on
whether the 'relax' input group was specified in the workchain inputs
or not.


sphuber:
I have removed the `skip_relax` context variable. You were right that in this case I can directly check for the presence of `relax` in the inputs of the workchain in the `should_do_relax` outline method. I hope I understood your comment correctly and this is what you meant.

I have also rebased the branch on the latest develop after the previous two PR's were merged. This to prevent the pointless merge commits to bring the base branch up to date.

PR #104, opened 2018-02-03
Use updated seekpath wrappers in aiida-core for PwBandsWorkChain

### LINKED ISSUE #83: Migrate PwBandsWorkChain to use seekpath wrappers of aiida_core
The release of `aiida_core` `v0.11.0` will contain wrappers for `seekpath` to generate k-points for a given structure. The `PwBandsWorkChain` currently still directly uses the wrappers in `seekpath` itself, which have been removed in the latest version

sphuber:
Fixes #83 

The kpoints_distance value was used to generate a kpoints mesh for the
scf step, when specified, but it did so on the input structure. However,
the structure that was actually used for the calculation, was the one
returned by Seekpath, which may have changed, since Seekpath tries to
reduce the structure to the primitive cell. To ensure that both the
kpoints mesh for the scf and the kpoints path (determined by Seekpath)
for the bands calculation are commensurate with the structure that is
used, we use the primitive structure returned by the Seekpath analysis

### DIFF
--- a/aiida_quantumespresso/workflows/pw/bands.py
+++ b/aiida_quantumespresso/workflows/pw/bands.py
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+from copy import deepcopy
+from aiida.common.extendeddicts import AttributeDict
 from aiida.orm import Code
 from aiida.orm.data.base import Str, Float, Bool
 from aiida.orm.data.parameter import ParameterData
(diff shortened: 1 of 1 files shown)


sphuber:
Thanks Nicolas, all very good points. I have added an update and here my answers:

1. The distance in the path could be set by `kpoints_distance` although this was also being used for the generation of the mesh for the scf calculation. I have added a separate input `kpoints_distance_bands` which will be passed to seekpath as the reference distance.

2. I have added the input `kpoints_mesh` that can be specified if one wants to run the bands step at a regulard mesh instead of a path

3. The `PwRelaxWorkChain` by default does not run a final scf after convergence and in general one does not want to restart the charge density of a `vc-relax` run to restart a bands calculation. Therefore we run an scf now always. We could improve this but I rather leave this for the next version, since this workchain will have to be changed and updated a lot anyway

aiida_quantumespresso/workflows/pw/bands.py
```diff
@@ -30,6 +30,8 @@ def define(cls, spec):
         spec.input('structure', valid_type=StructureData)
         spec.input('pseudo_family', valid_type=Str)
         spec.input('kpoints_distance', valid_type=Float, default=Float(0.2))
+        spec.input('kpoints_distance_bands', valid_type=Float, default=Float(0.2))
+        spec.input('kpoinst_mesh', valid_type=KpointsData, required=False)
```
nmounet: kpoinst -> kpoints

PR #105, opened 2018-02-03
Add parsing of electronic and ionic dipole for raw pw output

### LINKED ISSUE #56: Parsing Electronic and Ionic Dipole in pw calculations with applied external electric field
When performing a pw calculation in presence of an applied external electric field, the electronic and ionic dipoles are printed in pw.out for every iteration of the scf procedure.  See PW/examples/example10:

pw.in
```
&control
    calculation='scf'
    restart_mode='from_scratch',
    prefix='silicon',
    lelfield=.true.,
    nberrycyc=3
    pseudo_dir='$PSEUDO_DIR/',
    outdir='$TMP_DIR/'

 /
 &system
    ibrav= 1, celldm(1)=10.18, nat=  8, ntyp= 1,
    ecutwfc = 20.0
 /
 &electrons
    diagonalization='david',
    conv_thr =  1.0d-8,
    mixing_beta = 0.5,
    startingwfc='random',
    efield_cart(1)=0.d0,efield_cart(2)=0.d0,efield_cart(3)=0.001d0
```
pw.out
```
Electronic Dipole per cell (a.u.)  0.926395840102385
Ionic Dipole per cell (a.u.)   115.173552519665
Electronic Dipole on Carthesian axes
           1 -1.208125742863772E-005
           2 -4.814216075399945E-006
           3  0.926395840102385
Ionic Dipole on Carthesian axes
           1   115.173552519665
           2   115.173552519665
           3   115.173552519665
```
I would need these quantities to be parsed by AiiDA (and ideally to be returned also in the output_parameters of the PwBaseWorkChain)

sphuber:
Fixes #56 

When the boolean switch `lelfield` is defined in the `CONTROL` card of
the input parameters of a `PwCalculation`, the electronic and ionic
dipoles will be printed at the end of every scf iteration, both the
average over the cell as well as projected on the cartesian axes.

Add the parsing of these values for the final scf iteration of each
trajectory step under the following keys:

  * `electronic_dipole_cell_average`
  * `electronic_dipole_cartesian_axes`
  * `ionic_dipole_cell_average`
  * `ionic_dipole_cartesian_axes`

Each of these fields will be a list with as many entries as there were
scf steps in the trajectory.

### DIFF
--- a/aiida_quantumespresso/parsers/raw_parser_pw.py
+++ b/aiida_quantumespresso/parsers/raw_parser_pw.py
@@ -1728,13 +1728,28 @@ def parse_pw_text_output(data, xml_data={}, structure_data={}, input_dict={}, pa
         if lelfield is True:
 
             # For every property only get the last entry if possible
-            ed_cell = trajectory_frame['electronic_dipole_cell_average'].pop()
-            ed_axes = trajectory_frame['electronic_dipole_cartesian_axes'].pop()
-            id_cell = trajectory_frame['ionic_dipole_cell_average'].pop()
-            id_axes = trajectory_frame['ionic_dipole_cartesian_axes'].pop()
+            try:
+                ed_cell = trajectory_frame['electronic_dipole_cell_average'].pop()
+            except IndexError:
+                ed_cell = None
+
+            try:
+                ed_axes = trajectory_frame['electronic_dipole_cartesian_axes'].pop()
+            except IndexError:
+                ed_axes = None
+
+            try:
+                id_cell = trajectory_frame['ionic_dipole_cell_average'].pop()
+            except IndexError:
+                id_cell = None
+
+            try:
+                id_axes = trajectory_frame['ionic_dipole_cartesian_axes'].pop()
+            except IndexError:
+                id_axes = None
 
             # Only add them if all four properties were successfully parsed
-            if all([ed_cell, ed_axes, id_cell, id_axes]):
+            if all([value is not None for value in [ed_cell, ed_axes, id_cell, id_axes]]):
                 trajectory_data.setdefault('electronic_dipole_cell_average', []).append(ed_cell)
                 trajectory_data.setdefault('electronic_dipole_cartesian_axes', []).append(ed_axes)
                 trajectory_data.setdefault('ionic_dipole_cell_average', []).append(id_cell)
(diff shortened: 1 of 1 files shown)


aiida_quantumespresso/parsers/raw_parser_pw.py
```diff
@@ -1698,6 +1704,42 @@ def parse_pw_text_output(data, xml_data={}, structure_data={}, input_dict={}, pa
...
+                elif 'Ionic Dipole on Cartesian axes' in line:
+                    ionic_dipole = [float(data_step[count + i + 1].split()[1]) for i in range(3)]
+                    trajectory_frame.setdefault('ionic_dipole_cartesian_axes', []).append(ionic_dipole)
+
+
+        # End of trajectory frame, only keep last entries for dipole related values
+        if lelfield is True:
+
+            # For every property only get the last entry if possible
+            ed_cell = trajectory_frame['electronic_dipole_cell_average'].pop()
```
giovannipizzi: If by any reason none is found (e.g. 0 steps, or the syntax changes and the lines are not found anymore) this would crash I think... can we pop with a default value?

aiida_quantumespresso/parsers/raw_parser_pw.py
```diff
@@ -1698,6 +1704,42 @@ def parse_pw_text_output(data, xml_data={}, structure_data={}, input_dict={}, pa
...
+            ed_axes = trajectory_frame['electronic_dipole_cartesian_axes'].pop()
+            id_cell = trajectory_frame['ionic_dipole_cell_average'].pop()
+            id_axes = trajectory_frame['ionic_dipole_cartesian_axes'].pop()
+
+            # Only add them if all four properties were successfully parsed
+            if all([ed_cell, ed_axes, id_cell, id_axes]):
```
giovannipizzi: If the dipole is exactly zero, I imagine this would return False...

aiida_quantumespresso/parsers/raw_parser_pw.py
```diff
@@ -1698,6 +1704,57 @@ def parse_pw_text_output(data, xml_data={}, structure_data={}, input_dict={}, pa
...
+            except IndexError:
+                id_cell = None
+
+            try:
+                id_axes = trajectory_frame['ionic_dipole_cartesian_axes'].pop()
+            except IndexError:
+                id_axes = None
+
+            # Only add them if all four properties were successfully parsed
+            if any([value is None for value in [ed_cell, ed_axes, id_cell, id_axes]]):
```
giovannipizzi: Sorry to bother again, like this it is going to append *only* if at least one is *not* found, meaning that if they are all found, it's not going to append it, right?
Maybe you can just remove the if, and the missing values get set to None? Also because anyway you popped out some values, so if you don't put them back somewhere, you would lose it?
Or do you have different idea in mind?

sphuber:
The current logic should be correct, sorry for the sloppiness and delay.

PR #107, opened 2018-02-16
Fix the wrong parser option key for atomic occupations in PwParser

### LINKED ISSUE #106: PwParser looks for incorrect key in parser options for atomic_occupations
In PR #55  a new parser option was implemented 'parse_atomic_occupations' which when set would parse the atomic occupations that are printed for a DFT+U calculation. However, the parser checked for the key `atomic_occupations` instead.

sphuber:
Fixes #106 

The code looked for 'atomic_occupations' but the key should be
'parse_atomic_occupations' instead


sphuber:
Fair point, I noticed the same, but I think it is more intuitive to name the parser option with the `parse_` prefix. Since this is a recent feature that I implemented for Chiara, I think it is fine to keep the documented key and change the code

PR #113, opened 2018-03-01
Properly set do_break=True for ErrorHandlerReport when applicable

### LINKED ISSUE #112: Error handlers for PwBaseWorkChain and PhBaseWorkChain contain errors
The returned `ErrorHandlerReports` are not always consistent, having `do_break=False` when it shouldn't and in some cases `self.ctx.restart_calc` is not being set even though the calculation can be restarted from

sphuber:
Fixes #112 

Also set the restart calculation in one of the error handlers that
could definitely be restarted from


sphuber:
Yes that is correct

Issue #124, opened 2018-04-26
Parser of symmetry operations from PW selects wrong operations


borellim:
Method `parse_with_retrieved` of class `PwParser` has a problem when reducing raw symmetry operations to a list of indices to the internal matrix list. The loop at lines 111-113 of pw.py doesn't stop when a match is found, therefore operations with inversion symmetry (starting with "inv.") are selected even when not correct.

A simple fix is to add a `break` when a match is found, but this relies on the list being sorted with non-inversion operations placed first, so a comment would also be useful.

Furthermore, it could be useful to extend the check not only on the operation's name string, but also on the matrix itself, with an error if they don't match as expected.

--------------------

Attaching a PwCalculation node and a jupyter notebook that reproduce the problem:
[136529.aiida.tar.gz](https://github.com/aiidateam/aiida-quantumespresso/files/1950740/136529.aiida.tar.gz)
[symmetry_parser_issue.ipynb.tar.gz](https://github.com/aiidateam/aiida-quantumespresso/files/1950742/symmetry_parser_issue.ipynb.tar.gz)

giovannipizzi:
Things to add before closing the issue: 
- adding a version number to the parser output, or upping it (this is an important bug and people should be able to tell if they are seeing correct data or not).
- we should improve the check, to make sure the string is actually matched correctly
- we should add some parser tests for this (or improve the existing ones - probably it's enough)

sphuber:
Fixed in PR #171

PR #125, opened 2018-04-26
PW parser: fix erroneous symmetry operation detection

borellim:
Implements a partial fix to issue #124.


sphuber:
If this is a partial fix, what else is required to fix it?

borellim:
Ok, this change is an actual fix. But as an added safety measure, we should maybe also check that every symmetry matrix matches what we expect it to be according to the operation string. I don't know how to implement this: in particular I don't know how I should handle the error case where the matrices don't match. Finally, I believe that @giovannipizzi was looking into a possible bug in QE that would make this check not work at all. Giovanni, do you have any news?

giovannipizzi:
Closing, as this will be fixed in a more complete way in #171

PR #138, opened 2018-05-17
Fix get_dblogger_extra import

yakutovicha:
Due to some changes in AiiDA the line
`from aiida.utils.logger import get_dblogger_extra`

should be replaced with
` from aiida.common.log import get_dblogger_extra`


yakutovicha:
@giovannipizzi or @sphuber  can you review this PR? I guess this import issue should be fixed before the next release.

sphuber:
Thanks @yakutovicha

PR #150, opened 2018-05-26
Implement support for input flags that are multidimensional array

sphuber:
Certain input flags in for example pw.x are multidimensional arrays
as for example the `starting_ns_eigenvalue(m,ispin,I)` flag.
There was no way to build the python parameters datastructure such
that the automatic conversion to the input file could be done, which
uses the `aiida_quantumespresso.calculation.get_input_data_text`
method. We update this method such that we can use the following
notation:

    paramaters = {
        'SYSTEM': {
             'starting_ns_eigenvalue': [
                  [1, 1, 3, 3.5],
                  [2, 1, 1, 2.8]
             ]
         }
    }

which will result in the input card

    &SYSTEM
     starting_ns_eigenvalue(1, 1, 3) = 3.5
     starting_ns_eigenvalue(2, 1, 1) = 2.8
    /

which solves the problem for multidimensional array input flags.

### DIFF
--- a/aiida_quantumespresso/calculations/__init__.py
+++ b/aiida_quantumespresso/calculations/__init__.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import collections
 import os
 
 from aiida.common.exceptions import InputValidationError
@@ -14,6 +15,7 @@ from aiida.orm.data.singlefile import SinglefileData
 from aiida.orm.data.remote import RemoteData
 from aiida.common.datastructures import CodeInfo
 from aiida.common.links import LinkType
+from aiida_quantumespresso.utils.convert import convert_input_to_namelist_entry
 
 class BasePwCpInputGenerator(object):
     """
@@ -448,7 +450,7 @@ class BasePwCpInputGenerator(object):
             # empty namelist
             namelist = input_params.pop(namelist_name, {})
             for k, v in sorted(namelist.iteritems()):
-                inputfile += get_input_data_text(k, v, mapping=mapping_species)
+                inputfile += convert_input_to_namelist_entry(k, v, mapping=mapping_species)
             inputfile += "/\n"
     
         # Write cards now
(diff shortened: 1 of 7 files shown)


sphuber:
I have made the change you proposed and cleaned that code up a little bit. I have also added the documentation to explain how these variables can be defined. I tried the helper, and they don't work, but they never did. As far as I can understand, the helper will identify valid keywords and valid "dimension" which are keywords that use one index. That means that `starting_ns_eigenvalue`, which is marked as a normal keyword, will not work, because the passing of a nested list will raise. I don't see an easy way to fix this, because it will either involve changing the xmls, which seems like a really bad idea, or parsing the name tag for the `(m,ispin,I)` to detect keyword

giovannipizzi:
Thanks.

I realise that there is one case `hubbard_j(i,ityp)` where a species is present... should we accept there a string, and convert it to the correct species name?

Also, should we still support the old way (still working in the code) to pass a flat list,  and they get indices 1,2,3,... - in this case, should we also document this?

sphuber:
I will add the documentation for the single list case. Concerning the automatic conversion of ityp like indices, I think it may be useful. We could transform any string that matches a kind name in the mapping to the corresponding index and we could document this. The hypothetical case where this accidentally transforms a string index that should not have been changed seems highly unlikely, and this seems preferable than not being able to specify `hubbard_j` without manually trying to figure out the species index in advance

giovannipizzi:
Thanks, I agree!
I think we can 
- use the type to decide if it's a string or an integer, stop otherwise
- stop if the string does not match a kind_name
Warn users in the docs that there is no validation on whether the specific position is supposed to index kinds (and then all good) or other quantities (xyz, atoms) where then the conversion to an integer will give a wrong result. It's up to the user to use the string in the right place.

giovannipizzi:
For the XML problem (for the helpers) maybe it's a good idea to ask the QE folks to improve that so it's parsable? We could discuss this at the MaX Hackathon?

giovannipizzi:
Thanks!

PR #151, opened 2018-06-06
Fix update aiida core requirement to v1.0.0a1

sphuber:
This branch will become `aiida-quantumespresso v3.0.0` which will rely on
`aiida-core v1.0.0`, so here we update the requirement to the just released alpha version


giovannipizzi:
Could you put a requirement for aiida_core<1 in the main branch as well?

PR #156, opened 2018-06-08
 Add support for Hubbard V extension to PwCalculation

### LINKED ISSUE #155: Add support for Hubbard V in PwCalculation
The `HUBBARD` module will implement Hubbard parameters for `pw.x` and besides being able to specify either Hubbard U or V directly in the input file, it also allows for `pw.x` to read the parameters from a file that was produced in an earlier `hp.x` run. For this to work with the plugin, the `PwCalculation` class needs to get another optional use method to allow to pass this single file data in.

sphuber:
Fixes #155 

This change required the addition of an optional input for the
`PwCalculation class`, `hubbard_file`, through the use methods.
It can be used to pass in a single file produced by a `hp.x` calculation
for a +U+V calculation, which can be used to read the parameters
for a restart instead of having to define all Hubbard V parameters
manually in the input parameters.

The launch script for a `PwCalculation` has also been updated with
options to pass defined Hubbard parameters, either manually or through
the use of a Hubbard `SinglefileData` node, including validation of
these inputs

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,9 @@ This is the official Quantum ESPRESSO plugin for AiiDA.
 The version 3.x series ot the AiiDA-Quantum ESPRESSO plugins are compatible with the AiiDA 1.x series.
 For the plugin that supports AiiDA 0.9-0.12, check the earlier version of this plugin (versions 1.x and 2.x).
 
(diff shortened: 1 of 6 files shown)


sphuber:
@giovannipizzi could you please have a look at this PR when you find the time?

PR #159, opened 2018-06-13
[ONHOLD] Prepare release v2.0.2

sphuber:


sphuber:
I think it will be good to release v2.0.2 given that it greatly reduces the distribution size. It also includes two other bug fixes and there are no other pressing issues open for this milestone

giovannipizzi:
How urgent it is? I would also try to include #125 as it is a quite serious bug, probably? However I would try to solve it properly.
If it requires too much time, ok to release 2.0.2 and make a 2.0.3  with that

sphuber:
I am fine with including it, but I have no idea if that PR is good to be merged. If you think the changes are good, let's merge the PR and include it in this release

giovannipizzi:
I added some comments to issue #124 on what I think should  be added, and wanted to discuss with you first to see if we can already have the parser tests before merging

PR #161, opened 2018-06-13
Fix the requirements string for aiida-core and atomic_tools optional

### LINKED ISSUE #160: Incorrect requirements string for aiida_core and the `atomic_tools` optional
The requirements string in the `setup.json` is incorrect and therefore the optional requirements `atomic_tools` are not installed, causing the tests requiring e.g. `ase` to be skipped

sphuber:
Fixes #160 

The optional needs to directly follow the dependency name. I also
removed the `atomic_tools` option in `.travis.yml` when installing
the `aiida-quantumespresso` package, as it does not have this optional
requirement group

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,15 +35,15 @@ before_install:
 install:
     # Upgrade pip setuptools and wheel to be able to run the next command
     - pip install -U pip wheel setuptools reentry
-    # Install AiiDA with some optional dependencies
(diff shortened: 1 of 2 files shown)


sphuber:
Thanks! P.S. I would like to start using rebase merge's to avoid the merge commits, specially for these single commit PR's it significantly reduces the number of noisy unnecessary commits

PR #165, opened 2018-06-16
ProjwfcParser: fixed a bug when parsing pdos files

### LINKED ISSUE #166: PDOS files are associated with wrong projections
Because the list of pdos files is not sorted naturally, the order is wrong with that of the projections in the `<prefix>.out` file.

zhubonan:
Fixes #166 

Current implementation relying on the order of pdos files to be consistent with that of projections in `<prefix>.out`. However,  names of `<prefix>.pdos_atom#xxx` files with array of pdos were not sorted correctly, giving wrongly associated PDOS arrays. Natural sorting is now used to get the correct order of files.

Also, I removed one line of unnecessary if compare.

### DIFF
--- a/aiida_quantumespresso/parsers/projwfc.py
+++ b/aiida_quantumespresso/parsers/projwfc.py
@@ -219,6 +219,24 @@ def spin_dependent_subparcer(out_info_dict):
     # pdos=pdos_arrays
     return bands_data,  projection_data
 
+
+def natural_sort_key(sort_key, _nsre=re.compile('([0-9]+)')):
+    """
+    Pass to ``key`` for ``str.sort`` to achieve natural sorting.
+    For example, ``["2", "11", "1"]`` will be
+    sorted to ``["1", "2", "11"]`` instead of ``["1", "11", "2"]``
+    :param sort_key: Original key to be processed
+    :return: A list of string and integers.
+    """
+    keys = []
+    for text in _nsre.split(sort_key):
+        if text.isdigit():
+            keys.append(int(text))
+        else:
+            keys.append(text)
+    return keys
+
+
 def spin_dependent_pdos_subparcer(out_info_dict):
     """
     Finds and labels the pdos arrays associated with the out_info_dict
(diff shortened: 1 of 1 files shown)


sphuber:
@zhubonan thanks for the fix. As you noticed, there is an incompatibility with the current `develop` branch of `aiida-core` and the current release of `aiida-quantumespresso`. For this reason we have created the `release_v2.0.2` and `release_v3.0.0a1` branches. The former will be compatible with `aiida-core==0.12.*`, whereas `release_v3.0.0a1` will be compatible with `aiida-core v1.0.0a1` and up. So if you could please rebase this PR onto `release_v2.0.2` branch, then I will make sure that it gets released in the patch release and after it will also be merged into `release_v3.0.0a`. Thanks

zhubonan:
@sphuber No problem  I have rebased into the `release_v2.0.2` branch.

aiida_quantumespresso/calculations/__init__.py
```diff
@@ -872,8 +872,11 @@ def create_restart(self, force_restart=False, parent_folder_symlink=None,
             # add the restart flag
             old_inp_dict['CONTROL']['restart_mode'] = 'restart'
             inp_dict = ParameterData(dict=old_inp_dict)
-
-        remote_folders = self.get_outputs(node_type=RemoteData)
+        # Maintain compatibility with aiida_core 0.12.0 and older versions
+        try:
+            remote_folders = self.get_outputs(type=RemoteData)
```
sphuber: Please remove this try/catch. This version will only support 0.12.* and older anyway. In the `v3*` series we will make sure to update the signature for `aiida-core v1.0.0` and higher.

zhubonan: Sorry, I get confused when this change of method signature was made - it was there since `0.11.2`. The `try`  was for maintain compatibility with versions older than that and not related to `v1.0.0` and higher. 
Here is the commit that made the change https://github.com/aiidateam/aiida_core/commit/991596e911f7c91b847aa0293d39f885610757b5.

zhubonan: I have remove this commit.

sphuber:
@zhubonan thanks a lot for your contribution

PR #170, opened 2018-06-22
Reinstate the parser tests

sphuber:
After the migration of the plugin code from within `aiida-core` to
the current separate `aiida-quantumespresso` repository, the parser
tests are no longer being run. This is because the fixture, that was
dynamically creating tests based on the presence of tests data
folders in a specific location, was still in `aiida-core`, yet those
test data folders were moved here.

Here we migrate that tests fixture `parsers.py` to this repository
and register it as an entry point in `aiida.tests`. This makes sure
that the aiida tests will find this test.

Additionally, the test data in `tests/parsers/test_archives` had to
be updated, because essentially it is a folder generated by the
export functionality and its version was outdated. I migrated the
data with `verdi export migrate`.

A final change had to be made to the two tests that include the
parsing of the bands. Sometime ago, the logic in the PwCalculation
plugin was changed to use the new `retrieve_temporary_list` concept
to by default retrieve all the bands xml files and parse the bands
but then throw away the files. The old test data archives were
however generated with the old logic where if one wanted the bands
to be parsed, one passed `also_bands: True` in the settings input.
This would tell the execmanager to also retrieve the bands. To fix
this I had to change the data.json files of these folders, remove
the paths of the xml files from the `retrieve_list` and add it to
the `retrieve_temporary_list` of the calculation node attributes.
Finally, I updated the parsers.py file that dynamically runs the
tests to check for the `retrieve_temporary_list` key in the attrs
of the calculation node and if found construct a FolderData with
the contents of the retrieved folder, which should contain the
xml files necessary to parse the bands.

Note that this mechanism was changed slightly in aiida-core v1.0.0
and for the `retrieved_temporary_folder` a normal file system folder
is used as opposed to a FolderData. Once these changes are merged
into the branch that is meant to be compatible with aiida-core>=v1.0
we have to change this logic


giovannipizzi:
Maybe though we should rename this  2.1.0?

PR #171, opened 2018-06-25
Fix bug in the mapping of raw symmetries in PwParser

### LINKED ISSUE #124: Parser of symmetry operations from PW selects wrong operations

sphuber:
Fixes #124 

The raw output of a PwCalculation contains fairly verbose information
about the symmetry operations found for the input structure. Since
storing these data structures as parsed from the XML output would be
very heavy on the database, instead these are mapped onto a static
dictionary of rotation matrices. This incurs some minimal but acceptable
information loss.

The process of mapping the original symmetry operations onto the reduced
operations was implemented by comparing the string name of the operation
and using the corresponding index to obtain the correct rotation matrix.
However, since the string comparison did not use an exact match as the
matching criterion, and in addition no break was used after finding a
match, the inverse symmetry operation always ended up getting matched.

To solve this, we change the string comparison to be an exact match, and
we break out of the loop when a match is found, for good measure.

A parser test is added to test for the correct mapping of symmetries.


giovannipizzi:
Note that this also introduces a new parser_version key in the output, that now matches the aiida-quantumespresso package version. We should also add this info on the release notes and mention how to check (with a query?) if the parsed symmetries might be wrong.

PR #175, opened 2018-06-25
Update the version to v2.1.0 and add parser note to docs

### LINKED ISSUE #172: Update the version of release_v2.0.2 to v2.1.0
New functionality was added, the support of defining the `ATOMIC_FORCES` card for pw.x, which merits a minor version release.

sphuber:
Fixes #172 

From this point on, the parser version will be tied to the version of
the package. In addition, it will be added to the output parameters
node of the calculation under the key `parser_version`. A section is
added to the docs of pw.x to mention this change and how to retrieve
the parser version of a completed calculation.


sphuber:
I just realized two things:

* We have not made the same change in adding the parser version for all other parsers
* With this new convention, the parser version will change everytime the package version changes, even if there are literally no changes to the parser. I cannot see any immediate problems with this now, but it might be weird to users

sphuber:
Ok, I have updated the parser info for all parsers

PR #186, opened 2018-07-10
Fix .copy() vs. .clone()

giovannipizzi:
- Replace occurrences of .copy() for Data nodes with .clone()
- added a freestanding function to reimplement .copy() for calculations (inside the deprecated restart methods)


giovannipizzi:
This requires to release a 1.0.0a2 of aiida-core first to have the tests pass.
@ltalirz @sphuber should we do it and rename the various scheduled versions with a +1 in the alpha number?

ltalirz:
No problem with bumping the alpha versions.
But please don't break more code than you need to, i.e. keep the old function with a deprecation warning

giovannipizzi:
> But please don't break more code than you need to, i.e. keep the old function with a deprecation warning

I fully agree

PR #199, opened 2018-08-21
Improve pwimmigration

### LINKED ISSUE #197: Bug in tools/qeinputparser/str2val
The string '1e-08' is not accepted, i.e. the regex needs to accept an integer before the exponent. I believe this should do the trick:

```python
float_re = re.compile(r"""
        [-+]?               # optional sign
        (?:                   # either
        \d*[.]\d+           # 10.53 or .53
        |                       # or
        \d+[.]\d*           # 10.53 or 10.
        |                       # or
        \d+ )                 # an integer
    (?:[dEeE][-+]?[0-9]+)?  # optional exponent
    """, re.X)

```

chrisjsewell:
- Fixes #197 
- add text parsing of cells with fractional coordinates (crystal type)
- improved support for pwimmigration when no xml is available
- added hook to connect input nodes to previous calculations

An example use of the calculation hook is to connect the input structure to a source CIF file:

```python

def cifhook(struct):

    @workfunction
    def cif2struct(cif):
        return struct

    cifin = # find CIF based on structure
    cif2struct(cifin)

with transport as open_transport:

    calc1.create_input_nodes(open_transport, calc_hooks=[[cifhook, ["structure"]]])

```


aiida_quantumespresso/calculations/pwimmigrant.py
```diff
@@ -242,7 +268,14 @@ class or using the ``set_`` methods (e.g. ``set_remote_workdir``).
                                            self._OUTPUT_SUBFOLDER,
                                            '{}.save/'.format(self._PREFIX),
                                            fnm)
-                open_transport.get(remote_path, folder.abspath)
+                if open_transport.path_exists(remote_path):
+                    open_transport.get(remote_path, folder.abspath)
+                else:
```
sphuber: What if the pseudo's also cannot be found here? Would it perhaps be better to define a list of potential paths and loop over them and as soon as one is found to exist call `open_transport.get` and if none of the potential paths exist, we either raise or log an error. Given that QE allows to define a pseudo dir variable in the input to allow a single directory with pseudos, I imagine a lot of calculation folders will not have copies of the pseudo's. I am now thinking whether it makes sense to import a `PwCalculation` without the pseudo's as it won't be reproducible. So in the fall through case we should maybe always raise a specific exception, notifying that the pseudos could not be found. The user could then copy to the folder manually, or we can potentially provide a flag to ignore the error.

P.S. for the for loop you could use the [for-else construct in python](http://book.pythontips.com/en/latest/for_-_else.html#else-clause) and raise in the else clause: see

aiida_quantumespresso/calculations/pwimmigrant.py
```diff
@@ -115,6 +117,13 @@ class or using the ``set_`` methods (e.g. ``set_remote_workdir``).
             absolute path to the job's directory on that computer.
         :type remote_workdir: str
 
+        :param calc_hooks: one or more of the input nodes can be linked with a function which
+        takes as input those input nodes and returns them (in the same order).
+        In this way the input nodes can be linked to precursor Calculation node(s).
+        Allowed keys are "structure", "kpoints", "parameters", "settings" and can only be used once
+        an example would be calc_hooks=[[my_func1, ["structure", "parameters"]], [my_func2, ["kpoints"]]]
```
sphuber: I am not sure I understand the intended behavior of these hooks. When I take the example usage detailed in the PR message, you want to call a function that will be passed the `StructureData` that is generated from the completed calculation that is to be imported, and that function should then serve to connect that `StructureData` to a `CifData`, the latter which should be retrieved from the structure input. However, I don't see how that is possible, since the `StructureData` that will be passed is the reconstructed one from the completed calculation that we are importing and so will not yet exist in the graph. How should that be linked to an existing `CifData`? Similarly, in the documentation here you describe that these hooks should serve to link up the to be created/imported input nodes to previously existing calculations. However, the hook will be generic, correct? If you do not pass in the actual calculation, how is the hook going to determine which calculation to link to the input nodes, based on those input nodes? Should the calculation that the user wants to link everything to, not be passed as an input to the hook itself?

aiida_quantumespresso/parsers/pw.py
```diff
@@ -24,6 +24,7 @@ def __init__(self, calc):
 
         :param calculation: instance of the PwCalculation
         """
+        self._require_xml = calc._require_xml if hasattr(calc, "_require_xml") else True
```
sphuber: I don't quite understand why we should define `_require_xml` in the `Calculation` class, if it is only used in the parser?
Should we not just define it in the parser class, and optionally allow the user to override it with a parser option in the `settings` input node?

aiida_quantumespresso/parsers/raw_parser_pw.py
```diff
@@ -11,6 +11,8 @@
 import string
 import xml.dom.minidom
 
+import numpy as np
```
sphuber: I don't think this import is used

aiida_quantumespresso/parsers/pw.py
```diff
@@ -149,11 +158,12 @@ def parse_with_retrieved(self, retrieved):
             # We compute here the transpose and its inverse of the structure cell basis, which is needed to transform
             # the parsed rotation matrices, which are in crystal coordinates, to cartesian coordinates, which are the
             # matrices that are returned by the _get_qe_symmetry_list staticmethod
-            cell = structure_data['cell']['lattice_vectors']
-            cell_T = numpy.transpose(cell)
-            cell_Tinv = numpy.linalg.inv(cell_T)
 
             try:
+                cell = structure_data['cell']['lattice_vectors']
```
sphuber: Why is this placed inside the try block? The only case I can see this triggering is if `structure_data['cell']` does not have the element `lattice_vectors` (given that the presence of `cell` is guaranteed by the conditional of this block) and I am not sure if this can ever happen. In any case, if this can happen, it should be in its own try-except block, because if this triggers an `KeyError` a misleading error will be logged, related to missing symmetries.

aiida_quantumespresso/parsers/raw_parser_pw.py
```diff
@@ -1211,8 +1243,11 @@ def parse_pw_text_output(data, xml_data={}, structure_data={}, input_dict={}, pa
             if 'site n.' in data_lines[i] and 'atom' in data_lines[i]:
                 trajectory_data['atomic_species_name'] = \
                     [data_lines[i+1+j].split()[1] for j in range(nat)]
+                # strip numbers
+                symbols = [''.join([i for i in s if not i.isdigit()]) for s in trajectory_data["atomic_species_name"]]
+                trajectory_data["symbols"] = symbols
 
-        # parse the initialization time (take only first occurence)
+         # parse the initialization time (take only first occurence)
```
sphuber: wrong indentation

aiida_quantumespresso/parsers/raw_parser_pw.py
```diff
@@ -1211,8 +1243,11 @@ def parse_pw_text_output(data, xml_data={}, structure_data={}, input_dict={}, pa
             if 'site n.' in data_lines[i] and 'atom' in data_lines[i]:
                 trajectory_data['atomic_species_name'] = \
                     [data_lines[i+1+j].split()[1] for j in range(nat)]
+                # strip numbers
+                symbols = [''.join([i for i in s if not i.isdigit()]) for s in trajectory_data["atomic_species_name"]]
```
sphuber: I think we want to retain the species name and not reduce it to the elemental symbol, correct?

chrisjsewell:
Hey yeh no problem to split it into some separate PRs. These are basically all the changes that have allowed me to progress with my own work, so it ended up being a larger PR than I intended

aiida_quantumespresso/calculations/pwimmigrant.py
```diff
@@ -196,6 +205,23 @@ class or using the ``set_`` methods (e.g. ``set_remote_workdir``).
                 "`with` statement context guard. See the tutorial for more "
                 "information."
             )
+        
+        # check the calc_hooks are valid
+        if calc_hooks is not None:
+            try:
+                keys = [p for f, ps in calc_hooks for p in ps]
+            except:
```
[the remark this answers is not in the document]
chrisjsewell: Yep fair, I'll have a think

aiida_quantumespresso/calculations/pwimmigrant.py
chrisjsewell: yep makes sense

aiida_quantumespresso/calculations/pwimmigrant.py
chrisjsewell: So yes its a little complex lol. What I do is create the function as a closure.
This is what it looks like in my code:

```python
for sname in ["pyrite"]:
    for modname in ["AFM", "FM", "NM"]:     

       ... # setup calc with specific inputs

        cif = get_cif_node(sname, known_cod_ids.get(sname, None))
        # ParameterData which informs how to modify the CIF structure 
        # e.g. making a primitive cell and setting spin configuration
        cif_mod = get_cifmod_node(sname, modname)

        def cifhook(struct, params):

            class Cif2structDummy(WorkChain):
                @classmethod
                def define(cls, spec):
                    super(Cif2structDummy, cls).define(spec)
                    spec.input('cif')
                    spec.input('cif_mod')
                    spec.outline(
                        cls.results,
                    )
                    spec.output('struct')
                    spec.output('params')

                def results(self):
                    self.out('struct', struct)
                    params.label = modname
                    self.out('params', params)

            Cif2structDummy.run(cif=cif, cif_mod=cif_mod)

            calc.create_input_nodes(open_transport, calc_hooks=[[cifhook, ["structure", "parameters"]]])
```

and this is an example of the resulting graph:

example_image_tree

aiida_quantumespresso/parsers/pw.py
chrisjsewell: I only put it here because I couldn't work out how to do what you say! My goal was to make sure the standard behaviour of standard `PwCalculation`s (requiring the xml) wasn't changed, hence setting it in the `Calculation` class. But if we can use the `settings` node that makes more sense

aiida_quantumespresso/parsers/raw_parser_pw.py
chrisjsewell: its used to calculate the atom positions, if specified in `crystal` (fractional) coordinates (see line ~1468)

aiida_quantumespresso/parsers/pw.py
chrisjsewell: Before I made the changes in `raw_parser_pw.parse_raw_output`, the `structure_data` was empty if the xml file was not found, hence this was my initial quick fix. But yes it should not need to be in there now (or at least in its own try-except block)

aiida_quantumespresso/parsers/raw_parser_pw.py
chrisjsewell: I added this because, in the output `TrajectoryData` it was setting `symbols` == `atomic_species_name` (couldn't actually find in the code where this occurs) and then `trajnode.get_step_structure` failed because it could not, for example, reconcile 'Fe1' with an element. So shouldn't it be that `trajectory_data['atomic_species_name']` strores species name and trajectory_data['symbols'] stores only the symbols?

Note to self: if this is changed then `raw_parser_pw.parse_raw_output` (line 129) also needs to be changed.

aiida_quantumespresso/calculations/pwimmigrant.py
chrisjsewell: This is what has worked for me, but open to suggestions.

On a side note, do you know if it is possible to set a default label and/or description in a `WorkChain` subclass?

chrisjsewell:
This PR is covered in my recent PRs, and so is no longer required

PR #226, opened 2019-02-18
Added _blocked_keywords attribute (empty dict) to BasePwCpInputGenerator

asle85:
The blocked keywords are specified in the subclass.


sphuber:
Thanks @asle85 but this was taken care of in #235

PR #238, opened 2019-04-09
First implementation of adapting the `PwParser` to parse new XML

### LINKED ISSUE #7: Implement complete parsing based on new XML format and XSD schema files
To do (some of these might be deferred to a further release):
- [x] `lattice_parameter_xml` and `number_of_species`: take from ['input'] if not present in ['output'] -- WON'T DO unless we find cases where they are missing
- [x] parse exit status
- [x] parse output -> convergence_info (mandatory in schema 1.0, optional after that)
  - [ ] broken output? Ask Pietro
- [x] after calling xsd.to_dict(), check the results of 'errors' -- DONE as warnings
- [x] also parse bands from XML? -- Seems already done
- [x] fix various time_reversal flags
- [ ] confirm that it always nbnd_up == nbnd_dw (we are indirectly checking it with an assert, but it is not directly checked)
- [ ] update / add docstrings
- [x] parse Hubbard stuff (NB: mind schema changes, 'label' is now optional) -- WON'T DO: Seb&Iurii: nothing needs to be parsed from pw.x; all Hubbard-related outputs are identical to inputs.
- [ ] maybe related: #188 
- [x] rationalize the use of custom asserts, exceptions, and logging
- [ ] parse new timing type (optional) -- Deferred until I can see examples
- [x] parse nsym and nrot, and use them as extra validation
- [ ] Tests:
  - [ ] SCF, (vc-)relax, NSCF, BANDS, Berry phase, NEB (? it uses the xml parser)
  - [ ] with elementaries and compounds
  - [ ] no spin, spin-polarized, non-collinear spin polarized
  - [ ] try to automate the checks of equivalent output between old and new XML
  - [ ] Hubbard
  - [ ] a convoluted case where occupations = fixed and total_magnetizaion = (a small positive or negative integer): this triggers nbnd_up != nbnd_dw, and should be nbnd_up - nbnd_dw = total_magnetization.

(edited by @borellim)

sphuber:
Fixes #7 

I marked this as blocked, because I would like @borellim to quickly test this to see if it can run a `PwCalculation` that produces the new XML. I already checked that the old XML works.


sphuber:
Thanks @borellim I reverted the deletion in the second commit, so now there are two schema files again, one for 6.3 and one for 6.4

PR #261, opened 2019-04-12
Put new `pytest` based parser testing infrastructure in place

### LINKED ISSUE #259: Create a minimal infrastructure for testing parsers

sphuber:
Fixes #259 

The new system leverages the fixture manager of `aiida-core` that can
easily be run as a `pytest` fixture to generate and configure an entire
AiiDA instance on the fly, complete with repository and database.
In addition to this, some simple fixtures are provided to setup a
computer and a temporary working directory for it.

Finally, there is the `generate_calc_job_node` that given an entry point
and a name, will create a `CalcJobNode` instance as if it had been
created by running the process of the entry point and adding the files
contained in the directory corresponding to the test name as the
contents of its retrieved folder. This way, parser tests are really
simple to write as they require the minimum amount of static data and
for the tests themselves the developer has full access to the python and
pytest API.


sphuber:
@giovannipizzi note that I removed all the old tests for now as they were broken and skipped anyway. In the coming weeks, together with @borellim and @odarbelaeze we will be migrating the calculation and parser plugins. I created an issue for each one and we will require each PR to come with a set of tests based on the new framework added in this PR.

sphuber:
@giovannipizzi will do

PR #268, opened 2019-04-26
Add optional use_fractional flag in the function to generate pw.x and cp.x inputs

asle85:
In the function `_generate_PWCPinputdata` of `BasePwCpInputGenerator` class, to output atomic positions in fractional coordinates (optionally; the default behaviour is the previous one, of printing in Cartesian coordinates)


giovannipizzi:
Is this function `symop_fract_from_ortho` already part of aiida-core? What is it exactly doing?

sphuber:
Yes it is in there:
```
def symop_fract_from_ortho(cell):
    """
    Creates a matrix for conversion from fractional to orthogonal
    coordinates.

    Taken from
    svn://www.crystallography.net/cod-tools/trunk/lib/perl5/Fractional.pm,
    revision 850.

    :param cell: array of cell parameters (three lengths and three angles)
    """
    import math
    import numpy

    a, b, c, alpha, beta, gamma = cell
    alpha, beta, gamma = [math.pi * x / 180 for x in [alpha, beta, gamma]]
    ca, cb, cg = [math.cos(x) for x in [alpha, beta, gamma]]
    sg = math.sin(gamma)
    ctg = cg / sg
    D = math.sqrt(sg * sg - cb * cb - ca * ca + 2 * ca * cb * cg)

    return numpy.array([
        [1.0 / a, -(1.0 / a) * ctg, (ca * cg - cb) / (a * D)],
        [0, 1.0 / (b * sg), -(ca - cb * cg) / (b * D * sg)],
        [0, 0, sg / (c * D)],
    ])
```

Previously, @asle85 used a custom ad-hoc solution to convert to fractional, but I suggested to use a solution that already exists, like in `aiida-core`, such that we would have something that is tested. Except, I see that this function is not used at all in `aiida-core`, so I am not sure if this makes the situation any better.

giovannipizzi:
Mmm... this function seems potentially not fully correct to me... We'd need to double check (e.g. it uses only a,b,c,alpha,beta,gamma and not the full cell).
It might be used probably by the cod tools plugin? In this case we might want to move it out to that plugin only?

Instead I would feel more confident in using the function `structure_to_spglib_tuple`, for instance, from `aiida.tools.data.structure`, and taking the second element of the return tuple, or even just extracting the logic from there, it should be simply something like this:
```
    cell = np.array(structure.cell)
    abs_pos = np.array([_.position for _ in structure.sites])
    rel_pos = np.dot(abs_pos, np.linalg.inv(cell))
```
(apparently anyway now numpy is a dependency of aiida-core so we have it)

aiida_quantumespresso/calculations/__init__.py
```diff
@@ -379,12 +377,26 @@ def _generate_PWCPinputdata(self, parameters, settings, pseudos, structure, kpoi
...
-                "{0} {1:18.10f} {2:18.10f} {3:18.10f} {4}\n".format(
-                    site.kind_name.ljust(6), site.position[0], site.position[1],
-                    site.position[2], fixed_coords_string))
+        if use_fractional:
+            import numpy as np
+            atomic_positions_card_list = ["ATOMIC_POSITIONS crystal\n"]
+            cell = np.array(structure.cell)
+            abs_pos = np.array([_.position for _ in structure.sites])
+            rel_pos = np.dot(abs_pos, np.linalg.inv(cell))
+            for site, site_rel_coords, fixed_coords_string in zip(
```
sphuber: It looks like this for-loop can be abstracted as it is pretty much identical in both cases.
The only thing that you need to put inside the conditional is the `atomic_positions_card_list` and then a variable `coordinates` which contains either fractional or absolute.

asle85: Agreed, I moved the for loop outside the conditional.

PR #271, opened 2019-04-29
Fixes 243 migrate CP calculation

odarbelaeze:
Try this one again, to test:

```bash
verdi run examples/cp.py
```

The `cp.py` example might replace one of the existing examples or become a cli launcher alla. pw.

@sphuber, @borellim, @giovannipizzi I think this is heading "more" on the right direction, what do you think?


aiida_quantumespresso/parsers/raw_parser_cp.py
```diff
@@ -210,11 +188,12 @@ def parse_cp_raw_output(out_file,xml_file=None,xml_counter_file=None):
             raise AssertionError('%s found in both dictionaries' % key)
         # out_data keys take precedence and overwrite xml_data keys,
         # if the same key name is shared by both (but this should not happen!)
+
     final_data = dict(list(xml_data.items()) + list(out_data.items()) + list(xml_counter_data.items()))
 
     # TODO: parse the trajectory and save them in a reasonable format
 
-    return final_data,job_successful
+    return final_data, job_successful
```
sphuber: This was the old signature of return value for parsers, a tuple of a list of output nodes and a boolean, indicating success or failure. Since now we have the `self.out` and `self.exit_codes` respectively, we no longer need to return anything here

examples/cp.py
```diff
@@ -0,0 +1,94 @@
...
+        primitive_cell=True,
+    )
+    structure = StructureData(ase=ase_structure)
+    structure.store()
+
+    calculation = CalculationFactory("quantumespresso.cp")
+    builder = calculation.get_builder()
+
+    # get some code
+    code = Code.objects.all()[0]
```
sphuber: Note that this will horribly fail of course unless the first code so happens to be configured to run `cp.x`. If this a testing script just for yourself, I'd rather not put it in the repository (in any case this `example` directory is so old and outdated, we should remove it). If we want to make it more generally usable we could consider adding a launcher, however, we should think if we really want to add one for each code.

aiida_quantumespresso/calculations/cp.py
```diff
@@ -30,97 +34,176 @@ class CpCalculation(BasePwCpInputGenerator, CalcJob):
...
+    _FILE_XML_PRINT_COUNTER = os.path.join(
+        BasePwCpInputGenerator._OUTPUT_SUBFOLDER,
+        "{}_{}.save".format(BasePwCpInputGenerator._PREFIX, _CP_WRITE_UNIT_NUMBER),
+        _FILE_XML_PRINT_COUNTER_BASENAME,
+    )
+
+    # Input file "sections" that we are going to write by calculation type
+    # The term namelist is part of FORTRAN's jargon
+    _automatic_namelists = {
+        "scf": ["CONTROL", "SYSTEM", "ELECTRONS"],
```
sphuber: Even though not a hard convention, the custom is to use single quotes for string literal identifiers such as keys in dictionaries.

examples/cp.py
borellim: What's the problem with adding a launcher for every code? Just the effort required?

sphuber: The "problem" is that this adds an executable for each launcher in the environment of the user. Having a few may be ok, but polluting the space with dozens of executables may not be desirable. If we do want to support multiple launchers, we may want to refactor them to use `click` and its group functionality to create one end point, that allows the other commands to be reached through subcommands, e.g. `aiida-quantumespresso launch calculation pw --help`

borellim: I see. Makes sense.

aiida_quantumespresso/parsers/raw_parser_cp.py
sphuber: Scratch that, this is the raw parser, where it is fine.

odarbelaeze:
@sphuber comments addressed.

sphuber:
> @sphuber comments addressed.

I don't see a parser unit test?

odarbelaeze:
@sphuber all systems should be ready TM

sphuber:
Great work @odarbelaeze thanks. Although I do want it stated for the record that I still think that the second commit title would have greatly benefited from the reference `CpParser`. There are a lot more places in this repo where `TrajectoryData` is used and where this might potentially apply. From the message you might think it was done everywhere and not just in the `CpParser`

PR #272, opened 2019-05-01
Update PwBandStructureWorkChain to the latest aiida-core.

borellim:
These changes should help with the Jupyter notebook on band structure for the aiida tutorial.

### DIFF
--- a/aiida_quantumespresso/workflows/pw/band_structure.py
+++ b/aiida_quantumespresso/workflows/pw/band_structure.py
@@ -20,7 +20,7 @@ class PwBandStructureWorkChain(WorkChain):
         super(PwBandStructureWorkChain, cls).define(spec)
         spec.input('code', valid_type=orm.Code)
         spec.input('structure', valid_type=orm.StructureData)
-        spec.input('protocol', valid_type=orm.Dict, required=False, default=orm.Dict(dict={'name':'theos-ht-1.0'}))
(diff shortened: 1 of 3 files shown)


aiida_quantumespresso/workflows/pw/band_structure.py
```diff
@@ -20,8 +20,8 @@ def define(cls, spec):
         super(PwBandStructureWorkChain, cls).define(spec)
         spec.input('code', valid_type=orm.Code)
         spec.input('structure', valid_type=orm.StructureData)
-        spec.input('protocol', valid_type=orm.Dict)
-        spec.input('scf_options', valid_type=orm.Dict)
+        spec.input('protocol', valid_type=orm.Dict, required=False, default=orm.Dict(dict={'name':'theos-ht-1.0'}))
```
sphuber: specifying a default will automatically set `required=False` so explicitly defining it here is superfluous

borellim: I prefer it to be explicit, but I don't really mind. I'll remove it.

And we tested it on the tutorial VMs: it seems to work, although we couldn't complete the calculations for other reasons.

sphuber: Explicit is good, but in this case it is really superfluous. If you provide a default, it is automatically not required, since if not specified the default will be used, by definition

sphuber:
thanks @borellim

borellim:
Thank you @sphuber for the review and merge.

PR #275, opened 2019-05-06
Add sssp 1.1 protocol

borellim:
@sphuber : should I also change this line:
```
'pseudo_default': 'SSSP-efficiency-1.0',
```
to
```
'pseudo_default': 'SSSP-efficiency-1.1',
```
?
Do you see any other changes necessary?

I'm testing with the 4 structures of the PwBandStructureWorkChain tutorial, in the following way:
```python
results = launch.run(
    PwBandStructureWorkChain,
    code=code,
    structure=structure_GaAs,
    protocol=Dict(dict={'name':'theos-ht-1.0', 'modifiers':{'pseudo':'SSSP-efficiency-1.1'}})
)
```
and it's working fine for now.

### DIFF
--- a/aiida_quantumespresso/utils/protocols/pw.py
+++ b/aiida_quantumespresso/utils/protocols/pw.py
@@ -4,25 +4,36 @@ import json
 import os
 import six
 
+
+def _load_pseudo_metadata(filename):
+    """
+    Load from the current folder a json file containing metadata (incl. suggested cutoffs)
+    for a library of pseudopotentials.
+    """
+    with open(os.path.join(os.path.dirname(os.path.abspath(__file__)), filename)) as f:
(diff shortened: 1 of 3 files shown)


borellim:
@ltalirz I believe this is what you need.

sphuber:
Yes, I think `v1.1` should be the default if it is included. However, this will only work if you actually have a pseudo potential family uploaded to your database correct? Just to say that we *are* including the cut off dictionaries. This would make sense if with that it becomes self-contained but it is not. For now it is maybe OK, but we should look into making an `aiida-sssp` plugin and making that an optional dependency of `aiida-quantumespresso`.

borellim:
Still testing, do not merge.

PR #277, opened 2019-05-06
Forgot to add CLI launcher for dos.x

borellim:


aiida_quantumespresso/cli/calculations/dos.py
```diff
@@ -0,0 +1,62 @@
...
+from aiida_quantumespresso.cli.utils import validate
+
+# required inputs: code, input_parameters, (parent_calculation or parent_folder)
+# optional: settings
+
+@click.command()
+@options.CODE(required=True, type=types.CodeParamType(entry_point='quantumespresso.dos'))
+#@options.CALCULATION(required=True, type=types.CalculationParamType(sub_classes=('aiida.calculations:quantumespresso.pw',)))
+#@options.CALCULATION(required=True, type=CalcJobNode)
+@options.CALCULATION(required=True)
```
borellim: @sphuber do you have a suggestion on how to do ask for a calculation from a specific plugin? It seems that now all `CalcJobNode`s are equal, i.e. there's no such thing as a `PwCalculationNode`, and therefore I cannot ask for a specific class either directly or via the entry point system. Is this correct?

borellim: Passing `sub_classes=('aiida.calculations:quantumespresso.pw',)` tells me that `PwCalculation` is not a subclass of `CalculationNode`.

sphuber: Yes this is correct, because `PwCalculation` is a sub class of `Process` and not of `Node`. To be able to filter for nodes that were created by the `PwCalculation` process class, you would have to filter on the `process_type` of the nodes. This is currently not implemented in the `IdentifierParamType`.

borellim: Ok, thanks a lot.

sphuber:
Please don't add things to the `.pre-commit-config.yaml`, if anything we should be removing these exclusions. If you add code, it should pass the linter

borellim:
Fine.

aiida_quantumespresso/cli/calculations/dos.py
```diff
@@ -0,0 +1,65 @@
...
+# required inputs: code, input_parameters, (parent_calculation or parent_folder)
+# optional: settings
+
+
+@click.command()
+@options.CODE(required=True, type=types.CodeParamType(entry_point='quantumespresso.dos'))
+# Note: you cannot ask for a CalcJobNode created by a specific plugin, because you can get subclasses of Process via the
+# entry point system, but not subclasses of Node. You could filter by process_type of the CalcJobNode, but this is
+# currently not implemented in IdentifierParamType.
+#@options.CALCULATION(required=True, type=types.CalculationParamType(
```
sphuber: Instead of commented code, please put this in an issue to add and reference an issue on `aiida-core` for the feature request.

aiida_quantumespresso/cli/calculations/dos.py
```diff
@@ -0,0 +1,65 @@
...
+from __future__ import absolute_import
+import click
+
+from aiida.orm import RemoteData
+from aiida.cmdline.params import options, types
+from aiida.cmdline.utils import decorators
+
+from aiida_quantumespresso.cli.utils import options as options_qe
+
+# required inputs: code, input_parameters, (parent_calculation or parent_folder)
```
sphuber: Why are these comments here?

aiida_quantumespresso/cli/calculations/dos.py
```diff
@@ -0,0 +1,65 @@
...
+    expected_process_type = 'aiida.calculations:quantumespresso.pw'
+    if calculation.process_type != expected_process_type:
+        raise click.BadParameter('The input calculation node has a process_type: {}; should be {}'.format(
+            calculation.process_type, expected_process_type))
+    parent_folder = calculation.get_outgoing(node_class=RemoteData, link_label_filter='remote_folder').one().node
+
+    inputs = {
+        'code': code,
+        'parent_folder': parent_folder,
+        'parameters': Dict(dict={}),
```
sphuber: If `settings` is optional, better not pass one if it is going to be empty. Same goes for `parameters` but I think your comment above said it *is* required. Which does not make sense if it is allowed to be empty. If that is true, then the `CalcJob` should make it not required

borellim: Good points. I modeled the spec after `BasePwCpInputGenerator`, where `parameters` is required and `settings` is not. I'll make `parameters` and `settings` not required in `NamelistsCalculation` (which Dos inherits from), but I'm not sure if I should do the same for `BasePwCpInputGenerator`...

In general, is it better to add a `default=Dict(dict={})` or to just check for the presence of the input? The former is cleaner, but can add a useless empty input node.

aiida_quantumespresso/cli/calculations/dos.py
borellim: Self notes and documentation for the readers. They are not essential, I can remove them.

aiida_quantumespresso/cli/calculations/dos.py
sphuber: If the default is an empty dict, it is better not to specify it and just check for it in the code, assigning an empty dict there if not specified. This is because mutable defaults, such as dictionaries, are [dangerous](https://blog.florimond.dev/python-mutable-defaults-are-the-source-of-all-evil). The same story goes for AiiDA nodes and it has caused us a lot of problems in the past. We still use nodes as defaults for certain work chains, but we really shouldn't.

This is a more fundamental problem in `aiida-core` where we should forbid mutable defaults and only allow lambdas. They can then construct and store a default node each time it is invoked. This prevents the same node being passed around each time the process is launched. For the time being we should really limit using these defaults unless absolutely necessary.

borellim: Thanks, very interesting.
I'll keep 'parameters' as required in `BasePwCpInputGenerator` for now, but make it (and 'settings') optional without a default in `NamelistsCalculation`.

borellim:
Thanks for the review @sphuber. Is it better now?

PR #280, opened 2019-05-07
Migrate Pw2Wannier90

borellim:
* migrate Pw2Wannier90Calculation & Pw2Wannier90Parser;
* add CLI launcher for pw2wannier90.x;
* clean up `calculations/dos.py`;
* make a minor change to cp.x CLI launcher to make prospector happy.


borellim:
The second commit does not belong to the same topic, but I took the chance to fix it.

borellim:
By the way, thanks a lot to @normarivano for the help

aiida_quantumespresso/cli/calculations/pw2wannier90.py
```diff
@@ -0,0 +1,87 @@
...
+from __future__ import absolute_import
+import click
+
+from aiida.cmdline.params import options, types
+from aiida.cmdline.params.options import OverridableOption
+from aiida.cmdline.utils import decorators
+
+from aiida_quantumespresso.cli.utils import options as options_qe
+
+PARENT_FOLDER = OverridableOption(
```
sphuber: These will most likely be reusable so maybe we can put them in `aiida_quantumespresso.cli.utils.options`?

aiida_quantumespresso/cli/calculations/pw2wannier90.py
```diff
@@ -0,0 +1,87 @@
...
+    '-F',
+    '--parent_folder',
+    'parent_folder',
+    metavar='RemoteData/FolderData',  # placeholder in the help text
+    type=types.NodeParamType(sub_classes=('aiida.data:remote', 'aiida.data:folder')),
+    help='A RemoteData or FolderData node identified by its ID or UUID.')
```
sphuber: Just as an FYI, technically the `NodeParamType` also supports loading the node from its label if it has one and it is unique

aiida_quantumespresso/cli/calculations/pw2wannier90.py
```diff
@@ -0,0 +1,87 @@
...
+@options.CODE(required=True, type=types.CodeParamType(entry_point='quantumespresso.pw2wannier90'))
+@PARENT_FOLDER(required=True, help='RemoteData folder containing the output of a PW calculation')
+@SINGLE_FILE(required=True, help='SinglefileData containing the .nnkp file generated by a wannier90.x preprocessing')
+@options_qe.MAX_NUM_MACHINES()
+@options_qe.MAX_WALLCLOCK_SECONDS()
+@options_qe.WITH_MPI()
+@options_qe.DAEMON()
+@decorators.with_dbenv()
+def cli(code, parent_folder, single_file, max_num_machines, max_wallclock_seconds, with_mpi, daemon):
+    """Run a PwCalculation."""
```
sphuber: Incorrect docstring

aiida_quantumespresso/cli/calculations/pw2wannier90.py
```diff
@@ -0,0 +1,87 @@
...
+
+    Pw2wannierCalculation = CalculationFactory('quantumespresso.pw2wannier90')  # pylint: disable=invalid-name
+
+    parameters = {
+        'inputpp': {
+            'write_amn': True,
+            'write_mmn': True,
+            'write_unk': False,
+            'scdm_proj': True,
+            'scdm_entanglement': 'isolated',
```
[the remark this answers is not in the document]
sphuber: I agree that for now this is good enough and if the need arises this can be dealt with at a later stage

aiida_quantumespresso/parsers/pw2wannier90.py
```diff
@@ -1,51 +1,47 @@
...
+        
+        # In case of parsing error or calculation error, return an exit code
+        if not successful_raw:
+            if "Computation did not finish properly" in out_dict['warnings']:
+                return self.exit_codes.ERROR_JOB_NOT_DONE
+            elif 'error_message' in list(out_dict.keys()):
+                return self.exit_codes.ERROR_GENERIC_QE_ERROR
+            else:
+                return self.exit_codes.ERROR_GENERIC_PARSING_FAILURE
+        else:
```
sphuber: This else and `return None` is not necessary as it is implicit without a return statement

aiida_quantumespresso/cli/calculations/pw2wannier90.py
borellim: I don't understand what you mean, sorry. What's the label?

aiida_quantumespresso/cli/calculations/pw2wannier90.py
borellim: I understand that the short option names should be unique between all options defined in `cli.utils.options`, so I didn't want to add any more. For instance, `-S` for single_file would clash with `-S` for smearing (or `-s` for structure). If this is not a problem I can move them there, sure.

aiida_quantumespresso/cli/calculations/pw2wannier90.py
sphuber: In AiiDA, each `Node` has three identifiers: its `pk`, `uuid` and `label`. The `label` is the one you can set for example on the `ProcessNode` through the input `metadata.label`, but this is a property of every node. It is simply one of the columns of the node table. Now, since the label is not necessarily unique, unlike the pk and uuid, it might not be very useful to use, but *it can* be used for this. You could for example label a specific `FolderData` with a special label, which you could then use to reference it, instead of memorizing its pk or uuid.

borellim: Interesting, thanks. I tried it and it works already with this cli launcher.

aiida_quantumespresso/cli/calculations/pw2wannier90.py
sphuber: Fair point, but I would say that the `smearing` is way more specific and so the single file data might take precedence. In any case, for now it is not that important. As soon as we have another launcher that requires the same option, we can centralize it. Also, although sometimes useful, a short option is not necessary. For less common options like `smearing` we could just decide to only provide the long variant so there won't be a clash

borellim: ...or just accept that they won't be able to be used together?

borellim: On second thought, no, that's not a great idea. Ok thanks.

PR #286, opened 2019-05-09
Migrate `PhCalculation` and `PhParser` to `aiida-core==1.0.0`

sphuber:
Fixes #240 and fixes #255

Issue #279, opened 2019-05-07
Add parent calculation as alternative input to NamelistsCalculation classes?


borellim:
Notes taken from `calculations/dos.py` on how to possibly do this:

```python
    
    # TODO: if you want to be able to specify an input calculation (rather than a folder),
    #       you must add this as an input to the NamelistsCalculation with non_db=True, like this:
    # from aiida.orm import CalcJobNode
    # spec.input('parent_calculation', valid_type=CalcJobNode, required=False, non_db=True)
    # TODO: Then you must add the following logic to the prepare_for_submission method
    #       (note that you can't actually modify self.inputs, because it's a 'AttributesFrozendict')
    # def prepare_for_submission(self, folder):
    #     """
    #     Set the parent folder if necessary, or raise if neither parent_folder nor parent_calculation were given.
    #     Then call define() from the super class.
    #     """
    #     from aiida.common import InputValidationError
    #     if 'parent_folder' in self.inputs and 'parent_calculation' in self.inputs:
    #         raise InputValidationError('You cannot specify both inputs \'parent_folder\' and \'parent_calculation\'.')
    #     elif 'parent_folder' in self.inputs:
    #         pass
    #     elif 'parent_calculation' in self.inputs:
    #         parent_calc = self.inputs.parent_calculation
    #         try:
    #             # TODO: check that the calc's process_type is quantumespresso.pw 
    #             parent_folder = parent_calc.get_outgoing(node_class=RemoteData, link_label_filter='remote_folder').one().node
    #         except ValueError:
    #             raise InputValidationError('Parent calculation does not have a remote folder output node.')
    #         self.inputs['parent_folder'] = parent_folder
    #     else:
    #         raise InputValidationError('Either \'parent_calculation\' or \'parent_folder\' must be in the inputs.')
    # 
    #     super(DosCalculation, self).prepare_for_submission(folder)
    # 
    # TODO: ... except that it won't work, because I can't validate the process_type of the parent calc from the
    #  NamelistsCalculation class: the acceptable types may change according to what is the concrete
    #  type (e.g. DosCalculation, ...).
    # TODO: Maybe this can be done with a validator for the input port? (see plumpy.ports.InputPort.__init__())
```

odarbelaeze:
As a way to signal restarts, similar to the `pw` and `cp` ones?

borellim:
[...]

PR #287, opened 2019-05-09
Migrate `Q2rCalculation` and `Q2rParser` to `aiida-core==1.0.0`

### LINKED ISSUE #248: Migrate `Q2rCalculation` to be compatible with `aiida-core==1.0.0`

### LINKED ISSUE #257: Migrate `Q2rParser` to be compatible with `aiida-core==1.0.0`

sphuber:
Fixes #248 and fixes #257


borellim:
I assume this needs to be rebased after #286 is merged?

sphuber:
>  Sorry, what I mean is that it also contains the changes in that PR.

Yes, that is why I marked it blocked, because I will need to rebase after the other PR has been merged.

aiida_quantumespresso/calculations/dos.py
```diff
@@ -1,20 +1,13 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import
-from aiida.orm import RemoteData, FolderData, CalcJobNode
-from aiida.orm import Dict, XyData
+
+from aiida import orm
```
borellim: Is this just your preference or do you want to enforce this style across the code? Because I'm not a fan of it.

aiida_quantumespresso/cli/calculations/q2r.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # Check that the parent calculation node comes from quantumespresso.ph.
+    # I cannot move this check into the option declaration, because CalcJobNode is not subclassed by the specific
+    # calculation plugins (only Process is), and there is no feature yet to filter by the associated process_type.
+    expected_process_type = 'aiida.calculations:quantumespresso.ph'
+    if calculation.process_type != expected_process_type:
+        raise click.BadParameter('The input calculation node has a process_type: {}; should be {}'.format(
+            calculation.process_type, expected_process_type))
+
+    parent_folder = calculation.get_outgoing(node_class=orm.RemoteData, link_label_filter='remote_folder').one().node
```
odarbelaeze: Might be a good think to check that the calculation finished and finished ok before asking for the results.

aiida_quantumespresso/calculations/dos.py
sphuber: It is a personal preference yes, for the following reasons:
* It is less sensitive to internal path changes in the `aiida.orm` module
* It makes the import section less cluttered
* Adding code with new import requirements do not require new import statements
* Don't run the risk of forgetting to remove unused imports when removing code.

The one counter-argument could be that importing the whole module will be slower, but it is not as `aiida.orm` already import all lower lying modules anyway. At 30 - 60 micro seconds I anyway do not think that this weighs up to the advantages listed above.

borellim: Good points, however the one counter-argument to me is that now you have cluttered all mentions of orm classes with a `orm.` prefix, and to me there's no need since it's clear enough (I believe) that they belong to AiiDA already.

borellim: Unused imports and missing imports should be caught by the linter. And I don't get why it's less sensitive to internal path changes, as you will have to use the rest of path anyway when you use the class... Or not?

sphuber: The import robustness comes into play when importing like `from aiida.orm.nodes.data.str import Str`, which becomes more likely if we have multiple individual import statements. This would have to be changed if the internal path of `aiida-core` changes. For `v1.0.0` we have [specified a list](https://github.com/aiidateam/aiida_core/wiki/AiiDA-public-modules,-classes-and-functions) in the wiki with parts of the API that are intended to be public and as such we expose them on the second level packages. To me it only makes sense to use them as such, but as said before, indeed ultimately it is a preference

And, yes, the linter should catch those, but unfortunately for now almost all files are excluded because fixing the prospector warnings would take a lot of time. We definitely need to start uncommenting those though

aiida_quantumespresso/cli/calculations/q2r.py
sphuber: Good point, I think we can address this in #279 as a generic problem of validating input that is required for a restart. Although here your previous argument would also be important, that if we require the input be passed through a `calculation`, this would forbid the calculation to be launched from an input that was created otherwise, but valid nonetheless. Maybe then the option for the CLI would be better served with just directly accepting the data node itself.

aiida_quantumespresso/calculations/dos.py
borellim: > The import robustness comes into play when importing like from `aiida.orm.nodes.data.str import Str` [...]

Right, so let's just use `from aiida.orm import Str` instead. :)

And agreed on the prospector warnings.

Thanks a lot for the chat. I don't mind each using their own preferred style.

sphuber: Fair enough, although ideally we adopt a single style and be consistent. I can certainly see your point when you just import a few classes from the module, but as soon as you start to import more than a handful, it simply seems to become messy

PR #288, opened 2019-05-09
Migrate `MatdynCalculation` and `MatdynParser` to `aiida-core==1.0.0`

### LINKED ISSUE #241: Migrate `MatdynCalculation` to be compatible with `aiida-core==1.0.0`

### LINKED ISSUE #253: Migrate `MatdynParser` to be compatible with `aiida-core==1.0.0`

sphuber:
Fixes #241 and fixes #253


aiida_quantumespresso/calculations/matdyn.py
```diff
@@ -1,121 +1,72 @@
...
-        a different node) or the keys inside settings, using the 'pop' method,
-        so that inputdict and settings should remain empty at the end of 
-        _prepare_for_submission, if all flags/nodes were recognized
-        """
-        from aiida.common import InputValidationError
-        
+    @classmethod
+    def define(cls, spec):
+        super(MatdynCalculation, cls).define(spec)
+        spec.input('parent_folder', valid_type=ForceconstantsData, required=True)
```
borellim: Is it appropriate to call it 'parent_folder' when `ForceconstantsData` is a subclass of `SinglefileData`?

sphuber: No it is not really, I agree, but changing it would be a backwards incompatible change. I would still be in favor of doing it, but that should be done in a separate PR most likely.

borellim: Ok, but I would do it before the release of 3.0.0, since we are already making a new major release.

sphuber: I agree, let's see if we can convince Gio of this as well

PR #291, opened 2019-05-09
Create a single CLI script `aiida-quantumespresso`

### LINKED ISSUE #289: Bundle all CLI launchers in single click console script
Adding a single console script for each calculation and workflow launcher that we would like to make available is not tenable.

sphuber:
Fixes #289 

This new executable, installed through the `console_scripts` entry point
supersedes all the individual launcher scripts that were being added.
Having a single script for each calculation or workflow launching script
was not tenable as it was polluting the namespace of the environment.

The new `aiida-quantumespresso` has various advantages:

 * Only a single executable is required to host all CLI commands
 * Name makes immmediately clear that it belongs to this plugin
 * The `click` library is used to make a robust CLI tool
 * Tab-completion can be enabled with single line in `.bashrc`

Note that the `fastentrypoints.py` script is used to make the loading of
the entry points a lot faster. Without it, loading the tool takes 2
seconds instead of 0.2 seconds. Without it, tab-completion is unusable.


borellim:
How do you enable tab completion?

sphuber:
> How do you enable tab completion?

Just as you do for verdi, e.g. by adding this to your bashrc or environment activation script: `eval "$(_AIIDA_QUANTUMESPRESSO_COMPLETE=source aiida-quantumespresso)"`

I will update the `README.md` which I realize anyway still contained a reference to the launcher script I just removed here.

sphuber:
@borellim I have updated the `README.md` with instructions how to enable tab-completion

borellim:
Thanks Sebastiaan.

Since the `calculation` and `workflow` click groups have only `launch` as sub-command, wouldn't it be better to remove one level and use `launch_calculation` and `launch_workflow` instead? Seems more intuitive, faster and less prone to errors. Or are you figuring that we will add other commands beside `launch` in the future?

This is really cool stuff!

sphuber:
> Since the `calculation` and `workflow` click groups have only `launch` as sub-command, wouldn't it be better to remove one level and use `launch_calculation` and `launch_workflow` instead?

Exactly, I anticipate other useful commands and in order to not have to break backwards compatibility I organized it like this. Anyway, with the tab-completion it should not make a difference in terms of efficiency nor error-proneness

sphuber:
@borellim can we merge this?

borellim:
Ok. Thanks a lot Sebastiaan.

PR #293, opened 2019-05-09
Fix method _get_following_text of NamelistsCalculation 

borellim:
...that was broken in previous commit

### DIFF
--- a/aiida_quantumespresso/calculations/namelists.py
+++ b/aiida_quantumespresso/calculations/namelists.py
@@ -5,76 +5,67 @@ Plugin to create a Quantum Espresso input file for a generic post-processing
 afterwards).
 """
 from __future__ import absolute_import
+
 import os
-from aiida.common import InputValidationError
-from aiida.common import CalcInfo, CodeInfo
-from aiida.orm import Dict, Code
-from aiida.orm import RemoteData, FolderData, SinglefileData
-from aiida.engine import CalcJob
-from aiida.common import CodeInfo
-from aiida.plugins import ParserFactory
-from aiida_quantumespresso.calculations import _lowercase_dict, _uppercase_dict
-from aiida_quantumespresso.utils.convert import convert_input_to_namelist_entry
 import six
 
-# TODO: review usage of class variables and input names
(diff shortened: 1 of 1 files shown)


borellim:
Now pw2wannier90 works again.

@sphuber Why did we remove the settings_dict argument? The reasoning in the docstring made sense (that the function should pop from settings_dict). Is the idea now that it pops from self.settings instead?

borellim:
I changed the docstring too

sphuber:
I removed it because the only reference I could find was the one in `MatdynCalculation` and that did not use the settings at all. It was even worse at first because `NamelistsCalculation._get_following_text` specified `def _get_following_text(self, inputdict, settings):` at first. So where is `Pw2wannier90Calculation` using it, because I cannot find it.

borellim:
It's not using it directly, but it's calling prepare_for_submission of the base class NamelistsCalculation, which calls _get_following_text with the wrong arguments.

sphuber:
But if `MatdynCalculation` is the only one that actually implements the function, and it doesn't need the settings dictionary, then it seems more logical to simply remove that from the function signature instead.

aiida_quantumespresso/calculations/namelists.py
```diff
@@ -5,76 +5,67 @@
...
-from aiida.engine import CalcJob
-from aiida.common import CodeInfo
 from aiida.plugins import ParserFactory
+
 from aiida_quantumespresso.calculations import _lowercase_dict, _uppercase_dict
 from aiida_quantumespresso.utils.convert import convert_input_to_namelist_entry
-import six
 
-# TODO: review usage of class variables and input names
-# TODO: add exit codes
```
borellim: My idea here was to move some exit codes in the base class, to enforce max consistency of exit codes across subclasses. Bad idea?

sphuber: No good, idea! But we should not do this *just* for the `NamelistsCalculation` but we should homogenize as much as possible for all calculations. So let's open an issue and tackle this once all classes are migrated

PR #305, opened 2019-05-13
Add a test for DosParser

borellim:


tests/parsers/test_dos.py
```diff
@@ -0,0 +1,53 @@
...
+    data_regression.check({
+        'parameters': out_params,
+        'dos': {
+            'labels': dos_labels,
+            'units': dos_units,
+        }
+    })
+    num_regression.check({
+        'dos_val_{}'.format(i) : val for i,val in enumerate(dos_values)
+    }, default_tolerance=dict(atol=0, rtol=1e-18))
```
borellim: Do these tolerances make sense? The magnitudes of the values range from 83.0 to 7.77e-88, which is why I set the absolute tolerance to zero. But I'm not sure about the relative tolerance.
Docs: https://pytest-regressions.readthedocs.io/en/latest/api.html#num-regression

borellim:
Are these files too heavy?

sphuber:
> Are these files too heavy?

In any case, they need not be this heavy. If we generate an output file with just a few output lines, let's say 10, I would say that should test the output parsing just as much. We don't even necessarily care that the output file comes from a real calculation, i.e. you could just delete almost all the lines from the output, except for a couple. I don't think the parser does any consistency checks on the number of lines in the output, compared to the values in the input, so this should work. I would say that is testing more than enough for now and will save a lot of bytes

tests/parsers/test_dos.py
sphuber: Is there a reason to suspect that the parsed values will change significantly? Did you try to use the normal `data_regression` on the `output_dos` node? Did that not work due to float precision?

borellim: `data_regression` does not work on numpy arrays, as they are not yaml-serializable:
```
[...]
../../../venvs/aiida_py3.6/lib/python3.6/site-packages/yaml/representer.py:57: in represent_data
    node = self.yaml_representers[None](self, data)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <pytest_regressions.data_regression.RegressionYamlDumper object at 0x7fbc7cda0240>, data = array([-83.947, -83.937, -83.927, ...,  36.543,  36.553,  36.563])

    def represent_undefined(self, data):
>       raise RepresenterError("cannot represent an object: %s" % data)
E       yaml.representer.RepresenterError: cannot represent an object: [-83.947 -83.937 -83.927 ...  36.543  36.553  36.563]

../../../venvs/aiida_py3.6/lib/python3.6/site-packages/yaml/representer.py:229: RepresenterError
```

sphuber: We could deserialize them ourselves easily, before calling the `data-regression` no?

borellim: I guess we could, but why?

borellim:
> In any case, they need not be this heavy.

Gotcha. I reduced the data to 24+1 lines.

tests/parsers/test_dos.py
borellim: Let me clarify: if we serialize these numpy arrays ourselves, we still have to solve the problem of truncation and that of comparison tolerances. If all you want is to have zero tolerance, we can just set `atol=0` and `rtol=0` (it works on my machine, altough I suspect we might indeed incur in precision problems on different machines).

By the way, this is the comparison formula (from `numpy.isclose`): `absolute(a - b) <= (atol + rtol * absolute(b))`

sphuber: Fair enough. Honestly I don't have a good answer about the tolerance, but that is also not the focus of what we are testing here, so I think this is fine

tests/parsers/test_dos.py
```diff
@@ -0,0 +1,53 @@
...
+from aiida.orm import FolderData
+from aiida.common import AttributeDict
+
+
+@pytest.fixture
+def dos_inputs():
+    # The DosParser doesn't really need to access the parent folder, but we'd like to make the inputs as realistic
+    # as possible, so we create an empty FolderData and attach it as an input to the current CalcJobNode.
+    inputs = {
+        'parent_folder': FolderData().store(),
```
sphuber: Just for future reference, there is not need to call `store` manually here

sphuber:
Thanks @borellim

tests/parsers/test_dos.py
borellim: Ah, thanks.

Issue #306, opened 2019-05-13
Automatic parallelization of PW is broken


borellim:
- [x] run_init() crashes at `PwCalculation.process()` -- Addressed in #326 
- [x] XML schema validation fails with init-only calculations, because the XML file is incomplete (this error is only logged) -- Moved to #327 
- [x] The actual parser also fails for the same reason -- Moved to #327 
- [x] Finally: is the automatic parallelization algorithm still good? Moved to #328 

Example of how to reproduce: `aiida-quantumespresso workflow launch pw-base -X 9756 -s 9752 -p SSSP_eff_PBE_v1.1_b -a`

Example errors (items 2 and 3 above; this is after patching item 1): [failed_parse_init_pw_calc.txt](https://github.com/aiidateam/aiida-quantumespresso/files/3172936/failed_parse_init_pw_calc.txt)

sphuber:
I was just looking into this since I am rewriting the parser and I wanted to include support for the initialization calculations but I cannot even run an initialization run with pw.x 6.3 with new XML without it crashing. A normal calculation works fine with the compiled version I have on my work station, but with the initialization mode I get the following segmentation fault:
```
Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x7FE5CE331E08
#1  0x7FE5CE330F90
#2  0x7FE5CDA624AF
#3  0x7FE5CDB7AFD5
#4  0x4DDB71 in __pw_restart_new_MOD_pw_write_schema at pw_restart_new.f90:158 (discriminator 22)
#5  0x4D43B0 in punch_ at punch.f90:73
#6  0x4F5219 in run_pwscf_ at run_pwscf.f90:120
#7  0x407AFA in MAIN__ at pwscf.f90:77
_aiidasubmit.sh: line 6: 21705 Segmentation fault      (core dumped) '/home/sphuber/code/qe/qe-6.3/bin/pw_new_xml.x' '-in' 'aiida.in' > 'aiida.out'
```
Unfortunately there is not more info then that. The standard output is interrupted:
[stdout_pw6.3.txt](https://github.com/aiidateam/aiida-quantumespresso/files/3265271/stdout_pw6.3.txt)

sphuber:
[...]

PR #307, opened 2019-05-13
Fix run_init() method of PwBaseWorkChain

### LINKED ISSUE #264: run_init() in PwBaseWorkChain does not work in aiida 1.0
`PwCalculation.process()` does not exist.

Use as a model the method `run_calculation()` in `BaseRestartWorkChain`.

borellim:
Should fix #264


sphuber:
Before I merge this, does this actually make the work chain run with the automatic parallelization option turned on? Or does it just address this part? Because if that's the case, then I'd rather wait to fix the whole thing

borellim:
No, it only ticks the first box of #306, the automatic parallelization as a whole is still broken. However, it moves the error from the submission of the init calculation to its parsing, so it's a net improvement in my view.

sphuber:
Superseded by PR #326

PR #308, opened 2019-05-13
(WIP) NamelistsCalculation: allow to use a FolderData as parent folder

borellim:
It's currently not possible to use a `FolderData` (rather than a `RemoteFolder`) as parent folder for calculation classes derived from `NamelistsCalculation`. E.g. if I try the CLI launcher for `Pw2wannier90Calculation`, I get this:
```
  File "/home/aiidatests/repos/aiida_1.0/aiida-core/aiida/engine/processes/calcjobs/tasks.py", line 78, in do_upload
    raise Return(execmanager.upload_calculation(node, transport, calc_info, script_filename))
  File "/home/aiidatests/repos/aiida_1.0/aiida-core/aiida/engine/daemon/execmanager.py", line 165, in upload_calculation
    handle.write(data_node.get_object_content(filename))
  File "/home/aiidatests/repos/aiida_1.0/aiida-core/aiida/orm/nodes/node.py", line 690, in get_object_content
    return self._repository.get_object_content(key)
  File "/home/aiidatests/repos/aiida_1.0/aiida-core/aiida/orm/utils/repository.py", line 134, in get_object_content
    with self.open(key) as handle:
  File "/home/aiidatests/repos/aiida_1.0/aiida-core/aiida/orm/utils/repository.py", line 104, in open
    return io.open(self._get_base_folder().get_abs_path(key), mode=mode)
IsADirectoryError: [Errno 21] Is a directory: '/home/aiidatests/venvs/aiida/.aiida/repository-aiidatests/repository/node/2b/b6/0ee6-a2ea-4443-be85-e47d30687859/path/.aiida'
```
So I tried to fix it, but I think I'm facing a limitation of the engine.

The source of the error above is that, when we do `parent_calc_folder.list_object_names()`, some of the names are actually folders, which may have arbitrarily nested files inside. So my approach is now to traverse the folders and actually get all files, and add *them* to the `local_copy_list`.

The problem now is that some of the files I need to upload are binary files, but AiiDA tries to read them as unicode, so I get this:
```
Running a pw2wannier90.x calculation from parent FolderData<13638>... 
05/13/2019 06:44:29 PM <15313> aiida.engine.transports: [ERROR] Exception whilst using transport:
  File "/home/aiidatests/repos/aiida_1.0/aiida-core/aiida/engine/transports.py", line 106, in request_transport
    yield transport_request.future
  File "/home/aiidatests/repos/aiida_1.0/aiida-core/aiida/engine/processes/calcjobs/tasks.py", line 78, in do_upload
    raise Return(execmanager.upload_calculation(node, transport, calc_info, script_filename))
  File "/home/aiidatests/repos/aiida_1.0/aiida-core/aiida/engine/daemon/execmanager.py", line 165, in upload_calculation
    handle.write(data_node.get_object_content(filename))
  File "/home/aiidatests/repos/aiida_1.0/aiida-core/aiida/orm/nodes/node.py", line 690, in get_object_content
    return self._repository.get_object_content(key)
  File "/home/aiidatests/repos/aiida_1.0/aiida-core/aiida/orm/utils/repository.py", line 135, in get_object_content
    return handle.read()
  File "/home/aiidatests/venvs/aiida_py3.6/lib/python3.6/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe7 in position 8: invalid continuation byte
```

**My questions:**
* Is my approach sensible or is there a better way? (Maybe AiiDA should allow to pass folders in the `local_copy_list`?)
* Am I right that this needs a new feature in AiiDA?
* How should I prioritize this?

### DIFF
--- a/aiida_quantumespresso/calculations/namelists.py
+++ b/aiida_quantumespresso/calculations/namelists.py
@@ -19,6 +19,24 @@ from aiida_quantumespresso.calculations import _lowercase_dict, _uppercase_dict
 from aiida_quantumespresso.utils.convert import convert_input_to_namelist_entry
 
 
+def get_folderdata_files(folder_data, subpath=''):
+    from aiida.orm.utils.repository import FileType
+    
+    filepaths = []
+    for obj in folder_data.list_objects(subpath):
+        if obj.type == FileType.FILE:
+            filepaths.append(os.path.join(subpath,obj.name))
+        elif obj.type == FileType.DIRECTORY:
+            filepaths.extend(
+                get_folderdata_files(folder_data, os.path.join(subpath, obj.name))
+            )
+        else:
+            raise ValueError("Object of unexpected type {} in FolderData {}".format(
+                obj.type, folder_data.uuid
+            ))
+    return filepaths
+
+
 class NamelistsCalculation(CalcJob):
     """`CalcJob` implementation to serve as base class for simple post-processing tools of Quantum ESPRESSO."""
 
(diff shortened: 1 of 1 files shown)


sphuber:
The problem with binary files is a known issue and has an [open PR](https://github.com/aiidateam/aiida_core/pull/2748) on `aiida-core`.

The second problem really depends. I guess what we need to copy from the `parent_folder` depends on the situation. The logic is implemented in the base class, but the contents that need to be copied most likely depends on the type of "parent_folder" that is required by the sub class, i.e. `Pw2wannierCalculation` may need other files copied then `DosCalculation`. Or maybe not, that would make it a lot easier. In that case, we could just simply have the engine copy over everything in that node. If it is a folder type, copy the whole directory, or if it is a single file, just the only file, although the plugin will have to detect the type and use the correct tuple for the local copy list. Note that this might indeed need a new feature in `aiida-core` for the `local_copy_list` to support `FolderData` in addition to `SinglefileData` which makes sense to me.

In the former case, where *what* needs to be copied depends on the sub class, we might just have to delegate this to the sub classes. If they have specific files from a single `FolderData` node that need to be copied, then they should take the responsibility of correctly building the `local_copy_list`. They can add multliple directives for a single `FolderData` node

borellim:
I see. Thank you.
When I said "this requires a new feature in AiiDA" I actually meant copying binary files, but I'm glad that there's already an open PR. It seems to me that, when that PR is merged, we could at least proceed with this approach and just blindly copy the contents from "parent/out" into "workdir/out", as is done for `RemoteData`: this would bring at least feature-parity to `FolderData` inputs.

sphuber:
I think this functionality should just be made available by `aiida-core` so I opened an issue there #3869

PR #318, opened 2019-05-24
Move fastentrypoints.py to utils/ and add it to MANIFEST.in

### LINKED ISSUE #317: fastentrypoints is not found when installing from pip
This only occurs when installing directly from pip, while it works for installations with `pip -e`. 
This is probably because the fastentrypoints.py file is not included in the pip package (probably it's just sufficient to add it to the MANIFEST.in).

borellim:
Fix #317 

`fastentrypoints.py` is moved to `utils/` following the example of aiida-core, and is added to MANIFEST.in.


sphuber:
Thanks @borellim

PR #319, opened 2019-05-29
projwfc parser QE 6.3 support

eimrek:
Hi! When using QE 6.3 (installed with --disable-xml), the energy format in the output is changed slightly compared to older versions.

Before, the lines specifying energies started with

    ==== e(

In 6.3 they start with

    e =

The fix should support both versions.


eimrek:
Sorry for the merge conflict, don't know how it got through. I fixed it now and should be directly mergeable.

PR #325, opened 2019-06-06
`BaseRestartWorkChain`: improve and adapt to `aiida-core==1.0.0`

### LINKED ISSUE #220: Adapt the BaseRestartWorkChain to exponential backoff mechanism of aiida-core
In `aiida-core==1.0.0` an exponential backoff mechanism is implemented for transport tasks of calc jobs with them being paused if even that falls through. The current `BaseRestartWorkChain` implements some logic to deal with `SUBMISSIONFAILED` calculations and `RETRIEVALFAILED` is marked as not supported. Since these states should never be encountered, the logic should be updated to reflect this.

sphuber:
Fixes #220 

The base restart work chain was originally developed to be able to deal
with problems in the submission or retrieval of calculation jobs to the
remote clusters. However, with the addition of the exponential back off
retry mechanism for calculation jobs in the engine of `aiida-core` this
no longer needs to be dealt with by the restart work chain. The logic of
the calculation is inspected. In addition, the position of the sanity
check invocation is moved from failed calculations to finished
calculations. The whole concept of the sanity check is to add an
additional check on calculations that were marked as successful by the
parser. Calculations that were already marked failed do not have to have
a sanity check.

### DIFF
--- a/aiida_quantumespresso/common/workchain/base/restart.py
+++ b/aiida_quantumespresso/common/workchain/base/restart.py
@@ -1,34 +1,31 @@
 # -*- coding: utf-8 -*-
+# pylint: disable=inconsistent-return-statements,no-member
+"""Base implementation of `WorkChain` class that implements a simple automated restart mechanism for calculations."""
 from __future__ import absolute_import
 
-import six
-from six.moves import map
-
 from aiida import orm
-from aiida.common import EntryPointError
+from aiida.common import exceptions
 from aiida.common.lang import override
-from aiida.engine import CalcJob, WorkChain, ToContext, append_
+from aiida.engine import ExitCode, CalcJob, WorkChain, ToContext, append_
 from aiida.plugins.entry_point import get_entry_point_names, load_entry_point
 
 from aiida_quantumespresso.common.exceptions import UnexpectedCalculationFailure
 
 
 class BaseRestartWorkChain(WorkChain):
-    """
-    Base restart workchain
+    """Base restart work chain
 
-    This workchain serves as the starting point for more complex workchains that will be designed to
-    run a calculation that might need multiple restarts to come to a successful end. These restarts
-    may be necessary because a single calculation run is not sufficient to achieve a fully converged
-    result, or certain errors maybe encountered which are recoverable.
+    This work chain serves as the starting point for more complex work chains that will be designed to run a calculation
+    that might need multiple restarts to come to a successful end. These restarts may be necessary because a single
+    calculation run is not sufficient to achieve a fully converged result, or certain errors maybe encountered which
+    are recoverable.
 
-    This workchain implements the most basic functionality to achieve this goal. It will launch
-    calculations, restarting until it is completed successfully or the maximum number of iterations
-    is reached. It can recover from simple submission failures and will handle any other errors through
-    any of the error handlers that have been provided.
+    This work chain implements the most basic functionality to achieve this goal. It will launch calculations,
+    restarting until it is completed successfully or the maximum number of iterations is reached. It can recover from
+    errors through error handlers that can be attached dynamically through the `register_error_handler` decorator.
 
-    The idea is to subclass this workchain and leverage the generic error handling that is implemented
-    in the few outline methods. Part of the suggested outline would look something like the following::
+    The idea is to sub class this work chain and leverage the generic error handling that is implemented in the few
+    outline methods. The minimally required outline would look something like the following::
 
         cls.setup
         while_(cls.should_run_calculation)(
@@ -36,12 +33,11 @@ class BaseRestartWorkChain(WorkChain):
             cls.inspect_calculation,
         )
 
-    Each of these methods can of course be overriden but they should be general enough to fit most
-    calculation cycles. The run_calculation method will take the inputs for the calculation process
-    from the context under the key 'inputs'. The user should therefore make sure that before the
-    run_calculation method is called, that the to be used inputs are stored under self.ctx.inputs.
-    One can update the inputs based on the results from a prior calculation by calling an outline
-    method just before the run_calculation step, for example::
+    Each of these methods can of course be overriden but they should be general enough to fit most calculation cycles.
+    The `run_calculation` method will take the inputs for the calculation process from the context under the key
+    `inputs`. The user should therefore make sure that before the `run_calculation` method is called, that the to be
+    used inputs are stored under `self.ctx.inputs`. One can update the inputs based on the results from a prior
+    calculation by calling an outline method just before the `run_calculation` step, for example::
 
         cls.setup
         while_(cls.should_run_calculation)(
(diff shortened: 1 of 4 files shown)


aiida_quantumespresso/common/workchain/base/restart.py
```diff
@@ -310,13 +267,24 @@ def _handle_calculation_failure(self, calculation):
...
-        This is the format used by input groups as in for example the explicit pseudo dictionary where the key is
-        a tuple of kind to which the UpfData corresponds.
+    def _handle_unexpected_failure(self, calculation, exception=None):
+        """Handle an unexpected failure.
+
+        This occurs when a calculation excepted, was killed or finished with a non-zero exit status but no errors were
+        handled. If this is the second consecutive unexpected failure the work chain is aborted.
+
+        :param calculation: the calculation that failed in an unexpected way
+        :param exception: optional exception message to log to the report
```
borellim: Maybe 'optional exception **or error message** ...' ?

aiida_quantumespresso/common/workchain/base/restart.py
```diff
@@ -310,13 +267,24 @@ def _handle_calculation_failure(self, calculation):
...
+            self.report('{}'.format(exception))
+
+        if self.ctx.unexpected_failure:
+            self.report('failure of {}<{}> could not be handled for the second consecutive time'.format(
+                self.ctx.calc_name, calculation.pk))
+            return self.exit_codes.ERROR_SECOND_CONSECUTIVE_UNHANDLED_FAILURE
+
+        self.ctx.unexpected_failure = True
+        self.report('failure of {}<{}> could not be handled, restarting once more'.format(
+            self.ctx.calc_name, calculation.pk))
```
borellim: I would keep the "else:" clause and the indentation; it makes the code more readable IMO.

aiida_quantumespresso/common/workchain/base/restart.py
```diff
@@ -310,13 +267,24 @@ def _handle_calculation_failure(self, calculation):
...
+        :param calculation: the calculation that failed in an unexpected way
+        :param exception: optional exception message to log to the report
+        :return: `ExitCode` if this is the second consecutive unexpected failure
         """
-        from aiida_quantumespresso.utils.mapping import prepare_process_inputs
-        return prepare_process_inputs(process, inputs)
+        if exception:
+            self.report('{}'.format(exception))
+
+        if self.ctx.unexpected_failure:
```
borellim: I wonder: this flag is never reset, so technically we return `ERROR_SECOND_CONSECUTIVE_UNHANDLED_FAILURE` even if this happens twice non-consecutively. Can this really happen?

aiida_quantumespresso/common/workchain/base/restart.py
```diff
@@ -215,71 +200,43 @@ def on_terminated(self):
...
-            return self.exit_codes.ERROR_SECOND_CONSECUTIVE_UNHANDLED_FAILURE
+    def _handle_calculation_failure(self, calculation):
+        """Call the attached error handlers if any to attempt to correct the cause of the calculation failure.
 
-        else:
-            self.report('failure of {}<{}> could not be handled, restarting once more'
-                .format(self.ctx.calc_name, calculation.pk))
+        The registered error handlers will be called in order based on their priority until a handler returns a report
+        that instructs to break. If the last executed error handler defines an exit code, that will be returned to
+        instruct the work chain abort. Otherwise the work chain will continue the cycle.
```
borellim: 'to instruct the work chain **to** abort'

aiida_quantumespresso/common/workchain/base/restart.py
```diff
@@ -77,136 +74,124 @@ def _load_error_handlers(self):
...
-            message='the maximum number of iterations was exceeded')
-        spec.exit_code(102, 'ERROR_UNEXPECTED_CALCULATION_STATE',
-            message='the calculation finished with an unexpected calculation state')
-        spec.exit_code(103, 'ERROR_SECOND_CONSECUTIVE_SUBMISSION_FAILURE',
-            message='the calculation failed to submit, twice in a row')
-        spec.exit_code(104, 'ERROR_SECOND_CONSECUTIVE_UNHANDLED_FAILURE',
-            message='the calculation failed for an unknown reason, twice in a row')
+            message='The maximum number of iterations was exceeded.')
+        spec.exit_code(102, 'ERROR_SECOND_CONSECUTIVE_UNHANDLED_FAILURE',
+            message='The calculation failed for an unknown reason, twice in a row.')
```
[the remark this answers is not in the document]
sphuber: Absolutely agree. But while we are figuring this out we need to have the liberty to change these. That's also why I am changing most of this in the separate `3dd` branch. Ideally once I have a working situation, I discuss the proposal with other plugin developers to see if they can map the same guidelines on their work chains. Ideally input/output names and exit codes follow similar guidelines. Of course this will take time and combined with the idea that we don't want to change these lightly once released, this will push an official release far in the future. Not sure what the best balance will be here

aiida_quantumespresso/common/workchain/base/restart.py
sphuber: This is actually against common style guides, which is also the reason that pylint will complain about such an unnecessary else-clause

aiida_quantumespresso/common/workchain/base/restart.py
sphuber: Thanks, will correct it

aiida_quantumespresso/common/workchain/base/restart.py
sphuber: Will adapt

aiida_quantumespresso/common/workchain/base/restart.py
sphuber: Nice catch! This is indeed a bug. I have added the fix by resetting the flag in the handle failure method as soon as a handler returns that it handled an error

borellim: Nice, but shouldn't we also reset it if `calculation.is_finished_ok`? I would add it as first instruction here: https://github.com/aiidateam/aiida-quantumespresso/pull/325/files#diff-a075396b80333fe72d56cb4808da80caR137

sphuber: Yes, you are right, but we have to keep the other one as well. If we first have an unexpected failure, but then the second calculation has a failure but *it is* handled, then that should also reset the "unexpected failure" counter.

Originally, this was the only place where it should be reset, but since I changed the `is_finished_ok` branch to include the sanity checks, that can fail, which will cause another calculation to be run (whereas normally the work chain would terminate). In that case you are right, we should reset the flag, thanks for pointing that out! I will make the change

borellim: I see, yes. Thank you.

aiida_quantumespresso/common/workchain/base/restart.py
borellim: This guideline is not in PEP8, and its applicability [depends on context](https://dmerej.info/blog/post/else-after-return-yea-or-nay/). In this case it doesn't really matter (I can see your point of view: you could say that there are 2 precondition checks and a "main" code), but I would not let pylint's opinion about this override readability.

PR #326, opened 2019-06-07
`PwBaseWorkChain`: document spec and fix automatic parallelization

### LINKED ISSUE #264: run_init() in PwBaseWorkChain does not work in aiida 1.0

### LINKED ISSUE #306: Automatic parallelization of PW is broken

sphuber:
Fixes #264 and fixes #306 

The spec did not specify any help messages yet for the inputs and the
outputs which have now been added. Additionally, the automatic
parallelization functionality has been fixed. There were a few calls to
methods that have been removed in `aiida-core==1.0.0`. Note that this
functionality requires an initialization run, which is performed by
running `pw.x` with an exit file in the directory. This will cause the
code to perform the preamble and then exit. This functionality seems to
be broken for versions of `pw.x` that produce the XML output with the
new schema format. As a result the automatic parallelization only works
for codes that produce the XML in the old format.

### DIFF
--- a/aiida_quantumespresso/workflows/pw/base.py
+++ b/aiida_quantumespresso/workflows/pw/base.py
@@ -10,7 +10,7 @@ from aiida_quantumespresso.common.workchain.utils import ErrorHandlerReport
 from aiida_quantumespresso.common.workchain.utils import register_error_handler
 from aiida_quantumespresso.common.workchain.base.restart import BaseRestartWorkChain
 from aiida_quantumespresso.utils.defaults.calculation import pw as qe_defaults
-from aiida_quantumespresso.utils.mapping import update_mapping
+from aiida_quantumespresso.utils.mapping import update_mapping, prepare_process_inputs
 from aiida_quantumespresso.utils.pseudopotential import validate_and_prepare_pseudos_inputs
 from aiida_quantumespresso.utils.resources import get_default_options
 from aiida_quantumespresso.utils.resources import get_pw_parallelization_parameters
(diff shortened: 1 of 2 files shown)


aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -38,20 +38,41 @@ class PwBaseWorkChain(BaseRestartWorkChain):
...
-        spec.input('options', valid_type=orm.Dict, required=False)
-        spec.input('automatic_parallelization', valid_type=orm.Dict, required=False)
+        spec.input('code', valid_type=orm.Code,
+            help='The code to use for the calculation which is configured to a Quantum ESPRESSO `pw.x` executable.')
+        spec.input('structure', valid_type=orm.StructureData,
+            help='The input structure.')
+        spec.input('kpoints', valid_type=orm.KpointsData, required=False,
+            help='An explicit k-points list or mesh. Either this or `kpoints_distance` has to be provided.')
+        spec.input('kpoints_distance', valid_type=orm.Float, required=False,
+            help='The desired distance in 1/Å between k-points in reciprocal space. The explicit k-points will be'
```
borellim: Is this the minimum or maximum distance betweek k-points? I always forget...

borellim: Is this the minimum or maximum distance betweek k-points? I always forget...

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -162,17 +194,26 @@ def validate_inputs(self):
             self.report('{}'.format(exception))
             return self.exit_codes.ERROR_INVALID_INPUT_PSEUDO_POTENTIALS
 
-    def should_run_init(self):
+    def set_max_seconds(self, max_wallclock_seconds):
```
borellim: This method is almost identical to method `_prepare_process_inputs`: can we merge them?

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -139,6 +169,8 @@ def validate_inputs(self):
             if num_machines is None or max_wallclock_seconds is None:
                 return self.exit_codes.ERROR_INVALID_INPUT_RESOURCES_UNDERSPECIFIED
 
+            self.set_max_seconds(max_wallclock_seconds)
```
borellim: Why do this only if automatic_parallelization is enabled?

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -70,8 +70,8 @@ def define(cls, spec):
         spec.input('options', valid_type=orm.Dict, required=False,
             help='The metadata options that are to be passed to the calculation job.')
         spec.input('automatic_parallelization', valid_type=orm.Dict, required=False,
-            help='When specified, the work chain will first launch an initialization calculation to determine the '
-                 'dimensions of the problem and based on this will try to set optimal parallelization flags.')
+            help='When True, the work chain will first launch an initialization calculation to determine the '
```
sphuber: > When True

This is not correct, because the node type is not a `Bool` but a `Dict`. So the switching will happen if a value is passed at or not, not its content. I will fix this and then amend your changes in my original commit. Since you added a commit, I think that if I were to squash merge, the resulting commit would be attributed to you instead of me. In the future then I think it is better to just request these changes and the PR owner will have to apply them

borellim: Ah, sorry, my bad.
OK, just wanted to save you some time.

sphuber: No worries, appreciate the gesture in any case.

aiida_quantumespresso/workflows/pw/base.py
sphuber: I just forgot to remove `_prepare_process_inputs` as it is no longer used

aiida_quantumespresso/workflows/pw/base.py
sphuber: minimum, I added it to the message

aiida_quantumespresso/workflows/pw/base.py
sphuber: No this is done only when `automatic_parallelization` is *not* enabled. The reason is because in the case of the init run, we use the default options which sets half an hour, which should be more than enough for the init run.

sphuber:
@borellim I removed the redundant code and addressed the other comments. Should be good to go

sphuber:
Thanks for the review @borellim

PR #331, opened 2019-06-11
`BaseRestartWorkChain`: check for existence of `_error_handlers`

### LINKED ISSUE #330: `BaseRestartWorkChain` fails if handling a failure and no handlers are registered
This is because the `WorkChain` attribute `_error_handlers`, that contains a list of registered error handlers, is only created when the first handler is registered. However, the `_handle_calculation_failure` method assumes the attribute is there, resulting in an `AttributeError`

sphuber:
Fixes #330 

The `WorkChain._error_handlers` attribute is only initialized when a
handler gets registered through the `register_error_handler` decorator.
This is because it cannot be done on the `BaseRestartWorkChain` base
class since then handlers will be attached to all sub classes and not
just the one to which the decorator is applied. However, this means that
if *no* handlers are registered, the class will have no `_error_handlers`
attribute. The `_handle_calculation_failure` method, however, was
assuming its existance, resulting in an `AttributeError`. The fix is to
simply explicitly check for the existence of the attribute first.

### DIFF
--- a/aiida_quantumespresso/common/workchain/base/restart.py
+++ b/aiida_quantumespresso/common/workchain/base/restart.py
@@ -243,11 +243,11 @@ class BaseRestartWorkChain(WorkChain):
         is_handled = False
         handler_report = None
 
+        if not hasattr(self, '_error_handlers') or not self._error_handlers:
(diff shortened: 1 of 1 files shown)


sphuber:
Thanks for the reviews @borellim

PR #335, opened 2019-06-12
Non-zero exit code if a critical warning is encountered

chrisjsewell:
At present a 0 exit code is returned, even when `parse_raw_output` returns `job_successful=False`, which occurs for example when a critical message is encountered like `convergence NOT achieved`.

This feels wrong, and I think we can take advantage of the new exit code functionality to highlight these computation failures.


sphuber:
Thanks @chrisjsewell for the PR, it certainly is wrong that the parsers don't yet use the exit codes, but that's because I haven't had the time to migrate it yet. However, for the `pw.x` stack, I just started this work last week [in this branch](https://github.com/aiidateam/aiida-quantumespresso/tree/3dd). There are a lot more problems than the one you highlight here, however, so I am thoroughly overhauling the whole `PwParser` and the raw parsers. So if you don't mind I will close this PR for now

PR #336, opened 2019-06-12
Allow for optional XML parsing

chrisjsewell:
This is mainly for immigrating existing outputs, where the XML file is not available.
Mostly all the output data can still be extracted from the text output, except the input structure is required to generate the final structure (for `relax` calculations).

As you can see, in the `PwCalculation.define`, I have added an option `metadata.options.require_xml`. By default, this is True so, unless specifically set to False, this functionality will not be called on.

### DIFF
--- a/aiida_quantumespresso/calculations/pw.py
+++ b/aiida_quantumespresso/calculations/pw.py
@@ -66,7 +66,8 @@ class PwCalculation(BasePwCpInputGenerator):
         spec.input('metadata.options.input_filename', valid_type=six.string_types, default=cls._DEFAULT_INPUT_FILE)
         spec.input('metadata.options.output_filename', valid_type=six.string_types, default=cls._DEFAULT_OUTPUT_FILE)
         spec.input('metadata.options.parser_name', valid_type=six.string_types, default='quantumespresso.pw')
-        spec.input('metadata.options.require_xml', valid_type=bool, default=True)
+        spec.input('metadata.options.require_xml', valid_type=bool, default=True,
+            help='if True, output parsing will fail if an XML file cannot be found')
         spec.input('kpoints', valid_type=orm.KpointsData,
             help='kpoint mesh or kpoint path')
         spec.input('hubbard_file', valid_type=orm.SinglefileData, required=False,
--- a/aiida_quantumespresso/calculations/pw.py
+++ b/aiida_quantumespresso/calculations/pw.py
@@ -66,6 +66,7 @@ class PwCalculation(BasePwCpInputGenerator):
         spec.input('metadata.options.input_filename', valid_type=six.string_types, default=cls._DEFAULT_INPUT_FILE)
         spec.input('metadata.options.output_filename', valid_type=six.string_types, default=cls._DEFAULT_OUTPUT_FILE)
         spec.input('metadata.options.parser_name', valid_type=six.string_types, default='quantumespresso.pw')
+        spec.input('metadata.options.require_xml', valid_type=bool, default=True)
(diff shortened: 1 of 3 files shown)


sphuber:
As mentioned in the other PR, I am completely overhauling the `PwParser` including the raw parsers. I will keep this open since this contains some useful ideas that I haven't implemented yet, but I propose that we wait a bit until I have fleshed out my development in the other branch. As a side note, one thing that I am doing is to make the "raw" parsers AiiDA-agnostic. This makes it easier to unit tests them in an isolated way or just re-use it for other purposes. So adding the optional dependency on an AiiDA input structure here is not ideal.

sphuber:
Now that `3dd` is merged, it might be worth to revisit this. I take it you implemented this because you need it to immigrate completed jobs that do not have XML output? The problem I see with the current solution is that it injects AiiDA dependencies into the raw parser. I just refactored the raw stdout parser to be completely AiiDA agnostic and we should keep it that way. If it is possible to refactor your PR than it might be useful. The one thing I wanted to still change in the current parser, that I didn't get the chance to do in `3dd` is to also make the raw stdout parser independent of the XML parser. For historical reasons they made it dependent, or at least consume data parsed from the XML, which is then complemented by data parsed from the XML. I want to get rid of this and just have two independent parsers, one for the XML and one for stdout. This would naturally "solve" your problem as well, so maybe it makes more sense to tackle that problem first and then just make the XML parsing optional

chrisjsewell:
> I take it you implemented this because you need it to immigrate completed jobs that do not have XML output?

yep

> The one thing I wanted to still change in the current parser, that I didn't get the chance to do in `3dd` is to also make the raw stdout parser independent of the XML parser.

So my solution with `3dd` has been to use `immigrate_existing` from #350 (which I've just rebased).
The stdout parsing errors are returned before XML ones: https://github.com/aiidateam/aiida-quantumespresso/blob/dc43ac6a797087013cfe9afb8b21e3a927111cd2/aiida_quantumespresso/parsers/pw.py#L108

So I just replace `ERROR_OUTPUT_XML_MISSING` with a zero exit code. The dependencies from the XML are then the output structure and the bands. For the output structure, I just use the last structure from the trajectory data.

This worked about a month ago, I don't think there has been any changes that would invalidate it:

```python
    calc_node = immigrate_existing(builder, remote, seal_node=False)

    # the runs are missing the xml file
    pw_calc_cls = CalculationFactory("quantumespresso.pw")
    missing_xml = pw_calc_cls.exit_codes.ERROR_OUTPUT_XML_MISSING.status
    if calc_node.exit_status == missing_xml:
        calc_node.set_exit_status(0)

    # the output_structure node will not have been created but,
    # if output_trajectory is present, we can create it from there
    if ("output_structure" not in calc_node.outputs
            and "output_trajectory" in calc_node.outputs):
        structure = calc_node.outputs.output_trajectory.get_step_structure(
            -1, custom_kinds=calc_node.inputs.structure.kinds)
        structure.add_incoming(calc_node, link_type=LinkType.CREATE,
                               link_label='output_structure')
        structure.store()

    calc_node.seal()
```

chrisjsewell:
So yeh, could just replace https://github.com/aiidateam/aiida-quantumespresso/blob/dc43ac6a797087013cfe9afb8b21e3a927111cd2/aiida_quantumespresso/parsers/pw.py#L111

with:

```python
 if self.exit_code_xml and self.node.get_option('require_xml'):
```

Then the structure would be easy to add to the stdout parser, but not sure about the band information.

sphuber:
Superseded by #488

PR #348, opened 2019-06-26
Ensure the output structure has the same kinds as the input structure for PwParser

chrisjsewell:
The fix with test :)


sphuber:
Thanks @chrisjsewell

PR #350, opened 2019-06-27
Immigrate Existing CalcJobs

chrisjsewell:
@sphuber I've improved this, to make it completely generic to any `CalcJob`,
so actually it could be moved to `aiida-core`?

**TODO** update pwimmigrant_tutorial.rst


sphuber:
I am going to close this as I think we should go with the generic solution that will be imlemented in `aiida-core`.

PR #352, opened 2019-07-03
fix bandstructure workchain

unkcpz:
`aiida-quantumespresso/workflow/pw/base.py` expose inputs of `PwCalculation` by namespace 'pw'.
Which cause `PwBandStructureWorkChain` fail at running.

### DIFF
--- a/aiida_quantumespresso/workflows/pw/relax.py
+++ b/aiida_quantumespresso/workflows/pw/relax.py
@@ -184,7 +184,7 @@ class PwRelaxWorkChain(WorkChain):
         # Get the latest workchain, which is either the workchain_scf if it ran or otherwise the last regular workchain
         try:
             workchain = self.ctx.workchain_scf
-            structure = workchain.inputs.pw.structure
+            structure = workchain.inputs.pw__structure
         except AttributeError:
             workchain = self.ctx.workchains[-1]
             structure = workchain.outputs.output_structure
--- a/aiida_quantumespresso/workflows/pw/band_structure.py
+++ b/aiida_quantumespresso/workflows/pw/band_structure.py
@@ -146,24 +146,26 @@ class PwBandStructureWorkChain(WorkChain):
             }
 
             if 'scf_options' in self.inputs:
-                res['options'] = self.inputs.scf_options
+                res['metadata'] = {'options': self.inputs.scf_options.attributes}
 
             return res
 
         relax_inputs = {
-            'base': get_common_inputs(),
+            'base': {
+                'pw': get_common_inputs()
+            },
             'relaxation_scheme': orm.Str('vc-relax'),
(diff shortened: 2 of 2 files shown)


unkcpz:
I didn't know it is good to make these kinds of PR. 
I installed develop branch of this plugin, and have my own Workflows which nest `PwBaseWorkChain` and `PwBandStructureWorkChain`. However, when I update the plugin to newest version I have to spend much time to repair my Workflows to work properly. 
I want to know is there any good way to organize my workflows? (Where to put my own workflows? in separated repository or in the plugin's workflow folder? Whether it is good to write my workflow based on the existing workflow, or write every thing from scratch?)

Maybe it is better to ask the question in aiidauser google group?

sphuber:
Thanks @unkcpz . In principle PRs like this are fine. We are in a bit of a transition phase, adapting the code to `aiida-core==1.0.0` and I am taking the time to really improve the code as well, with some backwards incompatible changes as well. Since the calculation plugins and workchains are very much in flux, I don't bother to keep everything working a 100% on `develop` for each commit. Given that there are no tests for the work chains, errors like the one you found here will be common. Normally I would definitely encourage to re-use the workchains in this repository as opposed to redoing everything yourself. I am writing these base workflows exactly to be easily reusable by higher-level workflows. I will leave this PR open for now because I will be merging the `3dd` branch soon and that will change a lot again. I will be addressing the `PwBandsWorkChain` in the near future

aiida_quantumespresso/workflows/pw/relax.py
```diff
@@ -184,7 +184,7 @@ def results(self):
         # Get the latest workchain, which is either the workchain_scf if it ran or otherwise the last regular workchain
         try:
             workchain = self.ctx.workchain_scf
-            structure = workchain.inputs.pw.structure
+            structure = workchain.inputs.pw__structure
```
unkcpz: Inputs from nested workchain are access with double underscore `__` now? 
I can not find information in documentation. 
This will be the new way to access the inputs from nested workchain?

sphuber: Well, the link label for nested namespaces is generated by replacing each nesting with double underscore. Since a double underscore is valid in a python identifier, you can use it like this, yes

borellim: @sphuber, can you confirm that this is the *only* way to access a nested namespace now? It seems to be the case (at least for process inputs).

sphuber: through the `inputs` property, yes, that is the only way. You could call `self.node.get_incoming().nested()` which will return a nested dictionary with the reconstructed original nesting

sphuber:
Thanks for the PR @unkcpz . I had to address some other problems as well which I did in PR #363 so I am closing this. But thanks nonetheless!

Issue #225, opened 2019-01-29
Restart breaks if a calculation is killed by the scheduler after 'Maximum CPU time exceeded' but before finishing saving


borellim:
A PW calculation can be killed by the scheduler after printing 'Maximum CPU time exceeded' to aiida.out, but before it finishes to save the current state. In this case, the PwBaseWorkChain handles the error with `_handle_error_exceeded_maximum_walltime`, in which it sets a `restart_calc`, but the next restart calculation will fail as some data are missing.

This has happened with a very short total time (100s (scheduler) / 95s (PW input)).

We should try to check whether the output data of the previous calculation are correct (as far as reasonable), and if not, make a `from_scratch` calculation instead (not a restart) with a lower `delta_factor_max_seconds`.

Workarounds are to either set a larger max walltime (which will proportionally allow for a larger margin for saving) or change `delta_factor_max_seconds` to a smaller value here: https://github.com/aiidateam/aiida-quantumespresso/blob/79007b44b229170e9c12c4c679aa03e46e5499c6/aiida_quantumespresso/workflows/pw/base.py#L37-L43

See also the Google Groups discussion: https://groups.google.com/d/topic/aiidausers/dzzzDfo6qQ4/discussion

sphuber:
Addressed in PR #360

PR #360, opened 2019-07-31
Improvements to the `pw.x` stack

### LINKED ISSUE #6: Implement robust error codes for parser warnings
Currently, the parser writes warnings as strings to a dictionary in the `parser_warnings` key of the `ParameterData` node. Workflows have to match these strings in order to determine the fate of the calculation and what needs to be done. This is a fragile method and it would be much better if the error messages could be replaced with well defined error codes, potentially with the strings as error details. Error handling can then be done at the hand of well defined error codes.

### LINKED ISSUE #13: VC-relaxation restarts
VC-Relaxations now are not handled properly, it seems. The old Workflows restart from the output structure also of a failed calculation, the new workflow system restarts from the calculation without updating the structure. Without any parameters changing, this just makes the calculation crash again and the workflow not producing any output.

### LINKED ISSUE #223: Refactor pw parsing
`parse_with_retrieved` (pw.py) should call a function (which may still be named `parse_raw_output`), that calls 2 separate and independent functions: one parses the XML output (either old or new format), the other parses the raw text output. These two functions shouldn't talk to each other, and should just parse and return everything they can. `parse_raw_output` then will combine the two, discard duplicates and deprecated keys, check that everything required is present, and return dictionaries to `parse_with_retrieved` so that it can create output nodes.

The goal is to make individual parts of the parser more testable, and the overall parsing logic easier to follow.

sphuber:
Fixes #6 
Fixes #13 
Fixes #223 
Fixex #225


aiida_quantumespresso/utils/validation/trajectory.py
```diff
@@ -0,0 +1,77 @@
...
+
+    :param trajectory: the `TrajectoryData`
+    :param index: the frame index of the trajectory data to check, default is `-1` meaning the last frame
+    :param threshold: the stress threshold in kbar
+    :return: `None` if the `stress` array or given index does not exist, `True` if threshold is valid, `False` otherwise
+    """
+    if threshold is None:
+        return None
+
+    threshold /= 10  # Convert to GPa
```
giovannipizzi: You'll need a `from __future__ import division here to be safe (and actually, to be sure, put `10.` in case threshold happens to be an integer (I know that for the specific case it does not make much sense)

aiida_quantumespresso/utils/validation/trajectory.py
```diff
@@ -0,0 +1,77 @@
...
+
+    try:
+        stress = trajectory.get_array('stress')[index]
+    except (KeyError, IndexError):
+        return None
+
+    return (numpy.trace(stress) / 3) < threshold
```
giovannipizzi: `3` -> `3.`

aiida_quantumespresso/utils/validation/trajectory.py
sphuber: Addressed it now to be sure in 1913aa3

aiida_quantumespresso/utils/validation/trajectory.py
sphuber: Addressed it now to be sure in 1913aa3

PR #364, opened 2019-08-02
Add `qe-tools` dependency

### LINKED ISSUE #215: Remove old dependencies from quantum espresso input parser tools, move them to qe-tools package
The idea is to separate aiida-independent functionalities into a separate package "qe-tools".

sphuber:
Fixes #215 

This allows to remove the parsers in `aiida_quantumespresso.tools` that
have been moved to the `qe-tools` package, which is an `aiida-core`
independent package.


setup.json
```diff
@@ -74,6 +74,7 @@
     "install_requires": [
         "aiida_core[atomic_tools]>=1.0.0b5",
         "matplotlib<3.0.0; python_version<'3'",
+        "qe-tools==1.1.2",
```
borellim: So this commit does not depend on 1.1.3 yet?

sphuber: It should have, even though I haven't been able to release the package on PyPI since I don't have the credentials. Anyway, the PR I pushed was wrong. I have corrected it now. Just have to wait for `qe-tools==1.1.3` to be pushed.

sphuber:
@borellim this should be good for review now. I am having Elsa try it out for her use case to ensure it still works

sphuber:
@borellim I just confirmed with @asle85 that with these changes here use case still works, so this is good to go

PR #366, opened 2019-08-04
ensure the ERROR_INVALID_KPOINT_UNITS exit code is captured

chrisjsewell:
I noticed this, because it was highlighted by pylint (unused-variable). Then I noticed that you're currently excluding most of the files from the pylint pre-commit. I'm guessing that's on the todo list :)


chrisjsewell:
tests pass, but one has errored at install

sphuber:
> currently excluding most of the files from the pylint pre-commit

Yup, because all those files are a mess and fixing them would take a lot of time. I have been slowly uncommenting them as I find the time/courage to address the issues

PR #370, opened 2019-08-07
Turn `BasePwCpInputGenerator._generate_PWCPinputdata` into classmethod

### LINKED ISSUE #369: `BasePwCpInputGenerator._generate_PWCPinputdata` should be a class method
The logic that is contained within this method is essentially creating an QE input file based on some input nodes. This is often useful to be called without having to instantiate a process. But to be able to do that, it should be transformed into a class method.

sphuber:
Fixes #369 

The logic in this method does not depend on any specific instance and it
is in fact useful to be able to call it without having to instantiate
the process. Now the method can be called just to format the input file
for a given set of inputs.

### DIFF
--- a/aiida_quantumespresso/calculations/__init__.py
+++ b/aiida_quantumespresso/calculations/__init__.py
@@ -247,12 +247,9 @@ class BasePwCpInputGenerator(CalcJob):
 
         return calcinfo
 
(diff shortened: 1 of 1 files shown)


sphuber:
@borellim I needed to make this change for @asle85 for the QE input generator tool. If you could give it a quick review so I can merge, thanks

sphuber:
Thanks @borellim

PR #372, opened 2019-08-12
Fix the TCOD export plugins for `PwCalculation` and `PhCalculation`

### LINKED ISSUE #365: Fix the TCOD exporters
The `BaseTcodTranslator` has been moved from `aiida-core` to `aiida-tcod`

sphuber:
Fixes #365 

The base translator class was moved from `aiida-core` to `aiida-tcod`.
Add the latter as dependency, fix the imports and drop the directory
from the pre-commit exclude list.

### DIFF
--- a/aiida_quantumespresso/tools/dbexporters/tcod_plugins/pw.py
+++ b/aiida_quantumespresso/tools/dbexporters/tcod_plugins/pw.py
@@ -1,15 +1,22 @@
 # -*- coding: utf-8 -*-
+"""TCOD export plugin for `PwCalculations`."""
 from __future__ import absolute_import
+
 import re
 
 from aiida.orm.nodes.data.upf import UpfData
-from aiida.tools.dbexporters.tcod_plugins import BaseTcodtranslator
+
+try:
+    from aiida_tcod.tools.dbexporters.tcod import BaseTcodtranslator  # pylint: disable=import-error
+except ImportError:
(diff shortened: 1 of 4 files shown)


aiida_quantumespresso/tools/dbexporters/tcod_plugins/pw.py
```diff
@@ -45,26 +49,26 @@ def get_software_package(cls,calc,**kwargs):
...
         """
         Returns the energy of defined type in eV.
         """
         parameters = calc.outputs.output_parameters
         if energy_type not in parameters.attrs():
             return None
         if energy_type + '_units' not in parameters.attrs():
-            raise ValueError('energy units for {} are '
-                             'unknown'.format(energy_type))
+            raise ValueError('energy units for {} are ' 'unknown'.format(energy_type))
```
borellim: Maybe merge in a single string

borellim:
Maybe a stupid question, but what happens if one does not install aiida-tcod? Does the error instruct to install the [tcod] extras?

aiida_quantumespresso/tools/dbexporters/tcod_plugins/pw.py
sphuber: Done

sphuber:
> Maybe a stupid question, but what happens if one does not install aiida-tcod? Does the error instruct to install the [tcod] extras?
> 
> (PR ok otherwise for me)

Not a stupid question: I added a try except with a more informative error message.

borellim:
Cool, thanks!

PR #374, opened 2019-08-12
Add CLI command `aiida-quantumespresso data structure import`

sphuber:
This command allows one to parse a Quantum ESPRESSO pw.x input file into
a `StructureData` node.


aiida_quantumespresso/cli/data/structure.py
```diff
@@ -0,0 +1,34 @@
...
+
+from . import cmd_data
+
+
+@cmd_data.group('structure')
+def cmd_structure():
+    """Commands to create and inspect `StructureData` nodes."""
+
+
+@cmd_structure.command('import')
```
CasperWA: Just a general question, because I have never added click commands through a plugin before, but does this add the command `verdi data structure import` or does it add another possibility to that (already existing) `verdi` command?

aiida_quantumespresso/cli/data/structure.py
```diff
@@ -0,0 +1,34 @@
...
+@click.argument('filename', type=click.File('r'))
+@options.DRY_RUN()
+@decorators.with_dbenv()
+def cmd_import(filename, dry_run):
+    """Import a `StructureData` from a Quantum ESPRESSO input file."""
+    from aiida_quantumespresso.tools.pwinputparser import PwInputFile
```
CasperWA: My guess is that this parser already encompasses the old `get_structuredata_from_qetools` and `get_kpointsdata_from_qetools`? :)

aiida_quantumespresso/cli/data/structure.py
sphuber: No, this is a CLI that I designed especially for the `aiida-quantumespresso` plugin. It installs the `aiida-quantumespresso` binary that works exactly like `verdi` and also comes with autocompletion. I did it this way because extending `verdi` through entry points only really supports adding new commands for `verdi data`, but I wanted my CLI to be able to do much more, such as launching calculations and work chains.

aiida_quantumespresso/cli/data/structure.py
sphuber: Yes, the basic parsing functionality is now part of `qe-tools` and is fully AiiDA agnostic. In a recent PR in `aiida-quantumespresso` I extend that basic parsing functionality into transforming them into nodes. This simple CLI then simply exposes that through the CLI.

sphuber:
I am pinging @normarivano and @giovannipizzi so they know where the functionality from `verdi` went. I.e. to import a structure from a Quantum ESPRESSO input file, you can now do:
```
aiida-quantumespresso data structure import FILENAME
```

aiida_quantumespresso/cli/data/structure.py
CasperWA: Right - so for this particular thing, it could actually be added to `verdi`, but this is adding it to the `aiida-quantumespresso` CLI - very nice.
Now, however, I'm wondering whether it makes sense to do both, i.e. add it to the plugin CLI, but also to `verdi`?

aiida_quantumespresso/cli/data/structure.py
CasperWA: Very elegant.

aiida_quantumespresso/cli/data/structure.py
sphuber: I think it does not make sense to add it also to `verdi`. If it would be possible to add it to the existing `verdi data structure`, maybe. But it has to be a new subcommand, e.g. `verdi data qe-structure`, which I think is ugly and since there is already `aiida-quantumespresso data structure`, it is best to not add this redundancy

CasperWA: Right, I thought one could add to all sub-commands of `verdi data`. Yeah, then it doesn't make sense.

PR #376, opened 2019-08-13
Add default for structure option in certain CLI launchers

### LINKED ISSUE #375: Add default for structure option in `PwCalculation` CLI launcher
Launching a quick `PwCalculation` from the CLI is often the first thing one wants to do, but a structure is required, which currently the user needs to create or import first. The command should provide a default Silicon structure instead

sphuber:
Fixes #375 

The default is added for the launchers of:

 * `PhCalculation`
 * `PwCalculation`
 * `PwBaseWorkChain`

Since having a default structure automatically created is mostly useful
for demonstration purposes on clean databases, we only add them to the
most basic launchers.

### DIFF
--- a/aiida_quantumespresso/cli/calculations/cp.py
+++ b/aiida_quantumespresso/cli/calculations/cp.py
@@ -5,46 +5,15 @@ from __future__ import absolute_import
 from aiida.cmdline.params import options, types
 from aiida.cmdline.utils import decorators
 
+from ..utils import defaults
 from ..utils import launch
 from ..utils import options as options_qe
 from . import cmd_launch
 
 
-def silicon_structure():
-    """Return a `StructureData` representing bulk silicon.
-
-    The database will first be queried for the existence of the structure in case it might have already been created
-    before. If nothing is found, a new structure will be created
-
-    :return: a `StructureData` representing bulk silicon
-    """
-    from ase.spacegroup import crystal
-    from aiida.orm import QueryBuilder, StructureData
-
-    label = '__example__{name}'.format(name=__name__)
-    structure = QueryBuilder().append(StructureData, filters={'label': label}).first()
-    if not structure:
-        alat = 5.4
-        ase_structure = crystal(
-            'Si',
-            [(0, 0, 0)],
-            spacegroup=227,
-            cellpar=[alat, alat, alat, 90, 90, 90],
-            primitive_cell=True,
-        )
-        structure = StructureData(ase=ase_structure)
-        structure.label = label
-        structure.store()
-    else:
-        # .first() returns none if it doesn't exist or a list of one item if objects exist
-        structure = structure[0]
-
-    return structure.uuid
-
-
 @cmd_launch.command('cp')
 @options.CODE(required=True, type=types.CodeParamType(entry_point='quantumespresso.cp'))
-@options_qe.STRUCTURE(default=silicon_structure)
+@options_qe.STRUCTURE(default=defaults.get_structure)
 @options_qe.PSEUDO_FAMILY(required=True)
 @options_qe.MAX_NUM_MACHINES()
 @options_qe.MAX_WALLCLOCK_SECONDS()
--- a/aiida_quantumespresso/cli/calculations/pw.py
+++ b/aiida_quantumespresso/cli/calculations/pw.py
@@ -7,6 +7,7 @@ import click
 from aiida.cmdline.params import options, types
 from aiida.cmdline.utils import decorators
 
+from ..utils import defaults
 from ..utils import launch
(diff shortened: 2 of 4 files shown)


borellim:
@ sphuber Lol you edited my comment. :)

> If they are not specifying a structure, why would they be surprised that it is not using the one they intended? If you don't specify anything, it should be clear that it is using a default structure no?

Having a default input structure is certainly useful, but as a new user I would not *expect* it. It can be surprising in my opinion.
It's your choice though: if you disagree or you think it's not worth the effort, I'll approve the PR after you fix the commit message.

sphuber:
> @sphuber Lol you edited my comment. :)
Hahah, oops, I was wondering why it was behaving so weird. I must have miss clicked.

> Having a default input structure is certainly useful, but as a new user I would not _expect_ it. It can be surprising in my opinion.
My point is that there are way more inputs that get defaults than just the structure, such that as few options as possible have to be specified. For example the k-points and parameters and resource options etc. I am looking at this command from the perspective of its intended purpose and audience: people who just installed it and just want to run a calculation as quickly as possible. The code and pseudo family (hopefully I can automate this at some point with `aiida-sssp`) still require explicit definition, but the rest is arbitrary. Why would you as a new user not expect a default structure to be used if you do not specify anything? Put in another way: if you _would_ be surprised that a default structure would be used, than that means you understand that a structure data object _needs_ to be used and you probably already have a specific one in mind and you will try to pass it as an input.

Long story short, I don't think it is necessary. People who want to use a specific structure will find the option to pass it and for those who don't, they clearly don't care about what structure to use, so just using a default one is fine, just like the arbitrary default k-points and parameters.


Finally, regarding the commit message, I will adapt it when I squash merge, so I don't have to rerun the tests

borellim:
Thanks for your reply Seb.

> Put in another way: if you _would_ be surprised that a default structure would be used, than that means you understand that a structure data object _needs_ to be used and you probably already have a specific one in mind and you will try to pass it as an input.

Absolutely... unless I simply forgot, in which case it would take me from one second to several minutes to realize that I forgot it. A simple message may be enough to realize it.

In terms of implementation, it seems trivial (literally two lines):
```diff
--- a/aiida_quantumespresso/cli/utils/defaults.py
+++ b/aiida_quantumespresso/cli/utils/defaults.py
@@ -13,6 +13,7 @@ def get_structure():
 
     :return: a `StructureData` representing bulk silicon
     """
+    import click
     from ase.spacegroup import crystal
     from aiida.orm import QueryBuilder, StructureData
 
@@ -44,4 +45,5 @@ def get_structure():
     else:
         structure = results[0]
 
+    click.echo("Using sample {} input structure ({})".format(structure.get_formula(), structure.uuid))
     return structure.uuid
```
 
And if works perfectly from a simple test! However, I can't find any guarantees in click that this function will only be called when the default is actually used. So, to avoid any possible confusion, I agree to let this go, albeit with a sad heart. :)

> Finally, regarding the commit message, I will adapt it when I squash merge, so I don't have to rerun the tests

Ok cool.

sphuber:
It is not so much a question of implementation difficulty, it is that I don't think that notifying the user of this default being used should be implemented in this function. The function might be used elsewhere where the message does not make sense. I would have been fine with placing the message in the actual CLI command, but then detecting if the given structure was explicitly defined or came from the default is not trivial.

Anyway, I remain unconvinced that it is possible for someone to _forget_ using a specific structure for a test calculation if they actually do mean to use a specific one, and why the structure would need special treatment over the other options with default values.

Please don't take my stubbornness the wrong way, I do appreciate the review and comments :) and if in the future we get a question from a confused user, I will eat crow and apologize

borellim:
No worries, no offence taken at all. I also appreciate your thorough replies. :)

d-tomerini:
If I can add to the discussion, @sphuber , I have to say I agree with @borellim. 

I can imagine trying to run some bash script something like this:
  
     structurelist = #an array of structure node I have in the database   
     for s in structurelist;
       do
          aiida-quantumespresso calculation launch pw -X 1 -p efficiency -d 
      done 

and you end up submitting a bunch of silicon calculations just because you forgot to add the ``s``. I understand that is not the supposed to use the CLI, but I can see myself trying it!
I'd be more inclined to provide a mini-script to create the structure in the CLI section of the documentation, as we do for the pw. Or running a ``aiida-quantumespresso data structure import`` command on some example input, that also showcases the command

I'd rather use a flag like --use-default-structure, or similar, and fail on no structure provided, as a failsafe.

If this stays: if we're going to provide a ready-to-go example for an empty database, do you think it would be worthwhile to provide a small Si pseudo as well? Or would it be too big?

PR #377, opened 2019-08-13
Modernize ph stack

### LINKED ISSUE #371: Improve the ph.x stack
Modernize the `PhCalculation` and `PhParser` classes to use exit codes and then update the `PhBaseWorkChain`.

sphuber:
Fixes #371 

Will need input from @tsohier on how unconverged calculations can be detected by the parser.

Edit: consulted with tsohier and the base case of convergence not being reached is implemented correctly including a parser test. This is now good to go


aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -22,152 +21,126 @@ class PhBaseWorkChain(BaseRestartWorkChain):
...
+    """Calculations with an exit status below 400 are unrecoverable, so abort the work chain."""
+    if calculation.exit_status < 400:
+        self.report_error_handled(calculation, 'unrecoverable error, aborting...')
+        return ErrorHandlerReport(True, True, self.exit_codes.ERROR_UNRECOVERABLE_FAILURE)
+
+
+@register_error_handler(PhBaseWorkChain, 580)
+def _handle_out_of_walltime(self, calculation):
+    """In the case of `ERROR_OUT_OF_WALLTIME` calculation shut down neatly and we can simply restart."""
+    if calculation.exit_status == PhCalculation.spec().exit_codes.ERROR_OUT_OF_WALLTIME.status:
```
borellim: I believe you can just say `PhCalculation.exit_codes` if you want (without `spec()`).

aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -22,152 +21,126 @@ class PhBaseWorkChain(BaseRestartWorkChain):
...
-        settings = self.ctx.inputs.settings
 
-        factor = self.defaults.delta_factor_max_seconds
-        max_seconds = settings.get('max_seconds', inputs['INPUTPH']['max_seconds'])
-        max_seconds_reduced = int(max_seconds * factor)
-        self.ctx.inputs.parameters['INPUTPH']['max_seconds'] = max_seconds_reduced
+@register_error_handler(PhBaseWorkChain, 410)
+def _handle_convergence_not_achieved(self, calculation):
+    """In the case of `ERROR_CONVERGENCE_NOT_REACHED` decrease the mixing beta and restart from scratch."""
+    if calculation.exit_status == PwCalculation.spec().exit_codes.ERROR_CONVERGENCE_NOT_REACHED.status:
```
borellim: Idem.

aiida_quantumespresso/parsers/ph.py
```diff
@@ -16,44 +18,85 @@ class PhParser(Parser):
...
-            return self.exit_codes.ERROR_READING_OUTPUT_FILE
+        if filename_stdout not in self.retrieved.list_object_names():
+            return self.exit_codes.ERROR_OUTPUT_STDOUT_MISSING
 
-        xml_tensor_file = None
-        if filename_tensor in list_of_files:
-            xml_tensor_file = output_folder._repository._get_base_folder().get_abs_path(filename_tensor)
+        try:
+            stdout = self.retrieved.get_object_content(filename_stdout)
+        except IOError:
```
borellim: Should we also catch `OSError`?
https://stackoverflow.com/questions/29347790/difference-between-ioerror-and-oserror

aiida_quantumespresso/parsers/ph.py
```diff
@@ -16,44 +18,85 @@ class PhParser(Parser):
...
+            return self.exit_codes.ERROR_OUTPUT_STDOUT_READ
+
+        try:
+            tensor_file = self.retrieved.get_object_content(filename_tensor)
+        except IOError:
```
borellim: Idem

aiida_quantumespresso/parsers/ph.py
```diff
@@ -16,44 +18,85 @@ class PhParser(Parser):
...
+        self.out('output_parameters', orm.Dict(dict=parsed_data))
+
+        if 'ERROR_OUT_OF_WALLTIME' in logs['error']:
+            return self.exit_codes.ERROR_OUT_OF_WALLTIME
+
+        if 'ERROR_CONVERGENCE_NOT_REACHED' in logs['error']:
+            return self.exit_codes.ERROR_CONVERGENCE_NOT_REACHED
+
+    def emit_logs(self, *args):
+        """Emit the messages in one or multiple "log dictionaries" through the logger of the parser.
```
borellim: Since this method is the same in `PwParser` except for the ignored messages, does it make sense to factor it out as a utility function? We could keep these methods, but they would just pass the `ignore` list to the external function as an additional argument. If you want to do it later, that's fine.

aiida_quantumespresso/parsers/ph.py
```diff
@@ -16,44 +18,85 @@ class PhParser(Parser):
...
-            file.startswith(os.path.split(PhCalculation._OUTPUT_DYNAMICAL_MATRIX_PREFIX)[1]) and
-            not file.endswith('.freq')
-        ]
-
-        # Sort according to the number at the end of the dynamical matrix files, when there is one
+        dynmat_files = []
+        dynmat_folder = PhCalculation._FOLDER_DYNAMICAL_MATRIX
+        dynmat_prefix = os.path.split(PhCalculation._OUTPUT_DYNAMICAL_MATRIX_PREFIX)[1]
+
+        for filename in sorted(self.retrieved.list_object_names(dynmat_folder)):
```
borellim: Not sure how important this is, but if the filenames have a numerical suffix, this is not equivalent to the sorting that we did previously:
```
# New:
In [22]: sorted(['a-1','a-2','a-11'])
Out[22]: ['a-1', 'a-11', 'a-2']

# Old:
In [23]: sorted(['a-1','a-2','a-11'], key=lambda x: int(x.split('-')[-1]))
Out[23]: ['a-1', 'a-2', 'a-11']
```

aiida_quantumespresso/parsers/parse_raw/ph.py
```diff
@@ -191,29 +145,48 @@ def parse_xml_matrices(tagname,target_tags):
...
+    :return: dictionary with parsed values
     """
-    from aiida_quantumespresso.parsers.parse_raw.pw import parse_QE_errors
+    def detect_important_message(logs, line):
+
+        message_map = {
+            'error': {
+                'Maximum CPU time exceeded': 'ERROR_OUT_OF_WALLTIME',
+                'No convergence has been achieved': 'ERROR_CONVERGENCE_NOT_REACHED',
+            },
```
borellim: No `'%%%%%%%%%%%%%%'`?

aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -22,152 +21,128 @@ class PhBaseWorkChain(BaseRestartWorkChain):
...
+def _handle_out_of_walltime(self, calculation):
+    """In the case of `ERROR_OUT_OF_WALLTIME` calculation shut down neatly and we can simply restart."""
+    if calculation.exit_status == PhCalculation.spec().exit_codes.ERROR_OUT_OF_WALLTIME.status:
         self.ctx.restart_calc = calculation
-        self.report(
-            'PhCalculation<{}> terminated without reaching convergence, '
-            'setting alpha_mix to {} and restarting'.format(calculation.pk, alpha_mix_new)
-        )
-        return ErrorHandlerReport(True, True)
+        self.report_error_handled(calculation, 'simply restart from the last calculation')
```
borellim: Extra new line.

aiida_quantumespresso/workflows/ph/base.py
sphuber: correct, will adapt

aiida_quantumespresso/parsers/ph.py
sphuber: Agreed

aiida_quantumespresso/parsers/ph.py
sphuber: I kept it separate on purpose, but if it becomes useful for more than these two, then yeah we should refactor

borellim: Ok.

aiida_quantumespresso/parsers/parse_raw/ph.py
sphuber: I excluded this on purpose, just like for `PwParser`. I prefer to only add specific error messages, because only then can we do something useful with it.

aiida_quantumespresso/parsers/ph.py
sphuber: Very good catch. This is indeed important because the parser that consumes it depends on the order. This is bad design in itself and should really be changed, but for now I will just fix the sorting here

sphuber:
Thanks a lot for the review @borellim . I addressed the problems. Unfortunately I had to rebase because I do not want to squash merge.

aiida_quantumespresso/parsers/parse_raw/ph.py
borellim: Makes sense. But don't we risk ignoring possibly important messages? Don't you want to log them as warnings? Or are they too much of a pain to parse?

aiida_quantumespresso/parsers/ph.py
borellim: Ah, the tests now fail because we do the filtering after the sorting, and also we don't catch any exceptions from the lambda...

aiida_quantumespresso/parsers/parse_raw/ph.py
sphuber: They are more difficult to parse, because there could be anything in between. That is of course "solvable" by getting everything between the markers. The main argument I find though is that these are errors that are critical, i.e. they will stop the calculation immediately. They are not warnings as such.

borellim: Ok, so you're relying on 'JOB DONE' not being present in this case, which triggers 'ERROR_OUTPUT_STDOUT_INCOMPLETE'. Makes sense. Thanks.

PR #378, opened 2019-08-14
Fix and simplify the calculation restart builder utility

### LINKED ISSUE #136: Replace the old-syle create_restart methods with utility functions returning builders
Add these builders and deprecate the old methods:
- [x] builder for NEB
- [x] builder for PW + CP
- [x] builder for PH
- [x] Use the methods implemented in https://github.com/aiidateam/aiida_core/pull/1961 and https://github.com/aiidateam/aiida_core/pull/1962, e.g. here: https://github.com/aiidateam/aiida-quantumespresso/blob/fbac0993bb8b6cdeba85717453debcf0ab062b5a/aiida_quantumespresso/utils/restart.py#L100-L108

### LINKED ISSUE #146: Add tests for the builders
Add tests to check that the builders (introduced in #136) work:
- [x] NEB
- [x] PW
- [x] PH

sphuber:
Fixes #136 and fixes #146 

A lot of the functionality that was implemented in the restart utilities
in `aiida_quantumespresso.utils.restart` has become obsolete by the
`CalcJobNode.get_builder_restart` method which already returns a
`ProcessBuilder` instance with its inputs populated based on the source
node. The utility then merely needs to adapt some inputs such as setting
the correct restart parameter and `parent_folder` node.

### DIFF
--- a/aiida_quantumespresso/utils/restart.py
+++ b/aiida_quantumespresso/utils/restart.py
@@ -1,451 +1,57 @@
 # -*- coding: utf-8 -*-
-"""
-Utility functions to return builders ready to be submitted
-for restarting a Quantum ESPRESSO calculation (or to apply small
-modifications).
-"""
+"""Utility functions to return process builders ready to be submitted for restarting a Quantum ESPRESSO calculation."""
 from __future__ import absolute_import
-from aiida.common import InputValidationError
-from aiida.common.links import LinkType
+
+from aiida_quantumespresso.calculations.cp import CpCalculation
+from aiida_quantumespresso.calculations.neb import NebCalculation
+from aiida_quantumespresso.calculations.ph import PhCalculation
+from aiida_quantumespresso.calculations.pw import PwCalculation
 
 
-def clone_calculation(calculation):
-    """
-    Create a clone of a Calculation node.
-    This is a temporary workaround, it's better to create a new builder.
+def get_builder_restart(node, from_scratch=False, use_symlink=False):
+    """Create a restart `ProcessBuilder` from a completed `CalcJobNode`.
 
-    :returns: an unstored clone of this node. Note: links are not copied
-    """
-    clone = self.__class__()
-    clone.dbnode.dbcomputer = self._dbnode.dbcomputer
-    clone.dbnode.type = self._dbnode.type
-    clone.label = self.label
-    clone.description = self.description
-
-    for key, value in self.iterattrs():
-        if key != Sealable.SEALED_KEY:
-            clone.set_attribute(key, value)
-
-    for path in self.get_folder_list():
-        clone.add_path(self.get_abs_path(path), path)
-
-    return clone
-
-
-def _create_restart_pw_cp(
-    parent_calc, force_restart, parent_folder_symlink, use_output_structure, restart_from_beginning
-):
-    """
-    Function to restart a pw.x or cp.x calculation that was not completed before
-    (like max walltime reached...) i.e. not to restart a really FAILED calculation.
-
-    It returns a `builder` for a new calculation, with all links prepared
-    but not stored in DB.
-    To submit it, simply call::
+    The restart builder will be ready to be launched but of course one can further update the inputs before doing so.
+    To launch the restart calculation, simply run or submit it you would normally::
 
         from aiida.engine import submit
+        builder = get_builder_restart(node)
         submit(builder)
 
-    :param parent_calc: the calculation you want to restart
-    :param bool force_restart: restart also if parent is not in FINISHED
-       state (e.g. FAILED, IMPORTED, etc.).
-    :param bool parent_folder_symlink: if True, symlinks are used
-       instead of hard copies of the files.
-       Pass None for the default calculation behavior.
-    :param bool use_output_structure: if True, the output structure
-       of the restarted calculation is used if available, rather than its
-       input structure. Useful for nscf or bands calculations, but it
-       SHOULD NOT be used for the restart of a relaxation run.
-    :param bool restart_from_beginning: If set to True, creates a copy of
-       the parent calculation, without using the scratch for the restart.
-       Useful to restart calculations that have crashed on the cluster for
-       external reasons.
+    :param node: the `CalcJobNode` instance from which to create a restart builder
+    :param from_scratch: boolean, if True, will restart from scratch
+    :param use_symlink: boolean, if True, will symlink the parent folder instead of copying the contents
+    :return: a `ProcessBuilder` instance configured for a restart
     """
-    from aiida.plugins import DataFactory, CalculationFactory
-    from aiida_quantumespresso.utils.pseudopotential import get_pseudos_of_calc
+    from aiida.orm import Dict
 
-    Dict = DataFactory('dict')
-    RemoteData = DataFactory('remote')
+    supported = (CpCalculation, NebCalculation, PhCalculation, PwCalculation)
 
-    if not isinstance(
-        parent_calc, (CalculationFactory('quantumespresso.cp'), CalculationFactory('quantumespresso.pw'))
-    ):
-        raise TypeError(
-            'This function can only deal with restarts of PwCalculations or CpCalculations (QE pw.x or cp.x codes), '
-            'but I got {} instead'.format(parent_calc.__class__.__name__)
-        )
+    if node.process_class not in supported:
+        raise TypeError('calculation class `{}` of {} is not one of {}'.format(node.process_class, node, supported))
 
-    # Check the calculation's state using ``from_attribute=True`` to
-    # correctly handle IMPORTED calculations.
-    if not parent_calc.is_finished_ok:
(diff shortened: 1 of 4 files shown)


tests/utils/test_restart.py
```diff
@@ -0,0 +1,92 @@
...
+        'settings': orm.Dict(dict={})
+    }
+
+    return inputs
+
+
+class TestGetRestartBuilder(object):
+    """Tests for :py:mod:`~aiida_quantumespresso.utils.restart`."""
+
+    def test_restart(self, fixture_database, fixture_computer_localhost, generate_calc_job_node, generate_inputs):
```
borellim: This function tests DosCalculation, but the function name and the docstring say something else.

tests/utils/test_restart.py
```diff
@@ -0,0 +1,92 @@
...
+    def test_restart_neb(self, fixture_database, fixture_computer_localhost, generate_calc_job_node, generate_inputs):
+        """Test the `get_builder_restart` for a completed `NebCalculation`."""
+        entry_point_calc_job = 'quantumespresso.neb'
+        node = generate_calc_job_node(entry_point_calc_job, fixture_computer_localhost, 'default', generate_inputs)
+
+        builder = restart.get_builder_restart(node)
+        parameters = builder.parameters.get_dict()
+
+        assert isinstance(builder, ProcessBuilder)
+        assert parameters['PATH']['restart_mode'] is True
```
borellim: See other comment: I think this should be `'restart'`.

aiida_quantumespresso/utils/restart.py
```diff
@@ -1,451 +1,57 @@
...
-        parent_folder_symlink = parent_calc._default_symlink_usage
-    # Always set if it was already set. Otherwise, if it wasn't set, just set it if it's not the default
-    if ('PARENT_FOLDER_SYMLINK' in old_settings_dict or parent_folder_symlink != parent_calc._default_symlink_usage):
-        old_settings_dict['PARENT_FOLDER_SYMLINK'] = parent_folder_symlink
+    if node.process_class in (CpCalculation, PwCalculation):
+        parameters.setdefault('CONTROL', {})['restart_mode'] = 'from_scratch' if from_scratch else 'restart'
+    elif node.process_class is PhCalculation:
+        parameters.setdefault('INPUTPH', {})['recover'] = True
+    elif node.process_class is NebCalculation:
+        parameters.setdefault('PATH', {})['restart_mode'] = True
```
borellim: From the QE docs, this is supposed to be the string `'restart'` for NEB. Unless you tried it to run it and found otherwise.

aiida_quantumespresso/utils/restart.py
```diff
@@ -1,451 +1,57 @@
...
-    :param parent_calc: the calculation you want to restart
-    :param bool force_restart: restart also if parent is not in FINISHED
-       state (e.g. FAILED, IMPORTED, etc.).
-    :param bool parent_folder_symlink: if True, symlinks are used
-       instead of hard copies of the files.
-       Pass None for the default calculation behavior.
-    :param bool use_output_structure: if True, the output structure
-       of the restarted calculation is used if available, rather than its
-       input structure. Useful for nscf or bands calculations, but it
-       SHOULD NOT be used for the restart of a relaxation run.
```
borellim: Indeed the new code does not use the output structure of the previous calculation. But is this not appropriate to use it?

sphuber: I think that adding an option to do this is not worth it, because it would take the user literally one line to use it:
```
calculation = load_node()
builder = get_builder_restart(calculation)
builder.structure = calculation.outputs.output_structure
```
Adding a bunch of logic to deal with calculations that then might not have one and substituting with the input structure (which is done by default through the `node.get_builder_restart` call anyway) seems pointless

tests/utils/test_restart.py
sphuber: Good catch. No I didn't try it since NEB is not yet migrated, but I will change it

aiida_quantumespresso/utils/restart.py
sphuber: idem

tests/utils/test_restart.py
sphuber: This is testing generic parts of the function, i.e. the type check. I just picked `DosCalculation` as one of the arbitrary calculation classes that are not supported

aiida_quantumespresso/utils/restart.py
borellim: Ok, but I was wondering about this part of the docstring: `[...] it SHOULD NOT be used for the restart of a relaxation run`.
Does it mean that you should never use the output structure of a relax calculation to restart it? If so, why?

tests/utils/test_restart.py
borellim: Ah, I see. Then I would say that it specifically tests the failure case of an unsupported class, rather than "the generics". But ok, that's a minor complaint.

tests/utils/test_restart.py
```diff
@@ -0,0 +1,92 @@
...
+    def test_restart_neb(self, fixture_database, fixture_computer_localhost, generate_calc_job_node, generate_inputs):
+        """Test the `get_builder_restart` for a completed `NebCalculation`."""
+        entry_point_calc_job = 'quantumespresso.neb'
+        node = generate_calc_job_node(entry_point_calc_job, fixture_computer_localhost, 'default', generate_inputs)
+
+        builder = restart.get_builder_restart(node)
+        parameters = builder.parameters.get_dict()
+
+        assert isinstance(builder, ProcessBuilder)
+        assert parameters['PATH']['restart'] is True
```
borellim: No, I meant this:
```suggestion
        assert parameters['PATH']['restart_mode'] == 'restart'
```

aiida_quantumespresso/utils/restart.py
sphuber: Yeah it doesn't make sense. I guess what they were trying to say is that if you have a completed "relax" `PwCalculation` and then want to restart from it to do an "nscf" or "bands, then you want to use the output structure of course, which is what the function argument is for. However, even for a "relax" calculation, it might still make sense to reuse the output structure, e.g. if it did not converge yet, but already performed some relaxation steps. Another reason why I dropped the `use_output_structure` argument

tests/utils/test_restart.py
sphuber: Ah ok, so it is exactly like pw.x and cp.x then?

borellim: Yes.

tests/utils/test_restart.py
sphuber: Well the idea is that in the future more generics (behavior independent of process class of `node` argument) should be tested in this function. It is just that for the time being it is just testing the type check

tests/utils/test_restart.py
sphuber: should be good now

tests/utils/test_restart.py
borellim: Ok.

aiida_quantumespresso/utils/restart.py
borellim: Ah I see.
Ok, thanks (also for the offline discussion).

tests/utils/test_restart.py
```diff
@@ -0,0 +1,99 @@
...
+    def test_restart_neb(self, fixture_database, fixture_computer_localhost, generate_calc_job_node, generate_inputs):
+        """Test the `get_builder_restart` for a completed `NebCalculation`."""
+        entry_point_calc_job = 'quantumespresso.neb'
+        node = generate_calc_job_node(entry_point_calc_job, fixture_computer_localhost, 'default', generate_inputs)
+
+        builder = restart.get_builder_restart(node)
+        parameters = builder.parameters.get_dict()
+
+        assert isinstance(builder, ProcessBuilder)
+        assert parameters['CONTROL']['restart_mode'] == 'restart'
```
borellim:
```suggestion
        assert parameters['PATH']['restart_mode'] == 'restart'
```

tests/utils/test_restart.py
```diff
@@ -0,0 +1,99 @@
...
+
+        # Force `from_scratch`
+        builder = restart.get_builder_restart(node, from_scratch=True)
+        parameters = builder.parameters.get_dict()
+
+        assert isinstance(builder, ProcessBuilder)
+        assert parameters['CONTROL']['restart_mode'] == 'from_scratch'
```
borellim:
```suggestion
        assert parameters['PATH']['restart_mode'] == 'from_scratch'
```

tests/utils/test_restart.py
sphuber: Good catch, apologies for the sloppiness

tests/utils/test_restart.py
sphuber: Fixed

PR #380, opened 2019-08-14
Update the `Q2rWorkChain` and `MatdynWorkChain`

### LINKED ISSUE #379: Update the `Q2rWorkChain` and `MatdynWorkChain`

sphuber:
Fixes #379 

Leverage the expose functionality to expose the inputs and outputs of
the calculation jobs they wrap. Update the CLI launchers to make sure
that they work.

### DIFF
--- a/aiida_quantumespresso/cli/workflows/matdyn/base.py
+++ b/aiida_quantumespresso/cli/workflows/matdyn/base.py
@@ -12,7 +12,7 @@ from .. import cmd_launch
 
 @cmd_launch.command('matdyn-base')
 @options.CODE(required=True, type=types.CodeParamType(entry_point='quantumespresso.matdyn'))
-@options.CALCULATION(type=types.CalculationParamType(sub_classes=('aiida.calculations:quantumespresso.q2r',)))
+@options.CALCULATION()
 @options_qe.KPOINTS_MESH(default=[2, 2, 2])
 @options_qe.CLEAN_WORKDIR()
 @options_qe.MAX_NUM_MACHINES()
@@ -24,15 +24,19 @@ def launch_workflow(
     code, calculation, kpoints_mesh, clean_workdir, max_num_machines, max_wallclock_seconds, with_mpi, daemon
 ):
     """Run the `MatdynBaseWorkChain` for a previously completed `Q2rCalculation`."""
-    from aiida.orm import Bool, Dict
+    from aiida.orm import Bool
     from aiida.plugins import WorkflowFactory
     from aiida_quantumespresso.utils.resources import get_default_options
 
     inputs = {
-        'code': code,
-        'kpoints': kpoints_mesh,
-        'parent_folder': calculation.outputs.force_constants,
-        'options': Dict(dict=get_default_options(max_num_machines, max_wallclock_seconds, with_mpi)),
+        'matdyn': {
+            'code': code,
+            'kpoints': kpoints_mesh,
+            'parent_folder': calculation.outputs.forceconstants,
+            'metadata': {
+                'options': get_default_options(max_num_machines, max_wallclock_seconds, with_mpi),
+            }
+        }
     }
 
     if clean_workdir:
--- a/aiida_quantumespresso/cli/workflows/matdyn/base.py
+++ b/aiida_quantumespresso/cli/workflows/matdyn/base.py
@@ -2,6 +2,8 @@
 """Command line scripts to launch a `MatdynBaseWorkChain` for testing and demonstration purposes."""
 from __future__ import absolute_import
 
+import click
+
 from aiida.cmdline.params import options, types
 from aiida.cmdline.utils import decorators
 
@@ -12,7 +14,7 @@ from .. import cmd_launch
 
 @cmd_launch.command('matdyn-base')
 @options.CODE(required=True, type=types.CodeParamType(entry_point='quantumespresso.matdyn'))
-@options.CALCULATION()
+@options.CALCULATION(required=True)
 @options_qe.KPOINTS_MESH(default=[2, 2, 2])
 @options_qe.CLEAN_WORKDIR()
 @options_qe.MAX_NUM_MACHINES()
@@ -27,6 +29,14 @@ def launch_workflow(
     from aiida.orm import Bool
     from aiida.plugins import WorkflowFactory
     from aiida_quantumespresso.utils.resources import get_default_options
+
+    expected_process_type = 'aiida.calculations:quantumespresso.q2r'
+    if calculation.process_type != expected_process_type:
+        raise click.BadParameter(
+            'The input calculation node has a process_type: {}; should be {}'.format(
+                calculation.process_type, expected_process_type
+            )
+        )
 
     inputs = {
         'matdyn': {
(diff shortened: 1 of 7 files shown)


aiida_quantumespresso/workflows/q2r/base.py
```diff
@@ -21,46 +19,46 @@ class Q2rBaseWorkChain(BaseRestartWorkChain):
...
     def define(cls, spec):
+        # yapf: disable
         super(Q2rBaseWorkChain, cls).define(spec)
-        spec.input('code', valid_type=orm.Code)
-        spec.input('parent_folder', valid_type=orm.FolderData)
-        spec.input('parameters', valid_type=orm.Dict, required=False)
-        spec.input('settings', valid_type=orm.Dict, required=False)
-        spec.input('options', valid_type=orm.Dict, required=False)
+        spec.expose_inputs(Q2rCalculation, namespace='q2r')
+        spec.expose_outputs(Q2rCalculation, exclude=('remote_folder',))
```
borellim: Idem -- Do you also want to exclude the output `retrieved` (inherited from `CalcJob`)?

aiida_quantumespresso/workflows/q2r/base.py
```diff
@@ -21,46 +19,46 @@ class Q2rBaseWorkChain(BaseRestartWorkChain):
...
         spec.outline(
             cls.setup,
-            cls.validate_inputs,
             while_(cls.should_run_calculation)(
                 cls.run_calculation,
                 cls.inspect_calculation,
             ),
             cls.results,
         )
-        spec.output('force_constants', valid_type=ForceconstantsData)
```
borellim: This becomes `forceconstants` (no underscore). Just checking that it's what you want.

aiida_quantumespresso/cli/workflows/matdyn/base.py
```diff
@@ -12,7 +12,7 @@
 
 @cmd_launch.command('matdyn-base')
 @options.CODE(required=True, type=types.CodeParamType(entry_point='quantumespresso.matdyn'))
-@options.CALCULATION(type=types.CalculationParamType(sub_classes=('aiida.calculations:quantumespresso.q2r',)))
+@options.CALCULATION()
```
borellim: This should actually be:
```suggestion
@options.CALCULATION(required=True)
```
Also, I understand why you removed the type restriction (it does not work), but ideally we should check that the `process_type` of the CalcJobNode corresponds to  `Q2rCalculation`, either via a string comparison to the entry point `aiida.calculations:quantumespresso.q2r` or by actually loading it and checking the class object. However, if you want to defer this, that's fine.

aiida_quantumespresso/cli/workflows/q2r/base.py
```diff
@@ -12,7 +12,7 @@
 
 @cmd_launch.command('q2r-base')
 @options.CODE(required=True, type=types.CodeParamType(entry_point='quantumespresso.q2r'))
-@options.CALCULATION(type=types.CalculationParamType(sub_classes=('aiida.calculations:quantumespresso.ph',)))
+@options.CALCULATION()
```
borellim: Idem -- This should actually be:
```suggestion
@options.CALCULATION(required=True)
```
And same comment about checking the type.

aiida_quantumespresso/cli/workflows/matdyn/base.py
sphuber: Right, I will add the check in the function body for now, thanks

aiida_quantumespresso/cli/workflows/q2r/base.py
sphuber: will add type check and make it required

aiida_quantumespresso/workflows/q2r/base.py
sphuber: I actually prefer it with the underscore but since the `Q2rCalculation` for now uses without, I opted for consistency. I still think it would be good to change it, but then for both, before the release of `3.0.0`

aiida_quantumespresso/workflows/matdyn/base.py
```diff
@@ -21,48 +19,46 @@ class MatdynBaseWorkChain(BaseRestartWorkChain):
...
+        # yapf: disable
         super(MatdynBaseWorkChain, cls).define(spec)
-        spec.input('code', valid_type=orm.Code)
-        spec.input('kpoints', valid_type=orm.KpointsData)
-        spec.input('parent_folder', valid_type=ForceconstantsData)
-        spec.input('parameters', valid_type=orm.Dict, required=False)
-        spec.input('settings', valid_type=orm.Dict, required=False)
-        spec.input('options', valid_type=orm.Dict, required=False)
+        spec.expose_inputs(MatdynCalculation, namespace='matdyn')
+        spec.expose_outputs(MatdynCalculation, exclude=('remote_folder',))
```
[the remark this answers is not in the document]
sphuber: Good point. Thanks to your comment I actually realized that the `PwBaseWorkChain` has an incorrect exclude for the outputs. It currently excludes `retrieved_folder`, which does not exist. Should be `retrieved`. I am now thinking which of those two are actually useful to expose. Some of my higher level workflows use the `retrieved` output of the workchain as a shortcut to restart the next. It functions as the `parent_folder` input. However, for some calculations, it is the `remote_folder` that one wants to reuse. In the case of `PhCalculation` it is the `retrieved` that needs to be used as input for `Q2rCalculation`. But to start `PhCalculation` from `PwCalculation` one needs the `remote_folder`. In short, it changes a lot. So I am starting to think not exclude anything. Excluding one, just saves a single link in the database, which is relatively small and cheap. This prevents the situation that one actually does need an excluded output and then you have to first find the calculation, which if there are multiple run by the base work chain, it is not yet trivial to get the correct one (which is another usability problem to fixed in the future). So how about we do not exclude anything?

aiida_quantumespresso/workflows/q2r/base.py
borellim: I agree. I [made an issue](https://github.com/aiidateam/aiida-quantumespresso/issues/383) so we don't forget, but if you want to do it in this PR it's ok for me.

aiida_quantumespresso/cli/workflows/matdyn/base.py
borellim: Awesome, thanks!

aiida_quantumespresso/workflows/matdyn/base.py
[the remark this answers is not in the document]
borellim: Ok, sounds good.
Do you want to also remove `exclude=('retrieved_folder',)` from `PwBaseWorkChain`?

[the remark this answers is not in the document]
sphuber: Removed it

aiida_quantumespresso/workflows/q2r/base.py
sphuber: Thanks, let's do that one in a separate PR since it might affect other things

sphuber:
Thanks @borellim

PR #381, opened 2019-08-14
(WIP) Migrate NebCalculation and NebParser to aiida 1.0

### LINKED ISSUE #244: Migrate `NebCalculation` to be compatible with `aiida-core==1.0.0`

### LINKED ISSUE #254: Migrate `NebParser` to be compatible with `aiida-core==1.0.0`

borellim:
Changes:
* `NebCalculation` and `NebParser` seem to work fine.
* There are 2 tests for `NebParser`.
* A few other changes/fixes to other parts of the plugin (e.g. the XML parser for pw was ignoring parser options).

Issues:
* Due to the recent overhaul of the `PwParser` logic, on which `NebParser` was leaning, the `NebParser` logic itself is now a bit confusing and could probably use some refactoring.
* There's a bunch of TODOs and open questions, and cleanup is obviously needed.

Comments are welcome.

(Fixes #244, fixes #254)


aiida_quantumespresso/parsers/neb.py
```diff
@@ -95,168 +94,177 @@ def parse(self, **kwargs):
...
-                new_nodes_list.append((self.get_linkname_iterationarray(),arraydata))
+                for k, v in six.iteritems(iteration_data):
+                    arraydata.set_array(k, numpy.array(v))
+                self.out('iteration_array', arraydata)
 
         # Load the original and interpolated energy profile along the minimum-energy path (mep)
         try:
-            mep_file = os.path.join( out_folder.get_abs_path('.'),
-                                                self._calc._PREFIX + '.dat' )
+            mep_file = os.path.join( out_folder._repository._get_base_folder().abspath,
```
sphuber: Use the `out_folder.open` method here. `loadtxt` should also accept filelike objects

aiida_quantumespresso/parsers/neb.py
```diff
@@ -95,168 +94,177 @@ def parse(self, **kwargs):
...
             mep = numpy.loadtxt(mep_file)
         except Exception:
             self.logger.warning('Impossible to find the file with image energies '
                                 'versus reaction coordinate.')
             mep = numpy.array([[]])
 
         try:
-            interp_mep_file = os.path.join( out_folder.get_abs_path('.'),
-                                                self._calc._PREFIX + '.int' )
+            interp_mep_file = os.path.join( out_folder._repository._get_base_folder().abspath,
```
sphuber: same as above, use `out_folder.open`

aiida_quantumespresso/calculations/neb.py
```diff
@@ -5,438 +5,316 @@
...
+        # Create an empty dictionary for the compulsory namelist 'PATH' if not present
         if 'PATH' not in input_params:
             input_params['PATH'] = {}
 
         # In case of climbing image, we need the corresponding card
-        climbing_image = False
-        if input_params['PATH'].get('ci_scheme', 'no-ci').lower()  in ['manual']:
-            climbing_image = True
+        manual_climbing_image = False
+        if input_params['PATH'].get('ci_scheme', 'no-ci').lower() in ['manual']:
```
sphuber: `in ['manual']` could be simplified to `== 'manual'`?

aiida_quantumespresso/calculations/neb.py
```diff
@@ -5,438 +5,316 @@
...
-                                       'unrecognized: {}'.format(list(inputdict.keys())))
+            settings_dict = {}
+
+        parent_calc_folder = self.inputs.get('parent_folder', None)
+        vdw_table = self.inputs.get('pw.vdw_table', None)
 
         # Check that the first and last image have the same cell
-        if abs(np.array(first_structure.cell)-
-               np.array(last_structure.cell)).max() > 1.e-4:
+        if abs(np.array(self.inputs.first_structure.cell)-
```
sphuber: better to dereference `first_structure` and `last_structures` once and assign to variables, because each time you call it, it will perform database operation

aiida_quantumespresso/parsers/neb.py
```diff
@@ -95,168 +94,177 @@ def parse(self, **kwargs):
...
+                    except XMLParseError:
+                        return self.exit_codes.ERROR_OUTPUT_XML_PARSE
+                    except XMLUnsupportedFormatError:
+                        return self.exit_codes.ERROR_OUTPUT_XML_FORMAT
+                    except Exception:
+                        import traceback
+                        traceback.print_exc()
+                        return self.exit_codes.ERROR_UNEXPECTED_PARSER_EXCEPTION
+                    # this image is dealt with, go to next image
+                    break
```
sphuber: This should be a `continue` no? Not a break. With the break it will only ever parse one image

borellim: No, it breaks out of the inner loop (line 111), the one over the expected XML filenames. If none of the expected filenames are found, the 'break' is never executed and the 'else' clause is triggered.

aiida_quantumespresso/cli/calculations/neb.py
```diff
@@ -0,0 +1,134 @@
...
+    if parent_folder:
+        inputs['parent_folder'] = parent_folder
+
+    if dry_run:
+        if daemon:
+            # .submit() would forward to .run(), but it's better to stop here,
+            # since it's a bit unexpected and the log messages output to screen
+            # would be confusing ("Submitted PwCalculation<None> to the daemon")
+            raise click.BadParameter('cannot send to the daemon if in dry_run mode', param_hint='--daemon')
+        inputs.setdefault('metadata', {})['store_provenance'] = False
```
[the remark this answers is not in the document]
borellim: Ah right. I just copied the pw launcher; I'll change that too.

aiida_quantumespresso/calculations/neb.py
borellim: Good to know, I'll do it.
Is this the case for all types of nodes? And do they have all the same cost? The idiom `self.inputs.whatever` is clearer in my opinion, so I would try to use it when the cost is small.

aiida_quantumespresso/calculations/neb.py
borellim: Yeah

aiida_quantumespresso/calculations/__init__.py
```diff
@@ -618,3 +610,19 @@ def _case_transform_dict(d, dict_name, func_name, transform):
             'are repeated more than once when compared case-insensitively: {}.'
             'This is not allowed.'.format(dict_name, double_keys))
     return new_dict
+
+
+def _pop_parser_options(calc_job_instance, settings_dict):
+    # We need an instance of the parser class to get the parser options key (which is typically 'parser_options')
+    from aiida.plugins import ParserFactory
+    parser_name = calc_job_instance.inputs.get('metadata.options.parser_name', None)
```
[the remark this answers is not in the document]
borellim: Good catch. In fact, now that this function is working, I found that my approach is flawed: I can't make an instance of the ParserClass because I would need a CalcJobNode, which I don't have yet. To get around this, I'll turn the methods `get_parser_settings_key()` of `NebParser and `PwParser` into static methods. Alternatively, I could hardcode the 'parser_options' key (also fine by me if you prefer it).

aiida_quantumespresso/cli/calculations/neb.py
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    for structure in structures:
+        try:
+            validate.validate_starting_magnetization(structure, pw_parameters, starting_magnetization)
+        except ValueError as exception:
+            raise click.BadParameter(str(exception))
+
+        try:
+            validate.validate_smearing(pw_parameters, smearing)
```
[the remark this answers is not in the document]
borellim: Right.

borellim:
Alright, so:
1. I added back `tests/parsers/fixtures/neb/default/.gitignore`, so your builder tests should work.
2. I agree that we should try to minimize bloat in parser tests. I reduced the number of images, but I also added another test. Unfortunately I don't know how to identify the parts of the output that are not relevant. Since the most bloat seems to be in `PW.out` files, and you have already done such an exercise for `PwParser`, can you help me a bit here?
3. About the CLI. I understand that you want to keep the launcher as a minimal demo tool, but I *want* to have a script that launches a complete end-to-end calculation and parsing, because it's very useful for me to: (a) generate parser test fixtures, and (b) as a testing tool in general. I also believe that such a script would have value inside the repository, both as a demo tool and for other future developers (including my future self) who want to test with the actual HPC code in the loop. Can we try to think of a solution?

sphuber:
> * I agree that we should try to minimize bloat in parser tests. I reduced the number of images, but I also added another test. Unfortunately I don't know how to identify the parts of the output that are not relevant. Since the most bloat seems to be in `PW.out` files, and you have already done such an exercise for `PwParser`, can you help me a bit here?

What I did really was manual work. In essence I identify the parts of those output files that are not even considered by the parser and then removed them. Lots of information is repeated many times and is never parsed. Once you have the current stubs that the tests compares, you can simply start removing parts of the output file and run the test again. If you removed important information, the parsed results will be different and the test will fail. In principle then, as long as you don't change the parsed results you can keep removing big chunks. Of course this is not the cleanest and if the parser is improved/updated in the future to include some of the parts we remove now, these files are useless. However, while this is not the case I think this approach is worth it and necessary to prevent megabytes of files to be included.

> * About the CLI. I understand that you want to keep the launcher as a minimal demo tool, but I _want_ to have a script that launches a complete end-to-end calculation and parsing.

I understand your point, it is exactly what it is there for. Is this comment in reference to my question why there are so many hard coded parameters in the CLI? If so, I am not saying to remove the CLI, I was just wondering if all of these parameters are really necessary to just run a NEB calculation. That is to say, if you were to remove them and rely on the defaults, would the calculation not work at all?

borellim:
@sphuber I tried to improve the tests for NEB. Do they make sense now? Thanks.

sphuber:
@borellim as you can see I updated the branch to include the two fixes we merged into `develop` yesterday. After that I added some commits to solve some outstanding problems and questions, and I reduced the weight of the test fixture files. I would be happy to squash merge it like this. Let me know if you are on board as well.

sphuber:
Thanks a lot @borellim !

PR #386, opened 2019-08-27
Renaming of `ForceconstantsData` to `ForceConstantsData`

### LINKED ISSUE #383: Q2rCalculation and Q2rBaseWorkChain: replace output link 'forceconstants' with 'force_constants'
This seems more consistent with how we spell other link names.

sphuber:
Fixes #383 

This also indirectly changes `forceconstants` to `force_constants` both
in module name as well as the output link label used for calculations
and work chains.

### DIFF
--- a/aiida_quantumespresso/calculations/matdyn.py
+++ b/aiida_quantumespresso/calculations/matdyn.py
@@ -3,7 +3,7 @@ from __future__ import absolute_import
 
 from aiida import orm
 from aiida_quantumespresso.calculations.namelists import NamelistsCalculation
-from aiida_quantumespresso.data.forceconstants import ForceconstantsData
+from aiida_quantumespresso.data.force_constants import ForceConstantsData
 
 
 class MatdynCalculation(NamelistsCalculation):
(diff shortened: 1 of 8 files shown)


sphuber:
Thanks @borellim I will update the other PR

giovannipizzi:
Sorry to realise this only now.
I wanted to change the name earlier but didn't want to break backward-compatibility.
Since we are now changing it, I was wondering if we can give a name that says that these are force constants in QE format (QEForceConstantsData?) because IIRC it just stores the QE raw file.
Because we are planning to prepare instead a ForceConstantsData that is code-independent, and it would be nice to free up the namespace.

borellim:
Would it be sufficient to define them under different entry points ('quantumespresso.force_constants' versus 'force_constants')?

giovannipizzi:
Yes, this is also what @sphuber was mentioning to me shortly ago. Indeed, I was thinking to some time ago, where the entry point was simply `forceconstants`, but if it's under `quantumespresso.force_constants` I think this answers my question. Thanks!

PR #387, opened 2019-08-27
`MatdynCalculation`: change `parent_folder` to `force_constants`

### LINKED ISSUE #292: `MatdynCalculation`: replace input 'parent_folder' with some other name
The accepted type is `ForceconstantsData`, which is a subtype of `SinglefileData` (not `FolderData`) so the name is inappropriate. See #288.
@sphuber

sphuber:
Fixes #292 

The `MatdynCalculation` inherited the `parent_folder` input port from
the `NamelistCalculation` base class, except the name is a bit
unfortunate. All other namelist calculations indeed take a parent folder
as input, which can either be a remote or retrieved folder data.
However, for the `MatdynCalculation` this should actually be a
`ForceConstantsData` single file node, produced by a `Q2rCalculation`.


aiida_quantumespresso/cli/workflows/matdyn/base.py
```diff
@@ -14,7 +12,11 @@
 
 @cmd_launch.command('matdyn-base')
 @options.CODE(required=True, type=types.CodeParamType(entry_point='quantumespresso.matdyn'))
-@options.CALCULATION(required=True)
+@options.DATUM(
+    required=True,
+    type=types.DataParamType(sub_classes=('aiida.data:quantumespresso.forceconstants',)),
+    help='A ForceconstantsData node produced by a `Q2rCalculation`'
```
borellim: * After PR #386, these should become `'aiida.data:quantumespresso.force_constants'` and `ForceConstantsData` respectively.
* Using a `datum` instead of a `calculation` makes using this launcher a bit less easy, which is unfortunate, but it's certainly cleaner code.

tests/calculations/test_matdyn.py
```diff
@@ -0,0 +1,48 @@
...
+# pylint: disable=unused-argument
+"""Tests for the `MatdynCalculation` class."""
+from __future__ import absolute_import
+
+import os
+
+from aiida.common import datastructures
+from aiida.plugins import CalculationFactory
+
+from aiida_quantumespresso.data.forceconstants import ForceconstantsData
```
borellim: Same here and below: will require changes after PR #386 is in.

aiida_quantumespresso/cli/workflows/matdyn/base.py
sphuber: I also initially thought that it was less easy to use, but after some thinking I am not sure. Since anyway before being able to launch it you need to run a `Q2rCalculation`, if you do it through the CLI as well, at the end the `ForceConstantsData` output node will be printed with a pk. If you want to restart later from a `Q2rCalculation` pk, it only takes `verdi process show` to get the correct output node. So I think this is fine in the end

sphuber:
Thanks Marco for the reviews

PR #388, opened 2019-08-28
Import physical constants from `qe-tools` instead of `aiida-core`

### LINKED ISSUE #69: Inconsistent usage of conversion factors
The plugin uses several constants during parsing, which are currently defined and taken from `aiida_core`, see `aiida_quantumespresso.parsers.constants`. The constants `ry_to_ev` and `ry_si` are inconsistent but both are used. The first is used in general, but the latter is used for the parsing of the stress. These constants will be removed from `aiida_core` in the near future (see [issue on aiida_core](https://github.com/aiidateam/aiida_core/issues/161)) and will have to be moved here. When the parsing is updated to use the new XML schema, we should take the opportunity to also make the usage of the constants consistent

sphuber:
Fixes #69 

The physical constants, mostly used by the parsers, used to be defined
in `aiida-core` for the Quantum ESPRESSO plugins, before they were moved
to this repository. The physical constants remained, however, but will
be removed for `aiida-core==1.0.0`. The constants had already be copied
to `qe-tools` as well, which is where they will stay. Here we change the
import to use the constants from the latter instead of `aiida-core`.


sphuber:
I will open a PR on `aiida-core` as well to remove the constants from there. I was just thinking about compatibility though. Removing them from `aiida-core` will break existing `aiida-quantumespresso` releases. Given that they are alphas, that might be acceptable, especially if we make another release of `aiida-quantumespresso`, except we just released one yesterday. What do you think @giovannipizzi

sphuber:
I also opened [an issue on `qe-tools`](https://github.com/aiidateam/qe-tools/issues/25) to continue the discussion on the original problem of inconsistent use of constant sets.

sphuber:
The only problem is that I should release a new alpha of `aiida-quantumespresso` that depends on `qe-tools` for the constants _before_ a release is made of `aiida-core` where they are removed. Otherwise, if people upgrade their `aiida-core`, their current `aiida-quantumespresso` will fail. That is why I blocked the PR on `aiida-core` until after I made the latest beta release. That way we will have some period of leeway where we can apply the changes in `aiida-quantumespresso` and make yet another release there as well. It is unfortunate we just released an alpha on monday evening

PR #393, opened 2019-08-30
`PwParser`: restore definition of k-point labels in `BandsData` output

### LINKED ISSUE #392: Labels are not set on k-points of `BandsData` output of `PwCalculation`
This prevents plotting/export utilities to define labels on the x-axis for band structures produced from those `BandsData` output nodes

sphuber:
Fixes #392 

This got lost in the recent refactoring of the `PwParser` and prevents
the various export and plot utilities of `BandsData` nodes to plot a
nice band structure with the correct k-point labels on the x-axis.


ltalirz:
Great, that's exactly what we need!

PR #394, opened 2019-08-30
add pypi deployment to travis

ltalirz:
automatically deploy pypi package when tagging

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,3 +61,31 @@ script:
 
 git:
   depth: 3
+
+jobs:
+    include:
+        - stage: deploy
+          if: "tag =~ ^v[0-9]+\\.[0-9]+\\.[0-9]+(a|b|rc)?[0-9]*$"
+          services: ignore
+          addons: skip
+          python: 3.6
+          before_install: skip
+          install: skip
+          before_script: skip
+          script: skip
+          env: ignore
+          before_deploy:
+            - echo "We'll deploy $TRAVIS_TAG"
+          after_deploy:
+            - echo "Deployed $TRAVIS_TAG"
+            - echo "Could hook up further triggers here."
+          deploy: &pypi
+              provider: pypi
+              skip_existing: true
+              username: aiida-bot
+              password:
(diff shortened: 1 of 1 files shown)


ltalirz:
adding this, since I might ask for an a5 release ;-)

sphuber:
I think you copied this more or less from `aiida-core`. Does that mean it will only trigger for tags on the master branch? If so, then this won't help at all for the next alpha correct?

ltalirz:
it should trigger on the tag, independently of the branch

sphuber:
It didn't trigger for `aiida-core==1.0.0b6` though yesterday. Is this different from the `aiida-core` recipe?

ltalirz:
> It didn't trigger for aiida-core==1.0.0b6 though yesterday. Is this different from the aiida-core recipe?

The travis job is still missing from the `develop` branch (only present in `master`).
If you like, I can make a PR to fix this.

ltalirz:
hm wait... it seems I got confused

sphuber:
I don't understand on `aiida-core` in the develop branch I _do_ see the same deploy stage in `.travis.yml` except there we have
```
repo: aiidateam/aiida-core
all_branches: true
tags: true
```
whereas here you do no include `tags`.

ltalirz:
yes, perhaps I looked on my fork - it *is* there, and the deploy job *was* triggered:
https://travis-ci.org/aiidateam/aiida-core/builds/578424816?utm_source=github_status&utm_medium=notification

sphuber:
Ah it seems I already uploaded it manually to PyPI before the deploy stage got the chance. I forgot that this was already up and running even for pre-releases on develop

sphuber:
So final questions, don't we need to include to trigger on `tags` here? And why python 2 and not 3?

ltalirz:
just fixed the tags, happy to switch to py3 as well

ltalirz:
>  I forgot that this was already up and running even for pre-releases on develop

Just updated release instructions as well
https://github.com/aiidateam/aiida-core/wiki/How-to-make-a-new-release

sphuber:
Thanks @ltalirz

Issue #395, opened 2019-08-31
making new releases


ltalirz:
@sphuber While preparing a 3.0.0a5 release, I encountered some minor things:

 * when editing only the `setup.json`, the version consistency pre-commit hook rightfully complains, and then changes the version back to be consistent with the one in `__init__.py` (not what I intended, but ok)
 * when editing only the `__init__.py`, the pre-commit hook does *not* complain (bug)
 * the version number also appears in a wide range of test results. do these need to be updated for every release?

```
$ git grep -i "3.0.0a4"
aiida_quantumespresso/__init__.py:__version__ = '3.0.0a4'
setup.json:    "version": "3.0.0a4"
tests/parsers/test_ph/test_ph_default.yml:parser_info: aiida-quantumespresso parser ph.x v3.0.0a4
tests/parsers/test_ph/test_ph_default.yml:parser_version: 3.0.0a4
tests/parsers/test_ph/test_ph_not_converged.yml:parser_info: aiida-quantumespresso parser ph.x v3.0.0a4
tests/parsers/test_ph/test_ph_not_converged.yml:parser_version: 3.0.0a4
tests/parsers/test_ph/test_ph_out_of_walltime.yml:parser_info: aiida-quantumespresso parser ph.x v3.0.0a4
tests/parsers/test_ph/test_ph_out_of_walltime.yml:parser_version: 3.0.0a4
tests/parsers/test_pw/test_pw_default.yml:  parser_info: aiida-quantumespresso parser pw.x v3.0.0a4
tests/parsers/test_pw/test_pw_default.yml:  parser_version: 3.0.0a4
tests/parsers/test_pw/test_pw_default_xml_new.yml:  parser_info: aiida-quantumespresso parser pw.x v3.0.0a4
tests/parsers/test_pw/test_pw_default_xml_new.yml:  parser_version: 3.0.0a4
tests/parsers/test_pw/test_pw_failed_interrupted.yml:parser_info: aiida-quantumespresso parser pw.x v3.0.0a4
tests/parsers/test_pw/test_pw_failed_interrupted.yml:parser_version: 3.0.0a4
tests/parsers/test_pw/test_pw_failed_interrupted_stdout.yml:parser_info: aiida-quantumespresso parser pw.x v3.0.0a4
tests/parsers/test_pw/test_pw_failed_interrupted_stdout.yml:parser_version: 3.0.0a4
[...]

PR #396, opened 2019-08-31
bump version to 3.0.0a5

ltalirz:
partial fix to #395 

bump version number to 3.0.0a5 for new release.
also let version consistency pre-commit hook trigger when
aiida_quantumespresso/__init__.py is changed.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,7 +41,12 @@
     entry: python ./.ci/validate_version_number.py
     args: ['version']
     language: system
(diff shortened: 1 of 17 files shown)


ltalirz:
just fyi in case you're wondering: I created a v3.0.0a5 tag before (forgetting I need to increase the version number).
The deployment was triggered as planned - but the name of the  .tar.gz remained 3.0.0a4, so while uploading to pypi it realised that and didn't do anything, i.e. no harm done.
I deleted the v3.0.0a5 tag again.

PR #397, opened 2019-08-31
update encrypted pypi deploy pw

ltalirz:
This time, the pw was encrypted for this specific repo.


sphuber:
Is the user also correct? What does the `aiida-bot` reference. A PyPI account? And if so, does it have maintainance rights on the `aiida-quantumespresso` package?

PR #406, opened 2019-10-11
pw base workchain - restart from previous calculation if pw error = 410

### LINKED ISSUE #402: `PwBaseWorkChain`: investigate efficiency of non-convergence handler
Currently the error handler for `410` reduced the beta mixing and then restarts from scratch. @d-tomerini has suggested that probably restarting from the previous charge density in this case might be more successful.

d-tomerini:
Fixes #402
 Currently, when a ``quantumespresso.pw`` calculation return a 410 exit_status, the ``quantumespresso.pw.base`` workchain decreases the beta mixing parameters in ELECTRONS and restart the calculation from scratch.
A more effective solution is to restart the calculation from the previous charge density. 

I tested this strategy for 181 failing workflows, and the new strategy resulted in achieved convergence in 50% of the cases.
The same strategy was also applied to workchains that were successful after one or more the contained  ``quantumespresso.pw`` returned a 410 error. I observed failure in only 1 case out of 55 cases; the number of steps required to achieve convergence was also improved in 51 out of 54 cases.

### DIFF
--- a/aiida_quantumespresso/workflows/pw/base.py
+++ b/aiida_quantumespresso/workflows/pw/base.py
@@ -438,10 +438,11 @@ def _handle_electronic_convergence_not_achieved(self, calculation):
         mixing_beta = self.ctx.inputs.parameters.get('ELECTRONS', {}).get('mixing_beta', self.defaults.qe.mixing_beta)
         mixing_beta_new = mixing_beta * factor
 
-        self.ctx.restart_calc = None
+        self.ctx.restart_calc = calculation
         self.ctx.inputs.parameters.setdefault('ELECTRONS', {})['mixing_beta'] = mixing_beta_new
 
-        action = 'reduced beta mixing from {} to {} and restarting from scratch'.format(mixing_beta, mixing_beta_new)
+        action = 'reduced beta mixing from {} to {} and restarting from the last ' \
+            'calculation'.format(mixing_beta, mixing_beta_new)
         self.report_error_handled(calculation, action)
         return ErrorHandlerReport(True, True)
 


sphuber:
@d-tomerini thanks. I would ask you to read the [GIT style guidelines](https://github.com/aiidateam/aiida-core/wiki/Git-style-requirements) carefully. For the commit message format it states for example
> The message body can be widely varying, from empty, if the commit title really is all there is to say, or multiple paragraphs with again a maximum line length of 72 characters.

The reason for this can be seen on this page. Directly below your main post you see the commit. If you click on the three dots to expand to show the full message you see the text runs completely of screen. This is the reason that we require a line length of maximally 72 characters for the commit body. Note that this is not just for Github but also on the CLI when you do `git log`.

If you use `vim` to write your commit messages, it comes with automatic highlighting and linebreaks to respect these rules. It is advisable to **not** use the `-m` flag to write your commit message directly in the commit command

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -438,10 +438,11 @@ def _handle_electronic_convergence_not_achieved(self, calculation):
         mixing_beta = self.ctx.inputs.parameters.get('ELECTRONS', {}).get('mixing_beta', self.defaults.qe.mixing_beta)
         mixing_beta_new = mixing_beta * factor
 
-        self.ctx.restart_calc = None
+        self.ctx.restart_calc = calculation
         self.ctx.inputs.parameters.setdefault('ELECTRONS', {})['mixing_beta'] = mixing_beta_new
 
-        action = 'reduced beta mixing from {} to {} and restarting from scratch'.format(mixing_beta, mixing_beta_new)
+        action = 'reduced beta mixing from {} to {} and restarting from the last ' \
```
sphuber: Is there a reason you broke this on two lines? It fix fine on the 120 char limit we maintain

d-tomerini:
Hi @sphuber ,

Apologies for the VIM writing. I'll be more careful and recheck the guidelines for next time

aiida_quantumespresso/workflows/pw/base.py
d-tomerini: Prospector was complaining about the line length, so I split it

sphuber: Ah ok, I see you changed the string itself, so it became longer. I see. All good then

sphuber:
Thanks a lot @d-tomerini

PR #407, opened 2019-10-11
Added SCF convergence trajectory to the pw parser

### LINKED ISSUE #403: `PwParser`: parse the SCF convergence trajectory
This information is important to analyze the behavior of the SCF convergence and gives a good idea if it is going in the right direction in case convergence is not reached. Note, this is currently not available from the XML but only the stdout. @d-tomerini will write to the QE developers to request this information to be added.

d-tomerini:
Fixes #403

The pw parser now look for the SCF convergence in each SCF cycle of
a relaxation. The data is saved in two arrays of the output_trajectory. 

- ``scf_accuracy`` is a 1D array containing all the estimated errors
  found in the output;
- ``scf_accuracy_index`` acts as an index to identify different relax steps
  in ``scf_accuracy``. Each entry is the index of the last item of a specific relax step;
  i.e. the relaxation convergence in the latest relax step might be extracted to 
  the array ``xscf`` as:
```
  i1 = scf_accuracy_index[-2]
  i2 = scf_accuracy_index[-1]
  xscf = scf_accuracy[i1+1: i2]
```
The separation is two array is to avoid inhomogeneous arrays, see
https://github.com/aiidateam/aiida-core/issues/3411

Parsing of these quantities will allow for more complex logic in the workflows.

### DIFF
--- a/aiida_quantumespresso/parsers/parse_raw/pw.py
+++ b/aiida_quantumespresso/parsers/parse_raw/pw.py
@@ -16,6 +16,7 @@ from six.moves import range, zip
 from aiida_quantumespresso.parsers import QEOutputParsingError
 from aiida_quantumespresso.utils.mapping import get_logging_container
 from qe_tools.constants import ry_to_ev, bohr_to_ang, ry_si, bohr_si
+from six.moves import zip
 
 lattice_tolerance = 1.e-5
 units_suffix = '_units'
@@ -607,11 +608,17 @@ def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None)
                 except Exception:
                     logs.warning.append('Error while parsing scf accuracy.')
 
-
             elif 'convergence has been achieved in' in line or 'convergence NOT achieved after' in line:
                 try:
-                    scf_iterations = int(line.split('iterations')[0].split()[-1])
-                    trajectory_data.setdefault('scf_iterations', []).append(scf_iterations)
+                    value = int(line.split('iterations')[0].split()[-1])
+                    trajectory_data.setdefault('scf_iterations', []).append(value)
+                except Exception:
+                    logs.warning.append('Error while parsing scf iterations.')
+
+            elif 'Calculation stopped in scf loop at iteration' in line:
+                try:
+                    value = int(line.split()[-1])
+                    trajectory_data.setdefault('scf_iterations', []).append(value)
                 except Exception:
                     logs.warning.append('Error while parsing scf iterations.')
 
@@ -827,6 +834,20 @@ def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None)
                 trajectory_data.setdefault('electronic_dipole_cartesian_axes', []).append(ed_axes)
                 trajectory_data.setdefault('ionic_dipole_cell_average', []).append(id_cell)
                 trajectory_data.setdefault('ionic_dipole_cartesian_axes', []).append(id_axes)
+
+    # adding the final element to the scf_accuracy_index array and check consistency
+    if 'scf_accuracy' and 'scf_iterations' in trajectory_data:
+        value = len(trajectory_data['scf_accuracy'])
+        trajectory_data['scf_accuracy_index'].append(value)
+        # check consistency of accuracy_index vs scf_iterations
+        warning = 'scf_iterations and scf_accuracy arrays are not consistent'
+        if len(trajectory_data['scf_iterations']) + 1 != len(trajectory_data['scf_accuracy_index']):
+            logs.warning.append(warning)
+        value = trajectory_data['scf_accuracy'][0]
+        for i, j in zip(trajectory_data['scf_iterations'], trajectory_data['scf_accuracy_index'][1:]):
+            value += i
+            if value != j:
+                logs.warning.append(warning)
 
     # If specified in the parser options, parse the atomic occupations
     parse_atomic_occupations = parser_options.get('parse_atomic_occupations', False)
--- a/aiida_quantumespresso/parsers/parse_raw/pw.py
+++ b/aiida_quantumespresso/parsers/parse_raw/pw.py
@@ -843,7 +843,7 @@ def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None)
         warning = 'scf_iterations and scf_accuracy arrays are not consistent'
         if len(trajectory_data['scf_iterations']) + 1 != len(trajectory_data['scf_accuracy_index']):
             logs.warning.append(warning)
-        value = trajectory_data['scf_accuracy'][0]
+        value = trajectory_data['scf_accuracy_index'][0]
         for i, j in zip(trajectory_data['scf_iterations'], trajectory_data['scf_accuracy_index'][1:]):
             value += i
             if value != j:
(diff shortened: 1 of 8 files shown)


aiida_quantumespresso/parsers/parse_raw/pw.py
```diff
@@ -270,10 +270,15 @@ def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None)
...
     # now I create a bunch of arrays for every step.
-    for data_step in relax_steps:
 
+    # initialize the scf_index counter for the next cycle
+    scf_index = 0
+    for data_step in relax_steps:
         trajectory_frame = {}
-
+        try:
+            trajectory_data['scf_accuracy_index'].append(scf_index)
```
sphuber: There is a useful paradigm in python for this situation:
```
trajectory_data.setdefault('scf_accuracy_index', []).append(scf_index)
```
This will append the value to the list if it exists, or will create the list first if it does not yet exist

aiida_quantumespresso/parsers/parse_raw/pw.py
```diff
@@ -366,6 +371,22 @@ def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None)
                         parsed_data['dipole' + units_suffix] = default_dipole_units
                         break
 
+            # saving the SCF convergence accuracy for each SCF cycle
+            # this is not always present in the block starting with 'iteration #'
+            # so the array might be of different size for each relax cycle
+            elif 'estimated scf accuracy' in line:
+                try:
+                    value = float(line.split()[-2])* ry_to_ev
+                    scf_index += 1
```
sphuber: Instead of keep track of the `scf_index` through a variable and upping it in `scf_accuracy_index`, wouldn't it be nicer to just at the end of the frame, do something like:
```
current_index = len(trajectory_data.get('scf_accuracy', [])
trajectory_data.setdefault('scf_accuracy_index', []).append(current_index)
```
```

File: aiida_quantumespresso/parsers/parse_raw/pw.py (hunk 1 above)
Comment (d-tomerini) [reply aiida_quantumespresso/parsers/parse_raw/pw.py]: @sphuber That was actually my first idea, but I opted to keep consistency with the rest of the try...- except sections of the parser. I can go through those too in other places and change it if you prefer

Comment (sphuber) [reply aiida_quantumespresso/parsers/parse_raw/pw.py]: I see. Consistency is always good. But let's change them all for the more idiomatic version of `setdefault`.

File: aiida_quantumespresso/parsers/parse_raw/pw.py (hunk 2 above)
Comment (d-tomerini) [reply aiida_quantumespresso/parsers/parse_raw/pw.py]: Hi,

Your suggestion is an option, but in the way the code is now it needs to be at the beginning of the frame, and ``trajectory_data['scf_accuracy']`` does not before the first assignment. 

This would also change it so that the array starts from 0 and does not contain the last index (the length of the array). It could actually be an alternative.

Also, right now the ``scf_accuracy`` entry is not in the output unless it is assigned at least once, but we might want to pass an empty array anyway? Not sure.

Comment (sphuber) [reply aiida_quantumespresso/parsers/parse_raw/pw.py]: If you want to keep the index array to keep the starting index of each frame, then you can already create `trajectory_data['scf_accuracy_index'] = [0]` before the loop over the frames starts. Then at the end of each frame you can call `trajectory_data['scf_accuracy_index'].append(current_index)`

Comment (d-tomerini) [reply aiida_quantumespresso/parsers/parse_raw/pw.py]: I see what you mean now. I will change it, thanks for the suggestion

File: aiida_quantumespresso/parsers/parse_raw/pw.py [hunk 3]
```diff
@@ -270,10 +270,14 @@ def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None)
     relax_steps = [i.split('\n') for i in relax_steps]
 
     # now I create a bunch of arrays for every step.
-    for data_step in relax_steps:
 
+    # initialize the scf_index counter for the next cycle
```
sphuber: This comment is now obsolete

aiida_quantumespresso/parsers/parse_raw/pw.py
```diff
@@ -366,6 +370,20 @@ def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None)
...
                         break
 
+            # saving the SCF convergence accuracy for each SCF cycle
+            # this is not always present in the block starting with 'iteration #'
+            # so the array might be of different size for each relax cycle
+            elif 'estimated scf accuracy' in line:
+                try:
+                    value = float(line.split()[-2])* ry_to_ev
+                    try:
+                        trajectory_data['scf_accuracy'].append(value)
```
sphuber: Can you please replace this old idiom with the `setdefault` one. Also please replace the other uses of the old idiom in this function if you will

sphuber:
Thanks a lot @d-tomerini . It is looking good. I was trying this out locally:
```
In [16]: trajectory.get_array('scf_accuracy_index')
Out[16]: array([ 0,  6, 11, 18, 24])

In [17]: trajectory.get_array('scf_accuracy')
Out[17]: 
array([1.42905846e+00, 7.27112656e-02, 4.15286529e-03, 3.65965896e-03,
       1.44220332e-05, 6.80284586e-06, 8.58042949e-03, 9.61146881e-03,
       3.24944735e-03, 8.95254516e-05, 8.70764270e-06, 7.15332848e-03,
       7.15047129e-03, 2.31283154e-03, 6.35385804e-05, 5.03410594e-06,
       6.80284586e-07, 1.02042688e-08, 1.51567406e-04, 1.34288177e-04,
       4.14973598e-05, 9.52398421e-07, 2.72113835e-07, 4.89804902e-10,
       1.29673099e+00, 5.87516898e-02, 6.23140681e-04, 1.89119115e-04,
       1.37417486e-05, 5.30621977e-08])

In [18]: trajectory.get_array('scf_iterations')
Out[18]: array([6, 5, 7, 6, 6])
```
In a sense we are doubling up information a bit. The `scf_iterations` and `scf_accuracy_index` is a bit redundant. I don't think we need to drop either since they will never be very big and they might make working with the data a bit easier if one has both. Maybe it would be good however to put a small consistency check in? In the parser we check that the values in those arrays make sense and if not emit a warning?

Also maybe it would be better to add the final index to `scf_accuracy_index`? Currently if I wanted to make a function to automatically get the SCF accuracy for a given frame directly from the node, I would have to implement something like:
```
def get_scf_accuracy(node, index=0):
        trajectory = node.outputs.output_trajectory
        scf_accuracy = trajectory.get_array('scf_accuracy')
        scf_accuracy_index = trajectory.get_array('scf_accuracy_index')
        scf_iterations = trajectory.get_array('scf_iterations')
        return scf_accuracy[scf_accuracy_index[index]:scf_accuracy_index[index] + scf_iterations[index]]
```
if we had all information in `scf_accuracy_index` we could do:


```
def get_scf_accuracy(node, index=0):
        trajectory = node.outputs.output_trajectory
        scf_accuracy = trajectory.get_array('scf_accuracy')
        scf_accuracy_index = trajectory.get_array('scf_accuracy_index')
        return scf_accuracy[scf_accuracy_index[index]:scf_accuracy_index[index + 1]]
```

d-tomerini:
Hi @sphuber,
thanks for your comments.
The second suggestion makes indeed manipulations better. I'll add it now.

The consistency check is a bit more tricky. I've noticed that the 'scf accuracy' is not always there when too large, on some first iteration restart it prints just a warning (I have to look again what it looks like).
This means that the array can be shorter than the actual number of scf iterations, (but I would not consider it warning-worthy per se). I could add it a less-than check

d-tomerini:
@sphuber added your two suggestions and introduced an additional case where the scf_iterations array was not added (at the end of an SCF cycle interrupted for an out_of_walltime reasons)

PR #408, opened 2019-10-14
Soften monotonicity requiremnts occupations in `get_highest_occupied_band`

### LINKED ISSUE #399: Slightly negative occupations cause sanity check to fail
Running a calculation with ``occupations='tetrahedra'``, I obtain slightly negative (roughly 1e-8) occupations. 

This doesn't seem to be a problem, but causes the sanity check in the base workchain to fail with 
```
 BandsData<4895> has invalid occupations: Occupation of 0.0 at n=15 after lumo lkn<-7.536307628800891e-09,14,0>
```

Would it be appropriate to add a small margin of error on that test? For example in https://github.com/aiidateam/aiida-quantumespresso/blob/develop/aiida_quantumespresso/utils/bands.py#L48, change to ``if occupation > lumo_occupation + threshold:``?

sphuber:
Fixes #399 

The utility function, in addition to determining the HOMO, also checked
that the occupations are monotonically decreasing, since when broken, it
is a hint of non-sensical band structures. However, the check was that
once the LUMO was found, all subsequent bands had to have an occupation
that was lower. This gave problems for band structures that have
negative occupations just below the fermi level. This can happen with
both Methfessel-Paxton and Marzari-Vanderbilt smearing. The LUMO would
be marked as the band with negative occupations and any subsequent band
with occupation of zero, would trigger the exception. Although
physically meaningless, negative occupations are well-known effects of
these smearing methods and need to be accepted. The solution is to
change the criterion to compare unoccupied bands not to the occupation
of the LUMO but to the specified threshold. As long as all unoccupied
bands have an occupation less than twice the threshold (where the factor
two accounts for numerical noise) the monotonicity is preserved within
the bounds of numerical noise.

### DIFF
--- a/aiida_quantumespresso/utils/bands.py
+++ b/aiida_quantumespresso/utils/bands.py
@@ -3,17 +3,31 @@
 from __future__ import absolute_import
 
 
-def get_highest_occupied_band(bands, threshold=0.01):
+def get_highest_occupied_band(bands, threshold=0.005):
     """Retun the index of the highest-occupied molecular orbital.
 
     The expected structure of the bands node is the following:
+
+        * an array called `occupations`
+        * with 3 dimensions if spin polarized, otherwise 2 dimensions
+        * dimensions loop over, spin channel, kpoints, bands
+
+    .. note::
+
+        The threshold is used both as a limit below which a band is considered as unoccupied as well as a measure of
+        numerical noise in the occupancies. The first band to have an occupancy below the threshold at all kpoints is
+        marked as the LUMO. All subsequent bands are then expected to have an occupation that is less than twice the
+        threshold. The reason for the factor two is that in the worst case when the LUMO has an occupation exactly equal
+        to the threshold and a subsequent has twice that value, we can still consider that as not to break the
+        requirement of all bands above the LUMO to be empty. It can be considered as having the exact same value (equal
+        to the threshold) plus an numerical error also equal to the threshold.
 
     :param bands: the `BandsData` node
     :param threshold: raise a `ValueError` if the last band has an occupation above this threshold at any k-point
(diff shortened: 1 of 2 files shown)


aiida_quantumespresso/utils/bands.py
```diff
@@ -3,11 +3,25 @@
...
+        * dimensions loop over, spin channel, kpoints, bands
+
+    .. note::
+
+        The threshold is used both as a limit below which a band is considered as unoccupied as well as a measure of
+        numerical noise in the occupancies. The first band to have an occupancy below the threshold at all kpoints is
+        marked as the LUMO. All subsequent bands are then expected to have an occupation that is less than twice the
+        threshold. The reason for the factor two is that in the worst case when the LUMO has an occupation exactly equal
+        to the threshold and a subsequent has twice that value, we can still consider that as not to break the
+        monotonically decreasing behavior of occupations. It can be considered as having the exact same value (equal to
```
greschd: I think we shouldn't talk about "monotonicity" here - all that's checked is that the bands above LUMO are _also_ unoccupied (with a less strict threshold).

The 
``:raises ValueError: if the occupations are not monotonically decreasing at all k-points between LUMO and HOMO`` 
could probably be changed to something like 
``:raises ValueError: if any occupation above LUMO exceeds `2 * threshold``

sphuber: Good point. Fixed the docstrings and commit message, and added a unit test for the threshold parameter as well.

greschd:
Seems my approval is not sufficient to merge 😉

sphuber:
Done. Thanks Dominik

greschd:
Thanks for implementing it!

PR #412, opened 2019-10-16
Extend BaseRestartWorkChain to CalcJob+WorkChain

Tseplyaev:
BaseRestartWorkChain can also wrap another WorkChain due to similarities between CalcJob and WorkChain classes. There is no reason to restrict a user to wrap a CalcJob only, sometimes it might be helpful to automatically restart a WorkChain. Personally I found the same update for AiiDA-Fleur helpful.

PS. self.ctx.calculations and other variables could be renamed to self.ctx.processes or something like that. However, one needs to update error handlers then.

### DIFF
--- a/aiida_quantumespresso/common/workchain/base/restart.py
+++ b/aiida_quantumespresso/common/workchain/base/restart.py
@@ -58,8 +58,8 @@ class BaseRestartWorkChain(WorkChain):
     def __init__(self, *args, **kwargs):
         super(BaseRestartWorkChain, self).__init__(*args, **kwargs)
 
-        if self._calculation_class is None or not issubclass(self._calculation_class, CalcJob):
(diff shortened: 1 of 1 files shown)


Tseplyaev:
Thank you, Sebastiaan.
Sure, my implementation of wrapping a WorkChain via RestartWorkChain in AiiDA-Fleur was not tested in detail.
I just wanted to show that the restart workchain actually works for WorkChains too (at least I was able to catch an exit code and use the error_handler to solve it e.g. resubmit a workchain with a different set of parameters) and the only change that was needed is a small modification to the process class check.

sphuber:
I am closing this because we are in the process of moving the `BaseRestartWorkChain` functionality to `aiida-core` which will generalize it as you described here. This is the PR: https://github.com/aiidateam/aiida-core/pull/3654

PR #414, opened 2019-10-16
`PwParser`: allow parsing of incomplete XML output with new format

sphuber:
Fixes #324 

The parser of the XML output file through the new schema-based parsing
assumed the output would always be complete. However, for example during
an initialization run, where only the preamble of the code is run, this
is not the case and certain information specified in the schema as
required is not actually parsed. Since this is a valid use of `pw.x` our
parser should be able to deal with this.

The schema based XML parser will throw a lot of errors because of the
incomplete tags as it does not check whether it was an initialization
run. To prevent these verbose messages from flooding the log, all parser
error messages are suppressed.


sphuber:
>     * [procedural question] I didn't manage to fully check all the changes you made to `pars_xml/pw/parse.py`, because you moved around quite a bit of code. Should I make the effort or can I trust the tests?

This is always difficult when code "moves", but in this case it is mostly a huge block that go indented because it is now conditional upon the `bands` key being present in the output. The tests should of course cover these kinds of things but are not 100%. Ultimately it is up to you to decide if you trust it or you spot potential pitfalls.

>     * I opened a small PR against your own branch, which I expected to be more of substance, but ended up being very minor. I wanted to tackle #415 , but I couldn't figure it out so I just opened that issue.

I will merge that one first then so it is included here, thanks. I will try and have a look at the issue tomorrow.

PR #418, opened 2019-10-17
Add PdosWorkChain

chrisjsewell:
As it says in the docstring, to generate Total and Partial DoS data, the following calculations need to be run:

- SCF (pw.x), to generate the initial wavefunction
- NSCF (pw.x), to generate eigenvalues with a denser k-point mesh
- Total DoS (dos.x), to generate total densities of state
- Partial DoS (projwfc.x), to generate partial densities of state by projecting wavefunctions onto atomic orbitals

This workchain combines them.

To validate the input parameter dict, I've added the [jsonschema](https://github.com/Julian/jsonschema) package as a dependency. I can remove this if you want, but I do think it's the best way to validate dictionaries.

I think my testing solution for the `WorkChain` is pretty nice, would probably be good to apply it to some of the others :)


chrisjsewell:
Tested in the wild, and this is all working now. From the `test_default` inputs:

```python
fermi_energy = wkchain.outputs.nscf__output_parameters.dict.fermi_energy

xlabel, energy_tdos, xunits = wkchain.outputs.dos__output_dos.get_x()
yvals_tdos = {(n, u): v for n, v, u in wkchain.outputs.dos__output_dos.get_y()}

plt.plot(energy_tdos - fermi_energy, yvals_tdos[('dos', 'states/eV')])
plt.ylabel("DoS (states/eV)")
plt.xlabel("$E-E_f$ (eV)");
```
image

```python
xlabel, energy_pdos, xunits = wkchain.outputs.projwfc__Dos.get_x()
yvals_pdos = {(n, u): v for n, v, u in wkchain.outputs.projwfc__Dos.get_y()}

plt.plot(energy_pdos - fermi_energy, yvals_pdos[('Dos', 'states/eV')],
         label="Total PDOS", linestyle="dashed")

cum_pdos = None
for orbital, pdos, energy in wkchain.outputs.projwfc__projections.get_pdos():
    data =  orbital.get_orbital_dict()
    kind_name = data["kind_name"]
    orbital_name = orbital.get_name_from_quantum_numbers(
        data['angular_momentum'], data['magnetic_number'])
    old_pdos = 0 if cum_pdos is None else cum_pdos
    cum_pdos = pdos if cum_pdos is None else cum_pdos + pdos
    plt.fill_between(energy - fermi_energy, cum_pdos, old_pdos,
                     label=f"{kind_name} {orbital_name}")

plt.legend(ncol=2);
plt.ylabel("Cumulative PDoS (states/eV)")
plt.xlabel("$E-E_f$ (eV)");
```

image

chrisjsewell:
To explain the `serial_clean` option, added in 3f42c9e:

When I was running this for my work, I found that it was taking up massive amounts of memory for a single `WorkChain` (around 300 Gb), blitzing my 1 Tb disk quota on the HPC. Using `serial_clean=True` improves this situation, by cleaning work directories that are no longer necessary at the end of each step, and not running `DosCalculation` and `ProjwfcCalculation` in parallel.

With `serial_clean=False`:

```
2019-10-20 02:11:18 [18100 | REPORT]: [51091|PdosWorkChain|run_scf]: launching SCF PwBaseWorkChain<51094>
2019-10-20 02:11:19 [18101 | REPORT]:   [51094|PwBaseWorkChain|run_calculation]: launching PwCalculation<51097> iteration #1
2019-10-20 02:11:58 [18102 | REPORT]:   [51094|PwBaseWorkChain|_handle_calculation_sanity_checks]: calculation<51097> run with smearing and highest band is occupied
2019-10-20 02:11:58 [18103 | REPORT]:   [51094|PwBaseWorkChain|_handle_calculation_sanity_checks]: BandsData<51100> has invalid occupations: Occupation of 2.0 at last band lkn<0,0,4>
2019-10-20 02:11:59 [18104 | REPORT]:   [51094|PwBaseWorkChain|_handle_insufficient_bands]: PwCalculation<51097> had insufficient bands
2019-10-20 02:11:59 [18105 | REPORT]:   [51094|PwBaseWorkChain|_handle_insufficient_bands]: Action taken: increased the number of bands to 8 and restarting from scratch
2019-10-20 02:11:59 [18106 | REPORT]:   [51094|PwBaseWorkChain|inspect_calculation]: PwCalculation<51097> finished successfully, but sanity check failed, restarting
2019-10-20 02:11:59 [18107 | REPORT]:   [51094|PwBaseWorkChain|run_calculation]: launching PwCalculation<51105> iteration #2
2019-10-20 02:12:28 [18108 | REPORT]:   [51094|PwBaseWorkChain|inspect_calculation]: PwCalculation<51105> completed successfully
2019-10-20 02:12:28 [18109 | REPORT]:   [51094|PwBaseWorkChain|results]: work chain completed after 2 iterations
2019-10-20 02:12:28 [18110 | REPORT]:   [51094|PwBaseWorkChain|on_terminated]: remote folders will not be cleaned
2019-10-20 02:12:30 [18111 | REPORT]: [51091|PdosWorkChain|run_nscf]: launching NSCF PwBaseWorkChain<51118>
2019-10-20 02:12:31 [18112 | REPORT]:   [51118|PwBaseWorkChain|run_calculation]: launching PwCalculation<51121> iteration #1
2019-10-20 02:13:12 [18113 | REPORT]:   [51118|PwBaseWorkChain|inspect_calculation]: PwCalculation<51121> completed successfully
2019-10-20 02:13:12 [18114 | REPORT]:   [51118|PwBaseWorkChain|results]: work chain completed after 1 iterations
2019-10-20 02:13:13 [18115 | REPORT]:   [51118|PwBaseWorkChain|on_terminated]: remote folders will not be cleaned
2019-10-20 02:13:14 [18116 | REPORT]: [51091|PdosWorkChain|run_pdos_parallel]: launching DosCalculation<51128>
2019-10-20 02:13:14 [18117 | REPORT]: [51091|PdosWorkChain|run_pdos_parallel]: launching ProjwfcCalculation<51130>
2019-10-20 02:13:47 [18120 | REPORT]: [51091|PdosWorkChain|results]: workchain successfully completed
2019-10-20 02:13:56 [18121 | REPORT]: [51091|PdosWorkChain|on_terminated]: cleaned remote folders of calculations: 51130 51128 51121 51105 51097
```

With `serial_clean=True`:

```
2019-10-20 02:13:34 [18118 | REPORT]: [51142|PdosWorkChain|run_scf]: launching SCF PwBaseWorkChain<51144>
2019-10-20 02:13:35 [18119 | REPORT]:   [51144|PwBaseWorkChain|run_calculation]: launching PwCalculation<51147> iteration #1
2019-10-20 02:14:21 [18122 | REPORT]:   [51144|PwBaseWorkChain|_handle_calculation_sanity_checks]: calculation<51147> run with smearing and highest band is occupied
2019-10-20 02:14:21 [18123 | REPORT]:   [51144|PwBaseWorkChain|_handle_calculation_sanity_checks]: BandsData<51158> has invalid occupations: Occupation of 2.0 at last band lkn<0,0,4>
2019-10-20 02:14:21 [18124 | REPORT]:   [51144|PwBaseWorkChain|_handle_insufficient_bands]: PwCalculation<51147> had insufficient bands
2019-10-20 02:14:21 [18125 | REPORT]:   [51144|PwBaseWorkChain|_handle_insufficient_bands]: Action taken: increased the number of bands to 8 and restarting from scratch
2019-10-20 02:14:21 [18126 | REPORT]:   [51144|PwBaseWorkChain|inspect_calculation]: PwCalculation<51147> finished successfully, but sanity check failed, restarting
2019-10-20 02:14:21 [18127 | REPORT]:   [51144|PwBaseWorkChain|run_calculation]: launching PwCalculation<51163> iteration #2
2019-10-20 02:14:51 [18128 | REPORT]:   [51144|PwBaseWorkChain|inspect_calculation]: PwCalculation<51163> completed successfully
2019-10-20 02:14:51 [18129 | REPORT]:   [51144|PwBaseWorkChain|results]: work chain completed after 2 iterations
2019-10-20 02:14:51 [18130 | REPORT]:   [51144|PwBaseWorkChain|on_terminated]: remote folders will not be cleaned
2019-10-20 02:14:53 [18131 | REPORT]: [51142|PdosWorkChain|run_nscf]: launching NSCF PwBaseWorkChain<51170>
2019-10-20 02:14:54 [18132 | REPORT]:   [51170|PwBaseWorkChain|run_calculation]: launching PwCalculation<51173> iteration #1
2019-10-20 02:15:37 [18133 | REPORT]:   [51170|PwBaseWorkChain|inspect_calculation]: PwCalculation<51173> completed successfully
2019-10-20 02:15:37 [18134 | REPORT]:   [51170|PwBaseWorkChain|results]: work chain completed after 1 iterations
2019-10-20 02:15:37 [18135 | REPORT]:   [51170|PwBaseWorkChain|on_terminated]: remote folders will not be cleaned
2019-10-20 02:15:41 [18136 | REPORT]: [51142|PdosWorkChain|inspect_nscf]: cleaned remote folders of SCF calculations: 51163 51147
2019-10-20 02:15:41 [18137 | REPORT]: [51142|PdosWorkChain|run_dos_serial]: launching DosCalculation<51180>
2019-10-20 02:16:11 [18138 | REPORT]: [51142|PdosWorkChain|inspect_dos_serial]: cleaned remote folder of DosCalculation<51180>
2019-10-20 02:16:12 [18139 | REPORT]: [51142|PdosWorkChain|run_projwfc_serial]: launching ProjwfcCalculation<51186>
2019-10-20 02:16:41 [18140 | REPORT]: [51142|PdosWorkChain|inspect_projwfc_serial]: cleaned remote folder of ProjwfcCalculation<51186>
2019-10-20 02:16:41 [18141 | REPORT]: [51142|PdosWorkChain|results]: workchain successfully completed
2019-10-20 02:16:46 [18142 | REPORT]: [51142|PdosWorkChain|on_terminated]: cleaned remote folders of calculations: 51186 51180 51173 51163 51147
```

chrisjsewell:
@shuber this PR is now complete, and I have used it successfully to run all the computations I required. If you want to merge, I can also knock up a quick documentation tutorial?

Here's a bonus example of its use, combined with #425, to demonstrate PBE+U Mott insulation in iron sulfides :)

image
image

chrisjsewell:
@sphuber I was just trialing updating my environment with the new commits since the `aiida-core` `v1.0.0b6`. When I run pytest for `aiida-quantumespresso`, they all pass, except the new test in `tests/workflows/test_pdos.py`.

In [PdosWorkChain.define](https://github.com/chrisjsewell/aiida-quantumespresso/blob/e24878a625ac50cf1906ed1bcd4de983aee773c0/aiida_quantumespresso/workflows/pdos/base.py#L177), for the `nscf` namespace, I expose inputs for the `PwBaseWorkChain`, but only include `'pw.metadata.options'` (since I use the `code`, etc from the `scf` namespace). However, I'm now getting an error for not including the `code`. Is this an error in my implementation, or a bug that has been introduced in a recent commit (possibly in aiidateam/aiida-core@9ebb8cc17e2871e196598ef5ee2d6db820a156d8)? 

```python
>       wc_builder.nscf = {'kpoints': generate_kpoints_mesh(4), 'pw': {'metadata': {'options': get_default_options()}}}

tests/workflows/test_pdos.py:72: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../aiida_core_develop/aiida/engine/processes/builder.py:95: in __setattr__
    validation_error = port.validate(value)
//anaconda/envs/aiida_crystal17_b6/lib/python3.6/site-packages/plumpy/ports.py:685: in validate
    validation_error = self.validate_ports(port_values, breadcrumbs_local)
//anaconda/envs/aiida_crystal17_b6/lib/python3.6/site-packages/plumpy/ports.py:740: in validate_ports
    validation_error = port.validate(port_values.pop(name, UNSPECIFIED), breadcrumbs)
//anaconda/envs/aiida_crystal17_b6/lib/python3.6/site-packages/plumpy/ports.py:673: in validate
    message = self.validator(port_values)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

inputs = {'metadata': {'options': {'max_wallclock_seconds': 1800, 'resources': {'num_machines': 1}, 'withmpi': False}}}

    def validate_calc_job(inputs):
        """Validate the entire set of inputs passed to the `CalcJob` constructor.
    
        Reasons that will cause this validation to raise an `InputValidationError`:
    
         * No `Computer` has been specified, neither directly in `metadata.computer` nor indirectly through the `Code` input
         * The specified computer is not stored
         * The `Computer` specified in `metadata.computer` is not the same as that of the specified `Code`
         * The `metadata.options.parser_name` does not correspond to a loadable `Parser` class.
         * The `metadata.options.resources` are invalid for the `Scheduler` of the specified `Computer`.
    
        :raises `~aiida.common.exceptions.InputValidationError`: if inputs are invalid
        """
    
>       code = inputs['code']
E       KeyError: 'code'

../aiida_core_develop/aiida/engine/processes/calcjobs/calcjob.py:46: KeyError
```

sphuber:
Yeah, this was an intentional change. Part of the validation of `CalcJobs` was actually done on the node. This meant if invalid inputs were passed, the process validation would fail, but then storing the node would fail. I though it was more consistent to have this part of the process spec, since the validation on the node was just there for historical reasons. Since then I encountered a similar problem, where I could no longer make `metadata.options.resources` truely optional for the `PwBaseWorkChain` to enable automatic parallelization, so I [opened an issue](https://github.com/aiidateam/aiida-core/issues/3431) to propose splitting the top level validation to the actual specific port.

That still wouldn't really solve your problem here, because the `computer` validation is a joint validation on `code` and `metadata.options.computer`. Here the validation is already being triggered on the builder itself. I am wondering now if validating on value insertion into the builder is a good idea. The `validators` often depend on mulitiple values, so unless you specify everything in one shot, it is going to fail. Maybe we should remove builder validation until when it is submitted?

Then regarding the design of the `WorkChain`, part of the problem here is also that you are reusing the exposed ports of a single workchain into a namespace for another. I see the idea, but I did not really design it for that use. Even if we were to remove the validation on builder insert, it would still break on process submission. Now you _could_ remove the validator completely by doing `spec.inputs['nscf'].validator = None` but that is just working around the problem.

chrisjsewell:
Yeh this also breaks one of my workchains in `aiida-crystal17` (chrisjsewell/aiida-crystal17#28), so I definitely want to sort out what the best way of handling this is.

> Maybe we should remove builder validation until when it is submitted?

It would be a shame to remove validation for individual Ports. But Portnamespace validators, that cross validate against multiple ports, is obviously where it becomes an issue for exposing subsets of Ports. 

An alternative approach to fixing aiidateam/aiida-core#3431 may be to alter the behavior of [`PortNamespace.validate`](https://github.com/aiidateam/plumpy/blob/30005ae1e0fa229e7f80841a5f599ae935f801b7/plumpy/ports.py#L693).
Currently only a `dict` of inputs are pared to the validator, but if also the `Portnamespace` itself was passed to the validator (i.e. `message = self.validator(port_values_clone, self)`), then in 
[`validate_calc_job`](https://github.com/aiidateam/aiida-core/blob/9ebb8cc17e2871e196598ef5ee2d6db820a156d8/aiida/engine/processes/calcjobs/calcjob.py#L31) you could test whether, e.g. `inputs.code` or `metadata.options.resources` are specified as required, and validate accordingly. Something like:

```python
def validate_calc_job(inputs, spec):
    if not spec['metadata']['options'].required:
        return
```

sphuber:
I have actually been working on this on `plumpy`: https://github.com/aiidateam/plumpy/pull/130
Note that there it is the context of all inputs passed to the top level namespace and not so much the spec itself, but we may have to add the spec as well

chrisjsewell:
Yeh I noticed that but, to be smug, I think my solution is better than yours lol.
As is mentioned in aiidateam/plumpy#126, I'm not convinced that having the validation of individual Ports be coupled to other ports, is going to be particularly great design practice, in terms of complexity and reusability. Maintaining these cross port validation checks at the namespace level feels like a clearer design.

sphuber:
I have been thinking about your solution and although in some cases it can be better to rely on the spec than relying on values being passed, it suffers from the same problem. As soon as we allow and people use `validators` to do compound validation across ports, the expose functionality becomes problematic, except if and only if _all_ ports are exposed and on the top level, not in a new subspace. As soon as as either only a sub set is exposed validators cannot assume that the same ports will be present after exposing, and therefore also not that all values that they might expect will be past. Similarly, even when all ports are exposed, but they are exposed in a new namespace, the path of the old ports exposed in the namespace will have changed. The validator however will still expect them at the old path. In your example, for a validator:
```
def validate_calc_job(inputs, spec):
    if not spec['metadata']['options'].required:
        return
```
If the `CalcJob` is now exposed in namespace `pw`, the `options` port will now be in `pw.metadata.options`, so the check in `validate_calc_job` will now fail. The only solution I see is to either remove validators from ports automatically when exposing them, or document and warn people about the danger of exposing ports with validators that do cross port-validation. Final option is to leave things as is and just not allow all of this by keeping the current validator signature

greschd:
As another option to solve the original problem, wouldn't it be possible to just split the validator on the ``CalcJob`` into its constituent parts, and then (instead of ``spec.inputs['nscf'].validator = None``) disable only one part?

In general I think it would be nice to have a way to easily compose validator -- e.g. a ``CompositeValidator`` class, which just runs validators one after the other. You could name the validators, and then a method ``calcjob_validator.disable('resource_validation')`` would return a new [1] ``CompositeValidator`` without that specific validation step.

[1] The ``CompositeValidator`` should be "immutable" to avoid people down the line messing with it

sphuber:
> As another option to solve the original problem, wouldn't it be possible to just split the validator on the `CalcJob` into its constituent parts, and then (instead of `spec.inputs['nscf'].validator = None`) disable only one part?

This was my original plan and is why I opened [this issue](https://github.com/aiidateam/aiida-core/issues/3431) on `aiida-core`. However, two parts of the current `CalcJob.inputs` validator, the `computer` and the `resources`, require cross port information. I opened [this PR](https://github.com/aiidateam/plumpy/pull/130) on `plumpy` to allow the input values at top level be passed down as context, but @muhrin did not like the implementation for now and as described above this gives other problems.

> In general I think it would be nice to have a way to easily compose validator -- e.g. a `CompositeValidator` class, which just runs validators one after the other. You could name the validators, and then a method `calcjob_validator.disable('resource_validation')` would return a new [1] `CompositeValidator` without that specific validation step.
> 
> [1] The `CompositeValidator` should be "immutable" to avoid people down the line messing with it

The problem is that people can always just do `spec.inputs.validator = None`. This is another problem we were discussing. I recently moved the validation of the `CalcJobNode` to the `inputs` namespace validator of the `CalcJob` class. I felt this is a more natural place. Doing this on the node was just for historical reasons. However, this now allows users to remove the validation that `aiida-core` relies on. Of course one could argue if you remove the validation, then pass incorrect inputs and your calculation job crashes, well... what do you expect? But especially Gio is not convinced that this is the right approach.

I think the problem with validators that rely on anything else but their own values, break the intuition that exposed port(namespace)s continue to work as the originals. And I do not really see a way around this. So we are between a rock and a hard place:
* not allow compound validators, therefore limiting validation capabilities of the `PortNamespace`
* allow compound validators that will potentially unexpectedly break when their corresponding ports are exposed.

greschd:
> This was my original plan and is why I opened [this issue](https://github.com/aiidateam/aiida-core/issues/3431) on `aiida-core`. However, two parts of the current `CalcJob.inputs` validator, the `computer` and the `resources`, require cross port information. I opened [this PR](https://github.com/aiidateam/plumpy/pull/130) on `plumpy` to allow the input values at top level be passed down as context, but @muhrin did not like the implementation for now and as described above this gives other problems.

Yes, I think we agree that this validation needs to be at the higher namespace level.

> The problem is that people can always just do `spec.inputs.validator = None`. 

Here I probably wasn't clear enough: I'd still propose that people can change it in this way, but instead it would be (for example)
```
spec.inputs['nscf'].validator = spec.inputs['nscf'].validator.disable('resource_check')
```
assuming the original validator on the ``nscf`` namespace was something like
```
CompositeValidator(check_a=some_func, resource_check=resource_check_func, ...)
```

Another possibility (not sure if good) would be to somehow "formalize" which ports a validator depends on. If the port should exist in the spec but doesn't, it's not an issue with the input, but with the definition of the Process.

sphuber:
> I think this is good design though, a namespace validator **should** never rely on knowledge of _higher_ namespaces, because otherwise it is (a) no longer portable, and (b) prone to unexpected failures if the _higher_ namespaces are changed.

I fully agree with this now. Allowing to pass information to a validator of ports that are higher in the hierarchy might "solve" the initial problem of making it possible to implement resource validation on that port, but it introduces way more problems. Like you say, it is bad design and would break port validation in unexpected ways. So for me we can put that option to bed. I agree that your suggestion of passing the spec to the validator is still useful and we should probably add this. This should already allow us to adapt the `CalcJob` validator to check whether the resource port still exists and is required or otherwise skip the validation.

chrisjsewell:
Ah good 😀 I didn't want to come across to argumentative lol, but really wasn't a fan of the context design

sphuber:
Not at all. I like being shown wrong for the right reasons. Thanks for the feedback

giovannipizzi:
Hi @chrisjsewell also for this would you have time to rebase? Also here @ramirezfranciscof could help, or maybe @flavianojs and/or @mbercx

mbercx:
@chrisjsewell you didn't make any changes on top of the ones I made in the PR into your branch, right? So the work chain is all good for you?

@sphuber since this is blocking for the work for the MARVEL review on the QE app, I'd still like to add this to the next release.

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -0,0 +1,564 @@
...
+- `Quantum ESPRESSO tutorial slides <http://indico.ictp.it/event/7921/session/320/contribution/1261/material/0/0.pdf>`_
+
+.. warning::
+
+    For QE v6.1, there is an issue using ``tetrahedra`` occupations, as is recommended for ``nscf``,
+    and both ``dos.x`` and ``projwfc.x`` will raise errors when reading the xml file
+    (see `this post <https://lists.quantum-espresso.org/pipermail/users/2017-November/039656.html>`_).
+
+"""
+from __future__ import absolute_import
```
sphuber, line 47:
```suggestion
```

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -0,0 +1,564 @@
...
+
+import jsonschema
+from six.moves import map
+
+from plumpy import ProcessSpec  # pylint: disable=unused-import
```
sphuber, line 52: is this needed?

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -0,0 +1,564 @@
...
+    def define(cls, spec):
+        # type: (ProcessSpec) -> None
+        # yapf: disable
+        """Define the process specification."""
+        super().define(spec)
+        spec.input('structure')
+
+        spec.expose_inputs(
+            PwBaseWorkChain,
+            namespace='base_scf',
```
sphuber, line 203: I would call the namespace `scf` in line with the `PwRelax` and `PwBands`

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -0,0 +1,564 @@
...
+            namespace_options={
+                'help': 'Inputs for the `PwBaseWorkChain` to run `scf` and `nscf` calculations.',
+                'validator': validate_base_scf,
+                'required': False,
+                'populate_defaults': False,
+            }
+        )
+        spec.expose_inputs(
+            PwBaseWorkChain,
+            namespace='base_nscf',
```
sphuber, line 214:
```suggestion
            namespace='nscf',
```

aiida_quantumespresso/workflows/pdos.py
mbercx: Aren't the namespaces called `base` and `base_scf` for the `PwRelaxWorkChain`? Perhaps we should rename the `scf` and `bands` namespaces for the `PwBandsWorkChain`  to `base_scf` and `base_bands`? We could also rename those of the `PwRelaxWorkChain`, but then having e.g. `relax` instead of `base` would be confusing since the `relax` namespace for the `PwBandsWorkChain` corresponds to the `PwRelaxWorkChain`. So maybe we can consistently use namespaces of the entry points:

* `PwBaseWorkChain`: `base_<calculation_type>`
* `PwRelaxWorkChain`: `relax` (possibly add relax_type if there are multiple relax types in the work chain)
*  `PwBandsWorkChain`: `bands`
*  etc...

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -0,0 +1,564 @@
...
+        )
+        spec.input(
+            'parameters',
+            valid_type=orm.Dict,
+            serializer=to_aiida_type,
+            validator=validate_dos_parameters,
+            help='Combined input parameters for the `dos.x` and `projwfc.x` calculations.'
+        )
+        spec.expose_inputs(DosCalculation, namespace='dos', exclude=('parent_folder', 'parameters'))
+        spec.expose_inputs(ProjwfcCalculation, namespace='projwfc', exclude=('parent_folder', 'parameters'))
```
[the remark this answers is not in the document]
mbercx: I think this is because these parameters _need_ to be the same for these two calculations? @chrisjsewell correct me if I'm wrong. We could also add a validator to check this instead though, that would indeed be more consistent. There are also the `dos_only` and `proj_only` keys in the `parameters`, so clearly not all of them need to be the same.

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -0,0 +1,564 @@
...
+- Partial DoS (projwfc.x), to generate partial densities of state, by projecting wavefunctions onto atomic orbitals.
+
+Additional functionality:
+
+- Setting ``'align_to_fermi': True`` in the input ``parameters`` node,
+  will ensure that the energy range is centred around the Fermi energy.
+
+Storage memory management:
+
+The wavefunction file(s) created by the nscf calculation can get very large (>100Gb).
```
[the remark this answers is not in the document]
mbercx: Will do 👍

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -0,0 +1,564 @@
...
+        builder.dos.code = dos_code  # pylint: disable=no-member
+        builder.projwfc.code = projwfc_code  # pylint: disable=no-member
+        builder.clean_workdir = orm.Bool(inputs['clean_workdir'])
+        builder.parameters = orm.Dict(dict=inputs['parameters'])
+
+        return builder
+
+    def setup(self):
+        """Initialize context variables that are used during the logical flow of the workchain."""
+        self.ctx.clean_serial = 'serial_clean' in self.inputs and self.inputs.serial_clean.value
```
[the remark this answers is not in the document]
mbercx: Fair point, missed that one. :)

aiida_quantumespresso/workflows/pdos.py
mbercx: Don't think so, just missed this during my review. Will remove!

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -0,0 +1,564 @@
...
+        )
+        spec.input(
+            'clean_workdir',
+            valid_type=orm.Bool,
+            serializer=to_aiida_type,
+            default=lambda: orm.Bool(False),
+            help='If ``True``, work directories of all called calculation will be cleaned at the end of execution.'
+        )
+        spec.input(
+            'test_run',
```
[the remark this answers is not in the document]
mbercx: Agreed 👍

mbercx:
Thanks for the quick review, @sphuber! Just need some feedback on [this comment](https://github.com/aiidateam/aiida-quantumespresso/pull/418#discussion_r591248922) and then I can start making the required changes.

aiida_quantumespresso/workflows/pdos.py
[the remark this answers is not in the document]
greschd: Just FYI, in [this branch](https://github.com/greschd/aiida-quantumespresso/tree/feature/allow_symlink_pp) I've implemented symlinking for `pp.x`. It's depends on https://github.com/aiidateam/aiida-core/pull/4727 though, which is why I haven't opened a PR yet.

The reason why that `aiida-core` change is needed is because it could _add_ files into the `out` directory, so each file needs to be symlinked separately -- instead of symlinking `out` as a whole. I haven't checked, but `dos` and `projwfc` might be similar.

aiida_quantumespresso/workflows/pdos.py
sphuber: Fair point. I meant mostly the `PwBands` here as being a "real" higher level workflow, just like this `PdosWorkChain`. There we chose `scf` and `bands` (could have been `nscf`) which I think are succinct but still convey the entire message. The `PwRelaxWorkChain` is actually a bit of an exception. Ideally, I would have called it `relax` and `scf` for the two `PwBaseWorkChain` namespaces, but the weird thing is that when you then expose the `PwRelaxWorkChain` in a namespace called `relax`, you would have the namespace nesting `relax.relax`. So when looking at the `PwBandsWorkChain`, the `PwRelaxWorkChain` and the one of this PR, I think the relax is the exception and using the behavior of the bands one for the pdos one feels most natural. Also, this works nicely with the `dos` and `projwfc` namespaces. On a side note, they now directly expose the calcjob and I was going to suggest using a base workchain, except they don't exist yet and didn't want to put that effort on this PR. Also these calculations are unlikely to fail and need error handlers, but at some point it might become useful to expose the base workchain instead of the calcjob. Also then just keeping `dos` and `projwfc` would seem nicest to me.

mbercx: Thinking about this further, I still wonder if we shouldn't just use `base_<calculation_type>` consistently. In case a work chain wraps the `PwBandsWorkChain`, we again get a `bands.bands` namespace nesting. Is there a reason not to use `base_<calculation_type>`?

mbercx: Will stick with your suggestion for now, opened #662 to discuss the possibility of adding a consistent namespaces scheme.

aiida_quantumespresso/workflows/pdos.py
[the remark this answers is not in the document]
mbercx: Have adjust this, and made a separate input for the `align_to_fermi` input. I've also made some changes here, since I think setting this to True only makes sense in case the `Emin` and `Emax` are actually provided and not extracted from the `nscf` calculation. I've also added some validation for this point, as well as making sure they are the same for the `dos` and `projwfc` calculations.

mbercx:
@sphuber this is ready for another round of review! I haven't fixed the tests yet, but will do so once we agree on the work chain itself.

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -311,51 +327,53 @@ def get_builder_from_protocol(
...
+        builder.projwfc.metadata = inputs.get('projwfc', {}).get('metadata') # pylint: disable=no-member
 
         return builder
 
     def setup(self):
         """Initialize context variables that are used during the logical flow of the workchain."""
-        self.ctx.clean_serial = 'serial_clean' in self.inputs and self.inputs.serial_clean.value
-        self.ctx.is_test_run = 'test_run' in self.inputs and self.inputs.test_run.value
+        self.ctx.serial_clean = 'serial_clean' in self.inputs and self.inputs.serial_clean.value
+        self.ctx.is_dry_run = 'dry_run' in self.inputs and self.inputs.dry_run.value
```
sphuber, line 355: would suggest `self.ctx.dry_run` to be inline with `serial_clean`

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -0,0 +1,570 @@
...
+            exclude=('clean_workdir', 'pw.structure'),
+            namespace_options={
+                'help': 'Inputs for the `PwBaseWorkChain` of the `nscf` calculation.',
+                'validator': validate_nscf
+            }
+        )
+        spec.expose_inputs(
+            DosCalculation,
+            namespace='dos',
+            exclude=('parent_folder'),
```
sphuber, line 252:
```suggestion
            exclude=('parent_folder',),
```

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -0,0 +1,570 @@
...
+            namespace_options={
+                'help': ('Input parameters for the `dos.x` calculation. Note that the `Emin`, `Emax` and `DeltaE` '
+                         'values have to match with those in the `projwfc` inputs.'),
+                'validator': validate_dos
+            }
+        )
+        spec.expose_inputs(
+            ProjwfcCalculation,
+            namespace='projwfc',
+            exclude=('parent_folder'),
```
sphuber, line 262:
```suggestion
            exclude=('parent_folder',),
```

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -0,0 +1,570 @@
...
+
+class PdosWorkChain(ProtocolMixin, WorkChain):
+    """A WorkChain to compute Total & Partial Density of States of a structure, using Quantum Espresso."""
+
+    @classmethod
+    def define(cls, spec):
+        # yapf: disable
+        """Define the process specification."""
+        super().define(spec)
+        spec.input('structure')
```
sphuber, line 195: could do with some validation. At least a `valid_type` and maybe a `help`?

sphuber (commented):
New structuring and naming is fine with me :+1: just highlighted a few minor things that I noticed while skimming. Please finalize and I will give a final pass

mbercx:
> Please finalize and I will give a final pass

Thanks @sphuber, finalization complete, tests fixed! Let's get this baby merged.

sphuber (approved):
have at 'em

sphuber:
Thanks @chrisjsewell and @mbercx !

chrisjsewell:
better late than never lol 🎉

PR #420, opened 2019-10-18
Enable checking of log messages in parser unit tests

### LINKED ISSUE #419: Add checking of log messages to parser tests

sphuber:
Fixes #419 

The parsers configure a logger on the calculation node, which will
automatically insert `Log` entries in the database when a message is
emitted through that logger. Tests for successful calculations now check
that there are no log messages, whereas tests for failed calculations
verify the opposite, that there are in fact logs attached to the node.

Note that currently the `Log.objects.get_logs_for` method does not
provide a way to filter for log levels. Once this is implemented in
`aiida-core` the tests should be updated to only check for `warning` and
or `error` messages.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -28,6 +28,8 @@ def flatten_inputs(inputs, prefix=''):
 def fixture_environment():
     """Setup a complete AiiDA test environment, with configuration, profile, database and repository."""
     with fixture_manager() as manager:
+        from aiida.common.log import configure_logging
+        configure_logging(with_orm=True)
         yield manager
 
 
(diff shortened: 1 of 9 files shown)


sphuber:
Thanks @d-tomerini . Can you rebase your PR on `develop` and verify that the tests now fail. If that is the case, then correct the mistake in the name `accuracy_index` and push the changes. Then we can merge yours

d-tomerini:
I confirm that I had 5 pw tests failing, and all of them pass after I correct ``accuracy_scf_index``.
Thank you very much for spotting the problem!

PR #422, opened 2019-10-18
Implement `PwCalculationTools` tools entry point

### LINKED ISSUE #421: Add `PwCalculationTools` with `get_scf_accuracy` utility

sphuber:
Fixes #421 

This entry point will automatically make the methods of the
`PwCalculationTools` available to all `ProcessNodes` created by running
a `PwCalculation`.

### DIFF
--- a/aiida_quantumespresso/tools/calculations/pw.py
+++ b/aiida_quantumespresso/tools/calculations/pw.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8 -*-
+"""Tools for nodes created by running the `PwCalculation` class."""
+from __future__ import absolute_import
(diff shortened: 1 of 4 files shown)


sphuber:
This entry point will allow us to do:
```
pw_calc = load_node()
pw_calc.tools.get_scf_accuracy(2)
```
to get the SCF accuracies for the third cycle in let's say a relax run.

sphuber:
On being allowed to use negative indices, I like it so I will add that. And this was exactly what I was referring to earlier that we might not need the redundant `scf_accuracy_index` but not just for users, but also for the implementation of this function. I will look how "ugly" it will be though when using `scf_iterations` instead.

PR #424, opened 2019-10-19
Add default test for ProjwfcParser, using standard Si calculation

chrisjsewell:
Results, taken from Si calculation, using `Si.pbe-n-kjpaw_psl.1.0.0.UPF`.

I also added an `ERROR_PARSING_PROJECTIONS` exit code, so that the parser fails 'gracefully' for known exceptions in  `_parse_bands_and_projections`.


sphuber:
Looking good @chrisjsewell thanks a lot. The only thing that would be good to change is to see if we can reduce the size of the test files. There are thousands of lines being added and if we start doing this for each test, the repository is going to bloat enormously. Typically what I do is attempt to really strip down the test files to their bare essentials. One can do this either by running a calculation with certain parameters that produce very little data. In this case for example use enormous deltaE and small energy window. This will reduce the files from having 2000 lines to let's say 10. Times this with the number of files and we reduce the load greatly. For the functionality of testing parser this doesn't change anything, whether it parses 10 or 2000 lines. Would this be possible to do?

chrisjsewell:
> The only thing that would be good to change is to see if we can reduce the size of the test files.

Yep good point, all sorted in 4551235 (I've got a script setup to run #418, so easy peasy)

PR #428, opened 2019-10-23
Add parser for pp.x

### LINKED ISSUE #461: `PpCalculation` also needs to copy `pseudo` folder
This is required for certain post-processing modes but is currently not added to the `remote_copy_list`

### LINKED ISSUE #499: Verify units lookup table for pp.x parser
It's desirable to add the units when parsing the output of pp.x. However, many different quantities can be computed and plotted. This functionality is implemented via a lookup table which looks up the correct units for a given plot. Currently, only the quantities for which the units are clear have units. The rest return empty strings. 

To resolve this issue, one must:
1. Verify that the currently specified units are definately correct.
2. Add missing units.

The information is not clearly documented in the QE docs, so this requires specialist knowledge and/or a deep dive into the QE source code.

ConradJohnston:
Fixes #461 and fixes #499  

One challenge of pp calculations is that there is a choice of both dimensionality and of output format. As we want to produce AiiDA ArrayData output nodes, the pp CalcJob class is modified to enforce only Gnuplot (for 1D and 2D) and Cube (3D only) file formats, based on the dimensionality the user wants. The pp calculation class is still lightweight in the sense that the presence of 'iflag' (dimensionality) dependant required parameters are not detected automatically, but is improved over the previous 'free-form' input version in that the output will definitely be parsed by AiiDA and stored in the DB in a standard way. The parser collects the useful data from standard out and detects common problems.

Some things to look at during review:
- For 1D data, an XyData node is created, but for 2D and 3D an ArrayData node is created. I like the features of XyData (like get_x(), get_y()) but I think for the sake of consistency, it might be  better to use ArrayData in every case. I think it could be annoying for developers to have to remember to treat, or worse treat programmatically, the 1D case differently.
- I reuse the emit_logs() function from the Ph parser. It might be worth making into a common utility function and just importing it into the Ph parser and wherever else.


sphuber:
Thanks a million @ConradJohnston . I am a bit swamped right now with the release of `aiida-core==1.0.0` and papers, but I hope to be able to give this a look soon. Anyway we will include this in the release that follows `aiida-core==1.0.0`

ConradJohnston:
> Thanks a million @ConradJohnston . I am a bit swamped right now with the release of `aiida-core==1.0.0` and papers, but I hope to be able to give this a look soon. Anyway we will include this in the release that follows `aiida-core==1.0.0`

Beautiful. 
One thought about the versioning for you to chew on in the meantime- I've added a number of reserved keywords to the PP CalcJob (before it was basically totally free-form) so this PR is likely to backwards-incompatible with previous uses of PpCalculation. However, the PpCalculaton was quite limited before and with no parser, so I wonder how widely it is actually used 'in the wild'.

greschd:
I'm trying out this branch, and one issue I've encountered is that the `pseudo` directory from the `parent_folder` is not copied over.

Totally possible I'm doing something wrong (haven't used pp.x before), just thought I'd let you know.

ConradJohnston:
> I'm trying out this branch, and one issue I've encountered is that the `pseudo` directory from the `parent_folder` is not copied over.
> 
> Totally possible I'm doing something wrong (haven't used pp.x before), just thought I'd let you know.

Hi Dominik,
thanks for the feedback! 
In principle, I think this is fine as pp.x only postprocesses gridded data from PWSCF. So, on this basis, it will never use the pseudos as it only transforms the results of an existing calculation.

Let me know if there's any other oddities.

greschd:
Not sure why, but when trying to plot total potential, pp.x crashed complaining about missing potentials. Maybe it needs to add core charges?

ConradJohnston:
Hmm, could you share input/output?

It might be that in this case (plot_num = 1 ?) the pseudopotential is
needed in order to reconstruct the core charge density and add it, as you
say, to the valence density.

greschd:
So this was a bit of a facepalm on my side... the parent calculation was run with `disk_io='none'`. 

Anyway, it's a bit curious that `pp.x` complains about missing pseudo-potentials _before_ it complains about the missing charge density file. Maybe we can come up with a better error message / exit status for this kind of issue, but it doesn't have to be in this PR.

greschd:
FYI: I've used this branch for a while now, and haven't found any other issues (definitely not using the full set of possible inputs, though).

One feature I'd think would be convenient is adding unit information to the output data (could be later in a separate PR, of course).

aiida_quantumespresso/parsers/pp.py
```diff
@@ -0,0 +1,332 @@
...
+        # Note the unusual indexing: cube files run over the z index first, then y and x.
+        # E.g. The first volumetric data point is x,y,z = (0,0,0) and the second is (0,0,1)
+        for i in range(0, xdim):
+            for j in range(0, ydim):
+                for k in range(0, zdim):
+                    data_array[i, j, k] = (datalist[(i*ydim*zdim)+(j*zdim)+k])
+
+        arraydata = orm.ArrayData()
+        arraydata.set_array('voxel', voxel_array)
+        arraydata.set_array('dimensions', dimensions_array)
```
greschd: I'm not sure if it's necessary to set the `dimensions` explicitly as an array - is there a case where it's different from the shape of the `data` array? Otherwise, it will be the same as the content of the `array|data` attribute.

sphuber: Correct, it can be removed

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,240 @@
...
+        if parser_name is not None:
+            Parserclass = ParserFactory(parser_name)
+            parser = Parserclass(self)
+            try:
+                parser_opts = parser.get_parser_settings_key().upper()
+                settings.pop(parser_opts)
+            except (KeyError, AttributeError):
+                # the key parser_opts isn't inside the dictionary,
+                # or the parser doesn't have a method get_parser_settings_key().
+                pass
```
sphuber: I don't think your parser supports any options,  so best to remove this

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,240 @@
...
+        calcinfo.uuid = str(self.uuid)
+        calcinfo.codes_info = [codeinfo]
+        calcinfo.local_copy_list = local_copy_list
+        calcinfo.remote_copy_list = remote_copy_list
+
+        # Retrieve by default the output file and the xml file
+        calcinfo.retrieve_list = []
+        calcinfo.retrieve_list.append(self.inputs.metadata.options.output_filename)
+        settings_retrieve_list = settings.pop('ADDITIONAL_RETRIEVE_LIST', [])
+        calcinfo.retrieve_list += settings_retrieve_list
```
sphuber: Does this really make sense for pp.x or is this just copied from another calcjob implementation. If not necessary I prefer to remove it, and then also the `settings` input node as this is the only usage as far as I can tell

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,240 @@
...
+                        parent_calc_folder.uuid,
+                        filename,
+                        os.path.join(self._OUTPUT_SUBFOLDER, filename)
+                    ))
+            elif isinstance(parent_calc_folder, orm.SinglefileData):
+                single_file = parent_calc_folder
+                local_copy_list.append(
+                    (single_file.uuid, single_file.filename, single_file.filename)
+                )
+
```
sphuber: according to your spec this input cannot be a `SingleFileData`

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,240 @@
...
+        spec.exit_code(410, 'ERROR_OUTPUT_STDOUT_NOT_READ',
+            message='The stdout output file could not be read.')
+        spec.exit_code(411, 'ERROR_OUTPUT_DATAFILE_NOT_READ',
+            message='The formatted data output file could not be read.')
+        spec.exit_code(420, 'ERROR_OUTPUT_STDOUT_INCOMPLETE',
+            message='The stdout output file was incomplete.')
+        spec.exit_code(430, 'ERROR_UNSUPPORTED_DATAFILE_FORMAT',
+            message='The data file format is not supported by the parser')
+        spec.exit_code(500, 'ERROR_UNEXPECTED_PARSER_EXCEPTION',
+            message='The parser raised an unexpected exception.')
```
sphuber: I homogenized the exit codes across all calculation jobs in PR #479 . Please take a look at, especially `PwCalculation` and match exit code statuses and names.

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,240 @@
...
+        if dimension == 0 or dimension == 1:
+            parameters['PLOT']['output_format'] = 0    # Gnuplot, 1D
+        elif dimension == 2 or dimension == 4:
+            parameters['PLOT']['output_format'] = 7    # Gnuplot, 2D
+        elif dimension == 3:
+            parameters['PLOT']['output_format'] = 6    # Gaussian cube, 3D
+        else:
+            raise exceptions.InputValidationError(
+                "Unsupported value, {}, for 'plot_number' ('iflag' in pp.x docs)".format(dimension)
+            )
```
sphuber: if we keep `dimension` as a separate input, it is better to put the validation on the port itself

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,240 @@
...
+                if key in parameters[namelist]:
+                    raise exceptions.InputValidationError(
+                        "You cannot specify explicitly the '{}' key in the '{}' "
+                        'namelist.'.format(key, namelist))
+            else:
+                parameters[namelist] = {}
+            parameters[namelist][key] = value
+
+        # Set the values that have placeholders:
+        parameters['INPUTPP']['plot_num'] = self.inputs.plot_number.value
```
sphuber: does this need input validation? Also maybe best on the input port in the spec

aiida_quantumespresso/parsers/pp.py
```diff
@@ -0,0 +1,332 @@
...
+
+        # Detect any issues and detect job completion
+        for line in stdout_lines:
+            detect_important_message(logs, line)
+
+        # Parse useful data from stdout
+        for line in stdout_lines:
+            if 'Check:' in line:
+		if 'negative/imaginary' in line:    # QE6.1
+                    output_dict['negative_core_charge'] = float(line.split('=')[-1].split()[0])
```
sphuber: small optimizations in these lines is too call `line.split()` once and then use that instead of calling multiple times

aiida_quantumespresso/parsers/pp.py
```diff
@@ -0,0 +1,332 @@
...
+            coords = []
+            data = []
+            data_integral = []
+            for line in data_lines:
+                split_line = line.split()
+                coords.append(float(split_line[0]))
+                data.append(float(split_line[1]))
+            y_data_lists = [data]
+            y_array_names = ['data']
+            y_array_units = ['']
```
sphuber: is there no way of knowing the units?

greschd: AFAIK the units are determined by the input `plot_num` (different units for different kinds of calculations) - so we would just need a lookup table.

greschd:
@sphuber @ConradJohnston is the goal to include this in the upcoming 3.0 release? If needed, I can help resolve some of the outstanding issues.

sphuber:
Yes, ideally I would like to include this in the 3.0 release which I want to release this week. @ConradJohnston if you don't have the time for this, please let us know and I can take over from here with @greschd

ConradJohnston:
I'll get looking at these today.
Cheers!

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,240 @@
...
     def define(cls, spec):
         # yapf: disable
         super(PpCalculation, cls).define(spec)
-        spec.input('parent_folder', valid_type=(RemoteData, FolderData),
+        spec.input('parent_folder', valid_type=(orm.RemoteData, orm.FolderData), required=True,
             help='Output folder of a completed `PwCalculation`')
+        spec.input('plot_number', valid_type=orm.Int, required=True,
+            help='Plot number to compute')
+        spec.input('plot_dimension', valid_type=orm.Int, required=True,
+            help='Plot dimension to compute')
```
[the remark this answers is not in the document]
ConradJohnston: I don't recall the motivation other than to make it super straightforward to run the calculation - but happy to remove this and simplify.

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,240 @@
...
+            help='Use a node that specifies the input parameters for the namelists')
+        spec.input('settings', valid_type=orm.Dict, required=False,
+            help='Use an additional node for special settings')
+        spec.input('metadata.options.input_filename', valid_type=six.string_types, default=cls._DEFAULT_INPUT_FILE)
+        spec.input('metadata.options.output_filename', valid_type=six.string_types, default=cls._DEFAULT_OUTPUT_FILE)
+        spec.input('metadata.options.parser_name', valid_type=six.string_types, default='quantumespresso.pp')
+        spec.input('metadata.options.withmpi', valid_type=bool, default=True)
+
+        spec.output('output_parameters', valid_type=orm.Dict)
+        spec.output('output_data', valid_type=(orm.XyData, orm.ArrayData))
```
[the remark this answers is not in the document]
ConradJohnston: Glad you share this concern. I'll stick with ArrayData for all outputs.

aiida_quantumespresso/calculations/pp.py
ConradJohnston: Indeed it cannot. This was a copy-pasted stanza that I'll remove. Well spotted.

aiida_quantumespresso/calculations/pp.py
ConradJohnston: Matched up now.

aiida_quantumespresso/calculations/pp.py
ConradJohnston: Now we only have a 'parameters' input port and we do the main validation in the body of the class.

aiida_quantumespresso/calculations/pp.py
ConradJohnston: Dropped the separate input.

aiida_quantumespresso/calculations/pp.py
ConradJohnston: Done

aiida_quantumespresso/calculations/pp.py
ConradJohnston: Agreed.

ConradJohnston: Actually the only part of setting that really makes sense is that the output subfolder in the parent may be called something other than "./out"

greschd: AFAICT the output subfolder is not user-definable, it's hard-coded in the `cls._OUTPUT_SUBFOLDER` variable.

sphuber:
@ConradJohnston let me know when you are done with the fixes and I will give it a second pass.

ConradJohnston:
> @ConradJohnston let me know when you are done with the fixes and I will give it a second pass.

Good for a second pass, apologies for the delay.

Key changes:
1. Removed extra input ports - everything set in the parameters dict.
2. Removed settings node - it's not used in polite company.
3. Extra input validation
4. Now copies the pseudo folder
5. Updated the exit codes to conform.
6. Grabs the fixed folder names from the PW `BasePwCpInputGenerator` class. This is maybe a bit controversial because these are protected attributes in a different class. 
7. Added extra parser logic to deal with the polar coordinates case that wasn't handled 
before.
8. Added test for polar case
9. Slimmed down the datafiles to the minimum.

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,218 @@
...
 
-from aiida.orm import RemoteData, FolderData
-from aiida_quantumespresso.calculations.namelists import NamelistsCalculation
+import os
+import six
+from six.moves import range
 
+from aiida import orm
+from aiida.common import datastructures, exceptions
+from aiida.engine import CalcJob
```
sphuber, line 11: obsolete import

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,218 @@
...
+            else:
+                parameters[namelist] = {}
+            parameters[namelist][key] = value
+
+        # Check for essential keys
+        try:
+            plot_num = parameters['INPUTPP']['plot_num']
+            dimension = parameters['PLOT']['iflag']
+        except KeyError:
+            raise exceptions.InputValidationError("Both 'plot_num' and 'iflag' must be set")
```
sphuber, line 126: since they are in different namelists, would it maybe be better to add those explicitly in the exception message?

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,218 @@
...
+            parameters['PLOT']['output_format'] = 0    # Gnuplot, 1D
+        elif dimension == 2:
+            parameters['PLOT']['output_format'] = 7    # Gnuplot, 2D
+        elif dimension == 3:
+            parameters['PLOT']['output_format'] = 6    # Gaussian cube, 3D
+        else:
+            raise exceptions.InputValidationError(
+                "Unsupported value, {}, for 'iflag'. Must be in the range 0 to 4".format(dimension)
+            )
+
```
sphuber, line 151: These validation rules can be combined, no?

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,218 @@
...
+                parent_calc_folder.computer.uuid,
+                os.path.join(parent_calc_folder.get_remote_path(), self._INPUT_SUBFOLDER),
+                self._OUTPUT_SUBFOLDER
+            ))
+            remote_copy_list.append((
+                parent_calc_folder.computer.uuid,
+                os.path.join(parent_calc_folder.get_remote_path(), self._PSEUDO_SUBFOLDER),
+                self._PSEUDO_SUBFOLDER
+            ))
+        elif isinstance(parent_calc_folder, orm.FolderData):
```
sphuber, line 188: This is just an FYI: this solution for `FolderData` nodes is not comprehensive I am afraid, because it does not deal with nested folders. It is fine to leave this as is, because the required solution is messy and else where in the plugin these cases are also not dealt with correctly. I also opened a feature request on `aiida-core` to improve this use-case.

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,218 @@
...
+                    os.path.join(self._PSEUDO_SUBFOLDER, filename)
+                ))
+
+        codeinfo = datastructures.CodeInfo()
+        codeinfo.stdin_name = self.inputs.metadata.options.input_filename
+        codeinfo.stdout_name = self.inputs.metadata.options.output_filename
+        codeinfo.code_uuid = self.inputs.code.uuid
+
+        calcinfo = datastructures.CalcInfo()
+        calcinfo.uuid = str(self.uuid)
```
sphuber, line 207: this line is not necessary. Is done automatically by AiiDA as of v1.0

aiida_quantumespresso/parsers/pp.py
```diff
@@ -0,0 +1,383 @@
...
+""" Parser for pp.x calculations """
+from __future__ import absolute_import
+
+import traceback
+
+import numpy as np
+
+from aiida import orm
+from aiida.common import exceptions
+from aiida.parsers import Parser
```
sphuber, line 11: please also subclass this from the local base class i.e. `from .base import Parser`. This will give certain useful base functionality. I will point out below in the code what you can reuse

aiida_quantumespresso/parsers/pp.py
```diff
@@ -0,0 +1,383 @@
...
+        arraydata = orm.ArrayData()
+        arraydata.set_array('voxel', voxel_array)
+        arraydata.set_array('data', data_array)
+        arraydata.set_array('coordinates_units', np.array(coordinates_units))
+        arraydata.set_array('data_units', np.array(data_units))
+
+        return arraydata
+
+
+    def emit_logs(self, *args):
```
sphuber, line 350: This is now implemented on the base class `aiida_quantumespresso.parsers.base.Parser` and so you can remove it here

greschd:
As commented by @giovannipizzi [here](https://github.com/aiidateam/aiida-quantumespresso/pull/445#discussion_r348315540) there is a convention to convert to eV (and probably Angstrom?). I'm not sure if we should follow that here, also.

On the one hand, it's nice to be consistent within `aiida-quantumespresso`. But it could also be confusing to have different units than `pp.x` itself.

Opinions @sphuber @ConradJohnston @giovannipizzi?

ConradJohnston:
> As commented by @giovannipizzi [here](https://github.com/aiidateam/aiida-quantumespresso/pull/445#discussion_r348315540) there is a convention to convert to eV (and probably Angstrom?). I'm not sure if we should follow that here, also.
> 
> On the one hand, it's nice to be consistent within `aiida-quantumespresso`. But it could also be confusing to have different units than `pp.x` itself.
> 
> Opinions @sphuber @ConradJohnston @giovannipizzi?

My inclination is to be consistent.
Btw, @greschd, can you sanity check that units dict for me. A lot of these plot modes I've never used and the QE docs and source aren't too helpful.

aiida_quantumespresso/calculations/pp.py
ConradJohnston: Done.

aiida_quantumespresso/calculations/pp.py
ConradJohnston: I separated them into different try/excepts just.

aiida_quantumespresso/calculations/pp.py
ConradJohnston: Yes, combined into a lookup dict to set `output_format`. The actual validation of `iflag` was already done.

aiida_quantumespresso/calculations/pp.py
ConradJohnston: Just this singlular line?
`calcinfo.uuid = str(self.uuid)
`

aiida_quantumespresso/parsers/pp.py
ConradJohnston: Grand.

aiida_quantumespresso/parsers/pp.py
ConradJohnston: Dropped.

aiida_quantumespresso/parsers/pp.py
```diff
@@ -0,0 +1,348 @@
...
+    """
+    Parser for pp.x calculations
+    """
+
+    # Lookup for units
+    units_dict = {
+        0: 'e/bohr^3', # Electrons, electronic charge density
+        1: 'Ry', # Total potential
+        2: 'Ry', # Ionic potential
+        3: 'states/eV.bohr^3', # Density of states over an energy range
```
greschd, line 29: I think this is `states/Ry.bohr^3`

From the pp.x docs:
```
   3  = local density of states at specific energy or grid of energies
        (number of states per volume, in bohr^3, per energy unit, in Ry)
```

ConradJohnston: Beautiful

greschd:
To be honest, I also haven't used most of the plotting kinds - but I think we should do some more checking on these units (maybe ask someone who knows).

aiida_quantumespresso/parsers/pp.py
```diff
@@ -0,0 +1,348 @@
...
+        4: 'states/eV.bohr^3', # Local density of electronic entropy
```
greschd, line 30: Well, what is entropy in Rydberg units? Should be energy per temperature, right? So.. maybe Ry/K.bohr^3?

aiida_quantumespresso/parsers/pp.py
```diff
@@ -0,0 +1,348 @@
...
+        10: 'states/bohr^3', # Integrated local density of states (ILDOS)
+        11: 'Ry', # Bare + hartree potential
+        12: 'Ry', # Electric field
+        13: 'mu_B', # Noncollinear magnetisation, Bohr magnetons
+        17: 'e/bohr^3', # All electron charge density
+        18: 'T', # The exchange and correlation magnetic field in the noncollinear case
+        19: 'e/bohr^3', # Reduced density gradient - see dx.doi.org/10.1021/ct100641a, Eq.1
+        20: 'e/bohr^3', # Product of the electron density and the second eigenvalue of the electron-density Hessian matrix, see: dx.doi.org/10.1021/ct100641a
+        21: 'e/bohr^3', # All electron charge density, PAW case
+        22: 'Ry', # Kinetic energy density
```
greschd, line 45: I'd have guessed `Ry/bohr^3` here..

ConradJohnston: Agreed

aiida_quantumespresso/parsers/pp.py
```diff
@@ -0,0 +1,348 @@
...
+        3: 'states/eV.bohr^3', # Density of states over an energy range
+        4: 'states/eV.bohr^3', # Local density of electronic entropy
+        5: 'states/eV.bohr^3', # Simulated STM images from LDOS
+        6: 'e/bohr^3', # Spin density
+        7: 'e/bohr^3', # WFN contribution to charge density, assuming collinear spins
+        8: '1', # Electron localization function, dimensionless
+        9: 'e/bohr^3', # Charge density minus superposition of atomic densities
+        10: 'states/bohr^3', # Integrated local density of states (ILDOS)
+        11: 'Ry', # Bare + hartree potential
+        12: 'Ry', # Electric field
```
greschd, line 38: Here I think the unit is right, but the comment should be "the sawtooth electric field _potential_"

ConradJohnston: Agreed

greschd:
For the cases where we can't reliably figure out the units, it would be better to _not_ set the units IMO.

aiida_quantumespresso/parsers/pp.py
ConradJohnston: > To be honest, I also haven't used most of the plotting kinds - but I think we should do some more checking on these units (maybe ask someone who knows).

Yes, looking again, I'm seeing that it's a bit of a mess from rushing.

ConradJohnston: > Well, what is entropy in Rydberg units? Should be energy per temperature, right? So.. maybe Ry/K.bohr^3?

I'll make it this for now

ConradJohnston:
> For the cases where we can't reliably figure out the units, it would be better to _not_ set the units IMO.

Yes, this is the safe thing to do, with an issue for the rest.

aiida_quantumespresso/parsers/pp.py
```diff
@@ -0,0 +1,348 @@
...
+    """
+
+    # Lookup for units
+    units_dict = {
+        0: 'e/bohr^3', # Electrons, electronic charge density
+        1: 'Ry', # Total potential
+        2: 'Ry', # Ionic potential
+        3: 'states/Ry.bohr^3', # Density of states over an energy range
+        4: '', # Local density of electronic entropy - Ry/K.bohr^3?
+        5: 'states/Ry.bohr^3', # Simulated STM images from LDOS
```
greschd, line 31: I'm not sure about this one - does STM have energy resolution, or is it just integrated over an energy window?

greschd:
@ConradJohnston maybe we should write to the QE mailing list about this?

aiida_quantumespresso/parsers/pp.py
ConradJohnston: Not immediately clear to me.
So in pp.x, one cannot specify an energy range. That's only possible for plot_num=3 and 10, but one can specify a `sample_bias` in Ry.

ConradJohnston:
> @ConradJohnston maybe we should write to the QE mailing list about this?

Yes, this is probably the best way.
It perhaps should be an issue for pp.x on the QEF Git

aiida_quantumespresso/parsers/pp.py
greschd: Maybe best to remove it for now.

aiida_quantumespresso/calculations/pp.py
sphuber: yes, since this always has to be done, I just moved the responsibility to `aiida-core`

aiida_quantumespresso/parsers/pp.py
ConradJohnston: Removed.

sphuber:
@ConradJohnston and @greschd what is the final conclusion/status on the units?

ConradJohnston:
> @ConradJohnston and @greschd what is the final conclusion/status on the units?

Received a reply from Paolo on the QE issue confirming the units we had uncertainties about. Going to push the final change imminently.

aiida_quantumespresso/parsers/pp.py
ConradJohnston: Done.

aiida_quantumespresso/parsers/pp.py
ConradJohnston: Units are now looked up from a dictionary.

sphuber:
@ConradJohnston please give me a headsup when you're done with the changes and I can give this a final pass.

ConradJohnston:
@Sebastiaan - good to go!

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,218 @@
...
+        :param folder: an `aiida.common.folders.Folder` to temporarily write files on disk
+        :return: `aiida.common.datastructures.CalcInfo` instance
+        """
+
+        # Put the first-level keys as uppercase (i.e., namelist and card names) and the second-level keys as lowercase
+        if 'parameters' in self.inputs:
+            parameters = _uppercase_dict(self.inputs.parameters.get_dict(), dict_name='parameters')
+            parameters = {k: _lowercase_dict(v, dict_name=k) for k, v in six.iteritems(parameters)}
+        else:
+            parameters = {}
```
sphuber, line 103: since `parameters` is a required input, this conditional can be removed. If we get in `prepare_for_submission` it is guaranteed that `self.inputs.parameters` exists

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,218 @@
...
+        if plot_num in range(23) and plot_num not in [14, 15, 16]: # Must be integer in range 0-22, but not 14-16:
+            parameters['INPUTPP']['plot_num'] = int(plot_num) # If this test passes, we can safely cast to int
+        else:
+            raise exceptions.InputValidationError("'plot_num' must be an integer in the range 0-23")
+
+        # Check for valid plot dimension:
+        if dimension in range(5): # Must be in range 0-4:
+            parameters['PLOT']['iflag'] = int(dimension)
+        else:
+            raise exceptions.InputValidationError("'iflag' must be an integer in the range 0-4")
```
sphuber, line 141: Might it be a good idea to move this to a validator for the `parameters` input? This way a user won't end up with a failed calculation in the database but the problem will already be caught at process input validation. Writing the validator is easy:
```
def validate_parameters(value, ctx):
    try:
        plot_num = value['INPUTPP']['plot_num']
    except KeyError:
        return "parameter `INPUTPP.plot_num` must be explicitly set"


spec.input('parameters', valid_type=orm.Dict, validator=validate_parameters)
```

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,218 @@
...
 
     @classmethod
     def define(cls, spec):
         # yapf: disable
         super(PpCalculation, cls).define(spec)
-        spec.input('parent_folder', valid_type=(RemoteData, FolderData),
+        spec.input('parent_folder', valid_type=(orm.RemoteData, orm.FolderData), required=True,
             help='Output folder of a completed `PwCalculation`')
+        spec.input('parameters', valid_type=orm.Dict, required=False,
+            help='Use a node that specifies the input parameters for the namelists')
```
sphuber, line 56: Why is this not required even if the validation in `prepare_for_submission` seems to make it out as though it is required. I think it probably should be required, no?

aiida_quantumespresso/parsers/pp.py
```diff
@@ -0,0 +1,348 @@
...
+        # Scan logs for known errors
+        if 'ERROR_PARENT_XML_MISSING' in logs['error']:
+            return self.exit_codes.ERROR_PARENT_XML_MISSING
+        if 'ERROR_OUTPUT_STDOUT_INCOMPLETE' in logs['error']:
+            return self.exit_codes.ERROR_OUTPUT_STDOUT_INCOMPLETE
+
+        # Parse the post-processed-data according to what kind of data file was produced
+        if self.output_parameters['output_format'] == 'gnuplot':
+            if self.output_parameters['plot_type'] == '2D polar on a sphere':
+                parsed_data = self.parse_gnuplot_polar(data_raw)
```
sphuber, line 98: I am wondering now if we should maybe move the data file to the `retrieve_temporary_list`. The reason is that these files can be quite big (correct?) and we are parsing it essentially in its entirety into an `ArrayData`. In a sense we are then duplicating the content, because the original raw file is stored in the file repo of the calculation node as well as in a parsed version in the `ArrayData` output node. Do we really need the original raw file if we are storing it as a node as well? If not, then we can retrieve this file in the temporary retrieved folder, which still allows to parse it, but the engine will clean it after parsing and not store it in the repo. Maybe this is too much work for now and if you think it makes sense we can simply open an issue for this. @greschd what are your thoughts?

greschd: Hmm, I'm not sure about this. For my own use it would definitely be fine to just discard it. However, I think the main reason `pp.x` even supports different output formats is so that they can effortlessly be fed into different plotting tools. If we discard the file, the onus is on us to provide compatibility with these tools (from the `ArrayData`).

Maybe a good solution would be to discard the file by default, and provide a setting to keep it?

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,218 @@
...
+
+        # Restrict the plot output to the file types that we want to be able to parse
+        dimension_to_output_format = {
+            0: 0, # Spherical integration -> Gnuplot, 1D
+            1: 0, # 1D -> Gnuplot, 1D
+            2: 7, # 2D -> Gnuplot, 2D
+            3: 6, # 3D -> Gaussian cube
+            4: 0, # Polar on a sphere -> # Gnuplot, 1D
+        }
+        parameters['PLOT']['output_format'] = dimension_to_output_format[dimension]
```
greschd, line 151: Somewhat related to the data file discussion below: If we expect that people use the "raw" output file it would make sense to allow manually specifying the `output_format`, and just parse only the ones we understand.

That would be in the spirit of "allow everything the code itself can do", but I'd also be fine with keeping this as it is for now. If there is a need for different output formats, it's a relatively straightforward (and backwards-compatible) change.

aiida_quantumespresso/parsers/pp.py
sphuber: I think the discarding by default and providing a setting to override would be the best solution. But we can do this in a separate PR so that we can merge this one soon

greschd: Shouldn't we change this PR to _always_ discard then, so that the change of adding a setting is backwards compatible?

greschd:
Also fixes #499, correct?

aiida_quantumespresso/calculations/pp.py
ConradJohnston: What should the parser do in the first instance if it doesn't understand the format?
Just warn the user?

aiida_quantumespresso/parsers/pp.py
ConradJohnston: These can be GBs easily for a dense grid in a big supercell.  I'd agree that if we parse, we don't need the original. This is loosely analogous to the argument over what to do with MD trajectories.

aiida_quantumespresso/calculations/pp.py
ConradJohnston: Yep, good spot

aiida_quantumespresso/calculations/pp.py
ConradJohnston: Agreed.

aiida_quantumespresso/calculations/pp.py
ConradJohnston: Done.

aiida_quantumespresso/calculations/pp.py
greschd: Yeah, I guess either a `report` or `warning` - level message.

But again, I'd also be fine with just discarding and not allowing explicit `output_format` for now.

Long-term it might actually be nicer to have tools to convert `ArrayData` to whatever output format is needed - it seems silly to couple the storage to the visualization program.

ConradJohnston:
> Also fixes #499, correct?

Sure does.

aiida_quantumespresso/calculations/pp.py
ConradJohnston: I'd be inclined to leave this for the future. The idea of a warning or report is fine, but I think the argument against this is similar to what @sphuber says in his report - the user will only discover this after they've created the nodes in the DB and the calculation has effecively failed in that it did something unexpected.

Very supportive of an export module for common plotting/visualisation tools - this could be an `aiida-core` feature.

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,218 @@
...
+    if plot_num in range(23) and plot_num not in [14, 15, 16]:  # Must be integer in range 0-22, but not 14-16:
+        value['INPUTPP']['plot_num'] = int(plot_num)  # If this test passes, we can safely cast to int
+    else:
+        raise exceptions.InputValidationError("'plot_num' must be an integer in the range 0-23")
+
+    # Check for valid plot dimension:
+    if dimension in range(5):  # Must be in range 0-4:
+        value['PLOT']['iflag'] = int(dimension)
+    else:
+        raise exceptions.InputValidationError("'iflag' must be an integer in the range 0-4")
```
sphuber, line 41: you shouldn't raise here but just return the message

aiida_quantumespresso/parsers/pp.py
ConradJohnston: So as it stands, when one asks pp.x to write out to file in particular format, pp.x produces two files: 1. the 3D gridded quantity in a custom format, and 2. the post-processed data converted into a particular format, and reduced to the dimension requested by `iflag`. 
In the most recent implementation, we never use the first file, and so there is no need to retrieve it. The second, the file in a format we choose according to what we can parse, we temporarily retrieve, parse and discard.

aiida_quantumespresso/calculations/pp.py
ConradJohnston: Grand. Fixed.

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,221 @@
...
+        calcinfo.local_copy_list = local_copy_list
+        calcinfo.remote_copy_list = remote_copy_list
+
+        # Retrieve by default the output file and plot file
+        calcinfo.retrieve_list = []
+        calcinfo.retrieve_list.append(self.inputs.metadata.options.output_filename)
+        if self.inputs.metadata.options.keep_plot_file:
+            calcinfo.retrieve_list.append(self._FILEOUT)
+        else:
+            calcinfo.retrieve_temporary_list = [self._FILEOUT]
```
ConradJohnston, line 219: @sphuber  - Maybe you have some insight - this doesn't seem to do what I would expect. 
Even without the if/else block, all files are retrieved, rather than just those specified, as if retrieve_list is being ignored.

aiida_quantumespresso/calculations/pp.py
sphuber: Sorry that I wasn't more clear, but it should return just a string not an exception instance

aiida_quantumespresso/calculations/pp.py
```diff
@@ -1,27 +1,221 @@
...
+from aiida_quantumespresso.utils.convert import convert_input_to_namelist_entry
+from .base import CalcJob
+
+
+def validate_parameters(value, ctx):  # pylint: disable=unused-argument
+    """ Validate 'parameters' dict."""
+
+    # Check for essential keys
+    try:
+        plot_num = value['INPUTPP']['plot_num']
```
sphuber, line 22: Also the `value` here will be the actual input value, so it is a `Dict` node. That means you should probably first do `parameters = value.get_dict()` and then do checks on that normal dictionary

aiida_quantumespresso/calculations/pp.py
sphuber: Well for a kickoff, I cannot even find now where you are telling `pp.x` to write to these files. They are blocked keywords, so the plugin should add them to the parameters, correct? Something like
```
parameters = self.inputs.parameters.get_dict()
parameters['INPUTPP']['filplot'] = self._FILPLOT
parameters['INPUTPP']['filout'] = self._FILEOUT
```
or am I missing something here?

aiida_quantumespresso/calculations/pp.py
ConradJohnston: Fixed this, sorry. 
Added a test for the calculation class.

aiida_quantumespresso/calculations/pp.py
ConradJohnston: Yes, those variables are set right at the top of the class. and later the relevant keywords added to the  blocked list:
https://github.com/aiidateam/aiida-quantumespresso/blob/7a294c4be22ecaae62618d93aa77f9a7e6a362dd/aiida_quantumespresso/calculations/pp.py#L63-L66

I've added a `PpCalculation` test class. 

Retrieving the files works correctly now also.

sphuber:
@ConradJohnston I fixed the failing test, due to a compatibility issue in the validator signature and then took the liberty to clean up the `PpCalculation` tests and add some additional ones for the parameter validation. I hope you don't mind. I also rebased and the tests now all pass, so for me this would be good to go. Let me know if you agree.

ConradJohnston:
@sphuber, I don't mind at all! Your valuable experience is always welcome. I'm happy to go if you are.

sphuber:
Thanks a lot for the work and your patience @ConradJohnston !

PR #430, opened 2019-10-25
Add negative dexx exit code

### LINKED ISSUE #413: Order of checking for exit codes
I am trying to add an exit code to ``PwCalculation`` for the case where the code crashes with a ``negative dexx`` error. The XML file doesn't exist in that case, and the standard output terminates with 

```
     Error in routine electrons (1):
     dexx is negative!  Check that exxdiv_treatment is appropriate for the system, or ecutfock may be too low

     stopping ...
```

I've added the exit code to the spec (in the calculation class), and in the ``message_map`` (https://github.com/aiidateam/aiida-quantumespresso/blob/develop/aiida_quantumespresso/parsers/parse_raw/pw.py#L34) added ``'dexx is negative': 'ERROR_DEXX_IS_NEGATIVE'`` as an error. In the parser itself, I've added a clause to ``validate_premature_exit`` (https://github.com/aiidateam/aiida-quantumespresso/blob/develop/aiida_quantumespresso/parsers/pw.py#L125).

The issue is that while the error shows up in the log
```
+-> ERROR at 2019-10-16 11:40:32.443091+00:00
 | parser returned exit code<300>: Both the stdout and XML output files could not be read or parsed.
+-> ERROR at 2019-10-16 11:40:32.439404+00:00
 | Both the stdout and XML output files could not be read or parsed.
+-> ERROR at 2019-10-16 11:40:32.433249+00:00
 | ERROR_OUTPUT_STDOUT_INCOMPLETE
+-> ERROR at 2019-10-16 11:40:32.428673+00:00
 | ERROR_DEXX_IS_NEGATIVE
+-> WARNING at 2019-10-16 11:40:32.423150+00:00
 | Error while parsing for energy terms.
```
it's still exit code ``300`` that is returned. I think that's because the check whether both XML and stdout had a failure in parsing happens before the ``validate_premature_exit`` call (https://github.com/aiidateam/aiida-quantumespresso/blob/develop/aiida_quantumespresso/parsers/pw.py#L100). 

Is this correct, or should the order be switched here? Maybe my approach to adding the exit code is wrong?

greschd:
Fixes #413.

Adds an error code (551) for the "negative dexx" error QE sometimes likes to throw when calculating with hybrids / EXX.

Also changes the order in which errors are checked, running the "validate_premature_exit" before checking if there were errors in parsing stdout and XML (which raises a more generic error).


tests/parsers/fixtures/pw/failed_dexx_negative/aiida.out
```diff
@@ -0,0 +1,152 @@
...
+     http://www.quantum-espresso.org/quote
+
+# NOTE: roughly 3870 lines of output removed.
+
+ %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+     Error in routine electrons (1):
+     dexx is negative!  Check that exxdiv_treatment is appropriate for the system, or ecutfock may be too low
+ %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+     stopping ...
```
sphuber: Can you just keep this block and remove all the others, since they are just duplicates

tests/parsers/test_pw.py
```diff
@@ -597,3 +597,29 @@ def test_pw_vcrelax_failed_not_converged_nstep(fixture_database, fixture_compute
     assert 'output_parameters' in results
     assert 'output_structure' in results
     assert 'output_trajectory' in results
+
+def test_pw_hybrid_failed_dexx_negative(fixture_database, fixture_computer_localhost, generate_calc_job_node,
+        generate_parser, generate_inputs_default):
+    """Test the parsing of a calculation that was interrupted *after* convergence was achieved.
```
sphuber: The docstring is incorrect

aiida_quantumespresso/parsers/pw.py
```diff
@@ -103,16 +103,16 @@ def parse(self, **kwargs):
...
         # First check for specific known problems that can cause a pre-mature termination of the calculation
         exit_code = self.validate_premature_exit(logs_stdout)
         if exit_code:
             return self.exit(exit_code)
 
+        # If the both stdout and xml exit codes are set, there was a basic problem with both output files and there
+        # is no need to investigate any further.
+        if self.exit_code_stdout and self.exit_code_xml:
+            return self.exit(self.exit_codes.ERROR_OUTPUT_FILES)
+
```
sphuber: Is this change necessary for your new exit code to work correctly? I remember there being a problem with the order in evaluation, but I would rather deal with this then in separate PR first in order for it to not get lost in the rest of the changes in this PR.

aiida_quantumespresso/calculations/pw.py
```diff
@@ -134,6 +134,8 @@ def define(cls, spec):
         spec.exit_code(541, 'ERROR_SYMMETRY_NON_ORTHOGONAL_OPERATION',
             message='The variable cell optimization broke the symmetry of the k-points.')
 
+        spec.exit_code(551, 'ERROR_DEXX_IS_NEGATIVE', message='The code failed with negative dexx in the exchange calculation.')
```
sphuber: I was reserving the `5**` series for problems related to ionic calculations, such as `relax` and `md`. This is better suited as a 4xx. Since this is an error for hybrid functionals, let's reserve `46x` for those and make this one `461` (use `460` as the generic one in the future if needed).

aiida_quantumespresso/parsers/pw.py
greschd: Yeah, the test will fail otherwise. Should I pull out 296c601d4cd722663012dff9f3e39ea64a9be63f into a separate PR? We could merge it ahead of this one.

tests/parsers/fixtures/pw/failed_dexx_negative/aiida.out
greschd: Sure thing! Do you want to keep the standard QE header? Maybe just the line with the version number?

aiida_quantumespresso/calculations/pw.py
greschd: Looking at this section, I'm not sure if the comment ``# Significant errors but calculation can be used to restart`` applies to this error. Does that influence the choice of exit number, or is ``461`` good anyway?

tests/parsers/fixtures/pw/failed_dexx_negative/aiida.out
sphuber: Yes, let's keep the header but just remove the duplicated error message blocks

aiida_quantumespresso/parsers/pw.py
sphuber: Yeah, since potentially affects the behavior quite a bit, I want to make sure what it does and it doesn't mess with my production runs

aiida_quantumespresso/calculations/pw.py
sphuber: Yeah, the comment is a bit outdated and is not really what I intend anymore. Currently I maintain the following ranges, but that is still moving while we figure things out:
0 -  99: Reserved for `aiida-core` internal
100 - 199: Reserved for scheduler errors (will be implementing this soon, things like OOM, OOW and node failures
200 - 299: Extended input validation error (not sure if this is useful/necessary, but not everything is possible through spec validation, so some may have to be done in the process itself)
300 - 399: Critical errors where things really went tits up. Think any error that the program itself did not write. So typically output files will be missing, corrupt etc. for whatever reason.
400 - \inf: Any other error. Typically these will then be errors known and written by the program itself. Some of those can potentially be corrected by the calling work chain and some not. I am now using sub ranges for different types, e.g. 400 - 499 for electronic type calculations 'scf/nscf/bands' and 500 - 599 for ionic relaxation 'relax/vc-relax'. Mayb at some point 600 - 699 for 'md/vc-md'.

tests/parsers/test_pw.py
greschd: Indeed it is. Fixed now.

tests/parsers/fixtures/pw/failed_dexx_negative/aiida.out
greschd: Alright, done 👍

aiida_quantumespresso/calculations/pw.py
greschd: Makes sense. Should I remove the comment?

aiida_quantumespresso/parsers/pw.py
greschd: Ok, see #439. I'll rebase this PR once that's done.

sphuber: PR is merged, rebase away!

aiida_quantumespresso/calculations/pw.py
sphuber: Let's leave it for now. Pretty soon I will be doing a big update of all calculations to reflect a consistent use of exit codes

aiida_quantumespresso/parsers/pw.py
greschd: Done, ready for another review!

greschd:
Thanks for reviewing!

PR #433, opened 2019-11-01
Improvements to the pw2wannier90 CLI

normarivano:
Some improvements on the pw2wannier CLI. Now  the `single_file` has been renamed into `nnkp_file` (for the sake of clarity). Moreover, the user and can specify the  desired SCDM mode and if UNK matrices are to be written as output.


aiida_quantumespresso/cli/calculations/pw2wannier90.py
```diff
@@ -16,56 +18,75 @@
...
-            'scdm_entanglement': 'isolated',
         }
     }
 
+    if write_unk:
+        parameters['INPUTPP']['write_unk'] = True
+    else:
+        parameters['INPUTPP']['write_unk'] = False
+
+    if scdm_mode != 'no':
```
sphuber: it doesn't seem that the actual value of `scdm_mode` gets used so why does the option specify those other possible values `'isolated', 'erfc', 'gaussian'`?

aiida_quantumespresso/cli/calculations/pw2wannier90.py
```diff
@@ -16,56 +18,75 @@
...
-            'write_unk': False,
-            'scdm_proj': True,
-            'scdm_entanglement': 'isolated',
         }
     }
 
+    if write_unk:
+        parameters['INPUTPP']['write_unk'] = True
+    else:
+        parameters['INPUTPP']['write_unk'] = False
```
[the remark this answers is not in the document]
normarivano: Thank you!

aiida_quantumespresso/cli/calculations/pw2wannier90.py
normarivano: You are right! I've fixed it.

sphuber:
Thanks very much @normarivano

Issue #384, opened 2019-08-26
Backwards incompatible changes introduced in `aiida-quantumespresso==3.0.0`


sphuber:
- `PwCalculation`: the `fermi_energy` is no longer in the `output_parameters` when run with the new XML output format and without smearing
- `ForceconstantsData` has been renamed to `ForceConstantsData` [#386]
- `aiida.data:quantumespresso.forceconstants` entry point renamed to `aiida.data:quantumespresso.force_constants` [#386]
- `Q2rCalculation` output `forceconstants` has been renamed to `force_constants` [#386]
- `MatdynCalculation` input `parent_folder` has been renamed to `force_constants` [#387]

PR #445, opened 2019-11-16
`PwParser`: add `smearing_type` and `smearing` from new XML output

### LINKED ISSUE #415: Smearing type not always present in XML output
The XML output does not always contain the smearing type. In `aiida_quantumespresso/parsers/parse_xml/pw/parse.py`, we assume that the smearing type is always present when `occupations != 'from_input'`, however this is not true. Is this a bug in QE or is our logic incorrect? Even if it is a bug in QE, how can we work around it?

See: https://github.com/aiidateam/aiida-quantumespresso/blob/6fff0c2d241beadd2f9640d284b99692fd1071e4/aiida_quantumespresso/parsers/parse_xml/pw/parse.py#L153-L165

sphuber:
Fixes #415 

These keys were not yet being parsed from the output file with the new
XML schema. In addition, if an exception is thrown by the raw XML or
stdout parser, it is now caught and added as a critical warning. Finally
the test input generator for `PwParser` is simplified and only a single
generator is used for both scf and relax tests.

### DIFF
--- a/aiida_quantumespresso/parsers/parse_xml/pw/parse.py
+++ b/aiida_quantumespresso/parsers/parse_xml/pw/parse.py
@@ -2,13 +2,14 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
+from distutils.version import StrictVersion
 import numpy as np
 from xmlschema import XMLSchema
 from xmlschema.etree import ElementTree
 from xmlschema.exceptions import URLError
 
 from aiida_quantumespresso.utils.mapping import get_logging_container
-from qe_tools.constants import hartree_to_ev, bohr_to_ang
+from qe_tools.constants import hartree_to_ev, bohr_to_ang, ry_to_ev
 
 from .exceptions import XMLParseError
 from .legacy import parse_pw_xml_pre_6_2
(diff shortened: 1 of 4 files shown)


aiida_quantumespresso/parsers/parse_xml/pw/parse.py
```diff
@@ -340,6 +305,18 @@ def parse_pw_xml_post_6_2(xml, include_deprecated_v2_keys=False):
...
+        smearing_xml = None
+
+        if 'smearing' in outputs['band_structure']:
+            smearing_xml = outputs['band_structure']['smearing']
+        elif 'smearing' in inputs:
+            smearing_xml = inputs['smearing']
+
+        if smearing_xml:
+            xml_data['degauss'] = smearing_xml['@degauss']
+            xml_data['smearing_type'] = smearing_xml['$']
```
giovannipizzi: I guess this is one of the first cases where we are really changing the behaviour w.r.t. the old parser, right?

- Can we add already now a list of things that changed between version 2 and 3 (or even more precisely, that depend on how QE was compiled)? Possibly the list is already there, we should then just add also this

- I don't see any unit conversion - normally, in the parsed output, we try to use consistently `eV` for energy

sphuber: I have started a list in issue #384 . The units are not converted because it is not clear what they are. I sent an email for this to Delugas and put you in CC. The docs and file itself say it should be Hartree, but it is clearly Rydberg.

giovannipizzi: Ok, let's add this to the issue, and let's wait for the units (or add conversion from Rydberg since that's what they are and if they fix this, this will be only for future versions)

sphuber: I performed a calculation for 6.4, 6.4.1 and 6.5 and only 6.5 has the correct units. So for now I have just added the correct unit conversion in the parser. Everything 6.5 and above is assumed Hartree, for all version below I assume it is Rydberg. So @giovannipizzi this should be ready for review

PR #447, opened 2019-11-19
Move CI from Travis to Github

### LINKED ISSUE #446: Move CI from Travis to Github

sphuber:
Fixes #446


sphuber:
> I think you'll now first need to change branch protection rules.

Will do once we think this is good idea

> BTW, is the wait-for-it script not needed anymore?

No, since that was necessary for the machine with Torque that was setup through Docker, however, I am no longer using this in the tests. I now exclusively test through `pytest` which does not need this.

sphuber:
@giovannipizzi this also drops python2 which we probably shouldn't as we haven't released a py2 compatible version yet for `aiida-core==1.0.0`. I will add py2 to the matrix

giovannipizzi:
Indeed, better to have a py2+py3 version at least!

PR #451, opened 2020-01-06
`PwParser`: add new exit code `ERROR_OUT_OF_WALLTIME_INTERRUPTED`

### LINKED ISSUE #450: `PwParser` emits 400 even when killed by the scheduler
The `PwParser` just checks for the presence of the string `Maximum CPU time exceeded` in the stdout before it emits a `ERROR_OUT_OF_WALLTIME`. However, after that string, the job can still be killed by the scheduler while it is writing files to disk. In this case one can not restart from this job because the required output files are most likely corrupt. This should get another exit code, or the case of an OOW that does shutdown cleanly should get a different one.

sphuber:
Fixes #450 

If the `max_seconds` parameter is set in the input and this number is
exceeded, Quantum ESPRESSO will attempt to terminate the calculation
cleanly by writing all files to disk that are necessary for a restart.
However, during this phase the scheduler can still kill the job if the
total requested walltime has been exceeded. This will result in the
restart files being corrupt and not suitable for a restart. Therefore
this case needs a different exit code than the currently existing one
`ERROR_OUT_OF_WALLTIME` which should only be used for calculation that
shutdown cleanly.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -61,7 +61,8 @@ disable=bad-continuation,
         logging-format-interpolation,
         no-else-raise,
         useless-object-inheritance,
-        inconsistent-return-statements
+        inconsistent-return-statements,
+        import-outside-toplevel
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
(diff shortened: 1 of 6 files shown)


sphuber:
> Looks good. Do you want to indicate in the `ERROR_OUT_OF_WALLTIME` `message` that in this case the calculation managed to shut down cleanly?

I will leave it for now as is. All >=400 exit codes should be restartable in principle

PR #453, opened 2020-01-10
Set 2xx and 3xx exit codes in PwCalculation to invalidate cache

### LINKED ISSUE #452: Set "invalidates_cache" exit code keyword where appropriate
Since we implemented the "invalidates_cache" keyword on "exit_codes", I think it makes sense to use this here where appropriate.

@sphuber, do you think the 300's exit codes should have that flag? Or only some of them?

An important consideration is that the plugin should still work if the functionality is not present. I think a good way to do this would be something like
```python
if 'invalidates_cache' in ExitCode._fields:
    INVALIDATES_CACHE_INPUT = {'invalidates_cache': True}
else:
    INVALIDATES_CACHE_INPUT = {}
```
and then pass ``**INVALIDATES_CACHE_INPUT``. Where's a good place to put this code? Do we have a compatibility helper module somewhere?

greschd:
Fixes #452 

To maintain compatibility with AiiDA 1.0 (which does not contain the `invalidates_cache` functionality), a compatibility module is defined. When that AiiDA version is used, the behavior remains unchanged.

Marked as draft because it's not tested yet. We can probably just check `is_valid_cache` in the parser tests. Or do you prefer separate tests?


sphuber:
Instead of having to add this `INVALIDATES_CACHE` to all specific exit codes, I was thinking of just overriding the `ProcessSpec.exit_code` method and then sub classing `CalcJob` and `WorkChain` to automatically implement this. Something like:
```
from aiida.engine import ExitCode
from aiida.engine import CalcJob as BaseCalcJob
from aiida.engine import WorkChain as BaseWorkChain
from aiida.engine import ProcessSpec as BaseProcessSpec


class ProcessSpec(BaseProcessSpec):
    """Pass"""

    def exit_code(self, status, label, message, invalidates_cache=False):
        """Add an exit code to the ProcessSpec.

        .. note: for any status smaller than `400` the `invalidates_cache` is automatically set to `True`.

        :param status: the exit status integer
        :param label: a label by which the exit code can be addressed
        :param message: a more detailed description of the exit code
        :param invalidates_cache: when set to `True`, a process exiting
            with this exit code will not be considered for caching
        """
        if isinstance(status, int) and status < 400:
            invalidates_cache = True

        if 'invalidates_cache' in ExitCode._fields:
            super(ProcessSpec, self).(status, message, invalidates_cache=True)
        else:
            super(ProcessSpec, self).(status, message)


class CalcJob(BaseCalcJob):
    _spec_class = ProcessSpec

class WorkChain(BaseWorkChain):
    _spec_class = ProcessSpec
```
I haven't tested it, but I think this should work. Then we should just make sure that we respect our own convention of <400 being uncachable. What do you think?

greschd:
Yeah, I think that makes sense. Are the exit codes of all the other calculation classes already adapted to use this convention?

I also think we probably should still respect explicit override of the ``invalidates_cache`` keyword. We could change the default to ``None`` and then use it if anything else is passed, falling back to checking by status integer. 
If we want to be strict about _only_ invalidating cache for <400, we probably shouldn't accept the ``invalidates_cache`` keyword at all -- that would be preferable to not doing anything with it.

greschd:
@sphuber do you know what this error is about? Some incompatibility in dependency versions, maybe?

```
kwargs = {'help': 'Files that are retrieved by the daemon will be stored in this node. By default the stdout and stderr of the ... in `CalcInfo.retrieve_list`.', 'pass_to_parser': True, 'valid_type': <class 'aiida.orm.nodes.data.folder.FolderData'>}
namespace = [], port_name = 'retrieved'

    def _create_port(self, port_namespace, port_class, name, **kwargs):
        """
            Create a new Port of a given class and name in a given PortNamespace
    
            :param port_namespace: PortNamespace to which to add the port
            :param port_class: class of the Port to create
            :param name: name of the port to create
            :param kwargs: options for the port
            """
        if self.sealed:
            raise RuntimeError('Cannot add an output port after the spec has been sealed')
    
        namespace = name.split(self.namespace_separator)
        port_name = namespace.pop()
    
        if namespace:
            namespace = self.namespace_separator.join(namespace)
            port_namespace = port_namespace.create_port_namespace(namespace)
    
>       port_namespace[port_name] = port_class(port_name, **kwargs)
E       TypeError: __init__() got an unexpected keyword argument 'pass_to_parser'
```

sphuber:
> @sphuber do you know what this error is about? Some incompatibility in dependency versions, maybe?
> 
> ```
> kwargs = {'help': 'Files that are retrieved by the daemon will be stored in this node. By default the stdout and stderr of the ... in `CalcInfo.retrieve_list`.', 'pass_to_parser': True, 'valid_type': <class 'aiida.orm.nodes.data.folder.FolderData'>}
> namespace = [], port_name = 'retrieved'
> 
>     def _create_port(self, port_namespace, port_class, name, **kwargs):
>         """
>             Create a new Port of a given class and name in a given PortNamespace
>     
>             :param port_namespace: PortNamespace to which to add the port
>             :param port_class: class of the Port to create
>             :param name: name of the port to create
>             :param kwargs: options for the port
>             """
>         if self.sealed:
>             raise RuntimeError('Cannot add an output port after the spec has been sealed')
>     
>         namespace = name.split(self.namespace_separator)
>         port_name = namespace.pop()
>     
>         if namespace:
>             namespace = self.namespace_separator.join(namespace)
>             port_namespace = port_namespace.create_port_namespace(namespace)
>     
> >       port_namespace[port_name] = port_class(port_name, **kwargs)
> E       TypeError: __init__() got an unexpected keyword argument 'pass_to_parser'
> ```

Probably. This keyword was added in `aiida-core` in v1.0 or probably one of the betas already. It is so `Parser.parse_from_node` knows which output nodes to pass to the `Parser.parse` function. Don't understand why you would be hitting this though

greschd:
Yeah, I initially thought it was just on my machine.. but this is from the github actions run.

greschd:
Figured out the reason for this error: Instead of the bare `ProcessSpec`, I should have used `CalcJobProcessSpec` and `WorkChainProcessSpec` as base classes, respectively.

@sphuber, I have some design questions for you:
- Since caching is anyway implemented only for `CalcJob`s, does it still make sense to add this machinery also to the `WorkChain`? 
- I don't really like naming the new base classes just `CalcJob` and `WorkChain`, because that hides the fact that custom things are implemented. So for now I renamed it to `QuantumEspressoCalcJob` and `QuantumEspressoWorkChain` - but maybe it could be something more descriptive? What's your opinion on naming?

greschd:
@sphuber, what's your opinion on the two questions above?

sphuber:
Sorry for the delay, must have missed this in the flurry of mentions

> * Since caching is anyway implemented only for `CalcJob`s, does it still make sense to add this machinery also to the `WorkChain

Yes I think it is fine to only implement this for `CalcJobs`.

> * I don't really like naming the new base classes just `CalcJob` and `WorkChain`, because that hides the fact that custom things are implemented. So for now I renamed it to `QuantumEspressoCalcJob` and `QuantumEspressoWorkChain` - but maybe it could be something more descriptive? What's your opinion on naming?

I don't really like the `QuantumEspressoCalcJob` but cannot really think of a better one. Given that we are only going to implement this just for `CalcJob` I would actually be tempted to make `aiida_quantumespresso.calculations.base:CalcJob` which overrides the `aiida.engine.CalcJob` class (i.e. we keep the same name) and just change the imports for all calculations classes to import `from .base import CalcJob` instead of `from aiida.engine import CalcJob`. Since the import is in the same file as the sub classing I think that is more than clear enough.

greschd:
No worries, I had almost forgotten about this, too 😄 I'll get to implementing your suggestions, then.

greschd:
Alright, this is ready for review.

In the tests, I opted to go through all the entry points defined by `aiida-quantumespresso` instead of just a specific set, to also make sure any new `CalcJob` _uses_ the base class. Initially I wanted to test it through `is_valid_cache` on the parser tests, but there I ran into https://github.com/aiidateam/aiida-core/issues/3823, because the `parse_from_node` is a `calcfunction`.

aiida_quantumespresso/calculations/base.py
```diff
@@ -0,0 +1,48 @@
+# -*- coding: utf-8 -*-
+"""Defines a CalcJob base class for aiida-quantumespresso.
+
+The custom CalcJob base class automatically sets the'invalidates_cache'
```
sphuber:
```suggestion
The custom CalcJob base class automatically sets the `invalidates_cache`
```

aiida_quantumespresso/calculations/base.py
```diff
@@ -0,0 +1,48 @@
...
+attribute of exit codes based on the status integer.
```
sphuber: Maybe add also that all `CalcJob` implementation _have to_ subclass from this class and not `aiida.engine`.

tests/calculations/test_autoinvalidate_cache.py
```diff
@@ -0,0 +1,39 @@
...
+    """Test automatic 'invalidates_cache' attribute of exit codes.
+
+    Test that the 'invalidates_cache' attribute of exit codes is automatically
+    set according to the status integer.
+    """
+    calc_class = CalculationFactory(entry_point_name)
+    overrides = EXPLICIT_OVERRIDES.get(entry_point_name, {})
+    for exit_code in calc_class.exit_codes.values():
+        if exit_code.status in overrides:
+            assert exit_code.invalidates_cache == overrides[exit_code.status]
```
sphuber: What is the added value of also checking the value of `invalidates_cache` for the overrides? Since by adding an exit status to the override you are basically saying that it doesn't follow the normal convention, what actual value is set doesn't really matter no? Wouldn't a list of excluded exit statuses not be sufficient?

aiida_quantumespresso/calculations/base.py
```diff
@@ -0,0 +1,48 @@
...
+        :param status: the exit status integer
+        :param label: a label by which the exit code can be addressed
+        :param message: a more detailed description of the exit code
+        :param invalidates_cache: when set to `True`, a process exiting
+            with this exit code will not be considered for caching
+        """
+        if invalidates_cache is None:
+            invalidates_cache = (isinstance(status, int) and status < 400)
+
+        if 'invalidates_cache' in ExitCode._fields:
```
sphuber: Another unforeseen usage of `ExitCode` that would break with my [proposed change in `aiida-core](https://github.com/aiidateam/aiida-core/pull/3824) :)  can we turn this into a try-catch of `AttributeError`? That should work even if we decide to move `ExitCode` to a simple class, right?

greschd: Good point - it should be a catch of `TypeError`  when instantiating the `ExitCode` however, because we're checking on the class itself here, not the instance.

tests/calculations/test_autoinvalidate_cache.py
greschd: Yeah, I guess that would be sufficient.

tests/calculations/test_autoinvalidate_cache.py
```diff
@@ -0,0 +1,38 @@
...
+"""Test the automatic 'invalidates_cache' attribute for exit codes."""
+
+from __future__ import absolute_import
+import pytest
+
+import aiida
+from aiida.plugins import CalculationFactory
+from aiida.plugins.entry_point import get_entry_point_names
+
+if aiida.__version__ < '1.1':
```
sphuber: Is this string comparison always going to work?

greschd: That works as long as AiiDA doesn't put stuff at the beginning of its `__version__` it will work - because `.` comes before the numbers.

Anyway, should I wrap this in a `distutils.version.LooseVersion` or `StrictVersion`, to make it safe?
The difference between the two here would be that ``LooseVersion('1.1.0b1') < LooseVersion('1.1')`` is ``False``, while ``StrictVersion('1.1.0b1') < StrictVersion('1.1')`` is ``True``.

sphuber: Since the feature was added in `1.1.0` I would do `StrictVersion('1.1.0')`? Since we won't be releasing any `1.1.0` alphas or betas this should work

greschd: Done! Do you happen to know why `pylint` thinks there's no such thing as `distutils.version.StrictVersion`?

greschd:
CI error seems to be intermittent (apt resolution failure), can probably just be restarted.

tests/calculations/test_autoinvalidate_cache.py
sphuber: Remember having encountered this problem before, but don't remember or at least don't know the fix anymore. Probably just an exclude statement

PR #458, opened 2020-02-03
Split input file generation of 'NamelistsCalculation' from the  'prepare_for_submission' method

Crivella:
I've separated the input file generation starting from a generic python dictionary (which was included in the `prepare_for_submission`) into a separate staticmethod `_generate_input_file`.

The functionality to filter/set "blocked keywords", and filter "namelists to print" have also been separated into the respective methods: `_set_blocked_keywords` and `_filter_namelists`

In this way, another plugin that needs to generate input files for different type of calculations (eg: z2pack: scf, nscf, pw2wannier, wannier90) can more easily achieve this by simply calling this method, without having to subclass the original plugin and manually overwrite required input/outputs that are not actually needed by another plugin.


sphuber:
Thanks @Crivella , did you only have to do this for the `NamelistsCalculation` to get to your desired result? Everything looks fine, I would just change the `classmethod` and `staticmethod` to "public" ones, i.e. by removing the leading the underscore. They are supposed to be callable from outside the class after all.

PR #463, opened 2020-02-27
add a protocol for testing

qiaojunfeng:
Based on `theos-ht-1.0`, this `testing` protocol 
1. increase `kpoints_mesh_density` to `0.3`
2. divide `cutoff` by 2

So for debug/tutorial, the whole calculation can be quicker.


sphuber:
@qiaojunfeng if you fix the pre-commit, we can merge it

PR #464, opened 2020-02-27
Added calculation and parser plugin for pw2gw.x 

### LINKED ISSUE #484: Add calculation and parser for the `pw2gw.x` code

Crivella:
Fixes #484 

Added calculation and parser plugin (with tests) for the post-processing code `pw2gw.x`.


aiida_quantumespresso/calculations/pw2gw.py
```diff
@@ -0,0 +1,54 @@
...
+    def define(cls, spec):
+        # yapf: disable
+        super(Pw2gwCalculation, cls).define(spec)
+        spec.input('parent_folder', valid_type=RemoteData,
+            help='Output folder of a completed `PwCalculation`')
+
+        spec.exit_code(200, 'ERROR_NO_RETRIEVED_FOLDER',
+            message='The retrieved folder data node could not be accessed.')
+        spec.exit_code(300, 'ERROR_OUTPUT_FILES',
+            message='The eps*.dat output files could not be read or parsed.')
```
sphuber: In PR #479 I homogenized the exit codes across all `CalcJob` implementations. Please take on the same conventions here for consistency, i.e.
```
spec.exit_code(300, 'ERROR_NO_RETRIEVED_FOLDER')
    message='The retrieved folder data node could not be accessed.')
spec.exit_code(305, 'ERROR_OUTPUT_FILES',
    message='The eps*.dat output files could not be read or parsed.')

File: tests/calculations/test_pw2gw.py [hunk 2]
```diff
@@ -0,0 +1,56 @@
...
+from aiida.common import datastructures
+
+from aiida_quantumespresso.utils.resources import get_default_options
+
+
+def test_pw_default(
+    aiida_profile, fixture_localhost, fixture_sandbox, fixture_code, generate_calc_job, generate_remote_data, tmpdir,
+    file_regression
+):
+    """Test a default `PwCalculation`."""
```
sphuber:
```suggestion
    """Test a default `Pw2gwCalculation`."""
```

tests/calculations/test_pw2gw.py
```diff
@@ -0,0 +1,56 @@
+# -*- coding: utf-8 -*-
+"""Tests for the `PwCalculation` class."""
```
sphuber:
```suggestion
"""Tests for the `Pw2gwCalculation` class."""
```

aiida_quantumespresso/parsers/pw2gw.py
```diff
@@ -0,0 +1,57 @@
...
+        for name in Pw2gwCalculation._internal_retrieve_list:
+            content = retrieved.get_object_content(name)
+
+            try:
+                data =  np.loadtxt(StringIO(content))
+            except ValueError:
+                return self.exit(self.exit_codes.ERROR_OUTPUT_FILES)
+            if not len(data.shape) == 2:
+                return self.exit(self.exit_codes.ERROR_OUTPUT_FILES)
+            if not data.shape[0] > 0 or not data.shape[1] == 2:
```
sphuber: This can be simplified I think. Essentially you want to check that the shape is `(N, 2)` correct?
So you can make this:
```
if len(data.shape) !=2 or data.shape[0] == 0 or data.shape[1] != 2:
    return ....
```

aiida_quantumespresso/parsers/pw2gw.py
```diff
@@ -0,0 +1,57 @@
...
+                return self.exit(self.exit_codes.ERROR_OUTPUT_FILES)
```
sphuber: Looking a bit closer at the format of these output files and what you are doing in the parser, I am wondering two things:

1. If you are really only just validating the format of the files here, maybe we can make the exit code a bit more specific. Maybe `330 - ERROR_OUTPUT_FILES_INVALID_FORMAT - 'The eps*.dat output files do not have the expected shape (N, 2).' `
2. You are really just parsing these files to check that their content is valid. How are these files used in potential follow up calculations? Would it maybe make sense to turn one, or multiple in actual output nodes? Or do you think the additional cost on the database does not outweigh the benefit of having the data in node form, besides just as raw files in the retrieved folder.

tests/parsers/fixtures/pw2gw/default/epsTOT.dat
```diff
@@ -0,0 +1,100 @@
+       0.000500        0.000000
```
sphuber: Sorry to come back to this, but I think they can be reduced to a single line. Your parser is really just calling `numpy.loadtxt` and checking that all output files have the shape `(N, 2)`. To test this, just a single line suffices, no?

aiida_quantumespresso/calculations/pw2gw.py
```diff
@@ -0,0 +1,54 @@
...
+        ('INPUTPP', 'prefix', NamelistsCalculation._PREFIX),
+    ]
+
+    @classmethod
+    def define(cls, spec):
+        # yapf: disable
+        super(Pw2gwCalculation, cls).define(spec)
+        spec.input('parent_folder', valid_type=RemoteData,
+            help='Output folder of a completed `PwCalculation`')
+
```
sphuber: There are no output nodes whatsoever. Although this is possible is this really what we want? Is there no useful information to parse from the `stdout` that we can parse and put in a `Dict` node called `output_parameters`. For example, why not parse the `walltime` and `walltime_seconds`? This will make this queryable and easily accessible in analyses. Also see a comment related to the `eps*.dat` files

aiida_quantumespresso/parsers/pw2gw.py
Crivella: The most common post-processing to be applied to this files would be to apply a gaussian/lorentzian broadening or the KK-relations to get the real part of the dielectric function.

The size in .npy format of a possible output (using pw2gw default input values) file  would be around 25KB (I assume the size would be similar in the DB).
It could increase to close to 1MB depending on the energy range and sampling rate.

If you think it wouldn't be too cumbersome to store, I could add an output node 'eps' of type ArrayData containing the following:
{ 'energy': [...],    'epsX': [...],    'epsY': [...],    'epsZ': [...]}

tests/parsers/fixtures/pw2gw/default/epsTOT.dat
Crivella: I will reduce it to 2 lines in order t check the possibility of
```suggestion
       0.000500
       0.001000        0.000000
```

aiida_quantumespresso/parsers/pw2gw.py
sphuber: These are difficult but important decisions, but since I have never worked with `pw2gw` its difficult to judge. If you don't think it is useful to have the files as nodes, or at least the added weight to the database outweighs the benefits, it is better not to add them. Or if in doubt, we don't add them as an output node for now and if in the future it does seems like it is useful, we can always still add it then. That is easier then removing it again

aiida_quantumespresso/calculations/pw2gw.py
Crivella: I have added the output node `output_parameters` that for now contains only the result of the `walltime` parsing.
As far as i know there shouldn't really be any information in the stdout of pw2gw that cannot be read from the original pw.x calculation or input files

aiida_quantumespresso/parsers/pw2gw.py
Crivella: I decided to add the node output node `eps` as above described with the corresponding test.

aiida_quantumespresso/parsers/parse_raw/pw2gw.py
```diff
@@ -0,0 +1,69 @@
...
+def parse_stdout(stdout):
+    """Parses the stdout content of a Quantum ESPRESSO `pw2gw.x` calculation.
+
+    :param stdout: the stdout content as a string
+    :param input_parameters: dictionary with the input parameters
+    :param parser_options: the parser options from the settings input parameter node
+    :returns: tuple of two dictionaries, with the parsed data and log messages, respectively
+    """
+    # if parser_options is None:
+    #     parser_options = {}
```
sphuber: dead code, best to remove it

aiida_quantumespresso/parsers/pw2gw.py
```diff
@@ -0,0 +1,173 @@
...
+        """Parse the stdout file of pw2gw to build the `output_parameters` node."""
+        from aiida_quantumespresso.utils.mapping import get_logging_container
+        from aiida_quantumespresso.parsers.parse_raw.pw2gw import parse_stdout
+
+        logs = get_logging_container()
+        parsed_data = {}
+
+        filename_stdout = self.node.get_attribute('output_filename')
+        # print(" < ", filename_stdout, ' > ')
+        # return parsed_data, logs
```
sphuber: dead code

aiida_quantumespresso/parsers/parse_raw/pw2gw.py
```diff
@@ -0,0 +1,69 @@
...
+default_dipole_units = 'Debye'
+default_energy_units = 'eV'
+default_force_units = 'ev / angstrom'
+default_k_points_units = '1 / angstrom'
+default_length_units = 'Angstrom'
+default_magnetization_units = 'Bohrmag / cell'
+default_polarization_units = 'C / m^2'
+default_stress_units = 'GPascal'
+
+# def parse_stdout(stdout, input_parameters, parser_options=None):
```
sphuber: commented code

sphuber:
Thanks a lot @Crivella for the work and changes after review :+1:

PR #469, opened 2020-03-04
`PwParser`: use input parameters for ionic convergence thresholds

### LINKED ISSUE #468: `PwParser`: ionic convergence thresholds not properly taken from inputs
The method `PwParser.validate_ionic` is manually checking forces and stresses for convergence, however, it is always using the default thresholds. It tries to get them from the `parameters` but it thinks they are the input parameters, but they are the parsed output parameters. Since the thresholds are not there, the defaults are taken.

sphuber:
Fixes #468 

The `validate_ionic` method was accidentally using the output parameters
which does not contain the convergence thresholds parameters and so the
default values were always used, ignoring the actual values used in the
inputs of the calculation.

To bring this bug to light, we first replace the reference files for the
`vcrelax_success` test with those of a run with convergence thresholds
that are much higher than the default. Due to the bug, the parser would
mark the calculation incorrectly as unconverged, but the change in this
commit fixes that.


aiida_quantumespresso/utils/validation/trajectory.py
```diff
@@ -30,9 +31,6 @@ def verify_convergence_trajectory(trajectory, index=-1, threshold_forces=None, t
     else:
         converged_stress = True
 
-    if converged_forces is None or converged_stress is None:
-        return None
-
     return converged_forces & converged_stress
```
greschd: Any reason to use bitwise-and here, not `and`? They have the same behavior for booleans, but e.g.

```python
In [5]: True & 1
Out[5]: 1

In [6]: True & 2
Out[6]: 0

In [7]: True and 2
Out[7]: 2
```

sphuber: Note that that was already there and have to say I don't remember any specific reason. So happy to change it to normal boolean logic. Were there any other things you wanted to change in this PR? If not then I will make that change and then we can merge. I have two other more important PRs with bugs blocked by this

greschd: Yeah, just noticed it due to the adjacent change - mostly I was wondering if the `converged_*` variables might be some nontrivial type, but looking at the `verify_*` functions they're just bools. So, doesn't matter either way, although `and` is slightly more readable.

aiida_quantumespresso/utils/validation/trajectory.py
```diff
@@ -54,7 +53,7 @@ def verify_convergence_forces(trajectory, index=-1, threshold=None):
     try:
         forces = trajectory.get_array('forces')[index]
     except (KeyError, IndexError):
-        return None
+        return ValueError('the `forces` array does not exist or the given index exceeds the length.')
```
greschd: Should this be a `raise` instead of `return`?

aiida_quantumespresso/utils/validation/trajectory.py
```diff
@@ -75,6 +75,6 @@ def verify_convergence_stress(trajectory, index=-1, threshold=None):
     try:
         stress = trajectory.get_array('stress')[index]
     except (KeyError, IndexError):
-        return None
+        return ValueError('the `stress` array does not exist or the given index exceeds the length.')
```
greschd: Same here.

aiida_quantumespresso/utils/validation/trajectory.py
sphuber: Have made the change

aiida_quantumespresso/utils/validation/trajectory.py
sphuber: Damn straight :)

greschd: Almost missed it, too 😅 Seems almost like something `pylint` should warn about.

aiida_quantumespresso/utils/validation/trajectory.py
```diff
@@ -54,7 +53,7 @@ def verify_convergence_forces(trajectory, index=-1, threshold=None):
     try:
         forces = trajectory.get_array('forces')[index]
     except (KeyError, IndexError):
-        return None
```
greschd: If I understand this right, you could previously specify `threshold_forces` even if no forces were created, but now it will raise. 

Is that correct? Or was this combination caught somewhere else also? Anyway, I think the change in behavior is fine if intentional. There's not too much sense in specifying a threshold for forces if those aren't computed.

sphuber: Yeah that's true, but I figure the old behavior did not make much sense and made it difficult to use. On top of that I introduced these functions in one of the alphas and only used them internally, so I feel its fine to change the behavior

greschd: Yeah, I agree - the new behavior makes more sense.

sphuber:
Thanks a lot for the review @greschd

PR #470, opened 2020-03-04
`PwParser`: include external pressure in ionic convergence threshold

### LINKED ISSUE #449: `PwParser`: incorrectly assumes calculations to not be converged for external pressures
Dear AiiDa Team

I'm currently writing a workflow which makes use of the `PwRelaxWorkChain` workchain. I have done this in a Ubuntu18.04 machine with the following AiiDA version:

-  aiida-core:  v1.0.1 hash: e74ad1b36
-  aiida-quantumespresso: v3.0.0a5 hash: 6ef6af1

The volume relaxations are performed at fixed pressure (`press` keyword), making use of the `vc-relax`. The calculations go as one would expect with several bsgf steps being performed until convergence is reached. However, `PwRelaxWorkChain` reports the calculation as failed with exit code 500, `The ionic minimization cycle did not converge for the given thresholds.`.

I have noticed that this issue seems to be related to the verify_convergence_stress function found in `aiida-quantumespresso/aiida_quantumespresso/utils/validation/trajectory.py`.

This function checks 
```
(numpy.trace(stress) / 3.) < threshold
```
which is fine if there is no external pressure. However, as soon as an external pressure is applied this condition cannot be fulfilled, as the stress tensor would be shifted by the applied pressure. This becomes more apparent when one applies a negative pressure,  where the offset results in the condition being always fulfilled. 

Thus I guess that one should change the function to something akin to
```
numpy.abs(numpy.trace(stress)/3. - ext_pressure) < threshold
```

sphuber:
Fixes #449 

The parser was not properly including an external pressure when checking
the stress for convergence, leading the calculation to be unjustly as
not converged. By simply adding the external pressure from the input
parameters to the pressure convergence threshold this is solved.

### DIFF
--- a/aiida_quantumespresso/utils/validation/trajectory.py
+++ b/aiida_quantumespresso/utils/validation/trajectory.py
@@ -6,7 +6,9 @@ from __future__ import division
 import numpy
 
 
-def verify_convergence_trajectory(trajectory, index=-1, threshold_forces=None, threshold_stress=None):
+def verify_convergence_trajectory(
+    trajectory, index=-1, threshold_forces=None, threshold_stress=None, reference_pressure=0.
+):
     """Verify that the data of the given `TrajectoryData` is converged with respect to the given thresholds.
 
     There are up to three properties controlled for convergence: total energy, forces and stress.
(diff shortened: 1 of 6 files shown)


aiida_quantumespresso/utils/validation/trajectory.py
```diff
@@ -72,9 +75,14 @@ def verify_convergence_stress(trajectory, index=-1, threshold=None):
 
     threshold /= 10.  # Convert to GPa
 
+    if external_pressure:
+        threshold += external_pressure / 10  # Convert to GPa and add it to the threshold
```
giovannipizzi:
```suggestion
        reference_pressure = external_pressure / 10.  # Convert to GPa
    else:
        reference_pressure = 0.
```

aiida_quantumespresso/utils/validation/trajectory.py
```diff
@@ -72,9 +75,14 @@ def verify_convergence_stress(trajectory, index=-1, threshold=None):
...
+
     try:
         stress = trajectory.get_array('stress')[index]
     except (KeyError, IndexError):
         raise ValueError('the `stress` array does not exist or the given index exceeds the length.')
 
-    return (numpy.trace(stress) / 3.) < threshold
+    pressure = (numpy.trace(stress) / 3.)
+
+    return pressure < threshold
```
giovannipizzi:
```suggestion
    return abs(pressure - reference_pressure) < threshold
```

PR #474, opened 2020-03-06
 `PwBandsWorkChain`: make `nbands_factor` optional 

### LINKED ISSUE #441: Possibility for explicit "nbnds" in "PwBandsWorkChain"?
When calculating a band structure for comparison with a known reference, I sometimes want to explicitly specify ``nbnd`` for the bands step in ``PwBandsWorkChain``. Currently, ``nbnd`` in the ``bands`` step is calculated via ``nbands_factor`` - to explicitly set ``nbnd``, the factor has to be "reverse-engineered".

An alternative would be to make ``nbnds_factor`` optional, and use it only when there is no ``nbnd`` set in the ``bands.pw.parameters`` input. I'm not sure if that's a common enough use case, though. @sphuber, thoughts?

sphuber:
Fixes #441 

By making this a non required input, the caller has to opportunity to
explicitly specify a number of bands to be used for the bands step
through the `bands.pw.parameters` input node. Only if that key is not
set and the `nbands_factor` input is defined, does the workchain compute
the number of bands to use. Note that if neither are defined the
workchain will not set anything either and the eventual number of bands
that will be used is whatever the default is of Quantum ESPRESSO.

### DIFF
--- a/aiida_quantumespresso/workflows/pw/bands.py
+++ b/aiida_quantumespresso/workflows/pw/bands.py
@@ -16,8 +16,28 @@ PwBaseWorkChain = WorkflowFactory('quantumespresso.pw.base')
 PwRelaxWorkChain = WorkflowFactory('quantumespresso.pw.relax')
 
 
+def validate_inputs(inputs, ctx=None):  # pylint: disable=unused-argument
+    """Validate the inputs of the entire input namespace."""
+    if 'nbands_factor' in inputs and 'nbnd' in inputs['bands']['pw']['parameters'].get_attribute('SYSTEM', {}):
+        return PwBandsWorkChain.exit_codes.ERROR_INVALID_INPUT_NUMBER_OF_BANDS.message  # pylint: disable=no-member
+
+
 class PwBandsWorkChain(WorkChain):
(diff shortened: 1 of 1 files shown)


aiida_quantumespresso/workflows/pw/bands.py
```diff
@@ -141,25 +141,25 @@ def inspect_scf(self):
...
         inputs.pw.parameters.setdefault('ELECTRONS', {})
 
+        # The following flags always have to be set in the parameters, regardless of what caller specified in the inputs
         inputs.pw.parameters['CONTROL']['restart_mode'] = 'restart'
         inputs.pw.parameters['CONTROL']['calculation'] = 'bands'
-        inputs.pw.parameters['SYSTEM']['nbnd'] = nbands
+
+        # Only set the number of bands if not explicitly defined in the inputs and the `nbands_factor` input is defined
+        if 'nbnd' not in inputs.pw.parameters['SYSTEM'] and 'nbands_factor' in self.inputs:
+            factor = self.inputs.nbands_factor.value
```
giovannipizzi: Add a validator here - raise if both are set

PR #479, opened 2020-03-06
Homogenize the exit codes for all `CalcJob` implementations

### LINKED ISSUE #296: Homogenize exit codes across calculation classes
One way to do it is to add the exit codes to the spec of a base class instead of individual subclasses (`NamelistsCalculation`, `BasePwCpInputGenerator`).

sphuber:
Fixes #296 

Making sure that all calculation jobs use the same exit codes for the
same problem does not only make things easier for the user but it can
then also be used in code to make certain assertions based on the exit
code of a failed calculations. An example is the automatically
invalidating of caching for exit codes smaller than 400, which will be
implemented soon. This is only possible if all calculations respect the
convention that codes below 400 are irrecoverable errors and so those
should also not be used for caching.

### DIFF
--- a/aiida_quantumespresso/calculations/cp.py
+++ b/aiida_quantumespresso/calculations/cp.py
@@ -129,26 +129,30 @@ class CpCalculation(BasePwCpInputGenerator, CalcJob):
 
     @classmethod
     def define(cls, spec):
-        super(CpCalculation, cls).define(spec)  # noqa: D102 # yapf: disable
+        # yapf: disable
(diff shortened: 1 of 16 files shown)


sphuber:
It would be good to address this issue before merging #453 of @greschd since that functionality depends on the exit codes respecting the guidelines we set here.

sphuber:
It would be good to address this issue before merging #453 of @greschd since that functionality depends on the exit codes respecting the guidelines we set here.

PR #486, opened 2020-03-26
Add custom base `Parser` class that all parsers should use

sphuber:
The base class serves as a central point for useful functionality for
all parsers to be implemented. Currently it provides:

 * `emit_logs`: function to emit parsed messages to node logger
 * `exit`: exit with an exit code with the exit message automatically
   being emitted through the node's logger

### DIFF
--- a/aiida_quantumespresso/parsers/base.py
+++ b/aiida_quantumespresso/parsers/base.py
@@ -1,0 +1,65 @@
+# -*- coding: utf-8 -*-
+"""Defines a `Parser` base class for `aiida-quantumespresso`.
+
+All `Parser` implementations in `aiida-quantumespresso` must use this base class, not `aiida.parsers.Parser`.
+"""
+from __future__ import absolute_import
+
+from aiida.parsers import Parser as _BaseParser
+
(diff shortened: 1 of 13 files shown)


aiida_quantumespresso/parsers/base.py
```diff
@@ -0,0 +1,65 @@
...
+                        continue
+
+                    stripped = message.strip()
+
+                    if not stripped or stripped in ignore:
+                        continue
+
+                    try:
+                        getattr(self.logger, level)(stripped)
+                    except AttributeError:
```
greschd: In the case where `level` does not correspond to a known logging level, wouldn't it be safer to fall back on e.g. `warning`?

Is there a concrete use case where you want to suppress the logging, or is this just to not fail on typos?

sphuber: In principle this would be a programming error, because the programmer is supposed to use a dictionary that is created by `get_logging_container` which only supports `warning` and `error` keys for the moment. It is all a bit flimsy at the moment and just relies on the developers doing the right thing, but since that is basically me, I think it is fine for now.

PR #488, opened 2020-03-26
`PwParser`: allow parsing from only the stdout

### LINKED ISSUE #487: `PwParser`: make parsing from XML optional
This will be mostly useful for the immigration of existing calculations that no longer or never had the XML output file.

sphuber:
Fixes #487 

This is useful in cases where the XML output file is not available, for
example in older calculations run without AiiDA that need to be
imported. We define a new option on the `PwCalculation` class called
`without_xml` that is not required. We do not set a default otherwise
that would be set as an attribute on all nodes. If it is set to `True`
in the inputs, a missing XML output file will not cause the parsing to
fail as in the default case.

For certain calculation types, such as `relax` one of the required
outputs is the `structure`. Since the data for this output node is
nominally parsed from the XML data, this would normally fail. However,
we can use the input structure in this case, simply replacing the atomic
positions and optionally lattice vectors that are also parsed from the
stdout. Note that we do not do this in the raw stdout parsing function
because doing so would couple it more tightly to input data being passed
in through arguments, which is something we want to prevent.

### DIFF
--- a/aiida_quantumespresso/calculations/pw.py
+++ b/aiida_quantumespresso/calculations/pw.py
@@ -64,6 +64,8 @@ class PwCalculation(BasePwCpInputGenerator):
         # yapf: disable
         super(PwCalculation, cls).define(spec)
         spec.input('metadata.options.parser_name', valid_type=six.string_types, default='quantumespresso.pw')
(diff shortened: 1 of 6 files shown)


sphuber:
Would supersede #336

PR #490, opened 2020-03-26
`PwBandsWorkChain`: carry updated number of bands between steps

### LINKED ISSUE #382: Sanity check `PwBaseWorkChain` not carried by `PwBandsWorkChain`
If one of the `PwBaseWorkChains` of the `PwRelaxWorkChain` step in the `PwBandsWorkChain` is triggered, it will be properly carried through to the following by the relax workchain. However, the bands work chain does not and so it most likely will be triggered again in the scf step.

sphuber:
Fixes #382 

The `PwBaseWorkChain` has a sanity check to make sure the highest bands
are not occupied, unless we are running an insulator without smearing.
If this condition is not met, the number of bands are increased and the
calculation is restarted. We need to make sure that when this happens,
the new number of bands is carried to the next step, because otherwise
the next `PwBaseWorkChain` is most likely going to run in the same
problem. The `PwRelaxWorkChain` was already properly carrying the number
of bands so here we just need to make sure to update the value before
the scf ands bands steps.

### DIFF
--- a/aiida_quantumespresso/workflows/pw/bands.py
+++ b/aiida_quantumespresso/workflows/pw/bands.py
@@ -114,6 +114,7 @@ class PwBandsWorkChain(WorkChain):
     def setup(self):
         """Define the current structure in the context to be the input structure."""
         self.ctx.current_structure = self.inputs.structure
+        self.ctx.current_number_of_bands = None
         self.ctx.bands_kpoints = self.inputs.get('bands_kpoints', None)
 
     def should_run_relax(self):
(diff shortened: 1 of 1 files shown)


aiida_quantumespresso/workflows/pw/bands.py
```diff
@@ -168,8 +170,13 @@ def run_scf(self):
...
         inputs.pw.structure = self.ctx.current_structure
         inputs.pw.parameters = inputs.pw.parameters.get_dict()
-        inputs.pw.parameters.setdefault('CONTROL', {})
-        inputs.pw.parameters['CONTROL']['calculation'] = 'scf'
+        inputs.pw.parameters.setdefault('CONTROL', {})['calculation'] = 'scf'
+
+        # Make sure to carry the number of bands from the relax workchain if it was run and it wasn't explicitly defined
+        # in the inputs. One of the base workchains in the relax workchain may have changed the number automatically in
+        #  the sanity checks on band occupations.
+        if self.ctx.current_number_of_bands:
```
greschd: Maybe this check should be `if self.ctx.current_number_of_bands is not None:`? Right now all values that convert to `False` are ignored.

Then again, a `nbnd` of zero isn't really valid, so it probably doesn't matter 😄

sphuber: That was my thinking exactly

Issue #405, opened 2019-10-10
Promote parallelization settings to "proper" inputs?


greschd:
As far as I can tell, the parallelization options can currently only be specified using the ``settings.cmdline`` option.

To make this a bit more accessible, maybe it would make sense to instead specify them as "real" inputs, with their own input ports? This could also allow for some validation, e.g. that the options are compatible with the number of cores.

Related issue: #328.

sphuber:
Are you proposing an input port on the `PwCalculation` class. For example a `Dict` node with the keys:
* `ni`
* `nk`
* `nt`
* `nd`
which correspond to the available parallelization flags.
We are currently working with @zooks97 on understanding the parallelization options in Quantum ESPRESSO and how we can automate the selection of them.

greschd:
Yeah, something like that. I think this would be pretty independent of the automated parallelization, except for changing how the parameters are passed to the ``PwCalculation``.

I'm not sure if the "long" version of the flags  (``nimage``, ``npools``, ``nband`` ``ntg`` ``ndiag``) would be easier to use, or if both should be allowed.

sphuber:
No you are right, I think we should stick with the longer versions and only those. There being multiple version on QE itself is already confusing enough.

greschd:
Maybe it would be more appropriate to make it an input namespace, with each value set as an ``Int``? 
That way we wouldn't have to handle the default case in code, but you could still specify only one of the parameters with the others having the default.

Not handling the default in code also has benefits for caching, because no input and input=1 are actually the same.

greschd:
We should probably have a chat about parallelization options @zooks97 @sphuber. I'm currently running some tests, but my understanding is pretty limited still.

greschd:
@sphuber this seems relatively straightforward, I can have a crack at adding it. A few questions about how you want the interface to look:
- Should the parallelization inputs be a single ``Dict``, or would you prefer the namespace approach described above?
- If the parallelization options are also given in the ``CMDLINE`` argument, what should happen? Can we just raise ``InputValidationError``, or should we somehow preserve backwards compatibility by merging the options?

greschd:
Ping @sphuber on the design questions above.

greschd:
Another question: As far as I understand, different codes can have the same parallelization flags, but they are not _necessarily_ implemented.

How should we handle this? Implement them at the `BasePwCpInputGenerator` but then have a list of "allowed" parallelization flags per calculation class?

[...]

PR #491, opened 2020-03-26
`PwBaseWorkChain`: add handler for `ERROR_COMPUTING_CHOLESKY`

### LINKED ISSUE #213: `PwBaseWorkChain`: implement handler for cholesky decomposition problems
A `pw.x` calculation can throw the following error:
```
     Davidson diagonalization with overlap

     Error in routine  cdiaghg (668):
      problems computing cholesky

     stopping ...
```

from which the base workchain cannot recover and restarting is pointless. There should therefore be an error handler that exits the work chain with an appropriate exit code.

sphuber:
Fixes #213 

Currently there is no known cure for this problem so a handler is added
to abort the workchain.

### DIFF
--- a/aiida_quantumespresso/calculations/pw.py
+++ b/aiida_quantumespresso/calculations/pw.py
@@ -118,6 +118,8 @@ class PwCalculation(BasePwCpInputGenerator):
 
         spec.exit_code(461, 'ERROR_DEXX_IS_NEGATIVE',
             message='The code failed with negative dexx in the exchange calculation.')
+        spec.exit_code(462, 'ERROR_COMPUTING_CHOLESKY',
+            message='The code failed during the cholesky factorization.')
 
         spec.exit_code(481, 'ERROR_NPOOLS_TOO_HIGH',
             message='The k-point parallelization "npools" is too high, some nodes have no k-points.')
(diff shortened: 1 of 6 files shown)


greschd:
@sphuber from [this forum post](https://www.mail-archive.com/users@lists.quantum-espresso.org/msg32363.html), it seems at least sometimes the problem is fixed by setting `-ndiag 1`.

I've encountered at least one instance where that is also the case. Absent any better options, should we implement this handler - even if it could hurt performance?

Implementing this would be easier after https://github.com/aiidateam/aiida-quantumespresso/issues/405 is done - so maybe that's something to put higher on my priority list.

sphuber:
That is of course if the `-ndiag` is explicitly set to anything other than 1, because the default is 1, right?

greschd:
Ha, that's what I thought also! But actually it's (from [here](https://www.quantum-espresso.org/Doc/user_guide/node20.html))

> nd is set to 1 if ScaLAPACK is not compiled, it is set to the square integer smaller than or equal to half the number of processors of each pool

I was confused initially because I had _not_ set `ndiag` explicitly.

sphuber:
> I guess detecting whether ScaLAPACK is compiled is not really something we can do .. right? joy

Regardless of whether we could, I am not opening that can of worms :)

greschd:
Just to document some observations I've made:

> nd is set to 1 if ScaLAPACK is not compiled, it is set to the square integer smaller than or equal to half the number of processors of each pool

This is actually incorrect; as of [this QE MR](https://gitlab.com/QEF/q-e/-/merge_requests/626) the default goes up the the _full_ number of processors. [MR to fix the documentation](https://gitlab.com/QEF/q-e/-/merge_requests/1179) is filed.

Maybe a more interesting observation is that it's not always necessary to set `ndiag` to `1` to fix the Cholesky error - just reducing it can also solve it. I'm not sure if there's a good rule for how much `ndiag` is too much, though. In terms of adding a handler, maybe just switching to `1` is the safer option (I don't know _how much_ performance is expected to degrade due to that).

PR #492, opened 2020-03-27
`PwBaseWorkChain`: do not restart from scratch after bands sanity check 

### LINKED ISSUE #489: `PwBaseWorkChain`: sanity check on bands will always fail the restart for `nscf` or `bands` calculations
The reason is that the sanity check will force the restart to be from scratch, because if you change the number of bands, you cannot reuse the existing charge density or wavefunctions. However, starting from scratch in a bands or nscf calculation also does not make any sense and despite specifying `from_scratch` the code will attempt to read the XML to restart, which is not there and so the code crashes.

sphuber:
Fixes #489 

The base work chain has a sanity check that will make sure the highest
band has a neglible occupation. If this is not the case, unless the
calculation was explicitly ran with fixed occupations (indicating that
the system is insulating), the number of bands in the inputs is
increased and the calculation is restarted from scratch.

When first implemented, I erroneously assumed that one cannot restart
from the previous calculation because the number of bands will have
changed and so the old files will be incommensurate. This is the case
for certain parameters being changed, such as the wavefunction cutoffs,
but for the number of bands it is perfectly fine to restart. Therefore
we change the behavior of the bands sanity check to *not* restart from
scratch.

This also fixes a bug encountered when this sanity check was triggered
for non-scf calculations. There the restarts would crash horribly
because restarting from scratch in a non-scf calculation makes no sense.
Quantum ESPRESSO would try to read the files from the pr

### DIFF
--- a/aiida_quantumespresso/workflows/pw/base.py
+++ b/aiida_quantumespresso/workflows/pw/base.py
@@ -257,7 +257,7 @@ class PwBaseWorkChain(BaseRestartWorkChain):
         inputs = prepare_process_inputs(PwCalculation, inputs)
         running = self.submit(PwCalculation, **inputs)
 
-        self.report('launching initialization PwCalculation<{}>'.format(running.pk))
+        self.report('launching initialization {}<{}>'.format(running.pk, self._calculation_class.__name__))
 
         return ToContext(calculation_init=running)
 
@@ -327,7 +327,8 @@ class PwBaseWorkChain(BaseRestartWorkChain):
             bands = calculation.outputs.output_band
             get_highest_occupied_band(bands)
         except ValueError as exception:
-            self.report('calculation<{}> run with smearing and highest band is occupied'.format(calculation.pk))
+            args = [self._calculation_class.__name__, calculation.pk]
+            self.report('{}<{}> run with smearing and highest band is occupied'.format(*args))
             self.report('BandsData<{}> has invalid occupations: {}'.format(bands.pk, exception))
             return self._handle_insufficient_bands(calculation)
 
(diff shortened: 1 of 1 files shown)


greschd:
The one thing I don't understand yet with this problem is why the `PwBandsWorkChain` can increase the number of bands, to a value that is larger than the number of SCF bands. Is this just to force better convergence for the bands that _were_ included in the SCF run, but the newly added bands would be unphysical?

sphuber:
> The one thing I don't understand yet with this problem is why the `PwBandsWorkChain` can increase the number of bands, to a value that is larger than the number of SCF bands. Is this just to force better convergence for the bands that _were_ included in the SCF run, but the newly added bands would be unphysical?

You are absolutely right, it doesn't make sense. I think I made a mistake and what the real problem was is that the sanity check set `restart_mode=from_scratch`. But now that I think of it, restarting with more bands might be fine. It is changing the cutoffs for example that will invalidate it as one would need a different FFT grid etc.

However, that makes me wonder why we decided at some point to set `from_scratch` in the sanity check when increasing the number of bands. Maybe doing a `restart` is fine? I will do some testing.

sphuber:
Ok, I have confirmed that restarting a non-scf calculation with increased bands is perfectly fine. This makes the fix a lot easier. I have updated the PR. Should be good to go for review @greschd

greschd:
Maybe this should still emit a warning? I'm not sure if the newly added bands will be _physically_ correct.

sphuber:
I think they will be physically correct and meaningful. In the non-scf calculation the code just reconstructs the charge-density and potential from the previous calculation but then still solves the KS equations by diagonalizing the Hamiltonian for the given number of k-points and bands.  In the example band structure calculations of PW in the QE distribution they also increase the number of bands in the `bands` calculation compared to the `scf` one. @giovannipizzi can you confirm?

greschd:
I think my question is whether the charge-density is meaningful if the initial number of bands is _less than the actual number of electrons_. I guess it would make sense for that to be determined by the number of electrons instead of `nbnds`, but I don't know if that's what it does.

Maybe as a simple check we could do something ridiculous like solve SCF with way too few bands, and then see how the bands turn out?

sphuber:
Sure, but in this issue it is not the responsibility of the `PwBaseWorkChain` to check that the calculation of the `parent_folder` used enough bands. If that scf was run with the `PwBaseWorkChain` that also would not have been possible, because the same check would have increased the number of bands. Giovanni made the point that for non-scf calculations, however, sometimes (for whatever weird reason) you might only want to include occupied bands, and not even all of them, but just some. It is weird, but won't give a wrong answer.

greschd:
Hmm, doesn't that then mean we shouldn't even check for this here?

sphuber:
Shit you are right. I am doing too many things at the same time and so forgot about his reasoning. Only remembered after typing it out again in my last response :man_facepalming: 

I think I discussed with Giovanni that we would ignore the test _if_ `calculation_type in  ['nscf', 'bands']` and `'SYSTEM.nbnd'` is defined explicitly in the parameters. This would allow users that know what they are doing to run an NSCF with too few bands, whereas in the case where people simply didn't define anything (probably because they don't know) we automatically increase it.

Am I missing anything in this logic this time?

greschd:
Yeah, I think this is fine - in the higher-level workflows, `nbnd` is carried over from SCF to NSCF / BANDS, right? So it is fixed in the SCF, and then NSCF and BANDS take that value unless explicitly specified otherwise.

Maybe we still want to emit a warning, although that could also be annoying if you know what you're doing.

PR #503, opened 2020-04-17
Allow explicitly specifying 'ibrav'.

### LINKED ISSUE #347: Provide option to set an explicit `ibrav` value
Currently this parameter from the `CONTROL` list is explicitly forbidden from being set directly in the input parameters, as it is always set to `0` by the calculation plugin. However, for some purposes people want to enforce a specific `ibrav` if for example they know their structure is FCC and want to keep that symmetry even during volume relaxation for example.

@giovannipizzi are there any objections to providing this as an option through, for example, a key `force_ibrav` in the `settings` input?

greschd:
Fixes #347.

If an `ibrav` value other than zero is specified in the `SYSTEM` inputs, the cell of the input structure is converted into the appropriate A, B, C, cosAB, cosAC, cosBC values. As a check, the cell is reconstructed using qe_tools. The cells are compared element-wise, with an absolute tolerance controlled by the `IBRAV_CELL_TOLERANCE` setting.

Still needs testing and documentation. No need to check the correctness yet, but I'm soliciting feedback on the interface.

Note also: This adds `numpy` and `scipy` as _explicit_ dependencies. They're implicit dependencies already, so I don't think that should be problematic.

TODO:
- [x] ~add tests for the conversion between cell and parameters~ these will go into `qe-tools` directly
- [x] document the explicit `ibrav` feature
- [x] add test for immigration, possibly add conversion from parameters to cell
- [x] move cell -> parameter conversion to `qe_tools`?

### DIFF
--- a/aiida_quantumespresso/calculations/__init__.py
+++ b/aiida_quantumespresso/calculations/__init__.py
@@ -7,9 +7,12 @@ from aiida import orm
 from aiida.common import datastructures, exceptions
 from aiida.common.lang import classproperty
 
+from qe_tools.converters import get_parameters_from_cell
+
 from aiida_quantumespresso.utils.convert import convert_input_to_namelist_entry
 
 from .base import CalcJob
+from .helpers import QEInputValidationError
 
 
 class BasePwCpInputGenerator(CalcJob):
@@ -23,6 +26,7 @@ class BasePwCpInputGenerator(CalcJob):
     _DATAFILE_XML_PRE_6_2 = 'data-file.xml'
     _DATAFILE_XML_POST_6_2 = 'data-file-schema.xml'
     _ENVIRON_INPUT_FILE_NAME = 'environ.in'
+    _DEFAULT_IBRAV = 0
 
     # Additional files that should always be retrieved for the specific plugin
     _internal_retrieve_list = []
@@ -285,10 +289,15 @@ class BasePwCpInputGenerator(CalcJob):
 
         # ============ I prepare the input site data =============
         # ------------ CELL_PARAMETERS -----------
-        cell_parameters_card = 'CELL_PARAMETERS angstrom\n'
-        for vector in structure.cell:
-            cell_parameters_card += ('{0:18.10f} {1:18.10f} {2:18.10f}'
-                                     '\n'.format(*vector))
+
+        # Specify cell parameters only if 'ibrav' is zero.
+        if input_params.get('SYSTEM', {}).get('ibrav', cls._DEFAULT_IBRAV) == 0:
+            cell_parameters_card = 'CELL_PARAMETERS angstrom\n'
+            for vector in structure.cell:
+                cell_parameters_card += ('{0:18.10f} {1:18.10f} {2:18.10f}'
+                                        '\n'.format(*vector))
+        else:
+            cell_parameters_card = ''
 
         # ------------- ATOMIC_SPECIES ------------
         atomic_species_card_list = []
@@ -448,7 +457,18 @@ class BasePwCpInputGenerator(CalcJob):
         # Set some variables (look out at the case! NAMELISTS should be
         # uppercase, internal flag names must be lowercase)
         input_params.setdefault('SYSTEM', {})
-        input_params['SYSTEM']['ibrav'] = 0
+        input_params['SYSTEM'].setdefault('ibrav', cls._DEFAULT_IBRAV)
+        ibrav = input_params['SYSTEM']['ibrav']
+        if ibrav != 0:
+            try:
+                structure_parameters = get_parameters_from_cell(
+                    ibrav=ibrav,
+                    cell=structure.get_attribute('cell'),
+                    tolerance=settings.pop('IBRAV_CELL_TOLERANCE', 1e-6)
+                )
+            except ValueError as exc:
+                raise QEInputValidationError('Cannot get structure parameters from cell: {}'.format(exc)) from exc
+            input_params['SYSTEM'].update(structure_parameters)
         input_params['SYSTEM']['nat'] = len(structure.sites)
         input_params['SYSTEM']['ntyp'] = len(structure.kinds)
 
--- a/docs/source/user_guide/calculation_plugins/cp.rst
+++ b/docs/source/user_guide/calculation_plugins/cp.rst
@@ -35,7 +35,6 @@ Inputs
       'CONTROL', 'pseudo_dir': pseudopotential directory
       'CONTROL', 'outdir': scratch directory
       'CONTROL', 'prefix': file prefix
-      'SYSTEM', 'ibrav': cell shape
       'SYSTEM', 'celldm': cell dm
       'SYSTEM', 'nat': number of atoms
       'SYSTEM', 'ntyp': number of species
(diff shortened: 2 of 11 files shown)


greschd:
image
Ah, Python2, loving it 😅

greschd:
@giovannipizzi should the "cell to parameter" conversion go into `qe_tools`?

giovannipizzi:
@greschd yes, maybe as they are general they can be put into qe-tools?

In this case I agree testing should be done by qe-tools, and raise there.
If we want to do it, I suggest this could be a good moment to release a v2, with py3 only support, and maybe changing a bit the interface? (I don't like so much the way you initialise things in qe-tools now, where it guesses what you want depending if the string you pass is a valid absolute path, otherwise it falls back to parse the string itself). Suggestions welcome there.

greschd:
> I suggest this could be a good moment to release a v2, with py3 only support, and maybe changing a bit the interface?

AFAIK `aiida-quantumespesso` still maintains py2 compatibility for now. How about we proceed with it here, then polish it for inclusion in `qe_tools` and replace it here once that's released and compatible and `aiida-quantumespresso` drops py2?

For inclusion in `qe_tools`, probably it should also be able to change between the `celldm` and `A`, `B`, ... output formats.

sphuber:
Indeed, we first want to release `aiida-quantumespresso==3.0.0` with python 2 support, because that is the first official release that is both Python 2/3 and AiiDA 1.0 compatibile. However, soon after I was planning to drop Python 2 and require `aiida-core>=1.2` as I need certain features from those releases for advanced functionality. We will then only provide minimum support for `3.0.0` series and just continue with what will become `3.1` or `4.0`, to be decided

greschd:
Ok, what's your opinion on the interface of this PR, specifically the way specifying `ibrav` works with the single absolute tolerance on the cell?

If you're fine with this I will add tests and documentation.

sphuber:
I don't have the time for it today, unfortunately, sorry. I am live demo'ing AiiDA in two hours and then have to prepare a proposal. Will try to look at it soon

greschd:
No worries, it's not an urgent thing.

giovannipizzi:
@greschd yeah sorry, you're right.
Then it's just a matter of moving the relevant code into qe-tools that still has py2+py3 compatibility.
Feel free to do a PR there! If you want I can also give you admin rights there so you can make releases yourself.

giovannipizzi:
I guess we should first release qe_tools before we can finalise this, right?

greschd:
That's the plan :smile:

greschd:
@sphuber what's the best place to put the documentation for this feature? Maybe a subsection in https://aiida-quantumespresso.readthedocs.io/en/latest/user_guide/get_started/examples/pw_tutorial.html#structure could work, but I'm not sure if there might be a better place.

sphuber:
To be honest, I haven't yet been able to look at the documentation and it badly needs to be improved I feel. So I don't have an opinion really. Just put it wherever you feel is best for now.

greschd:
Ok, I've added a short paragraph in `docs/source/user_guide/calculation_plugins/pw.rst` and `docs/source/user_guide/calculation_plugins/cp.rst`.

greschd:
Rebased on the latest `develop` and ready to go!

aiida_quantumespresso/calculations/__init__.py
```diff
@@ -285,10 +288,16 @@ def _generate_PWCPinputdata(cls, parameters, settings, pseudos, structure, kpoin
...
         # ============ I prepare the input site data =============
         # ------------ CELL_PARAMETERS -----------
-        cell_parameters_card = 'CELL_PARAMETERS angstrom\n'
-        for vector in structure.cell:
-            cell_parameters_card += ('{0:18.10f} {1:18.10f} {2:18.10f}'
-                                     '\n'.format(*vector))
+
+        # Specify cell parameters only if 'ibrav' is either zero or
+        # unspecified.
+        if input_params.get('SYSTEM', {}).get('ibrav', 0) == 0:
```
sphuber, line 294: maybe define a `DEFAULT_IBRAV = 0` in the class and use that in this line as well as in line 460. Not likely that we will change the default, but it does make it a bit clearer what this `0` is.

docs/source/user_guide/calculation_plugins/cp.rst
```diff
@@ -48,6 +47,8 @@ Inputs
 
   Those keywords should not be specified, otherwise the submission will fail.
 
+  The `SYSTEM`, `ibrav` keyword is optional. If it is not specified, `ibrav=0` is used. When a non-zero `ibrav` is given, `aiida-quantumespresso` automatically extracts the cell parameters. As a consistency check, the cell is re-constructed from these parameters and compared to the input cell. The input structure needs to match the convention detailed in the `pw.x documentation <https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm199>`_. The tolerance used in this check can be adjusted with the `IBRAV_CELL_TOLERANCE` key in the `settings` dictionary. It defines the absolute tolerance on each element of the cell matrix.
```
sphuber, line 50: Just for my understanding, the consistency check on the parameters is done by the `get_parameters_from_cell` function as well, correct?

setup.json
```diff
@@ -92,7 +92,9 @@
         "aiida_core[atomic_tools]~=1.3",
         "packaging",
         "qe-tools~=2.0rc1",
-        "xmlschema~=1.2"
+        "xmlschema~=1.2",
+        "numpy",
+        "scipy"
```
sphuber, line 97: where are we using `scipy`?

tests/calculations/test_pw.py
```diff
@@ -30,3 +36,86 @@ def test_pw_default(fixture_sandbox, generate_calc_job, generate_inputs_pw, file
...
+
+    with fixture_sandbox.open('aiida.in') as handle:
+        input_written = handle.read()
+
+    # Checks on the files written to the sandbox folder as raw input
+    assert sorted(fixture_sandbox.get_content_list()) == sorted(['aiida.in', 'pseudo', 'out'])
+    file_regression.check(input_written, encoding='utf-8', extension='.in')
+
+
+def test_pw_wrong_ibrav(fixture_sandbox, generate_calc_job, fixture_code, generate_kpoints_mesh, generate_upf_data):
```
sphuber, line 93: Would it be an idea to add a test for a structure that fails the consistency check because the tolerance is not met and then you test that setting `IBRAV_CELL_TOLERANCE` with looser tolerance fixes it?

sphuber (changes requested):
Thanks @greschd . Looking good. Just some minor suggestions and request to add a test.

docs/source/user_guide/calculation_plugins/cp.rst
greschd: Exactly, that's here: https://github.com/aiidateam/qe-tools/blob/develop/qe_tools/converters/_structure.py#L68-L72

aiida_quantumespresso/calculations/__init__.py
greschd: Ah wait.. this logic actually works only if the default is zero - because for non-zero `ibrav` you can't specify the cell in the same way.

I agree that we want to make it clearer what this "magic zero" is, but I think just naming it `DEFAULT_IBRAV` is also dangerous: Simply changing the default will break the code.

setup.json
greschd: Whoops, that part has moved into `qe-tools` - removed it now.

aiida_quantumespresso/calculations/__init__.py
greschd: Hmm, I guess keeping the right-hand-side zero hard-coded and making the left one a variable works.

sphuber: good point. Just assigning it to variable to make its meaning clear is enough

tests/calculations/test_pw.py
greschd: Aand done!

sphuber (approved):

PR #508, opened 2020-04-22
Update dependency requirment `pgtest=~1.3,>=1.3.1`

### LINKED ISSUE #507: outdated pgtest dependency
the pgtest dependency is fixed to 1.2.0 which has issues with newer ubuntu versions
https://github.com/aiidateam/aiida-quantumespresso/blob/develop/setup.json#L71

please upgrade to the latest one

sphuber:
Fixes #507 

Older versions have bugs on Ubuntu systems.

### DIFF
--- a/setup.json
+++ b/setup.json
@@ -68,7 +68,7 @@
             "prospector==1.1.7",
             "pylint==1.9.4; python_version<'3.0'",
             "pylint==2.3.1; python_version>='3.0'",
-            "pgtest==1.2.0",
+            "pgtest~=1.3,>=1.3.1",
             "pytest==3.6.3",
             "pytest-regressions==1.0.6",
             "yapf==0.28.0"


setup.json
```diff
@@ -68,7 +68,7 @@
             "prospector==1.1.7",
             "pylint==1.9.4; python_version<'3.0'",
             "pylint==2.3.1; python_version>='3.0'",
-            "pgtest==1.2.0",
+            "pgtest~=1.3,>=1.3.1",
```
greschd, line 71: Are you sure ``1.3.1`` fixes all the bugs, and ``1.3.2`` is not required?

sphuber: This is what we use for `aiida-core` at least

greschd: @ltalirz should know

sphuber: Looking at the [commits](https://github.com/jamesnunn/pgtest/commits/master) after `1.3.1` they do not seem like critical fixes

greschd: Ah, from looking at the diff 1.3.1 could potentially raise some `ResourceWarning`s, but that shouldn't break anything.

ltalirz: yeah... 1.3.2 is safer but 1.3.1 will still work.

PR #515, opened 2020-05-07
Drop Python 2 support

### LINKED ISSUE #514: Remove Python 2 compatibility
Now that `v3.0.0` has been released as the first official release to be compatible with Python 2 and `aiida-core==1.0.0`, we can now proceed to drop Python 2 for what will become `v3.1.0`.

sphuber:
Fixes #514 

The following commands were executed to clean all source files:

    sed -i '/from __future__.*$/d' {} +
    sed -i 's/(object):/:/g' {} +
    sed -i 's/super([^)]*,[^)]*)/super()/g' {} +
    sed -i '/\(import six\|from six\).*$/d' {} +
    sed -i '/\(six\.add_metaclass\).*$/d' {} +
    sed -i 's/six.integer_types/int/g' {} +
    sed -i 's/six.string_types/str/g' {} +
    sed -i 's/six.text_type/str/g' {} +
    sed -i 's/six.binary_type/bytes/g' {} +
    sed -i 's/six.moves.StringIO/io.StringIO/g' {} +
    sed -i 's/six.StringIO/io.StringIO/g' {} +
    sed -i 's/six.moves.BytesIO/io.BytesIO/g' {} +
    sed -i 's/six.BytesIO/io.BytesIO/g' {} +
    sed -i 's/six.moves.range/range/g' {} +
    sed -i 's|six.iterkeys(\([^)]*\))|\1.keys()|' {} +
    sed -i 's|six.iteritems(\([^)]*\))|\1.items()|' {} +
    sed -i 's|six.itervalues(\([^)]*\))|\1.values()|' {} +

The operations were applied to all files using `find -exec`:

    find . -type f -not -path './.git*' -exec

Note the exclusion of the `.git` folder which is very important.


CHANGELOG.md
```diff
@@ -1,6 +1,4 @@
-## v3.0.0:
```
greschd, line 1: Shouldn't there still be a changelog entry for 3.0.0?

aiida_quantumespresso/parsers/parse_raw/neb.py
```diff
@@ -5,15 +5,10 @@
...
 decision doesn't have much structure encoded, [the values are simple ]
 """
-from __future__ import absolute_import
-import xml.dom.minidom
-import os
-import string
+from qe_tools.constants import bohr_to_ang
+
 from aiida_quantumespresso.parsers import QEOutputParsingError, get_parser_info
-from qe_tools.constants import ry_to_ev,hartree_to_ev,bohr_to_ang,ry_si,bohr_si
```
greschd, line 13: Here a few more imports vanished - were they just unused?

setup.json
```diff
@@ -86,14 +83,13 @@
     },
     "install_requires": [
```
greschd, line 84: I think we were missing the `six` requirement before 😅

tests/parsers/test_pw2gw.py
```diff
@@ -27,7 +25,7 @@ def test_pw2gw_default(
     data_regression.check({'output_parameters': results['output_parameters'].get_dict()},
                           basename='test_pw2gw_default_data')
 
-    num_regression.check({k: v for k, v in results['eps'].get_iterarrays()}, basename='test_pw2gw_default_eps')
+    num_regression.check(results['eps'].get_iterarrays(), basename='test_pw2gw_default_eps')
```
greschd, line 28: I think `.get_iterarrays()` is a generator - does `num_regression` iterate over it automatically? If not, this change might need reverting.

aiida_quantumespresso/parsers/parse_raw/neb.py
sphuber: yes

tests/parsers/test_pw2gw.py
sphuber: wrapping in `dict()` should also work perhaps no?

greschd:
It seems the `pre-commit` 2.X series supports only py36+

CHANGELOG.md
sphuber: I "forgot" for the release, but honesty, the changes were so enormous that it would be a) too much work and b) it should just be considered a new software package essentially. MIght fix it once we do a `v3.0.1` to at least state this instead of having the todo, but I wasn't planning on actually trying to write the changelog

greschd: Ha, fair enough 👍 Maybe you can just condense the "backwards-incompatible changes" issue?

tests/parsers/test_pw2gw.py
greschd: Absolutely.

sphuber:
> It seems the `pre-commit` 2.X series supports only py36+

yeah, in my other projects I therefore split the `dev` extra into `tests` and `pre-commit`. That way I can just run `pre-commit` on CI with python 3.8 with latest packages and still run `tests` with python 3.5. For now I just downgraded `pre-commit` because I don't want to mix those extra changes in this huge PR. I will do that straight after this is merged with some additional CI related stuff

greschd:
Yeah I think that's good practice anyway - the fewer extra dependencies are installed when testing, the less likely to forget something that should be `install_requires`.

CHANGELOG.md
greschd:
```
## v3.0.0:

The fire nation attacked.
```

sphuber: Yeah, I thought about including the items from that issues, but the problem is that it is so horribly incomplete, that maybe having it there does more damage than good. If people see that short list they think "oh great" and then there is a million more incompatible changes. Anyway, the previous version `v2.1.0` was python 2 and `aiida-core==0.12.0` only. Doubt that there are people on those versions that would really consider upgrading.

greschd:
image
Might need a change in branch protection settings.

CHANGELOG.md
greschd: Ooh that's right.. most people probably upgrade from some alpha version anyway.

PR #519, opened 2020-05-07
Replace builtin `BaseRestartWorkChain` with the one from `aiida-core`

### LINKED ISSUE #517: Replace `BaseRestartWorkChain` wtih version shipped with `aiida-core`
Now that the `BaseRestartWorkChain` has been moved to `aiida-core`, we should use that one instead of the original one defined in this plugin.

sphuber:
Fixes #517 

As of `aiida-core==1.1.0` it ships with the `BaseRestartWorkChain` built
in. This version is based directly on the original concept developed in
this plugin, with some minor interface changes and improvements. Here we
remove the old implementation and depend on the version that ships with
`aiida-core`. The base workchains are updated to respected the new
interface and to profit from the new features.

The dependency requirement is updated to `aiida-core~=1.2`, since anyway
soon we will have to rely on features introduced in that version. This
also allows us to simplify the backwards compatibility code that was put
in the base `CalcJob` class for this plugin.


sphuber:
@greschd here we start to notice the absence of tests for the workchains. I manually ran the `PwBaseWorkChain` and `PhBaseWorkChain` and they worked, but that is of course for a successful run. I will maybe experiment a bit now to try and create some workchain fixtures that allow to write tests easily

greschd:
Ah, that reminds me I still have quite a few TODOs on https://github.com/aiidateam/aiida-testing

sphuber:
@greschd I added some testing for the 4 base restart work chains.

aiida_quantumespresso/workflows/matdyn/base.py
```diff
@@ -54,10 +52,9 @@ def report_error_handled(self, calculation, action):
...
-    """Handle calculations with an exit status below 400 which are unrecoverable, so abort the work chain."""
-    if calculation.exit_status < 400:
-        self.report_error_handled(calculation, 'unrecoverable error, aborting...')
-        return ErrorHandlerReport(True, True, self.exit_codes.ERROR_UNRECOVERABLE_FAILURE)
+    @process_handler(priority=600)
+    def handle_unrecoverable_failure(self, node):
+        """Handle calculations with an exit status below 400 which are unrecoverable, so abort the work chain."""
+        if node.is_failed and node.exit_status < 400:
+            self.report_error_handled(node, 'unrecoverable error, aborting...')
+            return ProcessHandlerReport(True, self.exit_codes.ERROR_UNRECOVERABLE_FAILURE)
```
greschd, line 60: I think this would be a bit easier to understand if the parameters are passed by keyword (``do_break=True, exit_code=...``).

sphuber: In a sense yes, but since the `ProcessHandlerReport` is now the standard and shipped with `aiida-core`, it should be clear at some point. I will see how it looks, but merge this for now

sphuber: And thanks for the review! :+1:

PR #520, opened 2020-05-07
Move calcfunctions to `aiida_quantumespresso.calculations.functions`

### LINKED ISSUE #518: Move calculation functions to `aiida_quantumespresso.calculations`
Currently they are stored in `aiida_quantumespresso.workflows` but this is inapt since they are `calcfunctions` and not `workfunctions`.

sphuber:
Fixes #518 

Since they are calculation functions they do not belong in the workflow
module. To not break backwards-compatibility the original modules import
the functions from their new location and emit a `Warning`. Note that we
explicitly do not use a `DeprecationWarning` since those are
automatically filtered in Python 3.

### DIFF
--- a/aiida_quantumespresso/workflows/functions/create_kpoints_from_distance.py
+++ b/aiida_quantumespresso/workflows/functions/create_kpoints_from_distance.py
@@ -1,37 +1,12 @@
 # -*- coding: utf-8 -*-
 """Calculation function to compute a k-point mesh for a structure with a guaranteed minimum k-point distance."""
-from aiida.engine import calcfunction
+# pylint: disable=unused-import
+import warnings
+from aiida_quantumespresso.calculations.functions.create_kpoints_from_distance import create_kpoints_from_distance
 
-
-@calcfunction
-def create_kpoints_from_distance(structure, distance, force_parity):
-    """Generate a uniformly spaced kpoint mesh for a given structure.
-
-    The spacing between kpoints in reciprocal space is guaranteed to be at least the defined distance.
-
-    :param structure: the StructureData to which the mesh should apply
-    :param distance: a Float with the desired distance between kpoints in reciprocal space
-    :param force_parity: a Bool to specify whether the generated mesh should maintain parity
-    :returns: a KpointsData with the generated mesh
-    """
-    from numpy import linalg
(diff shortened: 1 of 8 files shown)


greschd:
I think that should be a `FutureWarning`, based on how it's described here: https://docs.python.org/3/library/exceptions.html#FutureWarning. Of course `DeprecationWarning` will still be shown e.g. by `pytest`, but I agree we are closer to an "end-user" scenario here, since we don't expect all code using it to run through a test suite.

greschd:
The test failure is probably on me, the test should probably filter out calcfunctions 😅

sphuber:
I think the way the words "user" and "developer" are being used by the Python docs is a bit different from our usage though. I want this warning to be seen by people using the `aiida-quantumespresso` plugin, but really they will be "developing" their own plugin or at least AiiDA-based code. I am guessing their reasoning is that developers will be reading the source code of libraries they are using (optimistic but okay) and so they will see it there and therefore it can be suppressed by default. My interpretation of the `FutureWarning` is that if you need to warn about deprecations that will even reach end-users. But for this change here, I need to reach everyone. But anyway, in the end it does not really matter. I will use `FutureWarning` given that it is more specific than just a general `Warning`. Will also fix the test :)

sphuber:
@greschd fixed the tests and change the `Warning` to `FutureWarning`

aiida_quantumespresso/workflows/functions/create_kpoints_from_distance.py
```diff
@@ -1,37 +1,10 @@
...
-        nkpoints = max(lengths_kpoint)
-        kpoints.set_kpoints_mesh([nkpoints, nkpoints, nkpoints])
-
-    return kpoints
+# pylint: disable=unused-import
+import warnings
+from aiida_quantumespresso.calculations.functions.create_kpoints_from_distance import create_kpoints_from_distance
+
+warnings.warn(
+    'This module is deprecated and will be removed soon. Please use thw module'
```
greschd, line 8: typo: thw -> the

Should we mention that the entry points could also be used?

sphuber: Done and done

PR #524, opened 2020-06-04
Fix import or `URLError` which has been removed in `xmlschema`

### LINKED ISSUE #523: Latest version of `xmlschema` breaks the `PwParser`
Due to a removed import `from xmlschema.exceptions import URLError`

sphuber:
Fixes #523 

As of `xmlschema==1.2.0` the `xmlschema.exceptions.URLError` exception
has been removed. This was just an alias for Python 2 and 3 compatibility
reasons, which now that Python 2 support is dropped can be removed.
Simply using `from urllib.error import URLError` addressed the problem.


sphuber:
I had already started this but then realized that `3.0.0` was released with `xmlschema==1.0.13` so it should not break

greschd:
Ah, perfect :+1:

PR #526, opened 2020-06-04
Remove `prospector` and run `yapf` and `pylint` from repo

### LINKED ISSUE #525: Change yapf requirement back to strict match?
In #513, the `yapf` requirement was changed to `~=0.29`  - which I didn't notice when reviewing 😅

I think in the case of `yapf`, the requirement should actually be a strict one. The problem with the compatible release clause is that formatting can change between different versions of `yapf`, which can lead to:
- CI installing a newer version than the developer, and subsequently failing the pre-commit step
- a "ping-pong" of formatting between different developers that don't have the same version installed

sphuber:
Fixes #525 

The `prospector` tools has been giving a lot of problems with dependency
incompatibilities since we have been using it and we really only use it
to run `pylint`. However, this is easy to run directly as well from a
remote, specifying the desired revision.

We do the same for `yapf` and remove the explicit dependency from the
`pre-commit` extra since the latter was not pinning a particular version
which could cause arbitrary style changes in different environments that
just happen to have a slightly different version.

### DIFF
--- a/setup.json
+++ b/setup.json
@@ -68,8 +68,7 @@
     "extras_require": {
         "pre-commit": [
             "pre-commit~=2.2",
-            "prospector~=1.2",
-            "yapf~=0.29"
+            "pylint~=2.5.0"
         ],
         "tests": [
             "pgtest~=1.3",
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,10 +12,10 @@ repos:
                 tests/.*.in$
             )$
 
--   repo: https://github.com/PyCQA/prospector
-    rev: 1.2.0
+-   repo: https://github.com/PyCQA/pylint
+    rev: pylint-2.5.2
     hooks:
-    -   id: prospector
+    -   id: pylint
         language: system
         exclude: &exclude_files >
             (?x)^(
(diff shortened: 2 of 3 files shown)


setup.json
```diff
@@ -68,8 +68,7 @@
     "extras_require": {
         "pre-commit": [
             "pre-commit~=2.2",
-            "prospector~=1.2",
-            "yapf~=0.29"
+            "pylint~=2.5"
```
greschd, line 71: Here, `pylint` also needs to be at a specific version (I guess ``~=2.5.0`` would also be ok, allowing bugfix-release changes).

The reason for this is that the ``language`` specified above is ``system``: This means that pylint won't actually be installed from the source specified in ``.pre-commit-config.yaml``, only the _hook_ specification from there is used. Instead, it uses the system-installed (or usually: virtualenv-installed) ``pylint``. You can check this by intentionally breaking ``.virtualenvs/<venv_name>/bin/pylint``.

The reason we actually _need_ to use the ``system`` language is because ``pylint`` needs access to all the other packages in the virtualenv to do its analysis properly.

greschd: You don't wanna know how long it took me to figure this out the first time around 😅

sphuber: Ah yes, I also broke my head over this and then forgot why I also needed to specify it in the `extras_require`. It really is a shitty solution. Do you know if this is a problem of `pylint` or just of `pre-commit` and so there will be no away around it?

greschd: I think that's a requirement of ``pylint``.. it's even mentioned in the docs of ``pre-commit`` as to why one might use the ``system`` language: https://pre-commit.com/#system

Why ``pylint`` itself has ``python`` as a language in its hook definition is beyond me.

PR #527, opened 2020-06-05
Use 'get_hash' instead of 'id' when comparing UpfData in test.

greschd:
In the 'immigrate' tests, the ``UpfData`` for Si was compared by id. When using an AIIDA_TEST_PROFILE, this led to errors because a different `UpfData` of the same kind could already be present, and the immigration did not necessarily pick the one created for the test.
To make the test independent of the exact PK, they are now compared by hash.

### DIFF
--- a/tests/tools/test_immigrate.py
+++ b/tests/tools/test_immigrate.py
@@ -39,7 +39,9 @@ def test_create_builder(
 
     assert builder['code'] == code
     assert builder['metadata'] == metadata
-    assert builder['pseudos']['Si'].id == si_upf.id
+    pseudo_hash = si_upf.get_hash()
+    assert pseudo_hash is not None
+    assert builder['pseudos']['Si'].get_hash() == pseudo_hash
     assert builder['parameters'].get_dict() == {
         'CONTROL': {
             'calculation': 'scf',


sphuber:
Shouldn't the database be empty when running the tests though? Where is this other `UpfData` coming from?

greschd:
Is it necessarily empty when using ``AIIDA_TEST_PROFILE``? I thought that didn't clear the DB, but might be wrong.

sphuber:
Well, test profiles should be expected to be cleaned. I think in this case it is better to just clear the database before the test since it should not be influenced by previously existing data. However, I just looked at the test and I think that your change is correct nonetheless, just for a different reason :smile:  In the real use case, there will probably be no existing upf and the point of the immigrator is that it should create by parsing it from the input files. So what we should test here is exactly that the reconstructed pseudo has the same content as the original one.

TLDR: this change is good just for a different reason

sphuber:
> I think wherever possible, we shouldn't _rely_ on the test DB to be cleaned, because it's expensive to do so between each test.

Fair enough, but I think it is a trade off. Cleaning an almost empty database is relatively fast still and often when having to take account that the db might not be clean, complicates the test logic making it more difficult to write and understand. But I agree, if it can reasonably be prevented to clean, then we shouldn't

greschd:
Absolutely agreed :+1:

PR #529, opened 2020-06-15
Update to qe-tools 2.0.0rc1.

greschd:
Updating the `qe-tools` dependency to the release candidate version `2.0.0rc1`.

Significant changes:
- The constants defined previously in `qe_tools.constants` are now a SimpleNamespace, importable as `qe_tools.CONSTANTS`. 
- The `PwInputFile` no longer accepts file handles or file names as input, just a single string containing the file contents. This is propagated through to the aiida-quantumespresso `InputFile` classes. 
@sphuber is it ok to change this also here, or should we add an ``__init__`` to ``pwinputparse.PwInputFile`` that implements the current behavior?

### DIFF
--- a/setup.json
+++ b/setup.json
@@ -91,7 +91,7 @@
     "install_requires": [
         "aiida_core[atomic_tools]~=1.2",
         "packaging",
-        "qe-tools~=1.1",
+        "qe-tools~=2.0rc1",
         "xmlschema~=1.2"
     ],
     "license": "MIT License",
--- a/aiida_quantumespresso/calculations/pwimmigrant.py
+++ b/aiida_quantumespresso/calculations/pwimmigrant.py
@@ -199,7 +199,7 @@ class PwimmigrantCalculation(PwCalculation):
             # Parse the input file.
             local_path = os.path.join(folder.abspath, self._INPUT_FILE_NAME)
             with open(local_path) as fin:
-                pwinputfile = pwinputparser.PwInputFile(fin)
+                pwinputfile = pwinputparser.PwInputFile(fin.read())
 
             # Determine PREFIX, if it hasn't already been set by the user.
             if self._PREFIX is None:
--- a/aiida_quantumespresso/cli/data/structure.py
+++ b/aiida_quantumespresso/cli/data/structure.py
@@ -21,7 +21,8 @@ def cmd_import(filename, dry_run):
     """Import a `StructureData` from a Quantum ESPRESSO input file."""
     from aiida_quantumespresso.tools.pwinputparser import PwInputFile
 
-    parser = PwInputFile(filename)
+    with open(filename, 'r') as input_file:
+        parser = PwInputFile(input_file.read())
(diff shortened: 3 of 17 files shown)


setup.json
```diff
@@ -87,7 +87,7 @@
     "install_requires": [
         "aiida_core[atomic_tools]~=1.2",
         "packaging",
-        "qe-tools~=1.1",
+        "qe-tools~=2.0rc1",
```
greschd, line 90: Note that this assumes we won't break backwards compatibility between the current release candidate and the "proper" release of `qe-tools`. I think that will be the case, but maybe we should pin `rc1` here just to be safe.

greschd:
@sphuber when you have time, could you take a look at this? If you'd prefer a fully released version of `qe-tools`, I can do that also. Mainly, I'm waiting with that to see if there's any feedback / changes needed to `qe-tools` from this PR - it's a chicken-egg problem 😄.

sphuber:
Hey @greschd sorry for the delay. I think it would be better to wait for `2.0.0`. If there really turns out to be a problem, it is easier to just release a patch version. We will just specify `qe-tools~=2.0` such that it will automatically get installed. Experience with relying on release candidates is not the best.

greschd:
Would ``~=2.0rc1`` also work? If we just specify ``~=2.0``, the release candidate isn't considered as a valid match.

As far as releasing ``qe-tools==2.0.0``, I don't think there is too much left to do. There's the constants issue https://github.com/aiidateam/qe-tools/issues/25, and then it would be nice to add a bit of documentation (but that's not necessarily blocking).

What do think about the interface changes in `qe-tools`, from the perspective of ``aiida-quantumespresso``?

sphuber:
@greschd what is the status of `qe-tools`? Is `2.0.0` already out? I was thinking of making the `3.2.0` release for `aiida-quantumespresso`, such that we can drop Python 3.5 for the next one. Just looking at what PRs we still want to include

greschd:
Well, this is a bit of a chicken-egg problem: I'd want to get feedback on this PR before releasing `2.0.0` at least. In principle there are then still [some issues labeled `2.0.0`](https://github.com/aiidateam/qe-tools/issues?q=is%3Aopen+is%3Aissue+milestone%3Av2.0), but I'm not sure if any of them are really blocking. 

I'm not sure if [the "inconsistencies in physical constants"](https://github.com/aiidateam/qe-tools/issues/25) issue would still need fixing before release. The interface is now adapted so that we can handle multiple QE versions, so we could also fix the inconsistency itself later.

greschd:
It would definitely be nice if we could get this PR and #503 (blocked only by this one) in though - if nothing else, so I can cross them off my mental to-do list 😄

sphuber:
Right, if you can update the branch and resolve the conflict I will merge this. Let's get this show on the road

greschd:
Ready to rock 🤘

sphuber (approved):
Rock on

PR #532, opened 2020-06-25
pp plugin: retrieve and parse multiple output files.

### LINKED ISSUE #530: PP plugin: retrieve multiple files.
Before #428 was merged it was possible to define an `additional_retrieve_list` setting to retrieve more files. This is needed, for example, if one wants to download multipe output files which are produced by plot orbitals run:

```
&INPUTPP
  filplot = 'aiida.filplot'
  kband(1) = 12
  kband(2) = 12
  kpoint(1) = 1
  kpoint(2) = 24
  outdir = './out/'
  plot_num = 7
  prefix = 'aiida'
/
&PLOT
  fileout = 'aiida.fileout'
  iflag = 3
  output_format = 6
/
```

In that case, `pp.x` will produce a number of files named as follows: `aiida.filplot_K***_B***aiida.fileout`. 

The possible solution for this is to either return `additional_retrieve_list` setting back or to define a wildcard instead of https://github.com/aiidateam/aiida-quantumespresso/blob/5e06a62756cdbae23f287a7ac9abe51a2a0492c6/aiida_quantumespresso/calculations/pp.py#L217-L220

yakutovicha:
fixes #530

### DIFF
--- a/aiida_quantumespresso/parsers/pp.py
+++ b/aiida_quantumespresso/parsers/pp.py
@@ -71,58 +71,66 @@ class PpParser(Parser):
         except (IOError, OSError):
             return self.exit_codes.ERROR_OUTPUT_STDOUT_READ
 
-        # The post-processed data should have been written to file, either in the retrieved or temp list
-        filename_data = PpCalculation._FILEOUT
-        if filename_data in self.retrieved.list_object_names():  # Retrieved list case
-            try:
-                data_raw = self.retrieved.get_object_content(filename_data)
-            except (IOError, OSError):
-                return self.exit_codes.ERROR_DATAFILE_READ
-        elif temp_folder_path is not None:  # Temp list case
-            data_file_path = os.path.join(temp_folder_path, filename_data)
-            if os.path.isfile(data_file_path):
-                try:
-                    with open(data_file_path, 'r') as fhandle:
-                        data_raw = fhandle.read()
-                except (IOError, OSError):
-                    return self.exit_codes.ERROR_DATAFILE_READ
-            else:
-                return self.exit_codes.ERROR_OUTPUT_DATAFILE_MISSING
-        else:
-            return self.exit_codes.ERROR_OUTPUT_DATAFILE_MISSING
-
-        # Parse stdout
+        # Parse stdout.
         try:
             logs, self.output_parameters = self.parse_stdout(stdout_raw)
         except Exception:
             self.logger.error(traceback.format_exc())
             return self.exit_codes.ERROR_UNEXPECTED_PARSER_EXCEPTION
 
-        # Print the logs
+        # Print the logs.
         self.emit_logs(logs)
 
-        # Scan logs for known errors
+        # Scan logs for known errors.
         if 'ERROR_PARENT_XML_MISSING' in logs['error']:
             return self.exit_codes.ERROR_PARENT_XML_MISSING
         if 'ERROR_OUTPUT_STDOUT_INCOMPLETE' in logs['error']:
             return self.exit_codes.ERROR_OUTPUT_STDOUT_INCOMPLETE
 
-        # Parse the post-processed-data according to what kind of data file was produced
-        if self.output_parameters['output_format'] == 'gnuplot':
-            if self.output_parameters['plot_type'] == '2D polar on a sphere':
-                parsed_data = self.parse_gnuplot_polar(data_raw)
-            else:
-                parsed_data = self.parse_gnuplot1D(data_raw)
-        elif self.output_parameters['output_format'] == 'gnuplot x,y,f':
-            parsed_data = self.parse_gnuplot2D(data_raw)
-        elif self.output_parameters['output_format'] == 'Gaussian cube':
-            parsed_data = self.parse_gaussian(data_raw)
-        else:
-            return self.exit_codes.ERROR_UNSUPPORTED_DATAFILE_FORMAT
-
-        # Create output nodes
-        self.out('output_data', parsed_data)
         self.out('output_parameters', orm.Dict(dict=self.output_parameters))
+
+        # The post-processed data should have been written to file, either in the retrieved or temp list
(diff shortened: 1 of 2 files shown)


aiida_quantumespresso/parsers/pp.py
```diff
@@ -71,58 +71,66 @@ def parse(self, **kwargs):
...
+        datafile_missing = True
+
+        for filename_data in self.retrieved.list_object_names():  # Retrieved list case
+            if PpCalculation._FILEOUT in filename_data:
+                datafile_missing = False
+                try:
+                    data_raw = self.retrieved.get_object_content(filename_data)
+                except (IOError, OSError):
+                    return self.exit_codes.ERROR_DATAFILE_READ
+            elif temp_folder_path is not None:  # Temp list case
```
greschd, line 102: That logic seems wrong - due to the for-loop above, it will only check the `filename_data` that are in the `retrieved` list, not the ones in the temp list.

greschd:
Not sure if I read the code right: is the output name going to be something like ``aiida_filplot_K***_B***aiida_fileout``? Since (I think) the start and end of that will always be the same, I think stripping that would allow for nicer names, e.g. ``output_data_K***_B***``.

Also, for backwards compatibility, we should probably keep the ``output_data`` output for the case of a single output file. It's a bit ugly, but I think (@sphuber may correct me there) the `3.x` series is supposed to be backwards-compatible.

aiida_quantumespresso/parsers/pp.py
yakutovicha: ah, right, thanks for catching

sphuber: Also, aren't you overriding `data_raw` every time you read a file?

sphuber:
Thanks @yakutovicha . I am trying to understand the logic for the generation of the output files by `pp.x` to understand what the correct and most flexible solution is, but I cannot find anything in the [documentation](https://www.quantum-espresso.org/Doc/INPUT_PP.html). As I understood it `pp.x` will first pre-process output data from a completed `pw.x` calculation and write it to a single file defined by `filplot`. The optional second step will then write the content of this file into another file with a specific format that is machine independent and that can be easily plotted with standard plotting tools.

So my first question is: why is `pp.x` generating multiple files here and is this the only case or are there other `plot_num` values that produce more than just the single file. Second, what do these files contain? Are these produced by the first step and so are they normal ascii files? Do they need to be parsed by the plugin, such as the `filout` currently is being done, or do they just need to be retrieved?

yakutovicha:
I am not a big expert of `pp.x`, but this is what I understand.

> Thanks @yakutovicha . I am trying to understand the logic for the generation of the output files by `pp.x` to understand what the correct and most flexible solution is, but I cannot find anything in the [documentation](https://www.quantum-espresso.org/Doc/INPUT_PP.html). As I understood it `pp.x` will first pre-process output data from a completed `pw.x` calculation and write it to a single file defined by `filplot`.

Not necessarily a single file, see for instance:
```shell
$ ls
aiida.filplot_K001_B012               aiida.filplot_K010_B012aiida.fileout  aiida.filplot_K020_B012
aiida.filplot_K001_B012aiida.fileout  aiida.filplot_K011_B012               aiida.filplot_K020_B012aiida.fileout
aiida.filplot_K002_B012               aiida.filplot_K011_B012aiida.fileout  aiida.filplot_K021_B012
```

> The optional second step will then write the content of this file into another file with a specific format that is machine independent and that can be easily plotted with standard plotting tools.

Sounds right to me


> So my first question is: why is `pp.x` generating multiple files here and is this the only case or are there other `plot_num` values that produce more than just the single file.

I guess this is done because it is very tedious to run pp.x for every orbital you want to export since most of the time is spent on reading the density matrix. I wouldn't like to run pp.x 100 times to export all orbitals that correspond to every combination of kpoint/band.

It is described in the documentation you were referring to:
```

kpoint(i), i=1,2 | INTEGER

    Unpolarized and noncollinear case: 
                k-point(s) to be plotted

            LSDA: 
               k-point(s) and spin polarization to be plotted
               (spin-up and spin-down correspond to different k-points!)

            To plot a single kpoint ikpt, specify kpoint=ikpt or kpoint(1)=ikpt
            To plot a range of kpoints [imin, imax], specify kpoint(1)=imin and kpoint(2)=imax
```


> Second, what do these files contain? Are these produced by the first step and so are they normal ascii files?

They are Gaussian-cube files that are machine-independent. They contain 3d representation of the orbitals.

> Do they need to be parsed by the plugin, such as the `filout` currently is being done, or do they just need to be retrieved?

I have no idea. I see that in the current state Hartree potential is also parsed (to save disk space?), while it can be used as is. I believe we should either parse everything or nothing.

mentioning @cpignedoli as he has more experience with `pp.x`.

sphuber:
ok, so if I understand correctly, here there are still "just" a `filplot` and a `filout` being created, but now there are multiple of these pairs. I just ran an example calculation and at least for `plot_num=7`, it seems like the data files get filenames where `filplot` is the prefix and then there is some suffix to indicate the kpoint and band number. The plot output files get the full filename of the data file as prefix, with the `filout` value as the suffix, e.g.:
```
filedata: aiida.filplot_K001_B001
fileplot: aiida.filplot_K001_B001aiida.fileout
```
It is difficult to say if this is the case for all combinations of parameters, but since the documentation is not clear about this, I think the only solution is to assume this for now.

greschd:
> 
> It is difficult to say if this is the case for all combinations of parameters, but since the documentation is not clear about this, I think the only solution is to assume this for now.

There _is_ another solution, it just involves looking at some Fortran code 🙄

sphuber:
> There _is_ another solution, it just involves looking at some Fortran code roll_eyes

So what you are saying is that there is no solution? :wink: 

Feel free to have a deep dive, but with all those combinations of flags I think getting a full picture might be quite tricky

PR #533, opened 2020-06-25
`PpCalculation`: add support for retrieving and parsing multiple files

### LINKED ISSUE #530: PP plugin: retrieve multiple files.

sphuber:
Fixes #530 

The implementation of the `PpCalculation` and `PpParser` assumed that
the code would always only produce on output file with pre-processed
data and one output file with the data formatted in a custom plot
format. However, for certain input parameter combinations, such as
`INPUTPP.plot_num = 7` where more than a single band is requested, a
pair of outputfiles is produced for each band. The filename of the data
file has `INPUTPP.filplot` as a prefix, with some kind of suffix to
distinguish it from the others. The corresponding plot file will use
that filename as a prefix with the `PLOT.fileout` value as a suffix.
For example, for the inputs:

    INPUTPP
        filplot = 'aiida.filplot'
    PLOT
        fileout = 'aiida.fileout'

The data files will be named `aiida.filplot_K1_B1` and the plot files
are formatted as `aiida.filplot_K1_B1aiida.fileout`.

To support this use case, the `PpCalculation` is updated to not just
retrieve a single file, but add a directive to the `retrieve_files` or
`retrieve_temporary_list` that contains the corresponding globbing
pattern. The `PpParser` then simply loops over the content of the
retrieved (temporary) folder and parses each file whose filename matches
the pattern described above. We assume that if there are more than one
file, they all have the exact same format and so can be parsed with the
same logic.

Since now there are potentially more than one parsed output `ArrayData`
node, the `output_data` port, which is not a namespace can not be used.
To keep backwards compatibility, we add the `output_data_multiple`
namespace, which is used if more than one output plot file is parsed.


sphuber:
@yakutovicha I created an alternative implementation including unit tests that now actually also test the retrieved temporary folder functionality. I also tested this locally, at least for `plot_num=7` which produces multiple files. But please try and give this a go.

@cpignedoli if you have the time, would be great if you could maybe give this branch a go as well to see if it fixes your use case, thanks.

cpignedoli:
I will for sure give it a try next week. Thanks a lot

aiida_quantumespresso/parsers/pp.py
```diff
@@ -71,34 +70,44 @@ def parse(self, **kwargs):
...
-                    with open(data_file_path, 'r') as fhandle:
-                        data_raw = fhandle.read()
-                except (IOError, OSError):
-                    return self.exit_codes.ERROR_DATAFILE_READ
-            else:
-                return self.exit_codes.ERROR_OUTPUT_DATAFILE_MISSING
+        data_raw = []
+
+        # Currently all plot output files should start with the `filplot` as prefix. If only one file was produced the
+        # prefix is the entire filename, but in the case of multiple files, there will
```
greschd, line 76: Seems like this comment is incompl

aiida_quantumespresso/parsers/pp.py
```diff
@@ -71,34 +70,44 @@ def parse(self, **kwargs):
...
+        filename_prefix = PpCalculation._FILPLOT
+        filename_suffix = PpCalculation._FILEOUT
+
+        # How to get the output filenames and how to open them, depends on whether they will have been retrieved in the
+        # `retrieved` output node, or in the `retrieved_temporary_folder`. Instead of having a conditional with almost
+        # the same loop logic in each branch, we apply a somewhat dirty trick to define an `opener` which is a callable
```
greschd, line 82: Dirty trick? That's fantastic! 👍

aiida_quantumespresso/parsers/pp.py
```diff
@@ -71,34 +70,44 @@ def parse(self, **kwargs):
...
+        # that will open a handle to the output file given a certain filename.
+        if retrieve_temporary_list:
```
greschd, line 84: Might be worth mentioning here that there can either be _all_ temporary files or _all_ retrieved, not a mix of the two.

I think this is a decent choice, but had to go check in the calculation that this is what's implemented there, too.

yakutovicha:
I tested with the fix you made in #534 and everythin went smooth. Thanks a lot @sphuber!

sphuber:
> Is there a case where multiple sub-output files are written but there is a single overall output file?
> If so, we may want to offer some user control over whether all sub-outputs are parsed.

I wasn't able to figure this out from the docs, but it looks like it should always be a pair of output files. So I guess for now this should work until we find another edge case I guess

yakutovicha:
I believe this code should definitely be changed:

```python
for line in data_lines:
     for i in range(0, len(line), 13):
        data_point = line[i:i + 13].strip()
        if data_point != '':
            datalist.append(float(data_point))
```

If `pp.x` decide to change the number formatting, the parser will fail.

A better approach would look something like:
```python
    data = np.empty(shape[0]*shape[1]*shape[2], dtype=float)
    cursor = 0
    for line in lies:
        ls = line.split()
        data[cursor:cursor+len(ls)] = ls
        cursor += len(ls)
```

sphuber:
@greschd this is ready for re-review

yakutovicha:
> @greschd this is ready for re-review

I am not @greschd but I believe this can be merged. I tested the plugin quite extensively with the nanoribbon work chain and it works nicely. Maybe merge #535 before (the changes there do not conflict with the current PR), as without it nanoribbon work chain wouldn't work.

PR #535, opened 2020-06-29
pp.x: fix gaussian cube file parsing.

yakutovicha:
The current implementation assumes that the number length is 13 digits, which is not
transferrable. In addition, it uses python loops, which are slow.

The PR replaces loops used to reshape the final array by the `reshape` method and
uses `line.split()` to split the data values in every line.


aiida_quantumespresso/parsers/pp.py
```diff
@@ -325,24 +325,13 @@ def parse_gaussian(self, data_file_str):
...
-                    datalist.append(float(data_point))
-
-        # Unpack the list and repack as a 3D array
-        # Note the unusual indexing: cube files run over the z index first, then y and x.
-        # E.g. The first volumetric data point is x,y,z = (0,0,0) and the second is (0,0,1)
-        for i in range(0, xdim):
-            for j in range(0, ydim):
-                for k in range(0, zdim):
-                    data_array[i, j, k] = (datalist[(i * ydim * zdim) + (j * zdim) + k])
+            ls = line.split()
```
sphuber, line 331: Is it possible to have empty lines in a Gaussian file? I.e. would it be good to put an
```
if not line.strip():
    continue
```

yakutovicha: This case is covered by the current implementation:
```python
In [1]: import numpy as np

In [2]: a = np.array([1,2,3])

In [3]: line = ''

In [4]: ls = line.split()

In [5]: a[2:2+len(ls)] = ls

In [6]: a
Out[6]: array([1, 2, 3])

In [11]: line = '100'

In [12]: ls = line.split()

In [13]: a[2:2+len(ls)] = ls

In [14]: a
Out[14]: array([  1,   2, 100])

In [15]:
```

sphuber:
@yakutovicha  Seems like one parser test is failing after I updated the branch to include the latest changes to support multiple file parsing for `PpCalculation`.

yakutovicha:
> @yakutovicha Seems like one parser test is failing after I updated the branch to include the latest changes to support multiple file parsing for `PpCalculation`.

The cube files were wrong. The number of data points per dimension is set to 1, but three data points instead of one were provided in the data array. I modified the cube files.

aiida_quantumespresso/parsers/pp.py
```diff
@@ -217,7 +217,7 @@ def detect_important_message(logs, line):
                 split_line = line.split('=')
                 if 'negative/imaginary' in line:  # QE6.1
                     output_dict['negative_core_charge'] = float(split_line[-1].split()[0])
-                    output_dict['imaginary_core_charge'] = float(split_line.split()[-1])
+                    output_dict['imaginary_core_charge'] = float(split_line[-1].split()[-1])
```
sphuber, line 220: This is apparently not being tested, correct? I guess that is because the output files we have are not from QE 6.1. However, why is there no `imaginary_core_charge` for the QE 6.4 branch?

aiida_quantumespresso/parsers/pp.py
```diff
@@ -335,10 +335,15 @@ def parse_gaussian(self, data_file_str):
         """
         lines = data_file_str.splitlines()
 
+        title = lines[0]
+        comment = lines[1]
```
sphuber, line 339: are these variables used at all?

aiida_quantumespresso/parsers/pp.py
```diff
@@ -352,34 +357,31 @@ def parse_gaussian(self, data_file_str):
...
         coordinates_units = 'bohr'
         data_units = self.units_dict[self.output_parameters['plot_num']]
 
         arraydata = orm.ArrayData()
         arraydata.set_array('voxel', voxel_array)
         arraydata.set_array('data', data_array)
-        arraydata.set_array('coordinates_units', np.array(coordinates_units))
         arraydata.set_array('data_units', np.array(data_units))
+        arraydata.set_array('coordinates_units', np.array(coordinates_units))
+        arraydata.set_array('coordinates', coordinates)
```
sphuber, line 384: why are we adding this? These are the coordinates of the atoms? This is redundant information, because it can be retrieved from the structure that was input of the parent calculation

aiida_quantumespresso/parsers/pp.py
yakutovicha: They aren't, thanks. I thought prospector should catch this kind of issues, though.

aiida_quantumespresso/parsers/pp.py
yakutovicha: > This is apparently not being tested, correct?

Looks like it wasn't.

> However, why is there no imaginary_core_charge for the QE 6.4 branch?

I don't know why, I can just say that there are no imaginary charges for QE 6.4, indeed:
https://github.com/aiidateam/aiida-quantumespresso/blob/a52266d096afe48dbc6b38b63ac17a9989dd12fe/tests/parsers/fixtures/pp/default_3d/aiida.out#L32

I should also mention, that in `QE 6.5` this line is not present at all, while for  `6.3` it is present, and looks as follows:
```
Check: negative/imaginary core charge=   -0.000001    0.000000
```

aiida_quantumespresso/parsers/pp.py
yakutovicha: In nanoribbon work chain we reduce the size of the cube files before parsing them, as they contain a lot of data points with values close to zero. This reduction involves changes in the structure's positions.

> why are we adding this? This is redundant information ...

In the view what I wrote above it would be nice for the standard parser to be able to extract the coordinates so that we don't need to write our own parser for that.

yakutovicha:
> Changes to parsing of volumetric data is fine, but not sure why you added the other changes. I would remove them because they are beyond the scope of this PR and they are just adding redundant information.

Thanks for the review @sphuber. In the reply to your comments I provided the reasoning why, I believe, parsing of the structure should be introduced. It is fine for me to add it in a separate PR if you agree with this change.

yakutovicha:
not sure why some tests are failing now, but I believe it is unrelated to the changes in this PR.

aiida_quantumespresso/parsers/pp.py
sphuber: Wait, how do you reduce the size of the cube files before parsing them? They are generated by `pp.x` and parsed straight away, no? Wouldn't the more reasonable approach be to have `pp.x` generate less points in the first place?

sphuber:
Yeah, this also happened on `aiida-core` and is due to a bug on Github actions. I will merge a fix and then you can rebase.

aiida_quantumespresso/parsers/pp.py
yakutovicha: > Wait, how do you reduce the size of the cube files before parsing them?

I crop them: https://github.com/nanotech-empa/aiida-nanotech-empa/blob/nanoribon_workchain/aiida_nanotech_empa/parsers.py#L21-L27

> They are generated by pp.x and parsed straight away, no?

I redefine `parse_gaussian` function and crop the file before giving it to the standard `parse_gaussian` 

> Wouldn't the more reasonable approach be to have pp.x generate less points in the first place?

How do you do that? One cannot modify the cell size in the pp.x, which leaves us with changing the density of points as the only option, which leads to a poor data quality. Instead, we take the whole cube file, crop it,  keeping only the nanoribbon's surrounding, and give it to the standard parser afterwords.

yakutovicha: After a discussion, we agreed to leave this part out for another PR.

sphuber:
Thanks a lot @yakutovicha

PR #537, opened 2020-07-09
PP plugin, add back 'settings' input

### LINKED ISSUE #536: Put back `settings` input to pp.x plugin.
This is needed to provide `-npools` argument as without it the calculations with a little number of atoms and a large number of k-points might fail.

yakutovicha:
fixes #536

### DIFF
--- a/tests/calculations/test_pp.py
+++ b/tests/calculations/test_pp.py
@@ -25,7 +25,7 @@ def generate_inputs(fixture_localhost, fixture_sandbox, fixture_code, generate_r
             'parameters':
             orm.Dict(dict=parameters),
             'settings':
-            orm.Dict(dict=settings) or orm.Dict(),
+            orm.Dict(dict=settings),
             'metadata': {
                 'options': get_default_options()
             }
--- a/tests/calculations/test_pp.py
+++ b/tests/calculations/test_pp.py
@@ -11,7 +11,7 @@ from aiida.common import datastructures, AttributeDict
 def generate_inputs(fixture_localhost, fixture_sandbox, fixture_code, generate_remote_data):
     """Return only those inputs that the parser will expect to be there."""
 
-    def _generate_inputs(parameters=None):
+    def _generate_inputs(parameters=None, settings=None):
         from aiida_quantumespresso.utils.resources import get_default_options
 
         if parameters is None:
@@ -24,6 +24,8 @@ def generate_inputs(fixture_localhost, fixture_sandbox, fixture_code, generate_r
             generate_remote_data(fixture_localhost, fixture_sandbox.abspath, 'quantumespresso.pw'),
             'parameters':
             orm.Dict(dict=parameters),
+            'settings':
+            orm.Dict(dict=settings) or orm.Dict(),
             'metadata': {
                 'options': get_default_options()
             }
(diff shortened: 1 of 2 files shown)


sphuber:
If you update the branch, the tests should now also run because I pushed a fix to `develop`

tests/calculations/test_pp.py
```diff
@@ -24,6 +24,8 @@ def _generate_inputs(parameters=None):
             generate_remote_data(fixture_localhost, fixture_sandbox.abspath, 'quantumespresso.pw'),
             'parameters':
             orm.Dict(dict=parameters),
+            'settings':
+            orm.Dict(dict=settings) or orm.Dict(),
```
sphuber, line 28: I don't think this works as expected. The `or` won't be applied to the value of `settings` but `orm.Dict(dict=settings)`, so no matter the value of `settings` you will always hit the first clause. Since `orm.Dict(dict=None)` is fine and is exactly equal to `orm.Dict()` you might as well get rid of `or orm.Dict()`

tests/calculations/test_pp.py
```diff
@@ -79,6 +81,14 @@ def test_pp_keep_plot_file(aiida_profile, fixture_sandbox, generate_calc_job, ge
         assert element in calc_info.retrieve_list
 
 
+def test_pp_cmdline_setting(aiida_profile, fixture_sandbox, generate_calc_job, generate_inputs):
+    """Test a `PpCalculation` with user-defined cmdline setting."""
```
sphuber, line 85:
```suggestion
    """Test a `PpCalculation` with user-defined cmdline settings."""
```

tests/calculations/test_pp.py
```diff
@@ -79,6 +81,14 @@ def test_pp_keep_plot_file(aiida_profile, fixture_sandbox, generate_calc_job, ge
...
+    entry_point_name = 'quantumespresso.pp'
+    inputs = generate_inputs(settings={'cmdline': ['-npools', '2']})
+    calc_info = generate_calc_job(fixture_sandbox, entry_point_name, inputs)
+    assert ['-npools', '2'] == calc_info.codes_info[0].cmdline_params
```
sphuber, line 89:
```suggestion
    assert calc_info.codes_info[0].cmdline_params == ['-npools', '2']
```
This is a nitpick, but since you have to make a small update in the `_generate_inputs` anyway, I though you could include this as well. I think when testing it reads more natural to put the unknown value that you are testing as the first operand.

tests/calculations/test_pp.py
yakutovicha: ah, didn't know. And you are right, empty AiiDA `Dict` object is not falsy:
```python
In [3]: if Dict(dict={}):
   ...:     print("Hello")
   ...:
Hello
```

Maybe one should fix this here as well:
https://github.com/aiidateam/aiida-quantumespresso/blob/0031475aff4aac8677c8da31c90e5498aad24b37/tests/parsers/test_pw.py#L25

sphuber: Ah, yes, please do

yakutovicha: also, just a side note:
```python
In [6]: Dict(dict=None) == Dict(dict=None)
Out[6]: False
```
not sure if this is what we expect.

PR #540, opened 2020-07-14
Build system: unvendor `fastentrypoints` and use `pyproject.toml`

sphuber:
By using the new `pyproject.toml` file to define the build requirements,
we can unvendor the `fastentrypoints`.

This commit also cleans up some other minor things:

 * Remove the unused `.prospector.yaml`, as prospector is no longer used
 * Update the years in `LICENSE.txt`
 * Remove unnecessary `docs/.gitignore`
 * Minor fix in `tests/parsers/test_pw.py` that doesn't affect behavior


sphuber:
Since this touches the build mechanism, I wanted to make sure this works properly in order to not release a broken version. That's why I prefer to have @greschd give a look. @greschd will you have time to give a quick check or are you out-of-office or don't have time?

greschd:
Yeah sorry, I'm OOF until Monday. From looking at it it looks fine, if you want I can check some more then.

Unrelated to this change, but in the `open` calls in `setup.py` I think we should specify an encoding.

PR #546, opened 2020-08-17
Projwfc parser noncollinear - Fixes issue #542

### LINKED ISSUE #542: Projwfc parser fails for non-collinear calculations
When running a `ProjwfcCalculation` from a parent that has `noncolin = .true.` (and happens to have an odd number of k-points), the parser fails with

```
aiida.parser.ProjwfcParser: [ERROR] Error parsing bands and projections: Internal formatting error regarding spin
```

This is caused by this bit in the parser: https://github.com/aiidateam/aiida-quantumespresso/blob/1c9e55d51e800cc360d16f38cfcbff83b09831dd/aiida_quantumespresso/parsers/projwfc.py#L363-L366

The trouble is that the parser assumes the k-points are duplicated for spin-up and spin-down when `nspin != 1`, but this is not the case for `nspin = 4` (`noncolin = .true.`).

I will see if just changing the conditional  in https://github.com/aiidateam/aiida-quantumespresso/blob/1c9e55d51e800cc360d16f38cfcbff83b09831dd/aiida_quantumespresso/parsers/projwfc.py#L352-L355

from ``nspin != 1`` to ``nspin == 2`` solves the issue, and report back.

The non-collinear code path seems to need a little love in general, see also #443.

lbotsch:
Fixes #542

This PR was intended to fix issue #542, but I ended up also implementing the projwfc.x parser for spin-orbit calculations.

This adds two new types of orbitals:

- In the non-collinear case, orbitals are saved as `noncollinearhydrogen` orbitals and include the spin projection s_z along the z-axis. Spin-up/down states are treated as separate orbitals.

- In the spin-orbit case, orbitals are saved as `spinorbithydrogen` orbitals and include the `total_angular_momentum` quantum number `j`, the `angular_momentum` quantum number `l` and the total magnetic_number `m_j`.

The PR includes tests for each of the four cases supported by the parser, namely nspin=1 (default), nspin=2 (spin), noncolin=.true. (noncollinear) and lspinorb=.true. (spinorbit)

Note: I also modified some regular expressions in order to make them work across qe versions.


Crivella (commented):
I've tested some projwfc calculations with the new parser and it works fine with all the cases (nspin = 1/2/4/4+spinorb)
Also the parsed data matches the one in the files.

The new orbitals have been implemented as separate subclasses of "RealhydrogenOrbital", by adding or overriding some of its validations (concerning the discussion aiidateam/aiida-core#4285). 

A minor suggestion would be to reduce the size of the test files, by running the calculations with reduced kpoints and bands (they don't need to be exact or converged, just to produce a valid output)

sphuber (commented):
Thanks a lot @lbotsch , this is great. I have to say that I have little experience with non-orbital calculations and projwfc so I haven't gone into the implementation details. I do have two comments and a question.

1) As @Crivella mentioned, it would be indeed good to really try and reduce the test file content to the bare minimum. Since we already have quite some test files but will have to be adding a lot more, we need to try and keep it down in order to not bloat the repo. You can either try and run a smaller example calculation, with less bands/kpoints/energies or simply manually remove the data. Note that for example we are not testing that the numerical data from the `*pdos_tot` files are correctly parsed, so even a single line of values would be sufficient there. Let me know if you need help with this point.

2) Regarding the naming of the new orbitals. I realize that the current orbital classes in `aiida-core` use inconsistent capitalization (`RealhydrogenOrbital` as opposed to `RealHydrogenOrbital`) which is still there for historical reasons, and you chose to stick to that convention. Still, I wonder if we should just bite the bullet and use the "correct" naming, e.g. `spin_orbit_hydrogen = aiida_quantumespresso.tools.data.orbital.spin_orbit_hydrogen:SpinOrbitHydrogenOrbital`.

3) This is more directed at @greschd and @qiaojunfeng : does the solution of the `SpinorbitHydrogenOrbital` and `NoncollinearHydrogenOrbital` solve your use case as well, as described in [the issue on `aiida-core`](https://github.com/aiidateam/aiida-core/issues/4285)? If so, should we maybe add this to `aiida-core` to be with the rest of the orbitals? Then again, we are planning to actually move those out of `aiida-core` for v2.0 to a plugin package (to be decided which), so maybe including them here for now is just as well. Wanted to have your opinion on this before deciding.

lbotsch:
Thanks, @Crivella and @sphuber, for the suggestions:

1. That makes sense! I tried to reduce the test fixture size by rerunning the calculations with only one k-point. Manually editing the pdos files and removing all but the first data line+header, as suggested by @sphuber, results in an IndexError [here](https://github.com/lbotsch/aiida-quantumespresso/blob/2cd8bf454c0c6b0a968edde4b4df89f4b754d02f/aiida_quantumespresso/parsers/projwfc.py#L323), which is probably a bug (need to investigate that). Any suggestions on how to further reduce the size? I would like to keep the test data as realistic as possible and not manually edit the output files.

2. I would prefer the "correct" capitalization, but chose to be consistent with the naming in aiida-core. BTW, the naming of the class does not affect the database structure. The type of orbital used in the projection is saved using the entry point names which are all lower-case (i.e. `realhydrogen`, `spinorbithydrogen` and `noncollinearhydrogen`). Therefore, changing the class names at a later time, maybe once they are merged into another plugin, should not be too dramatic.

Note: I rebased the branch onto `develop`

Crivella:
@lbotsch 
The IndexError you mention is due to the fact that np.genfromtext, when reading a 1 line file, generates a 1D array and not a 2D one.
I don't know if there could ever be the need to generate a only one energy point in the PDOS files (maybe if someone is only interested in the projwfc stdout?)

The problem could be fixed by changing the code as such

```
    pdostot_array = np.atleast_2d(np.genfromtxt(pdostot_file))
```

lbotsch:
@Crivella: Yes indeed, good catch! Thank you

greschd:
Regarding 3.: I think this is _exactly_ the use case described in that `aiida-core` issue. Because I didn't end up using that approach (for unrelated reasons), I don't currently have an easy way to run it. @qiaojunfeng maybe you could try running this with the `aiida-wannier90-workflows`?

sphuber:
Thanks @greschd . Ok @qiaojunfeng if you can confirm this works for `aiida-wannier90-workflows` I will merge this.

qiaojunfeng:
Hi guys, the PR works smoothly with `aiida-wannier90-workflows`, thanks a lot!

sphuber (approved):
Thanks a lot @lbotsch and all the others for checking!

PR #549, opened 2020-08-27
Fix the incorrect usage of repository interface of `aiida-core`

### LINKED ISSUE #548: Fix incorrect usage of repository interface of `aiida-core`
Especially:

 * `Node.open` has to be use in a context manager
 * `Node._get_base_folder` will soon be removed and cannot be used.

sphuber:
Fixes #548 

The following illegal uses have been fixed:

 * `Node.open` has to be use in a context manager
 * `Node._get_base_folder` will soon be removed and cannot be used.

### DIFF
--- a/aiida_quantumespresso/parsers/cp.py
+++ b/aiida_quantumespresso/parsers/cp.py
@@ -44,11 +44,10 @@ class CpParser(Parser):
             # TODO: Add an error for this counter
             return self.exit(self.exit_codes.ERROR_MISSING_XML_FILE)
 
-        # Let's pass file handlers to this function
-        out_dict, _raw_successful = parse_cp_raw_output(
-            out_folder.open(stdout_filename), out_folder.open(xml_files[0]),
-            out_folder.open(self.node.process_class._FILE_XML_PRINT_COUNTER_BASENAME)
-        )
+        output_stdout = out_folder.get_object_content(stdout_filename)
+        output_xml = out_folder.get_object_content(xml_files[0])
+        output_xml_counter = out_folder.get_object_content(self.node.process_class._FILE_XML_PRINT_COUNTER_BASENAME)
+        out_dict, _raw_successful = parse_cp_raw_output(output_stdout, output_xml, output_xml_counter)
 
         # parse the trajectory. Units in Angstrom, picoseconds and eV.
         # append everthing in the temporary dictionary raw_trajectory
@@ -96,7 +95,8 @@ class CpParser(Parser):
 
         # =============== EVP trajectory ============================
         try:
-            matrix = numpy.genfromtxt(out_folder.open('{}.evp'.format(self._node.process_class._PREFIX)))
+            with out_folder.open('{}.evp'.format(self._node.process_class._PREFIX)) as handle:
+                matrix = numpy.genfromtxt(handle)
             # there might be a different format if the matrix has one row only
             try:
                 matrix.shape[1]
--- a/aiida_quantumespresso/parsers/neb.py
+++ b/aiida_quantumespresso/parsers/neb.py
@@ -60,13 +60,12 @@ class NebParser(Parser):
         # load the neb input parameters dictionary
         neb_input_dict = self.node.inputs.parameters.get_dict()
 
-        stdout_abspath = os.path.join(out_folder._repository._get_base_folder().abspath, filename_stdout)
-
         # First parse the Neb output
         try:
-            neb_out_dict, iteration_data, raw_successful = parse_raw_output_neb(stdout_abspath, neb_input_dict)
+            stdout = out_folder.get_object_content(filename_stdout)
+            neb_out_dict, iteration_data, raw_successful = parse_raw_output_neb(stdout, neb_input_dict)
             # TODO: why do we ignore raw_successful ?
-        except QEOutputParsingError as exc:
+        except (OSError, QEOutputParsingError):
             return self.exit(self.exit_codes.ERROR_OUTPUT_STDOUT_READ)
 
         for warn_type in ['warnings', 'parser_warnings']:
--- a/aiida_quantumespresso/parsers/parse_raw/cp.py
+++ b/aiida_quantumespresso/parsers/parse_raw/cp.py
@@ -149,30 +149,27 @@ def parse_cp_xml_counter_output(data):
     return parsed_data
 
 
-def parse_cp_raw_output(out_file, xml_file=None, xml_counter_file=None):
+def parse_cp_raw_output(stdout, output_xml=None, output_xml_counter=None):
 
     parser_info = get_parser_info(parser_info_template='aiida-quantumespresso parser cp.x v{}')
 
-    # analyze the xml
-    if xml_file is not None:
-        xml_data = parse_cp_xml_output(xml_file.read())
-    else:
+    if output_xml is None:
         parser_info['parser_warnings'].append('Skipping the parsing of the xml file.')
         xml_data = {}
-
-    # analyze the counter file, which keeps info on the steps
-    if xml_counter_file is not None:
-        xml_counter_data = parse_cp_xml_counter_output(xml_counter_file.read())
     else:
+        xml_data = parse_cp_xml_output(output_xml)
+
+    # XML counter file, which keeps info on the steps
+    if output_xml_counter is None:
         xml_counter_data = {}
+    else:
+        xml_counter_data = parse_cp_xml_counter_output(output_xml_counter)
 
-    # analyze the standard output
-    out_lines = out_file.readlines()
-
+    stdout = stdout.split('\n')
     # understand if the job ended smoothly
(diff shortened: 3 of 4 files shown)


aiida_quantumespresso/parsers/cp.py
```diff
@@ -45,10 +45,10 @@ def parse(self, **kwargs):
             return self.exit(self.exit_codes.ERROR_MISSING_XML_FILE)
 
         # Let's pass file handlers to this function
-        out_dict, _raw_successful = parse_cp_raw_output(
-            out_folder.open(stdout_filename), out_folder.open(xml_files[0]),
-            out_folder.open(self.node.process_class._FILE_XML_PRINT_COUNTER_BASENAME)
-        )
+        output_stdout = out_folder.get_object_content(stdout_filename)
+        output_xml = out_folder.get_object_content(xml_files[0])
```
greschd, line 49: Is there a particular reason to use `get_object_content` instead of 

```python
with out_folder.open(stdout_filename) as stdout_handle:
    parse_cp_raw_output(stdout_handle, ...)
```
If the files are large, I'd think the latter could potentially be more efficient (depending on what `parse_cp_raw_output` does) because it doesn't need to load the whole file into memory at once.

Or are these files generally small enough that it doesn't matter?

sphuber: It wouldn't matter because `parse_cp_raw_output` would load them into memory anyway and has to because it cannot parse it in a streaming manner. At least not currently.

sphuber:
I am putting this on hold. This actually serves as an example for other plugin developers for fixes they most likely have to apply to their plugins for them to be compatible with the upcoming new version of `aiida-core` that will have the new repository implementation. Since the behavior I am correcting here will break with the new version (and we cannot provide a backwards compatible transition period) we are going to already alert them and suggest they update the plugins ahead of time.

The current plan we are discussing is to add deprecation warnings in `aiida-core==1.4` and then add the new repository implementation in `aiida-core==1.5`. If all plugins can update between v1.4 and v1.5, there should be no problem.

aiida_quantumespresso/parsers/cp.py
greschd: Ok, yeah in that case it doesn't matter 👍

greschd:
Is it just the `_repository._get_base_folder().abspath` construction that would be deprecated, or also using `Node.open()` without a context manager?

Of course a bare `Node.open` is not nice because it leaves the file open (well, until the garbage collector gets around to it), but since you _can_ manually `.close()` it I'd probably be in favor of still allowing it (similar to the regular Python `open`).

As for `_repository._get_base_folder()`... I mean, with two leading underscores you're kinda asking for that to be broken 😄

sphuber:
> Of course a bare `Node.open` is not nice because it leaves the file open (well, until the garbage collector gets around to it), but since you _can_ manually `.close()` it I'd probably be in favor of still allowing it (similar to the regular Python `open`).

Unfortunately, the way the new repository is currently implemented, this will be difficult. The reason is that the `disk-object-store` library, implemented by Giovanni, which provides our efficient long term storage, requires use of context manager, because it allows it to make the implementation more efficient. We would have to ask @giovannipizzi if it is possible and worth reverting this requirement and allowing also non-contextual use of `open`.

> As for `_repository._get_base_folder()`... I mean, with a double protected access you're kinda asking for that to be broken

Sure, except that `aiida-quantumespresso` was also using it, which many plugin developers have used as the example over the years :sweat_smile: 
I guess we should have clearly stated "Do as I say, not as I do" ;)

greschd:
> 
> Unfortunately, the way the new repository is currently implemented, this will be difficult. The reason is that the `disk-object-store` library, implemented by Giovanni, which provides our efficient long term storage, requires use of context manager, because it allows it to make the implementation more efficient. We would have to ask @giovannipizzi if it is possible and worth reverting this requirement and allowing also non-contextual use of `open`.

Ah, I see. Well, I don't really see a use case where it would be _necessary_ to manually open / close instead of using the context manager, so I guess it's fine to deprecate that.
We should probably make sure that manual open / close gives a reasonable traceback, though: Since the new implementation (I presume) will still have a `.open()` method that just returns a context manager, it might result in a confusing error when e.g. trying to read or iterate over that object without `__enter__`ing it first.

> 
> Sure, except that `aiida-quantumespresso` was also using it, which many plugin developers have used as the example over the years 😅
> I guess we should have clearly stated "Do as I say, not as I do" ;)

Yeah, true.. definitely best to proactively check who is using it and warn them.

> and we cannot provide a backwards compatible transition period

That just means the _old interface_ won't work with the _new repository_, right? As in, the updated code will still be compatible with older AiiDA versions.
I had somehow interpreted this as "there is no interface that works with new and old AiiDA versions simultaneously" before, which I think would be more serious.

sphuber:
> We should probably make sure that manual open / close gives a reasonable traceback, though: Since the new implementation (I presume) will still have a .open() method that just returns a context manager, it might result in a confusing error when e.g. trying to read or iterate over that object without __enter__ing it first.

Yeah, currently you can get something like:
```
 AttributeError: '_GeneratorContextManager' object has no attribute 'read'
```

> That just means the old interface won't work with the new repository, right? As in, the updated code will still be compatible with older (1.0+) AiiDA versions?
I had somehow interpreted this as "there is no interface that works with new (>=1.5) and old (<=1.4) AiiDA versions simultaneously" before, which would be more serious.

If with "updated code" you mean the updated `aiida-quantumespresso` after this PR, then yes. The changes here will make the code compatible with current `aiida-core` as well as `aiida-core~=1.5` once we merge the new interface. There are some other minor changes, such as moving some imports, changing some methods argument names and changing type from enum to class (I know, we did this before thinking it was easy ;) ). I will make a PR soon that already adds these changes with just the deprecation warnings such that it can be merged in v1.4.

Unfortunately, due to the code fixed in this PR that many plugins will have, it will be impossible to have older plugin versions backward compatible with the new repository interface. The use of `_get_base_folder` to get the `abspath` will be impossible to provide with the new repository.

Would you be interested in giving a look at my repo branch to see if you can spot potential backwards incompatibilities that I introduced that I forgot about? In any case, I think once I finalized it, I will send a mail to plugin developers notifying them about the plans, this PR and the changes they should prepare themselves and then have them test their plugin against my branch.

All input and comments to the plan are welcome to try and make this as smooth as possible.

greschd:
> 
> If with "updated code" you mean the updated `aiida-quantumespresso` after this PR, then yes. The changes here will make the code compatible with current `aiida-core` as well as `aiida-core~=1.5` once we merge the new interface. There are some other minor changes, such as moving some imports, changing some methods argument names and changing type from enum to class (I know, we did this before thinking it was easy ;) ). I will make a PR soon that already adds these changes with just the deprecation warnings such that it can be merged in v1.4.
> 
> Unfortunately, due to the code fixed in this PR that many plugins will have, it will be impossible to have older plugin versions backward compatible with the new repository interface. The use of `_get_base_folder` to get the `abspath` will be impossible to provide with the new repository.

Yeah, I mean at least there's a happy updating path, and with a deprecation period.. not ideal, but I'd say good enough :) At least it doesn't force users to upgrade `aiida-core` if they want newer plugins (unless the plugin itself decides so).

> 
> Would you be interested in giving a look at my repo branch to see if you can spot potential backwards incompatibilities that I introduced that I forgot about? In any case, I think once I finalized it, I will send a mail to plugin developers notifying them about the plans, this PR and the changes they should prepare themselves and then have them test their plugin against my branch.
> 

Sure, I can give it a glance. Which branch is that?

sphuber:
> Sure, I can give it a glance. Which branch is that?

https://github.com/sphuber/aiida_core/tree/fix/3445/disk-object-store-repository

Note that I wrote this without trying to be fully backwards compatible. Since we now decided that we still want to try, in the latest commit I started "adding" backward compatibility stuff. But what I will have to do soon, is take those affected resources and already put those deprecations on `develop` in a separate PR. The list probably is not yet complete, would be good if you find some more.

greschd:
I've added a few comments here: https://github.com/greschd/aiida_core/pull/5#pullrequestreview-477688727

By no means a guarantee that I found all backwards-incompatible bits.

greschd:
I think it would be ok to merge this now - it should be compatible with 1.3+ and (as far as we know now) 2.0, right?

If we want to point other plugins here as an example, there's no harm in it being merged already.

sphuber:
Yeah this could be merged. I was just keeping it open as an example, but we could just as easy refer to it when it is closed. Will update it

aiida_quantumespresso/parsers/cp.py
```diff
@@ -45,10 +45,10 @@ def parse(self, **kwargs):
             return self.exit(self.exit_codes.ERROR_MISSING_XML_FILE)
 
         # Let's pass file handlers to this function
```
greschd, line 47: Guess this comment is now outdated.

greschd (approved):

greschd:
Good to go for me, just one minor comment.

PR #550, opened 2020-09-02
Fix parsing problem with van der Waals functionals+stress

giovannipizzi:
Previously, the flag `vdw_correction` was set in the raw PW parser
only if the string 'Carrying out vdW-DF run using the following parameters:'
was found.
However, `vdW-DF is not the only vdW functional, there are many more.
Therefore, e.g. in my case using a rVV10 functional, the string was more like
'Carrying out RVV10 run...', and the capitalization might also differ.

Since this flag is used later to parse lines having the text
'Non-local correlation energy' in it, I now change the setting of the flag
to the existence of this string.

As a consequence, the actual manifestation of this bug was that
(beside the key `energy_vdw` missing in the output) that a generic
'Error while parsing stress tensor.' was raised.
After debugging the cause was that tsince the vdw flag was not
set, the search for `P=` was failing as it was searching not far
enough (was: `for k in range(10 + 5 * vdw_correction):`), and
this was not creating a `count2` variable, and a NameError was raised,
then swallowed and hidden by the generic `except Exception`.

I now:
- replace directly the search on the next 15 lines in all cases
  (this should be safe in all cases, and shouldn't be a major performance hit)
- set `count2 = None` before, and issue a Warning if `P=` is not found
- print additional output (traceback) in the warning, in case
  an exception is raised.

The latter might need to be done elsewhere, but for now I think it's safe
and a good thing to start doing it here.


giovannipizzi:
This could be a good PR to be reviewed by @mbercx 
Also @sphuber could you remind me where are the instructions to add tests, please? I have some runs so it should be easy (?) to add one more test for my rVV10 case

sphuber:
> This could be a good PR to be reviewed by @mbercx
> Also @sphuber could you remind me where are the instructions to add tests, please? I have some runs so it should be easy (?) to add one more test for my rVV10 case

There are no instructions yet, but it is really straightforward. Since this concerns `pw`, create a new folder in `tests/parsers/fixtures/pw` , something like `specific_vdw_correction`, where you put the required output files. Should be just the `aiida.out` and `data-file-schema.xml`. Then you just create a test in `tests/parsers/test_pw.py`. You can look at one of the examples there how it works. The trick here is to keep the fixture files you will be adding to a minimum. You are adding this test just to test the correct parsing of the vanderWaals flags, so there is no need to check the entire output of all produced output nodes. Normally I even cut the output manually removing all lines that are not really necessary to test what we're testing. It would also be good to try and get a file as small as possible. So if we are testing the parsing of this flag, doesn't make sense to use a big vc-relax run with lots of steps. Instead maybe run a quick dummy SCF calculation on Si.

mbercx (commented):
LGMT! The logic seems solid, and a little test run on the output file you set me works like a charm. Once the test is added, this is good to go.

Maybe it'd be a good exercise for me to add the test here? What's the conventional way of making changes in *someone else's* PR? Open a PR into your `fix_vdW_parsing` branch?

giovannipizzi:
Hi, I thought i had left a message, but apparently I didn't? Anyway, you should have write permissions in my fork now, so you can commit a test! Let me know if this is OK

mbercx:
> Hi, I thought i had left a message, but apparently I didn't? Anyway, you should have write permissions in my fork now, so you can commit a test! Let me know if this is OK

Yes, I saw the message for the permissions, thanks! I'll work on adding the test asap.

mbercx:
I've added a test run for rVV10 applied to a very short vc-relax run for silicon. The change in parsing logic was causing the `test_pw_vcrelax_success_external_pressure` test to fail, but I think this is simply related to the fact that there aren't enough lines after the `Computing stress` line due to manually removed output. I've added some lines with `--- OUTPUT REMOVED ---` to fix this.

@sphuber let me know if you have any comments!

tests/parsers/test_pw.py
```diff
@@ -595,6 +595,32 @@ def test_pw_vcrelax_success_fractional(
...
+    assert calcfunction.is_finished_ok, calcfunction.exit_message
+    assert not orm.Log.objects.get_logs_for(node), [log.message for log in orm.Log.objects.get_logs_for(node)]
+    assert 'output_parameters' in results
+    assert 'output_structure' in results
+    assert 'output_trajectory' in results
+    data_regression.check({
+        'output_parameters': results['output_parameters'].get_dict(),
+        'output_structure': results['output_structure'].attributes,
+        'output_trajectory': results['output_trajectory'].attributes,
+    })
```
sphuber, line 621: Since we are checking a very specific part of the parsing code here, I would argue that it is not necessary to check the entire parsed output, but just the part that is relevant here. The parsing of the rest of the information is already done in the default tests. So what we could check here is:
* the `energy_vdw` key is in `output_parameters`
* the `output_trajectory` contains the parsed `stress`.

That would be sufficient and saves a lot of unnecessary lines in the comparison files. Other than that everything looks good.

aiida_quantumespresso/data/force_constants.py
```diff
@@ -5,6 +5,8 @@
 from qe_tools.constants import bohr_to_ang
 from aiida.orm import SinglefileData
 
+# pylint: disable=too-many-ancestors, raise-missing-from
```
sphuber, line 8: Prefer not to disable these here. The former I am disabling project wide in #562 so will soon no longer be necessary and the second would be better to actually fix

sphuber (changes requested):
Thanks @mbercx  Two minor requests

tests/parsers/test_pw.py
mbercx: Just to check: Since the stress array cannot be represented in `.yml` format, it's ok just check the array size stored in the attributes of the `output_trajectory`, right?

mbercx:
Thanks, @sphuber! I trimmed down the parts of the output that are checked to the relevant parts as you requested, and removed more lines from the output file. The pylint disables are also removed, and I fixed the `raise-missing-from` issue. The pre-commit still fails because of the `too-many-ancestors` check, once #562 is merged I'll rebase the PR.

sphuber (approved):

sphuber:
Thanks @giovannipizzi and @mbercx

PR #552, opened 2020-09-02
`PwParser`: add support for output format of pw.x v6.6

### LINKED ISSUE #551: Add support for new XML schema of QE v6.6

sphuber:
Fixes #551 

Quantum ESPRESSO pw.x v6.6 not only comes with a new schema for the XML
output file, which is added here such that it can be parsed, it also has
changes to the standard output format. In a simple SCF calculation for a
silicon structure as a test, one change turned out to be problematic.
The line containing the Harris-Foulkes estimate that used to be printed
directly after the computed total energy has been removed. This line
itself was never parsed anyway, but the estimated SCF accuracy that is
printed afterwards was, and the parser relied on it being two lines
after the total energy. The parser is now adapted to, starting from the
total energy line, scout 5 lines ahead for the line containing the
marker of the estimate SCF accuracy. This once more shows how we should
not rely on parsing textual output.


giovannipizzi (approved):
Thanks! Looks good to me.
@mbercx I don't remember if you wanted to do more tests here?
Otherwise for me it's ready to merge, could you merge it?

mbercx:
> @mbercx I don't remember if you wanted to do more tests here?

I just wanted to do a little test run for qe v6.6 myself, but ran into some issues compiling the code on my laptop. I'll see if I can fix that and (squash and 😉) merge.

mbercx:
LGMT! :rocket:

PR #554, opened 2020-09-18
Implement parallelization options as explicit inputs.

### LINKED ISSUE #405: Promote parallelization settings to "proper" inputs?

greschd:
Fixes #405.

If any parallelization option is already specified in the `settings['CMDLINE']`, a `DeprecationWarning` is issued. However, the value from `settings['CMDLINE']` still takes precedence. This allows us to implement the defaults at the level of the input
port, without having to care about it later on.

The `BasePwCpInputGenerator` is given three class attributes:
- `_ALLOWED_PARALLELIZATION_FLAGS` is a list of tuples (flag_name, default, help), of all possible flags
- `_ENABLED_PARALLELIZATION_FLAGS` is a list of flag names that are implemented in a particular code.
- `_PARALLELIZATION_FLAG_ALIASES` is used to detect all possible variations on a flag name.

TODO:
- [x] Testing
- [x] Documentation: Not sure where best to put this - the auto-generated help will be there in any case.
- [x] Check if flag names are case sensitive. If not, this needs to be taken into account when checking for existing flags. **Update:** If flags are capitalized differently (e.g. `-nPools`) QE will simply ignore them.
- [x] Set `_ENABLED_PARALLELIZATION_FLAGS` for codes other than pw.x. **Help needed:** I have no idea what flags other codes support.
- [x] PW accepts `-nimage`, but I'm not sure if passing that makes any sense (a PW calculation only treats one "image"). If it doesn't, we should probably disable `nimage` in the `PwCalculation`. **Update:** Disallowed `-nimage` for `pw.x` for now - if there is in fact a use case, we can always re-enable it.

### DIFF
--- a/aiida_quantumespresso/calculations/__init__.py
+++ b/aiida_quantumespresso/calculations/__init__.py
@@ -2,10 +2,16 @@
 """Base `CalcJob` for implementations for pw.x and cp.x of Quantum ESPRESSO."""
 import abc
 import os
+import copy
+import numbers
+import warnings
+from functools import partial
+from types import MappingProxyType
 
 from aiida import orm
 from aiida.common import datastructures, exceptions
 from aiida.common.lang import classproperty
+from aiida.common.warnings import AiidaDeprecationWarning
 from aiida.plugins import DataFactory
 from qe_tools.converters import get_parameters_from_cell
 
@@ -29,6 +35,37 @@ class BasePwCpInputGenerator(CalcJob):
     _DATAFILE_XML_POST_6_2 = 'data-file-schema.xml'
     _ENVIRON_INPUT_FILE_NAME = 'environ.in'
     _DEFAULT_IBRAV = 0
+
+    # A mapping {flag_name: help_string} of parallelization flags
+    # possible in QE codes. The flags that are actually implemented in a
+    # given code should be specified in the '_ENABLED_PARALLELIZATION_FLAGS'
+    # tuple of each calculation subclass.
+    _PARALLELIZATION_FLAGS = MappingProxyType(
+        dict(
+            nimage="The number of 'images', each corresponding to a different self-consistent or "
+            'linear-response calculation.',
+            npool="The number of 'pools', each taking care of a group of k-points.",
+            nband="The number of 'band groups', each taking care of a group of Kohn-Sham orbitals.",
+            ntg="The number of 'task groups' across which the FFT planes are distributed.",
+            ndiag="The number of 'linear algebra groups' used when parallelizing the subspace "
+            'diagonalization / iterative orthonormalization. By default, no parameter is '
+            'passed to Quantum ESPRESSO, meaning it will use its default.',
+            nhw="The 'nmany' FFT bands parallelization option."
+        )
+    )
+
+    _ENABLED_PARALLELIZATION_FLAGS = tuple()
+
+    _PARALLELIZATION_FLAG_ALIASES = MappingProxyType(
+        dict(
+            nimage=('ni', 'nimages', 'npot'),
+            npool=('nk', 'npools'),
+            nband=('nb', 'nbgrp', 'nband_group'),
+            ntg=('nt', 'ntask_groups', 'nyfft'),
+            ndiag=('northo', 'nd', 'nproc_diag', 'nproc_ortho'),
+            nhw=('nh', 'n_howmany', 'howmany')
+        )
+    )
 
     # Additional files that should always be retrieved for the specific plugin
     _internal_retrieve_list = []
@@ -87,6 +124,35 @@ class BasePwCpInputGenerator(CalcJob):
             help='Optional van der Waals table contained in a `SinglefileData`.')
         spec.input_namespace('pseudos', valid_type=(LegacyUpfData, UpfData), dynamic=True,
             help='A mapping of `UpfData` nodes onto the kind name to which they should apply.')
+        spec.input(
+            'parallelization',
+            valid_type=orm.Dict,
+            required=False,
+            help=(
+                'Parallelization options. The following flags are allowed:\n' + '\n'.join(
+                    f'{flag_name:<7}: {cls._PARALLELIZATION_FLAGS[flag_name]}'
+                    for flag_name in cls._ENABLED_PARALLELIZATION_FLAGS
+                )
+            ),
+            validator=cls._validate_parallelization
+        )
+
+    @classmethod
+    def _validate_parallelization(cls, value, port_namespace):  # pylint: disable=unused-argument
+        if value:
+            value_dict = value.get_dict()
+            unknown_flags = set(value_dict.keys()) - set(cls._ENABLED_PARALLELIZATION_FLAGS)
+            if unknown_flags:
+                return (
+                    f"Unknown flags in 'parallelization': {unknown_flags}, "
+                    f'allowed flags are {cls._ENABLED_PARALLELIZATION_FLAGS}.'
+                )
+            invalid_values = [val for val in value_dict.values() if not isinstance(val, numbers.Integral)]
+            if invalid_values:
+                return (
+                    f'Parallelization values must be integers; got invalid values {invalid_values}.'
+                )
+
 
     def prepare_for_submission(self, folder):
         """Create the input files from the input nodes passed to this instance of the `CalcJob`.
@@ -203,8 +269,12 @@ class BasePwCpInputGenerator(CalcJob):
         calcinfo = datastructures.CalcInfo()
 
         calcinfo.uuid = str(self.uuid)
-        # Empty command line by default
-        cmdline_params = settings.pop('CMDLINE', [])
+        # Start from an empty command line by default
+        cmdline_params = self._add_parallelization_flags_to_cmdline_params(
+            cmdline_params=settings.pop('CMDLINE', [])
+        )
+
+
         # we commented calcinfo.stin_name and added it here in cmdline_params
         # in this way the mpirun ... pw.x ... < aiida.in
         # is replaced by mpirun ... pw.x ... -in aiida.in
@@ -243,6 +313,55 @@ class BasePwCpInputGenerator(CalcJob):
             raise exceptions.InputValidationError(f'`settings` contained unexpected keys: {unknown_keys}')
 
         return calcinfo
+
+    def _add_parallelization_flags_to_cmdline_params(self, cmdline_params):
+        """Get the command line parameters with added parallelization flags.
+
+        Adds the parallelization flags to the given `cmdline_params` and
+        returns the updated list.
+
+        Raises an `InputValidationError` if multiple aliases to the same
+        flag are given in `cmdline_params`, or the same flag is given
+        both in `cmdline_params` and the explicit `parallelization`
+        input.
+        """
+        cmdline_params_res = copy.deepcopy(cmdline_params)
+        # The `cmdline_params_normalized` are used only here to check
+        # for existing parallelization flags.
+        cmdline_params_normalized = []
+        for param in cmdline_params:
+            cmdline_params_normalized.extend(param.split())
+
+        if 'parallelization' in self.inputs:
+            parallelization_dict = self.inputs.parallelization.get_dict()
+        else:
+            parallelization_dict = {}
+        # To make the order of flags consistent and "nice", we use the
+        # ordering from the flag definition.
+        for flag_name in self._ENABLED_PARALLELIZATION_FLAGS:
+            all_aliases = list(self._PARALLELIZATION_FLAG_ALIASES[flag_name]) + [flag_name]
+            aliases_in_cmdline = [alias for alias in all_aliases if f'-{alias}' in cmdline_params_normalized]
+            if aliases_in_cmdline:
+                if len(aliases_in_cmdline) > 1:
+                    raise exceptions.InputValidationError(
+                        f'Conflicting parallelization flags {aliases_in_cmdline} '
+                        "in settings['CMDLINE']"
+                    )
+                if flag_name in parallelization_dict:
+                    raise exceptions.InputValidationError(
+                        f"Parallelization flag '{aliases_in_cmdline[0]}' specified in settings['CMDLINE'] conflicts "
+                        f"with '{flag_name}' in the 'parallelization' input."
+                    )
+                else:
+                    warnings.warn(
+                        "Specifying the parallelization flags through settings['CMDLINE'] is "
+                        "deprecated, use the 'parallelization' input instead.", AiidaDeprecationWarning
+                    )
+                    continue
+            if flag_name in parallelization_dict:
+                flag_value = parallelization_dict[flag_name]
+                cmdline_params_res += [f'-{flag_name}', str(flag_value)]
+        return cmdline_params_res
 
     @staticmethod
     def _generate_PWCP_input_tail(*args, **kwargs):
--- a/tests/calculations/test_pw.py
+++ b/tests/calculations/test_pw.py
@@ -5,6 +5,8 @@ import pytest
 
 from aiida import orm
 from aiida.common import datastructures
+from aiida.common.warnings import AiidaDeprecationWarning
+from aiida.common.exceptions import InputValidationError
 from aiida_quantumespresso.utils.resources import get_default_options
 from aiida_quantumespresso.calculations.helpers import QEInputValidationError
 
@@ -25,7 +27,7 @@ def test_pw_default(fixture_sandbox, generate_calc_job, generate_inputs_pw, file
     # Check the attributes of the returned `CalcInfo`
     assert isinstance(calc_info, datastructures.CalcInfo)
     assert isinstance(calc_info.codes_info[0], datastructures.CodeInfo)
-    assert sorted(calc_info.codes_info[0].cmdline_params) == sorted(cmdline_params)
+    assert sorted(calc_info.codes_info[0].cmdline_params) == cmdline_params
     assert sorted(calc_info.local_copy_list) == sorted(local_copy_list)
     assert sorted(calc_info.retrieve_list) == sorted(retrieve_list)
     assert sorted(calc_info.retrieve_temporary_list) == sorted(retrieve_temporary_list)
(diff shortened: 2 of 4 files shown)


aiida_quantumespresso/calculations/__init__.py
```diff
@@ -264,12 +345,12 @@ def _generate_PWCPinputdata(cls, parameters, settings, pseudos, structure, kpoin
             if namelist in input_params:
                 # The following lines is meant to avoid putting in input the
                 # parameters like celldm(*)
-                stripped_inparams = [re.sub('[(0-9)]', '', _)
```
greschd, line 267: Ah crap, that's because of the `# yapf: enable` above. Ignore everything in this file below this  - I'll put it back in its original state later.

greschd:
Side note: I noticed the `# yapf: disable` statement in the `define` classmethod applies to the whole file after that, not just `define`. I doubt that's intentional, but should probably be fixed in a separate PR such as to not clutter the changes.

greschd:
@sphuber this is ready for review, but not quite for merging yet.

greschd:
Aliases are taken from https://gitlab.com/QEF/q-e/-/blob/develop/Modules/command_line_options.f90#L93

The choice of which alias is used in the interface is somewhat arbitrary - now it's the one that matches the internal variable name in QE.

sphuber:
> Side note: I noticed the # yapf: disable statement in the define classmethod applies to the whole file after that, not just define. I doubt that's intentional, but should probably be fixed in a separate PR such as to not clutter the changes.

Is that a bug in `yapf` or intended behavior? I thought those statements only applied to the scope to which they were applied. So I assumed that was only being applied to the define method. How did you discover this? And where should it be placed then? On the line of method definition?

aiida_quantumespresso/calculations/__init__.py
```diff
@@ -24,6 +26,44 @@ class BasePwCpInputGenerator(CalcJob):
     _DATAFILE_XML_POST_6_2 = 'data-file-schema.xml'
     _ENVIRON_INPUT_FILE_NAME = 'environ.in'
 
+    _ENABLED_PARALLELIZATION_FLAGS = []
+    # Tuples (flag_name, default, help) of parallelization flags possible
+    # in QE codes. The flags that are actually implemented in a given
+    # code should be specified in the '_ENABLED_PARALLELIZATION_FLAGS'
+    # list of each calculation subclass.
+    _ALLOWED_PARALLELIZATION_FLAGS = (  # yapf: disable
```
sphuber, line 34: Two things:
1. Find the prefix "allowed" a bit weird. Why not just `_PARALLELIZATION_FLAGS`?
2. Why not make this a dictionary with the flag name as the key? I think you only use it in line 125 and there you transform it in a dict. Might as well define it as one then and save those lines

aiida_quantumespresso/calculations/__init__.py
```diff
@@ -82,6 +122,29 @@ def define(cls, spec):
...
+                ) from exc
+
+            if default is None:
+                extra_kwargs = {'required': False}
+            else:
+                # We use 'functools.partial' here because lambda: `orm.Int(default)` would
+                # bind to the name `default`, and look up its value only when the lambda is
+                # actually called. Because `default` changes throughout the for-loop, this
+                # would give the wrong value.
+                extra_kwargs = {'default': partial(orm.Int, default)}
```
sphuber, line 145: Not sure about specifying these defaults. I think this would mean that all these calculations will now always have a bunch of int nodes as inputs even if the user doesn't specify anything. I am not sure we want this, at least not until we have proper automatic node deduplication in the database. If we still need the defaults to be set for consistency, we can maybe just do this on `prepare_for_submission` of the base class, but without going through input nodes

tests/calculations/test_pw.py
```diff
@@ -30,3 +34,36 @@ def test_pw_default(fixture_sandbox, generate_calc_job, generate_inputs_pw, file
...
+    Test that passing parallelization flags in the `settings['CMDLINE']
+    emits a `DeprecationWarning`.
+    """
+    Dict = DataFactory('dict')  # pylint: disable=invalid-name
+    entry_point_name = 'quantumespresso.pw'
+
+    inputs = generate_inputs_pw()
+    inputs['settings'] = Dict(dict={'CMDLINE': ['-{}'.format(flag_name), '2']})
+    with pytest.warns(DeprecationWarning):
+        generate_calc_job(fixture_sandbox, entry_point_name, inputs)
```
sphuber, line 69: maybe still check here that the value of the settings is kept and not overridden?

aiida_quantumespresso/calculations/__init__.py
greschd: I added these defaults mostly for caching: Two calculations could otherwise have different inputs but be functionally equivalent. They wouldn't use caching, just because the inputs are different.

I'll leave it up to you to decide if that's an important reason - I can also see the argument that the extra input nodes are unnecessary. Otherwise I can just remove the defaults.

aiida_quantumespresso/calculations/__init__.py
greschd: Yeah good point, will change that.

greschd:
Thanks for the review @sphuber, I'll make the changes tomorrow. Do you think we should aim to get this into the release?

aiida_quantumespresso/calculations/__init__.py
sphuber: Hmm, didn't think about the caching. Then again, maybe if we miss a cache hit because of the parallelization flags is not too bad as it should not occur too often. That reminds me though: didn't you mention somewhere that the actual defaults in QE are a bit surprising, i.e. not all 1 or something like that?

sphuber:
> Thanks for the review @sphuber, I'll make the changes tomorrow. Do you think we should aim to get this into the release?

Honestly, since this is a bit of an experiment, would it maybe make sense to not include it so that we have it in develop for some time and can get others to test it to see if they like it?

greschd:
Sure, I don't mind either way.

aiida_quantumespresso/calculations/__init__.py
greschd: The QE defaults are 1 for all inputs _except_ `ndiag` - that's why we don't have a default for that here. QE determines it based on the number of available processors and some of the other parallelization levels.

Should I remove the defaults for the other parallelization options, then?

greschd:
> 
> 
> > Side note: I noticed the # yapf: disable statement in the define classmethod applies to the whole file after that, not just define. I doubt that's intentional, but should probably be fixed in a separate PR such as to not clutter the changes.
> 
> Is that a bug in `yapf` or intended behavior? I thought those statements only applied to the scope to which they were applied. So I assumed that was only being applied to the define method. How did you discover this? And where should it be placed then? On the line of method definition?

I'm not sure, but I'd lean towards a yapf bug. The way I discovered it is that I added a `# yapf: enable` statement somewhere else, and it ended up reformatting a large part of the file.

tests/calculations/test_pw.py
greschd: Done!

aiida_quantumespresso/calculations/__init__.py
greschd: Done!

greschd:
Alright, I've addressed the comments above. One thing to still decide is whether we want to keep the ability to set defaults on the parallelization options.

If not, I would remove the corresponding code. I think it's still prudent to give the parallelization options set via `settings['CMDLINE']` higher precedence though, but only because it's difficult to reliably strip them, since e.g. `['-ndiag', '', '2']` is also technically valid.

Note that I've removed the py3.5 test because we're targeting post-release, so I went with py3.6 compatibility only. I will rebase after we removed py3.5 compatibility.

aiida_quantumespresso/calculations/__init__.py
```diff
@@ -202,6 +258,35 @@ def prepare_for_submission(self, folder):
...
+            # to make sure a `DeprecationWarning` is emitted whenever
+            # a flag is specified in the `settings['CMDLINE']`.
+            all_aliases = list(self._PARALLELIZATION_FLAG_ALIASES[flag_name]) + [flag_name]
+            if any(f'-{alias}' in cmdline_params_normalized for alias in all_aliases):
+                # To preserve backwards compatibility, we ignore the
+                # `parallelization.flag_name` input if the same flag is
+                # already present in `cmdline_params`. This is necessary
+                # because the new inputs can be set by their default,
+                # in user code that doesn't explicitly specify them.
+                warnings.warn(
```
greschd, line 281: While dog-fooding this, I noticed this warning is pretty invisible. I guess it's raised somewhere in the daemon, and then promptly discarded. Maybe it'd show up if using `run` instead of `submit`.

So my question is: Is ``self.logger.warning`` the right way to warn here? Or maybe warn both via the logger and ``warnings.warn``?

In the tests, do we have a way to capture the logging and check for it?

sphuber: > While dog-fooding this

Interesting, didn't know that term yet :smile: 

> I noticed this warning is pretty invisible. I guess it's raised somewhere in the daemon, and then promptly discarded. Maybe it'd show up if using `run` instead of `submit`.

Warnings are not shown by default in Python (used to be the opposite but was changed long time ago, since it usually just annoys end-users of libraries). This is also why we have the `AiidaDeprecationWarning` in `aiida-core` to make sure that doesn't get ignored by default. You could use that as well.

> So my question is: Is ``self.logger.warning`` the right way to warn here? Or maybe warn both via the logger and ``warnings.warn``?

I wouldn't use the logger because then it will end up in the database and I don't think this deprecation warning should go on the node. We should just use the special deprecation warning class from `aiida-core` I think.

> In the tests, do we have a way to capture the logging and check for it?

I think you can use the `warning` module itself, something like:
```
with warnings.catch_warnings(record=True) as captured:
    assert 'some warning' in captured
```

greschd: Ah, fantastic. Those end up in the daemon log I presume?

In that case the testing is also easy, we can use ``pytest.warns`` - I was just unsure about the `logging` case.

sphuber: > Ah, fantastic. Those end up in the daemon log I presume?

They are simply written to stdout, which when `run`ning is written to terminal, but for submitted jobs ends up in the daemon log yes.

> In that case the testing is also easy, we can use `pytest.warns`

Yes, this is also possible since it is just using Python's logging system

greschd: > Yes, this is also possible since it is just using Python's logging system

I think ``warnings.warn`` and ``logging`` are two different things, no? It seems for `logging` the `caplog` fixture needs to be used: https://docs.pytest.org/en/6.1.1/logging.html

Anyway, the code is updated to use `AiidaDeprecationWarning` now.

greschd:
Do you want to keep the option of setting defaults, or should I remove that code?

aiida_quantumespresso/calculations/__init__.py
sphuber: Yeah, I meant the `warnings` system of course.

aiida_quantumespresso/calculations/__init__.py
greschd: Right, I'm asking whether the _code_ allowing for defaults should also be removed.

sphuber: > the default is not None code path is dead now.

Not quite. A subclass could still override `_PARALLELIZATION_FLAGS` to define default for one or more flags and automatically get the spec to work correctly. Not sure if this would be useful, but could be the case. So maybe keeping the code is ok. That being said, what is this `MappingProxyType` and why do you use that to create the class attribute?

giovannipizzi:
A few comments on my side:
- ok to keep the settings higher priority for now; however I'm not sure `['-ndiag', '', '2']` is equivalent, when creating the submissions script we escape everything so I think that QE really gets an empty string for the value after `-ndiag` (if then fortran is weird and skips it, I don't know)

- is there a strong reason to use different ports (in a namespace) instead of a `parallelisation` Dict input node, and do validation later? I agree having multiple ports helps for having defaults, docs etc., but exactly because every code might be using different parallelisation options, and these might change between QE versions (see e.g. `-npools`, `-nks`, `-nk` are (or used to be?) all equivalent I think), isn't it better to just put the complex logic in a single python file?

greschd:
> 
> 
> A few comments on my side:
> 
>     * ok to keep the settings higher priority for now; however I'm not sure `['-ndiag', '', '2']` is equivalent, when creating the submissions script we escape everything so I think that QE really gets an empty string for the value after `-ndiag` (if then fortran is weird and skips it, I don't know)

Hm yeah, I have to admit I didn't test this. The point I was trying to make is that I'm not sure I'd get "stripping out" the explicitly passed parameters right, exactly because it's not very clear how these edge cases are handled.

> 
>     * is there a strong reason to use different ports (in a namespace) instead of a `parallelisation` Dict input node, and do validation later? I agree having multiple ports helps for having defaults, docs etc., but exactly because every code might be using different parallelisation options, and these might change between QE versions (see e.g. `-npools`, `-nks`, `-nk` are (or used to be?) all equivalent I think), isn't it better to just put the complex logic in a single python file?

No _particularly_ strong reason. In https://github.com/aiidateam/aiida-quantumespresso/issues/405 I proposed both variants. The advantage of separate inputs is that it plays nicer with defaults:
- it's not necessary to handle the defaults / missing parameters in code
- works nicer with caching: explicitly specifying the default value is the same as not specifying anything
Since we now don't specify the defaults anyway, I'm not sure how much weight that argument carries.

The "equivalent aliases" are handled: Only one variant is allowed, see the `_PARALLELIZATION_FLAG_ALIASES`.

As far as different QE versions go: To be honest, I hadn't thought of that. But I think it wouldn't be too hard to add. Since we would still want to support "older" QE versions, at the most we will have to add _new_ parallelization flags. We can always add a validation that checks if the used flags are compatible with the used QE version. We'd first have to introduce an input for passing the QE version, but that's the same for both variants.

If there's preference for the `Dict` version, I can switch that around, no problem.

greschd:
> If there's preference for the Dict version, I can switch that around, no problem.

@sphuber @giovannipizzi what's your preference here?

> Honestly, since this is a bit of an experiment, would it maybe make sense to not include it so that we have it in develop for some time and can get others to test it to see if they like it?

Would now-ish be a good time to push it into `develop` to get some exposure, or do you have a release planned soon?

sphuber:
I am planning a release soon which is need for the common workflows project. Not sure if that should stop us though. I am not sure how many people are actually working of `develop` and are willing to experiment with it. I have to go now, but will try to come back to this later today and see what I think we should do

sphuber:
I think going with the single `Dict` node is preferable for now:

* I am not sure we can justify creating around 4 additional base type nodes for each `PwCalculation`. Especially by default.
* We could still provide a default for the `Dict` node, specifying a dict with all flags set to 1 except `ndiag` because it is not necessarily 1. This would essentially solve the caching problem wouldn't it? The `ndiag` is more or less included indirectly through the `options` where the number of procs per cpu  and number of machines are defined. I am not sure if this is fully correct, because the default number of procs is set on the computer and maybe can be changed. That all being said, I am not sure we should create this node by default. I am not sure how many people are actually specifying parallelization flags currently and so potentially we will be adding a lot of extra nodes. If that is just to help with the unlikely caching hit, I am not sure if that is worth it.
* You mentioned that the individual nodes is also useful to automatically deal with default handling, but you still need to add code to the `prepare_for_submission` right? If we go to a single optional dict, you can simply add the handling code there, couldn't you? Or am I missing something.

TLDR: I think for now we should just be conservative and make a single optional `Dict` node called `parallelization`.

greschd:
Alright, switched the implementation to a single `Dict` input. The main TODOs now are documentation (where should I put this?) and adding `_ENABLED_PARALLELIZATION_FLAGS` to other codes (I don't know which flags each code takes..).

greschd:
Note to self: the (largely undocumented) `-n_howmany` flag is missing from this PR.

aiida_quantumespresso/calculations/__init__.py
```diff
@@ -203,8 +267,12 @@ def prepare_for_submission(self, folder):
         calcinfo = datastructures.CalcInfo()
 
         calcinfo.uuid = str(self.uuid)
-        # Empty command line by default
-        cmdline_params = settings.pop('CMDLINE', [])
+        # Start from an empty command line by default
+        cmdline_params = self._add_parallelization_flags_to_cmdline_params(
+                cmdline_params=settings.pop('CMDLINE', [])
```
sphuber, line 272: overindented no?
```suggestion
            cmdline_params=settings.pop('CMDLINE', [])
```

aiida_quantumespresso/calculations/__init__.py
```diff
@@ -87,6 +122,35 @@ def define(cls, spec):
...
+    @classmethod
+    def _validate_parallelization(cls, value, port_namespace):  # pylint: disable=unused-argument
+        if value:
+            value_dict = value.get_dict()
+            unknown_flags = set(value_dict.keys()) - set(cls._ENABLED_PARALLELIZATION_FLAGS)
+            if unknown_flags:
+                raise exceptions.InputValidationError(
+                    f"Unknown flags in 'parallelization': {unknown_flags}, "
+                    f'allowed flags are {cls._ENABLED_PARALLELIZATION_FLAGS}.'
+                )
```
sphuber, line 147: port validators should just return an error message string, not raise

aiida_quantumespresso/calculations/__init__.py
```diff
@@ -244,6 +312,55 @@ def prepare_for_submission(self, folder):
...
+        else:
+            parallelization_dict = {}
+        # To make the order of flags consistent and "nice", we use the
+        # ordering from the flag definition.
+        for flag_name in self._ENABLED_PARALLELIZATION_FLAGS:
+            all_aliases = list(self._PARALLELIZATION_FLAG_ALIASES[flag_name]) + [flag_name]
+            aliases_in_cmdline = [alias for alias in all_aliases if f'-{alias}' in cmdline_params_normalized]
+            if aliases_in_cmdline:
+                if len(aliases_in_cmdline) > 1:
+                    raise exceptions.InputValidationError(
```
sphuber, line 344: I prefer `ValueError`s here personally, because `InputValidationError` does not inherit from it and so you will have to catch it explicitly. Same for line 349

sphuber (changes requested):
Thanks @greschd , apart from some minor requests, this is ready to be merged. I don't think we need to specify the `_ENABLED_PARALLELIZATION_FLAGS` for any other plugins yet. We can check how this goes for pw.x and then apply elsewhere when needed and we figure out how it works. As for the docs, those are a bit of a mess. If you can add a stub anywhere close to the `PwCalculation` for now, at some point hopefully we can work on the docs with @mbercx and we will incorporate it in the correct place then

aiida_quantumespresso/calculations/__init__.py
greschd: Yeah.. we should fix the issue that makes the formatting not work on this file.

aiida_quantumespresso/calculations/__init__.py
greschd: I was following the example of the surrounding code - should I change those also?

aiida_quantumespresso/calculations/__init__.py
greschd: Done!

aiida_quantumespresso/calculations/__init__.py
sphuber: Maybe let's keep it the same then for now for consistency.

greschd:
For the documentation, I just added a description to the `Inputs` section in `pw.rst` - since this isn't all that complicated, I don't think it needs a dedicated section.

greschd:
Had to add an `__init__.py` to `aiida_quantumespresso/common`; otherwise `types.py` can shadow the standard library `types` module - at least in the eyes of `pylint`.

sphuber (approved):

sphuber:
THanks a lot @greschd

greschd:
Thanks for reviewing @sphuber!

greschd:
@sphuber I just noticed the squashed commit message for this PR is incorrect / outdated.

```
    `PwCalculation`: add explicit `parallelization` input port (#554)
    
    This new input port accepts a dictionary that allows one to specify the
    parallelization flags that `pw.x` accepts. This supersedes the old method
    of defining the flags in the `settings `input node. The latter is
    deprecated but still takes precedence if the same flag is specified
    through both input ports.
```

Specifically, we _raise_ "if the same flag is specified through both input ports.". Just FYI, not sure if it makes sense to go back and modify the commit message - but we should probably make sure it's correct in an eventual changelog.

sphuber:
So the only incorrect part is that it doesn't silently take the `settings` value in case of conflict but it actually raises? I think it is not that bad of a mistake and given that it has been in develop for a while I don't think I should correct the message and force push now

greschd:
Yeah, makes sense - force-pushing to `develop` seems like a bit of a drastic measure. At the end of the day, there's _always_ the possibility that old commit messages are now outdated.

PR #555, opened 2020-09-18
Possibility to run a NamelistCalculation using symlinks to the parent_folder

Crivella:
In most use-cases, the calculations  that are a subclass of `NamelistCalculation`, are used for post-processing that only reads data from the ./out/ folder of a scf/nscf calculation.

I find it useful to have the option of symlinking the ./out/ folder instead of straight out copying it as it avoids, in some cases, to have to copy 10s to 100s of GB of data, slowing down the starting of the job, and having to clear the remote_data afterwards to avoid huge disk consumption on the cluster.

With this pull request, the option to symlink the ./out/ folder is by default off (to be consistent with older inputs) and is turned on by specifying `'PARENT_FOLDER_SYMLINK': True` in the `settings` input Dict node.


sphuber (changes requested):
Thanks @Crivella , this looks fine. Could you just please add a quick test?

Crivella:
@sphuber 
I have added the tests to the pw2gw Calcjob tests (now a parametrized fixture runs the test with and w/o the symlynk option)

sphuber (approved):

sphuber:
Thanks a lot @Crivella

PR #558, opened 2020-09-23
Patch to the XSD for QE 6.6 to support mixed-case vdW functional names

### LINKED ISSUE #553: Problem with the XSD of QE
The [XSD (XML schema)](https://github.com/aiidateam/aiida-quantumespresso/blob/61fe3b4ec0b90df4dbc5dad7c452b47b83eef071/aiida_quantumespresso/parsers/parse_xml/pw/schemas/qes_191206.xsd) that is currently in the code has a problem when the functional is not provided all upper case.

E.g., if I specify `input_dft = 'rvv10'` (lower case), then In the output XML only one of the two values is converted to upper case:
```
$ verdi calcjob outputcat 2273 data-file-schema.xml | grep -i rvv10 
      <functional>rvv10</functional>
      <functional>RVV10</functional>
```

And then the parser fails with this error:
```
 | failed validating 'rvv10' with XsdEnumerationFacets("['PZ', 'BP', 'PBE', 'REVPBE', 'PBESOL', ...]"):
 | 
 | Reason: invalid value 'rvv10', it must be one of ['PZ', 'BP', 'PBE', 'REVPBE', 'PBESOL', 'BLYP', 'OLYP', 'PW91', 'WC', 'SOGGA', 'EV93', 'B3LYP', 'GauPBE', 'PBE0', 'HSE', 'VDW-DF', 'VDW-DF-CX', 'VDW-DF-C09', 'VDW-DF-OB86', 'VDW-DF-OBK8', 'VDW-DF2', 'VDW-DF2-C09', 'VDW-DF2-B86R', 'RVV10']
 | 
 | Schema:
 | 
 |   <enumeration xmlns="http://www.w3.org/2001/XMLSchema" value="PZ" />
 | 
 | Instance:
 | 
 |   <functional>rvv10</functional>
 | 
 | Path: /{http://www.quantum-espresso.org/ns/qes/qes-1.0}espresso/input/dft/functional
```

[This file](https://github.com/QEF/qeschemas/blob/912e059ae52c076616f132367b763fa0741a3fc7/PW_CPV/qes_devel_030920.xsd) fixes the capitalisation issue

Should we just overwrite the XSD version 191206 that is in the repo? This file is picked when parsing the output , since the XSD version is written by QE in the output of each run.
And how do we keep track of the 'version' of the patch fix of the XSD? Pinging @sphuber and @mbercx  for comments

giovannipizzi:
Fixes #553

The QES comes from [this file and commit](https://github.com/QEF/qeschemas/blob/912e059ae52c076616f132367b763fa0741a3fc7/PW_CPV/qes_devel_030920.xsd) and has been fixed by @pietrodelugas

I am just patching the qes_200420 xsd (used, I believe, by QE 6.6)

We could also 'manually' patch earlier QE versions, but I'm not currently
planning to do this to minimise problems in understanding which version of the schema
we are using.
In any case, wrong case just reflects (I think) in a warning from the parser and not
a complete failure, with the current implementation.

### DIFF
--- a/aiida_quantumespresso/parsers/parse_xml/pw/schemas/qes_200420.xsd
+++ b/aiida_quantumespresso/parsers/parse_xml/pw/schemas/qes_200420.xsd
@@ -263,30 +263,36 @@ Authors: Antonio Zambon, Paolo Giannozzi, Pietro Delugas
   </complexType>
   <simpleType name="functionalType">
     <restriction base="string">
-    <enumeration value="PZ"/>
-    <enumeration value="BP"/>
-    <enumeration value="PBE"/>
-    <enumeration value="REVPBE"/>
-    <enumeration value="PBESOL"/>
-    <enumeration value="BLYP"/>
-    <enumeration value="OLYP"/>
-    <enumeration value="PW91"/>
-    <enumeration value="WC"/>
-    <enumeration value="SOGGA"/>
-    <enumeration value="EV93"/>
-    <enumeration value="B3LYP"/>
-    <enumeration value="GauPBE"/>
-    <enumeration value="PBE0"/>
-    <enumeration value="HSE"/>
-    <enumeration value="VDW-DF"/>
-    <enumeration value="VDW-DF-CX"/>
-    <enumeration value="VDW-DF-C09"/>
-    <enumeration value="VDW-DF-OB86"/>
-    <enumeration value="VDW-DF-OBK8"/>
-    <enumeration value="VDW-DF2"/>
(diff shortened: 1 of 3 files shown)


sphuber:
Would be good to add a test that would have failed with the old schema but is fixed by this update. You know how to trigger it right? Was it just running with a specific vdW functional or in combination with some other setting?

sphuber:
@giovannipizzi I added the test. Hope you don't mind

sphuber (approved):
Thanks @giovannipizzi

PR #560, opened 2020-09-25
Add context argument to validator in PwBandStructureWorkChain

qiaojunfeng:
When submitting the `PwBandStructureWorkChain`, it complains
```
/home/junfeng/.virtualenvs/aiida/lib/python3.8/site-packages/plumpy/ports.py:194: UserWarning: the validator `validate_protocol` has a signature that only takes a single argument.
    This has been deprecated and the new signature is `validator(value, port)` where the `port` argument will be the
    port instance to which the validator has been assigned.
  warnings.warn(VALIDATOR_SIGNATURE_DEPRECATION_WARNING.format(self.validator.__name__))
```
Because the validator
https://github.com/aiidateam/aiida-quantumespresso/blob/59b957abdcb4b02e1745e262f8e2230566c65cc3/aiida_quantumespresso/workflows/pw/band_structure.py#L15-L24
lacks the context argument as required by this commit https://github.com/aiidateam/plumpy/commit/2b1a268f8d4fbcbdc40a319fee0e2cae91278382


greschd:
``TypeError: fname must be a string, filehandle, list of strings, or generator. Got <class 'aiida.orm.nodes.node.WarnWhenNotEntered'> instead.``

@sphuber I guess you know what that means - we messed up :see_no_evil:

sphuber:
So the problem is when we pass our wrapped `open` to `numpy.genfromtext`. They do some type checking and don't expect our wrapped class. The problem is the line `iter(fd)`. Shown by the following:
```
import io
node = Node()
node.put_object_from_filelike(io.BytesIO(b'test'), 'test', 'wb')
iter(node.open('test'))
```
will raise
```
TypeError                                 Traceback (most recent call last)
<ipython-input-12-ae0521db3156> in <module>
----> 1 iter(node.open('test'))

TypeError: 'WarnWhenNotEntered' object is not iterable
```
Could implement `__iter__` but what should it return? The content of the file line by line?

greschd:
Yeah - I had tested `__iter__`, but only when it _is_ used within the context manager. Which works because at that point we get the "bare" fileobj.

I don't really understand why it doesn't fall back on getting the `__iter__` from the `_fileobj`.

greschd:
I think this pre-commit failure is because the base classes in `aiida-core` changed, and can just be ignored with a ``pylint: diable`` statement:
```
aiida_quantumespresso/data/force_constants.py:9:0: R0901: Too many ancestors (8/7) (too-many-ancestors)
```

sphuber:
I was actually just working on #561 and was anyway going to add that to the ignore list. So no need to do anything. Once that is merged, will update this PR and it can be merged

sphuber (approved):
Thanks @qiaojunfeng

PR #562, opened 2020-09-29
Pre-commit: move `pytest` and `pylint` config to `pyproject.toml`

### LINKED ISSUE #561: Move tool configuration to `pyproject.toml` as much as possible

sphuber:
Fixes #561 

Requirement of `pylint` is upped as support for `pyproject.toml` was
only added in v6.0.


tests/parsers/test_pw.py
```diff
@@ -29,9 +29,8 @@ def _generate_inputs(calculation_type='scf', settings=None, metadata=None):
     return _generate_inputs
 
 
-def test_pw_default(
-    aiida_profile, fixture_localhost, generate_calc_job_node, generate_parser, generate_inputs, data_regression
-):
+@pytest.mark.usefixture('aiida_profile')
```
greschd, line 32: Just so you're aware, I think `pytestmark` can apply the fixture to all tests in a test module (haven't tried this, only looked at the documentation): https://docs.pytest.org/en/stable/reference.html#globalvar-pytestmark

pyproject.toml
```diff
@@ -1,3 +1,65 @@
...
+    'too-many-locals',
+]
+
+[tool.pylint.basic]
+good-names = [
+    'i',
+    'j',
+    'k',
+    'Run',
+    'UpfData',
```
greschd, line 32: Maybe we should just add `'*Data', '*Calculation', '*WorkChain'` here? It's a bit less strict, but we won't have to tediously maintain this list.

setup.json
```diff
@@ -76,7 +76,7 @@
         ],
```
greschd, line 76: With ``pylint==2.5.2`` I get the following error: 

```
$ pylint --version
  File "/home/a-dogres/.virtualenvs/aiida-qe/bin/pylint", line 8, in <module>
    sys.exit(run_pylint())
  File "/home/a-dogres/.virtualenvs/aiida-qe/lib/python3.7/site-packages/pylint/__init__.py", line 22, in run_pylint
    PylintRun(sys.argv[1:])
  File "/home/a-dogres/.virtualenvs/aiida-qe/lib/python3.7/site-packages/pylint/lint/run.py", line 306, in __init__
    linter.load_config_file()
  File "/home/a-dogres/.virtualenvs/aiida-qe/lib/python3.7/site-packages/pylint/config.py", line 785, in load_config_file
    for option, value in parser.items(section):
  File "/usr/lib/python3.7/configparser.py", line 857, in items
    return [(option, value_getter(option)) for option in d.keys()]
  File "/usr/lib/python3.7/configparser.py", line 857, in <listcomp>
    return [(option, value_getter(option)) for option in d.keys()]
  File "/usr/lib/python3.7/configparser.py", line 854, in <lambda>
    section, option, d[option], d)
  File "/usr/lib/python3.7/configparser.py", line 394, in before_get
    self._interpolate_some(parser, option, L, value, section, defaults, 1)
  File "/usr/lib/python3.7/configparser.py", line 411, in _interpolate_some
    p = rest.find("%")
AttributeError: 'list' object has no attribute 'find'
```

I think it is this issue, which is fixed in ``2.5.3``: https://github.com/PyCQA/pylint/issues/3538

Since the pre-commit appears to pin ``2.5.2``, I don't know why it doesn't show up in CI.

greschd: I think it would make sense to _pin_ the pylint version also here (probably to 2.6.0), because the messages change between versions.

tests/parsers/test_pw.py
sphuber: How does this work though? Following the docs, I added `pytestmark = pytest.mark.usefixture('aiida_profile')` at the top of the module file, but that gives me a bunch of warnings:
```
tests/parsers/test_q2r.py:13
  /home/sph/code/aiida/env/dev/aiida-quantumespresso/tests/parsers/test_q2r.py:13: PytestUnknownMarkWarning: Unknown pytest.mark.usefixture - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/mark.html
    @pytest.mark.usefixture('aiida_profile')

```

pyproject.toml
sphuber: I did this, but there might be a bug here. See this build failed: https://github.com/aiidateam/aiida-quantumespresso/pull/562/checks?check_run_id=1192909400 among other things because it contains
```
************* Module aiida_quantumespresso.calculations.pw
aiida_quantumespresso/calculations/pw.py:159:12: C0103: Variable name "HpCalculation" doesn't conform to snake_case naming style (invalid-name)
```
Why would `HpCalculation` here not match the `*Calculation` regex? Many other cases are accepted just fine

greschd: It seems my testing for whether this works was wrong.. but the only ones that are actually needed are ``'UpfData', 'PwCalculation', 'HpCalculation'``.

Note that this is only for the case where a _variable_ is called like this, i.e. when the factory is used to get the data / calculation / workchain type.

We could override the `variable-rgx` setting to make the regex work, but that's a bit ugly.

tests/parsers/test_pw.py
greschd: Seems it's called `pytest.mark.usefixtures`, not `pytest.mark.usefixture`. I have no idea why it worked without the `pytestmark = ...` though.

greschd: Ah, the `aiida_profile` fixture is used by default when using the AiiDA pytest plugin. We _could_ just remove all those usefixture marks:

https://github.com/aiidateam/aiida-core/blob/1e1bdf2dee779970654ed9b22eb996b78e9c4149/aiida/manage/tests/pytest_fixtures.py#L27

sphuber:
@greschd should be good now

greschd (approved):

PR #568, opened 2020-10-05
CI: add nightly build against `aiida-core/develop`

### LINKED ISSUE #567: Add nightly build that runs the test suite of `develop` against the `develop` branch of `aiida-core`
This will give us a chance to catch breaking changes in `aiida-core`. Ideally, we would be able to run it against the latest release of `aiida-quantumespresso` as well, although this might become more tricky. If, for example, `aiida-core/develop` is updated to become the next major release, e.g. v2.0.0, the last stable release _will_ most likely fail, but that is expected. Getting failing builds in this case will be false positives.

sphuber:
Fixes #567 

This should alert us whenever something is merged to the `develop`
branch of `aiida-core` that will break our current `develop` branch.

If either the Python install fails, or if the test suite fails, a
message will be sent to the `dev-aiida-qe` channel of the AiiDA slack
board.


sphuber:
@greschd with this setup, in case of failure, it just sends a notification to the `aiida-quantumespresso` dev channel on the AiiDA slack. I figured this is the least intrusive way that will make false positives not so bad. Other options, like opening issues, seemed worse in that hypothetical case.

sphuber:
@greschd this is blocked by PR #562 which will fail the failing pylint job

greschd:
Ah yeah, let me look at #562.

This is only for the `develop` version of `aiida-quantumespresso`, not the latest released one, correct?

>  will be sent to the `dev-aiida-qe` channel

Wasn't aware of that channel - joined it now.

greschd (approved):

Issue #443, opened 2019-11-15
Why are occupations always double the value printed by QE?


greschd:
In my PW calculations, the "occupations" value of the output bands is always 2 for occupied bands. This makes some sense for ``nspin=1`` calculations, because the physical number of electrons is of course twice what the calculation gives.

However, I get the same behavior for ``lnoncollinear=.true.`` calculations, too. That seems strange, because the sum of occupations is then twice the physical number of electrons. Is that an issue in the parser, or what is the reasoning behind this?

Personally, I think it would be more convenient to just have the same occupations as QE gives them (even for ``nspin=1``), but I understand that such a change would be highly backwards-incompatible.

greschd:
I'm using QE version 6.4.1, in case the change of output format has something to do with this.

greschd:
Note also that I didn't check for ``nspin=2``, which (from looking at the code), might be an exception.

PR #570, opened 2020-10-06
PwBandsWorkChain generating twice the amount of expected band for noncolin case (nspin=4)

Crivella:
In the case of a  `PwBandsWorkChain` run with nspin=4 and using the input `nbands_factor`, the amount of band requested from the calculation, is double that of the expected one.


greschd:
I guess this is related to #443?

Crivella:
@greschd  I don't think it is strictly related. This fix is not about the occupations, but just the number of bands that the workchain is setting for the PW bands calculations.

The unexpected behavior that this PR would fix arise from the fact that thenumber of bands is being calculated as `0.5 * nelectron * nspin * factor`. 
The problem is that also when nspin=4, only `nelectron * factor` (the same for nspin=2) should be requested and not twice that amount.

greschd:
Right - I didn't mean to imply it's exactly the same problem, just based in the same root cause: `lspinorb = True` being represented by `nspin = 4`, while there are actually not twice as many occupied bands compared to `nspin = 2`.

sphuber:
So I take it that for `nspin=4` one has the same amount of bands as for `nspin=2`? I guess this ok then for a quick fix, but relying on an integer that is being used as a "mode" by QE to represent a literal multiplication factor by us, is really fragile. Not sure if its likely that QE will be adding more values for `nspin` but that could really mess with all the code that makes that assumption

Crivella:
@sphuber 
Yes both `nspin=2` and `nspin=4` have 1 electron per band in contrast with `nspin=1` with 2 electrons per band.

I don't think it make sense for QE developers to add cases that requires more bands to be calculated. (So the max(2, ...) should hold)
A problem would arise if they ever add a case `nspin =  X > 1` that has 2 electrons per band.

In any case I don't think right now there is a solution without hard-coding every possibility into the workchain.
Maybe in that regard a dict containing a conversion from nspin to a second factor (instead of the 0.5 currently present) would be more elegant?

aiida_quantumespresso/workflows/pw/bands.py
```diff
@@ -217,7 +217,7 @@ def run_bands(self):
         if 'nbands_factor' in self.inputs:
             factor = self.inputs.nbands_factor.value
             parameters = self.ctx.workchain_scf.outputs.output_parameters.get_dict()
-            nspin = int(parameters['number_of_spin_components'])
+            nspin = min(2, int(parameters['number_of_spin_components']))
```
sphuber, line 220: Although this fixes it, I think this is a bit cryptic.
Could we instead make it:
```python
if int(parameters['number_of_spin_components']) > 1:
    nspin_factor = 2
else:
    nspin_factor = 1
....
nbnd = max(int(0.5 * nelectron * nspin_factor * factor), int(0.5 * nelectron * nspin_factor) + 4 * nspin_factor, nbands)
```

Crivella: This would indeed improve the readability of the code.
I have added the proposed changes.

sphuber (approved):
Thanks @Crivella

PR #571, opened 2020-10-07
Fix for required outputs that should be optional in PwBandsWorkChain

Crivella:
The outputs `primitive_structure` and `seekpath_parameters` should be optional.
Using manually specified kpoints (with the input `bands_kpoints`) right now causes the workchain to finish with non-zero exit status:  11 and exit_message `The process did not register a required output.`


sphuber (approved):
Thanks @Crivella

PR #579, opened 2020-10-09
Replace old format string interpolation with f-strings

### LINKED ISSUE #576: Replace formatted strings with f-string where possible

sphuber:
Fixes #576 

Since Python 3.5 is no longer supported, format string interpolations
can now be replaced by f-strings, introduced in Python 3.6, which are
more readable, require less characters and are more efficient.

Note that `pylint` issues a warning when using f-strings for log
messages, just as it does for format interpolated strings. The reasoning
is that this is slightly inefficient as the strings are always
interpolated even if the log is discarded, but also by not passing the
formatting parameters as arguments, the available metadata is reduced.
I feel these inefficiencies are premature optimizations as they are
really minimal and don't weigh up against the improved readability and
maintainability of using f-strings. That is why the `pylint` config is
update to ignore the warning `logging-fstring-interpolation` which
replaces `logging-format-interpolation` that was ignored before.

The majority of the conversions were done automatically with the linting
tool `flynt` which is also added as a pre-commit hook. It is added
before the `yapf` step because since `flynt` will touch formatting,
`yapf` will then get a chance to check it.

### DIFF
--- a/aiida_quantumespresso/calculations/__init__.py
+++ b/aiida_quantumespresso/calculations/__init__.py
@@ -6,7 +6,6 @@ import os
 from aiida import orm
 from aiida.common import datastructures, exceptions
 from aiida.common.lang import classproperty
-
 from qe_tools.converters import get_parameters_from_cell
 
 from aiida_quantumespresso.utils.convert import convert_input_to_namelist_entry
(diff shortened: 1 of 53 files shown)


aiida_quantumespresso/calculations/__init__.py
```diff
@@ -332,7 +331,7 @@ def _generate_PWCPinputdata(cls, parameters, settings, pseudos, structure, kpoin
                 )
 
             kind_names.append(kind.name)
-            atomic_species_card_list.append('{} {} {}\n'.format(kind.name.ljust(6), kind.mass, filename))
+            atomic_species_card_list.append(f'{kind.name.ljust(6)} {kind.mass} {filename}\n')
```
greschd, line 334: Should we change this to ``{kind.name:<6}``? Maybe it's just me, but I'm much more familiar with that way of specifying the format length.

greschd (approved):
Can't say I carefully checked each change, but upon a cursory glance I didn't see anything off here.

aiida_quantumespresso/calculations/__init__.py
sphuber: Me too, but I will leave that to another PR

PR #580, opened 2020-10-09
`Parsers`: remove explicit check for the retrieved folder's existence

### LINKED ISSUE #577: Remove explicit checking of existence of retrieved folders in the parsers
As of `aiida-core==1.4.0`, the engine itself will check for the presence of the retrieved folder. If it is missing, which is really just a bug in `aiida-core` the engine will terminate the process and not call the parser. Anyway there would be no point since there is nothing to parse. This means that now if the parser is called, it is guaranteed that the retrieve folder exists and so we no longer have to explicitly check for it.

sphuber:
Fixes #577 

As of `aiida-core==1.4.0`, the engine itself will check for the presence
of the retrieved folder. If it is missing, which is really just a bug in
`aiida-core` the engine will terminate the process and not call the
parser. Anyway there would be no point since there is nothing to parse.
This means that now if the parser is called, it is guaranteed that the
retrieve folder exists and so we no longer have to explicitly check for
it.

Since this is only valid from `aiida-core>=1.4` we update the dependency
requirement. We also require `aiida-core>=1.4.2` because `1.4.0` and
`1.4.1` contain a bug where the content of files in `local_copy_list`
are copied to the repository even though they shouldn't. For this plugin
this means that mostly the content of the pseudo potentials are
duplicated everytime in the repository, which will bloat the repository
unnecessarily.

### DIFF
--- a/aiida_quantumespresso/calculations/cp.py
+++ b/aiida_quantumespresso/calculations/cp.py
@@ -132,8 +132,6 @@ class CpCalculation(BasePwCpInputGenerator):
         spec.output('output_parameters', valid_type=orm.Dict)
         spec.default_output_node = 'output_parameters'
 
-        spec.exit_code(300, 'ERROR_NO_RETRIEVED_FOLDER',
-            message='The retrieved folder data node could not be accessed.')
         spec.exit_code(301, 'ERROR_NO_RETRIEVED_TEMPORARY_FOLDER',
             message='The retrieved temporary folder could not be accessed.')
         spec.exit_code(303, 'ERROR_MISSING_XML_FILE',
--- a/aiida_quantumespresso/calculations/dos.py
+++ b/aiida_quantumespresso/calculations/dos.py
@@ -26,8 +26,6 @@ class DosCalculation(NamelistsCalculation):
         spec.output('output_parameters', valid_type=orm.Dict)
         spec.output('output_dos', valid_type=orm.XyData)
         spec.default_output_node = 'output_parameters'
-        spec.exit_code(300, 'ERROR_NO_RETRIEVED_FOLDER',
-            message='The retrieved folder data node could not be accessed.')
         spec.exit_code(310, 'ERROR_OUTPUT_STDOUT_READ',
             message='The stdout output file could not be read.')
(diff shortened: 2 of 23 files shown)


greschd:
Hmm, I feel it's a bit early to remove ``1.3`` compatibility.. ``1.4`` is still relatively new. I may be biased though, since I haven't yet updated my production environment :smile: 

Anyway, maybe we could just add ``!=1.4.0,!=1.4.1`` to the requirements now, and punt the rest of the changes for a bit? We can leave the PR as is, just put it on hold for a while.

sphuber:
> Hmm, I feel it's a bit early to remove 1.3 compatibility.. 1.4 is still relatively new.

Fair enough, but this change requires 1.4. However, it is not pressing, so we can keep it open until we feel 1.4 is stable enough

greschd:
Alright, thanks. Did I understand it right that the current code will work fine with 1.3 and 1.4, but at 1.4 what we're removing here is essentially dead code?

And by the way: Does `aiida-core` also use the `300` exit code for this, or something else?

sphuber:
> Did I understand it right that the current code will work fine with 1.3 and 1.4, but at 1.4 what we're removing here is essentially dead code?

Yes

> And just out of curiosity: Does aiida-core also use the 300 exit code for this, or something else?

It does define it, but it uses `100`, as `1**` is reserved for `aiida-core` internal exit codes:

https://github.com/aiidateam/aiida-core/blob/bd6903d88a4d88077150763574784a1bc375c644/aiida/engine/processes/calcjobs/calcjob.py#L197

I think it is fine, because the exit codes defined here in the plugin stem from the alpha days where I put them as a pre-caution.

sphuber:
@greschd although this change is not really pressing, the other PR #581 to add support for `aiida-pseudo` *is* pressing, for another project we are running. This also requires `aiida-core~=1.4` for the support of extras on groups. I see your point that v1.4 is still relatively young, but this will just make `develop` require it. We still have `aiida-quantumespresso==3.2` that is compatible with lower versions. So I would argue that it is fine to make develop require v1.4.

greschd:
Right then, let's do this. I guess if I need to use `develop` in production I might as well update to ``1.4.2``.

greschd (approved):

PR #581, opened 2020-10-11
Add support for `aiida-pseudo`

### LINKED ISSUE #512: Add support for `aiida-pseudo`
Note that `aiida-pseudo` requires `aiida-core>=1.4` so this will also become the minimum requirement for `aiida-quantumespresso`

sphuber:
Fixes #512 

This new library provides functionality to make it easier to install
pseudo potential families, either from an archive of pseudo potential
files on the local file system, or by directly downloading from the web.
There is special functionality to install an SSSP family which will also
install the corresponding recommended cutoffs that can be easily
retrieved from a family for a given set of elements or a structure.

For now, the library is just plumbed into the CLI commands that launch a
process that requires pseudos in the input. The plugin also provides a
new implementation of `UpfData` that is different from the one that
ships with `aiida-core` so the base calculation is updated to accept
both classes for the `pseudos` input namespace. They should work the
same from the process' perspective. At some point the data plugin from
the `aiida-pseudo` plugin should supersede that of `aiida-core` which
will be deprecated.


mbercx:
Thanks @sphuber! I tried running a `PwBaseWorkChain` using this branch and `aiida-pseudo==0.1.0`, and ran into an issue:

``` bash
$ verdi process report 94
2020-10-26 11:59:03 [1 | REPORT]: [94|PwBaseWorkChain|validate_pseudos]: pseudo for element Si is not of type UpfData
2020-10-26 11:59:03 [2 | REPORT]: [94|PwBaseWorkChain|on_terminated]: remote folders will not be cleaned
```
This is because during the `validate_pseudos` step, the `validate_and_prepare_pseudos_inputs` function is called:

https://github.com/aiidateam/aiida-quantumespresso/blob/5f46f51e2fe41f59e9db5d50aea7664d6467acfa/aiida_quantumespresso/workflows/pw/base.py#L162-L163

Which only imports the `aiida-core` version of `UpfData`:

https://github.com/aiidateam/aiida-quantumespresso/blob/5f46f51e2fe41f59e9db5d50aea7664d6467acfa/aiida_quantumespresso/utils/pseudopotential.py#L3

and then raises a `ValueError` in case the pseudo isn't an instance of this type:

https://github.com/aiidateam/aiida-quantumespresso/blob/5f46f51e2fe41f59e9db5d50aea7664d6467acfa/aiida_quantumespresso/utils/pseudopotential.py#L42-L43

I'd be happy to work on fixing this and doing some more testing, if I can get push permissions to the branch. :)

sphuber:
This branch is actually waiting for `aiida-pseudo==0.2.0` to be released. I already added some functionality that requires that new release and the unit tests are now failing because of a bug in `0.1.0` that is already fixed. Thanks for the offer, but I will wrap this PR up myself once I have released the new `aiida-pseudo` version

mbercx:
@sphuber is this ready for review?

sphuber:
Almost, haven't had the time to go through to the code that isn't unit tested yet. For example `validate_and_prepare_pseudos_inputs` has not been adapted yet. WIll give you a sign when its ready for review

sphuber:
ok, this is ready for review @mbercx . Note that there is still some code that deals with only legacy `UpfData`, such as the protocols for the `PwBandStructureWorkChain` and the immigrator code, but both are outdated and should be removed. Don't really feel like spending time fixing that and making it compatible with old and new pseudos

aiida_quantumespresso/tools/pwinputparser.py
```diff
@@ -59,7 +61,7 @@ def get_kpointsdata(self):
         return kpoints
 
 
-def create_builder_from_file(input_folder, input_file_name, code, metadata, pseudo_folder_path=None, use_first=False):
+def create_builder_from_file(input_folder, input_file_name, code, metadata, pseudo_folder_path=None):
```
mbercx, line 64: This function works just fine, and might be handy in some cases. However, now it's a little hard to find (I wasn't aware of it). Would it be a good idea to implement a `.from_folder()` method for each process class that replicates this behaviour?

aiida_quantumespresso/utils/pseudopotential.py
```diff
@@ -1,6 +1,10 @@
 # -*- coding: utf-8 -*-
 """Utilities for pseudo potentials."""
-from aiida.orm.nodes.data.upf import UpfData, get_pseudos_from_structure
+from aiida.orm.nodes.data.upf import UpfData as LegacyUpfData
```
mbercx, line 3: In the `aiida_quantumespresso/calculations/__init__.py` file, you just load the `aiida.orm` package and use `orm.UpfData` to get what here is referred to as `LegacyUpfData`. Why not consistently use the `DataFactory` for both:

```Python
LegacyUpfData = DataFactory('upf')
UpfData = DataFactory('pseudo.upf')
```

Also in the `aiida_quantumespresso/calculations/__init__.py` file?

mbercx (changes requested):
Thanks @sphuber! I've left some (nitpicky) comments and questions.

README.md
```diff
@@ -34,9 +34,10 @@ To enable it, add the following to your shell loading script, e.g. the `.bashrc`
 The tool comes with various sub commands, for example to quickly launch some calculations and workchains
 For example, to launch a test `PwCalculation` you can run the following command:
 
-    aiida-quantumespresso calculation launch pw -X pw-v6.1 -p SSSP_v1.1_eff_PBE
+    aiida-quantumespresso calculation launch pw -X pw-v6.1 -F SSSP/1.1/PBE/efficiency
 
-Note that this requires the code `pw-v6.1` and pseudo potential family `SSSP_v1.1_eff_PBE` to be configured.
+Note that this requires the code `pw-v6.1` and pseudo potential family `SSSP/1.1/PBE/efficiency` to be configured.
```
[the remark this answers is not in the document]
sphuber: Aye aye cap'n. Fair point

aiida_quantumespresso/cli/utils/options.py
```diff
@@ -4,15 +4,46 @@
...
+from aiida.cmdline.utils import decorators
+from aiida.common import exceptions
 
 from . import validate
 
-STRUCTURE = OverridableOption(
-    '-s', '--structure', type=types.DataParamType(sub_classes=('aiida.data:structure',)), help='StructureData node.'
-)
+
+class PseudoFamilyType(types.GroupParamType):
```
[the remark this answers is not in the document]
sphuber: Not sure, in this case I made a custom type in order to print a custom message. This will help people that pass an old `Group` or `UpfFamily` that they used to use as pseudopotential family, but will now no longer work. At least this way they can quickly install a new group that *will* work with the new interface. I am not sure whether all other plugins need this.

aiida_quantumespresso/cli/utils/options.py
```diff
@@ -4,15 +4,46 @@
...
+                raise click.BadParameter(
+                    f'`{value}` is not a supported pseudo potential family type.\nTo install a supported pseudo family,'
+                    ' use the `aiida-pseudo` plugin. See the following link for detailed instructions:\n\n'
+                    '    https://github.com/aiidateam/aiida-quantumespresso#pseudo-potentials'
+                ) from exception
+
+        return group
+
 
 PSEUDO_FAMILY = OverridableOption(
```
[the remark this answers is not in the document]
sphuber: Potentially yes, but the types are plugin specific. We require `upf` or `sssp`. Of course a plugin could use the option and simply override the `type`, but then there is not much code that is actually reused. Still, if other people find it useful, happy to  move it to `aiida-pseudo`.

aiida_quantumespresso/tools/pwinputparser.py
sphuber: Feel free to open a feature request. At some point we would need something similar anyway in order to implement immigrator functionality. I still have an open AEP for this with an implementation in `aiida-core` which I need to get back to at some point, but it would still require the functionality that constructs a builder from a folder of input files. I won't address this in this PR of course.

aiida_quantumespresso/cli/utils/options.py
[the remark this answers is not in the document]
mbercx: I figured the warning message could also be used more generally, e.g. in case someone tries to use a upf family for running a Siesta `Calcjob`. We can change the link to the `aiida-pseudo` README and add the supported types there.

There also seems to be the `PseudoPotentialFamilyParam` class in the `aiida-pseudo` package, which has a similar purpose? I suppose its usage in the `show` command could also benefit from the warning in case someone tries to apply it to an old `Group` or `UpfFamily`.

Nothing critical though, just a thought.

aiida_quantumespresso/cli/utils/options.py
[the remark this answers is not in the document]
mbercx: Here I definitely see your point. We'd require quite a few imports to avoid a bit of code duplication. For some reason I thought the upf format could be used in a lot more DFT codes. (U does _not_ stand for universal. 😅 )

aiida_quantumespresso/cli/utils/options.py
```diff
@@ -4,15 +4,46 @@
...
+                from aiida.orm import load_group
+                load_group(value)
+            except exceptions.NotExistent as exception:  # pylint: disable=try-except-raise
+                raise
+            else:
+                raise click.BadParameter(
+                    f'`{value}` is not a supported pseudo potential family type.\nTo install a supported pseudo family,'
+                    ' use the `aiida-pseudo` plugin. See the following link for detailed instructions:\n\n'
+                    '    https://github.com/aiidateam/aiida-quantumespresso#pseudo-potentials'
+                ) from exception
```
mbercx, line 32:
```suggestion
                )
```
`exception` is not defined here, only in the previous `except`.

sphuber: Fixed

sphuber:
Maybe we should hold off with merging this until we fixed `aiida-pseudo` because now it will be impossible to install the SSSP through it, or has the legacy server been put back into action.

mbercx:
> Maybe we should hold off with merging this until we fixed `aiida-pseudo` because now it will be impossible to install the SSSP through it, or has the legacy server been put back into action.

The legacy server is back online, and can stay online until we change the URL's.

sphuber:
> The legacy server is back online, and can stay online until we change the URL's.

So do you propose we merge this then?

mbercx:
> So do you propose we merge this then?

Sure, I don't see why not? Perhaps wait with including it in a release until we've updated and tested the new URL's in `aiida-psuedo`.

sphuber:
Alright, feel free to approve it then so I can merge, or is there still a remark outstanding

mbercx:
> Alright, feel free to approve it then so I can merge, or is there still a remark outstanding

Ah, yes: there was still the idea of moving the `PseudoFamilyType` to `aiida-pseudo`, see [here](https://github.com/aiidateam/aiida-quantumespresso/pull/581#discussion_r514232642). I also still spot a few two-word pseudopotentials in the README. 😅 

Out of curiosity: is there a reason you don't commit the suggestions on github, but instead fix them manually?

sphuber:
> Ah, yes: there was still the idea of moving the `PseudoFamilyType` to `aiida-pseudo`, see [here](https://github.com/aiidateam/aiida-quantumespresso/pull/581#discussion_r514232642).

As I said in my response, I think it is better to wait to see weather other plugin packages also want to use it. Anyway, we would have to make a new release first before we could merge this, and prefer to merge it now.

> I also still spot a few two-word pseudopotentials in the README. sweat_smile

Apologies, they should all have been squashed into correctness now

> Out of curiosity: is there a reason you don't commit the suggestions on github, but instead fix them manually?

Out of habit mostly, I guess. Just prefer to edit locally and use the cli. In addition, I think you would receive an email for each suggestion I accept, wouldn't it? I personally find it quite annoying if my inbox is flooded with emails.

mbercx:
> As I said in my response, I think it is better to wait to see weather other plugin packages also want to use it. Anyway, we would have to make a new release first before we could merge this, and prefer to merge it now.

👍 
 
> Out of habit mostly, I guess. Just prefer to edit locally and use the cli. In addition, I think you would receive an email for each suggestion I accept, wouldn't it? I personally find it quite annoying if my inbox is flooded with emails.

I don't think you receive any emails, and I just love pressing that "Add suggestion to batch" button. 😄  To each their own, I guess!

aiida_quantumespresso/cli/utils/options.py
```diff
@@ -4,15 +4,46 @@
...
+            group = super().convert(value, param, ctx)
+        except click.BadParameter:
+            try:
+                from aiida.orm import load_group
+                load_group(value)
+            except exceptions.NotExistent:  # pylint: disable=try-except-raise
+                raise
+            else:
+                raise click.BadParameter(  # pylint: disable=raise-missing-from
+                    f'`{value}` is not a supported pseudo potential family type.\nTo install a supported pseudo family,'
```
mbercx, line 29:
```suggestion
                    f'`{value}` is not of a supported pseudopotential family type.\nTo install a supported pseudofamily,'
```

😄

aiida_quantumespresso/cli/utils/options.py
```diff
@@ -4,15 +4,46 @@
...
+        return group
+
 
 PSEUDO_FAMILY = OverridableOption(
-    '-p', '--pseudo-family', 'pseudo_family', type=click.STRING, help='Pseudo potential family name.'
+    '-F',
+    '--pseudo-family',
+    type=PseudoFamilyType(sub_classes=('aiida.groups:pseudo.family.upf', 'aiida.groups:pseudo.family.sssp')),
+    required=True,
+    help='Select a pseudo potential family.'
```
mbercx, line 42:
```suggestion
    help='Select a pseudopotential family.'
```
😄

aiida_quantumespresso/cli/utils/options.py
sphuber: In my defense, you said there were some cases remaining in the README not anywhere else ;)

mbercx:
Eh, oops. I guess my suggestion broke the pylint `line-too-long` rule. 😅

sphuber:
> Eh, oops. I guess my suggestion broke the pylint `line-too-long` rule. sweat_smile

guess there we have another reason for using the CLI :)

mbercx (approved):

PR #584, opened 2020-10-13
Add the `Framework :: AiiDA` trove classifier to the package

### LINKED ISSUE #583: AiiDA registry points out issues with setup.json
```
  >> WARNING: Missing classifier 'Framework :: AiiDA'
  >> WARNING: Entry point 'spinorbithydrogen' does not start with 'quantumespresso'
  >> WARNING: Entry point 'noncollinearhydrogen' does not start with 'quantumespresso'
```

The first one is easy to fix; and I also suggest to fix the other ones - unless there is a very good reason to put those entry points into the global namespace without a `quantumespresso` prefix

sphuber:
Fixes #583

### DIFF
--- a/setup.json
+++ b/setup.json
@@ -2,6 +2,7 @@
     "author": "The AiiDA team",
     "author_email": "developers@aiida.net",
     "classifiers": [
+        "Framework :: AiiDA",
         "License :: OSI Approved :: MIT License",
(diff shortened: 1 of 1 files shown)


ltalirz (approved):
thanks @sphuber !

PR #585, opened 2020-10-14
Remove duplicate line of setting cmdline arguments

qiaojunfeng:
The `cmdline_params` should be in `codeinfo`, not `calcinfo`, am I right?


greschd:
I _think_ you're right, but the tests explicitly test the calcinfo. Maybe the tests need to be adapted, but I'll let @sphuber give his opinion on this.

sphuber:
It should indeed be on the `CodeInfo`

https://github.com/aiidateam/aiida-core/blob/bd6903d88a4d88077150763574784a1bc375c644/aiida/common/datastructures.py#L150

The tests are just checking the wrong object and because of the duplicated line they are actually passing.

sphuber (approved):
Thanks @qiaojunfeng

PR #587, opened 2020-10-15
Pp parser memory fix

eimrek:
Hello!

This PR considerably reduces the memory usage of the pp.x parser. Before, all the retrieved files were loaded into memory and then parsed. This quickly ate up all the RAM and made our system crash for some of our use cases (e.g. we were parsing many orbital cubes files). Now parsing happens right after each retrieved file is read (and the memory is freed when the next retrieved file is read).

### DIFF
--- a/aiida_quantumespresso/parsers/pp.py
+++ b/aiida_quantumespresso/parsers/pp.py
@@ -138,6 +138,7 @@ class PpParser(Parser):
                 try:
                     key = get_key_from_filename(filename)
                     data_parsed.append((key, parsers[iflag](data_raw)))
+                    del data_raw
                 except Exception:  # pylint: disable=broad-except
                     return self.exit_codes.ERROR_OUTPUT_DATAFILE_PARSE.format(filename=filename)
 
--- a/aiida_quantumespresso/parsers/pp.py
+++ b/aiida_quantumespresso/parsers/pp.py
@@ -123,7 +123,7 @@ class PpParser(Parser):
             pattern = r'{}_(.*){}'.format(filename_prefix, filename_suffix)
             matches = re.search(pattern, filename)
             return matches.group(1)
-        
+
         for filename in filenames:
             if filename.endswith(filename_suffix):
                 try:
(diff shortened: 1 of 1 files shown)


aiida_quantumespresso/parsers/pp.py
```diff
@@ -140,12 +124,25 @@ def get_key_from_filename(filename):
             matches = re.search(pattern, filename)
             return matches.group(1)
 
-        for filename, data in data_raw:
-            try:
-                key = get_key_from_filename(filename)
-                data_parsed.append((key, parsers[iflag](data)))
-            except Exception:  # pylint: disable=broad-except
-                return self.exit_codes.ERROR_OUTPUT_DATAFILE_PARSE.format(filename=filename)
+        for filename in filenames:
```
sphuber, line 127: Would be good to add a comment here to explain that with this setup, the `data_raw` is parsed directly and then overwritten in the next iteration, which prevents all raw data from being read into memory. Took me a while to realize this from just reading the code

sphuber (changes requested):
Thanks @eimrek . Change looks all good, but would be good to add comment as described in my comment.

aiida_quantumespresso/parsers/pp.py
```diff
@@ -140,12 +124,25 @@ def get_key_from_filename(filename):
...
+            if filename.endswith(filename_suffix):
+                try:
+                    with file_opener(filename) as handle:
+                        data_raw = handle.read()
+                except OSError:
+                    return self.exit_codes.ERROR_OUTPUT_DATAFILE_READ.format(filename=filename)
+
+                try:
+                    key = get_key_from_filename(filename)
+                    data_parsed.append((key, parsers[iflag](data_raw)))
```
greschd, line 137: Probably adding a `del data_raw` here could reduce the memory usage further. I think because `handle.read()` is executed before the assignment on `data_raw` above, you can still end up with a case where two objects are in memory simultaneously.

aiida_quantumespresso/parsers/pp.py
eimrek: Good point, will add a comment!

sphuber (approved):

sphuber:
Thanks @eimrek !

PR #588, opened 2020-10-16
projwfc parser allow digit in element name

eimrek:
Hi again, I commonly use digits in the species names to mark the initial spin guess (e.g. C1, C2). Recently the projwfc parser stopped supporting these names. I reverted back to the old regex that was used.

### DIFF
--- a/aiida_quantumespresso/parsers/projwfc.py
+++ b/aiida_quantumespresso/parsers/projwfc.py
@@ -46,7 +46,7 @@ def find_orbitals_from_statelines(out_info_dict):
 
     out_file = out_info_dict['out_file']
     atomnum_re = re.compile(r'atom\s*([0-9]+?)[^0-9]')
-    element_re = re.compile(r'atom\s*[0-9]+\s*\(\s*(.*?)\s*\)')
+    element_re = re.compile(r'atom\s*[0-9]+\s*\(\s*([A-Za-z0-9_-]+?)\s*\)')
     if out_info_dict['spinorbit']:
         # spinorbit
         lnum_re = re.compile(r'l=\s*([0-9]+?)[^0-9]')
--- a/aiida_quantumespresso/parsers/projwfc.py
+++ b/aiida_quantumespresso/parsers/projwfc.py
@@ -46,7 +46,7 @@ def find_orbitals_from_statelines(out_info_dict):
 
     out_file = out_info_dict['out_file']
     atomnum_re = re.compile(r'atom\s*([0-9]+?)[^0-9]')
-    element_re = re.compile(r'atom\s*[0-9]+\s*\(\s*([A-Za-z]+?)\s*\)')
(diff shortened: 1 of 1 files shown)


aiida_quantumespresso/parsers/projwfc.py
```diff
@@ -46,7 +46,7 @@ def find_orbitals_from_statelines(out_info_dict):
 
     out_file = out_info_dict['out_file']
     atomnum_re = re.compile(r'atom\s*([0-9]+?)[^0-9]')
-    element_re = re.compile(r'atom\s*[0-9]+\s*\(\s*([A-Za-z]+?)\s*\)')
+    element_re = re.compile(r'atom\s*[0-9]+\s*\(\s*(.*?)\s*\)')
```
sphuber, line 49: Instead of matching all characters, could we not just add digits, to be a bit more safe on what is matched?
```suggestion
    element_re = re.compile(r'atom\s*[0-9]+\s*\(\s*([A-Za-z0-9]+?)\s*\)')
```

greschd:
To prevent regressing on this again, there should probably be a test for this?

aiida_quantumespresso/parsers/projwfc.py
eimrek: Okay i guess if we want to be strict, we should allow everything that pw.x allows for species labels:

```
X | CHARACTER
-- | --
label of the atom. Acceptable syntax: chemical symbol X (1 or 2 characters, case-insensitive) or chemical symbol plus a number or a letter, as in "Xn" (e.g. Fe1) or "X_*" or "X-*" (e.g. C1, C_h; max total length cannot exceed 3 characters)
```

I'll make a commit with `[A-Za-z0-9_-]`

eimrek:
> To prevent regressing on this again, there should probably be a test for this?

I can take a look, but honestly the whole testing process is not super clear to me

aiida_quantumespresso/parsers/projwfc.py
sphuber: I wouldn't add the hard limit of three characters though. This is an arbitrary limitation they have because of memory allocation, and we have to work around this in some of our projects by updating the source and recompiling it.

eimrek: ah sure, well the latest commit shouldn't impose that. i just added underscores and dashes to your suggestion.

sphuber:
@eimrek I started to work on writing a small patch for you to include to add a test, but ended up refactoring all the tests for the `ProjwfcParser` quite a bit. I have opened #591 that includes those refactorings as well as add a regression test and the fix you proposed here. I have included a co-authored-by in the commit message to acknowledge your contribution. If its ok with you, we could merge that PR instead.

eimrek:
@sphuber Looks great! Yes, i'm perfectly fine with merging the #591 instead. Thanks a lot!

PR #589, opened 2020-10-16
CLI: add tests for the process launch commands

sphuber:
The CLI had no tests whatsoever, and especially for the process launch
commands this was becoming critical, as they contain quite some code and
they are used a lot for demonstration purposes. Here we add tests for
all commands that launch a process, just by passing the required inputs.
This tests that the command can be invoked and that the code path for
the default inputs can be executed without problem.

To prevent the process from actually being submitted, the utility method
`cli.utils.launch.launch_process` is monkey patched to be a no-op. The
code of this utility should be tested once separately.

### DIFF
--- a/tests/cli/conftest.py
+++ b/tests/cli/conftest.py
@@ -1,0 +1,63 @@
+# -*- coding: utf-8 -*-
+# pylint: disable=redefined-outer-name
+"""Fixtures for the command line interface."""
+import pytest
+
+
+def mock_launch_process(*_, **__):
+    """Mock the :meth:`~aiida_quantumespresso.cli.utilslaunch.launch_process` to be a no-op."""
+    return
+
+
+@pytest.fixture
+def run_cli_command():
+    """Run a `click` command with the given options.
+
+    The call will raise if the command triggered an exception or the exit code returned is non-zero.
+    """
+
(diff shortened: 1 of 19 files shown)


sphuber:
@greschd I started to work on something that touched the CLI and broke it without being noticed, so I thought it best to finally add some minimal testing for it.

tests/cli/conftest.py
```diff
@@ -0,0 +1,63 @@
...
+        :param raises: optionally an exception class that is expected to be raised
+        """
+        import traceback
+        from click.testing import CliRunner
+
+        runner = CliRunner()
+        result = runner.invoke(command, options or [])
+
+        if raises is not None:
+            assert result.exception is not None, result.output
```
greschd, line 33: Shouldn't there be a way to check for the error message that is printed here?

greschd:
The pytest fixtures look ok  - I'm not really familiar with the `aiida-quantumespresso` CLI, so maybe someone else should review the individual tests.

tests/cli/conftest.py
sphuber: We could add this later yeah, where you can specify a `match` argument that can match the error like in `pytest.raises`.

PR #592, opened 2020-10-21
Fix bug in pw protocols

Crivella:
Right now, the assignment at line
https://github.com/aiidateam/aiida-quantumespresso/blob/216015807adb255e95df8a0c80a2379ea4834818/aiida_quantumespresso/utils/protocols/pw.py#L66
makes it so that the protocols 'default' and 'scdm' are the same dictionary.
For this reason, the successive change in 'scdm' are also done on 'default' which I believe is not the intended behavior.


sphuber:
Thanks @Crivella . Note that these protocols will soon be deprecated and removed by a new system. This `PwBandStructureWorkChain` was once added as a quick proof-of-principle of a turn-key solution but will soon be superseded.

Issue #582, opened 2020-10-12
Remove parser version from output_parameters


mbercx:
Now that `aiida-core` adds both the AiiDA and plugin version to `CalcJob` node attributes, it is no longer necessary to add this information to the `output_parameters` dictionary in the `parser_info` and `parser_version` keys.

Currently, we have to update the `yml` files produced by `pytest`'s `data_regression` check for every release, which is a little tedious. Since adding the version to the `output_parameters` just duplicates information that is already stored on the `CalcJob` node, it would be better to remove it.

Although this change isn't backwards compatible, it seems unlikely to me that this output is heavily used. So I would be fine with just removing it.

mbercx:
Pinging @sphuber and @greschd for comments.

greschd:
> Now that `aiida-core` adds both the AiiDA and plugin version to `CalcJob` node attributes, it is no longer necessary to add this information to the `output_parameters` dictionary in the `parser_info` and `parser_version` keys.

Do you know exactly which `aiida-core` version added this? If it's added in all versions we currently support I agree we can remove this output.

sphuber:
Was add in [this commit](https://github.com/aiidateam/aiida-core/commit/5058e23c93f1f5f7fbcdffe82ab5dd7dd1f01829) which has been part of `aiida-core` since `v1.0.0b5`.

greschd:
So, good to remove then?

mbercx:
Yes, seems we are all agreed! :) I'll start working on removing them and open a PR.

PR #593, opened 2020-10-21
Release v3.2.1

sphuber:


mbercx:
Those failing tests reminded me of my promise in #582. 😅 I'll do this later this week, so we don't have to fix this for the next release.

PR #596, opened 2020-10-23
`PwCalculation`: fix bug when `hubbard_file` is specified

### LINKED ISSUE #595: `PwCalculation`: bug in `prepare_for_submission` when `hubbard_file` is specified
The line:
https://github.com/aiidateam/aiida-quantumespresso/blob/6924ce13b6d02caa8d0ca5e2dc238a1d72ae650e/aiida_quantumespresso/calculations/__init__.py#L125
is incorrect, it needs to use the `.uuid` property obviously.
Also the `HpCalculation.input_file_name_hubbard_file` has been renamed to `HpCalculation.filename_input_hubbard_parameters`

sphuber:
Fixes #595 

The `prepare_for_submission` was using the `filename` of the
`hubbard_file` input as the UUID in the `local_copy_list` which
obviously fails. In addition, the `input_file_name_hubbard_file`
classproperty of the `HpCalculation` clas was changed and adapted here
to `filename_input_hubbard_parameters`.

Note that a test cannot be added because it would have to depend on the
`aiida-quantumespresso-hp` plugin, which itself depends on
`aiida-quantumespresso`, creating a circular dependency. The test will
be added to the former to make sure this behavior works.

### DIFF
--- a/aiida_quantumespresso/calculations/__init__.py
+++ b/aiida_quantumespresso/calculations/__init__.py
@@ -122,9 +122,9 @@ class BasePwCpInputGenerator(CalcJob):
             local_copy_list.append((uuid, src_path, dst_path))
 
         if 'hubbard_file' in self.inputs:
-            uuid = self.inputs.hubbard_file.filename
+            uuid = self.inputs.hubbard_file.uuid
             src_path = self.inputs.hubbard_file.filename
-            dst_path = self.input_file_name_hubbard_file
+            dst_path = self.filename_input_hubbard_parameters
             local_copy_list.append((uuid, src_path, dst_path))
 
         arguments = [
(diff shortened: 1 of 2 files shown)


mbercx (approved):
Thanks @sphuber! It looks fine to me, but I haven't been able to do a test run yet since I'm not very familiar with `hp.x` and `aiida-quantumespresso-hp`.

Perhaps we can spend a little time during our next meeting to discuss this and then I can provide the test for `aiida-quantumespresso-hp`?

sphuber:
> Perhaps we can spend a little time during our next meeting to discuss this and then I can provide the test for `aiida-quantumespresso-hp`?

Sure, good idea. I already added a test to the other plugin, but we can have a look. Thanks for the review

PR #597, opened 2020-10-24
Remove parser information from output_parameters

### LINKED ISSUE #582: Remove parser version from output_parameters

mbercx:
Fixes #582 

Now that aiida-core [adds both the AiiDA and plugin version to the
CalcJob node attributes](https://github.com/aiidateam/aiida-core/commit/5058e23c93f1f5f7fbcdffe82ab5dd7dd1f01829), as well as the parser name, it is no longer
necessary to add this information to the output_parameters dictionary.
This PR removes the get_parser_info() function that generates this
output.

The `parser_warnings` are also removed for the base/pw/ph parsers, since
they were not used here. We leave the `parser_warnings` for the neb/cp
parsers, but these should be replaced in a future PR by the use of a
logging container.

### DIFF
--- a/aiida_quantumespresso/parsers/__init__.py
+++ b/aiida_quantumespresso/parsers/__init__.py
@@ -5,24 +5,3 @@ from aiida.common import OutputParsingError
 class QEOutputParsingError(OutputParsingError):
     """Exception raised when there is a parsing error in the QE parser."""
     pass
-
-
-def get_parser_info(parser_info_template=None):
-    """Return a template dictionary with details about the parser such as the version.
-
-    :param parser_info_template: template string with single placeholder to be replaced by current version number
-    :returns: dictionary with parser name, version and empty list for warnings
-    """
-    import aiida_quantumespresso
-
-    parser_version = aiida_quantumespresso.__version__
-    parser_info = {}
-    parser_info['parser_warnings'] = []
-    parser_info['parser_version'] = parser_version
-
-    if parser_info_template is None:
-        parser_info['parser_info'] = f'aiida-quantumespresso parser v{parser_version}'
-    else:
-        parser_info['parser_info'] = parser_info_template.format(parser_version)
-
-    return parser_info
--- a/aiida_quantumespresso/parsers/parse_raw/base.py
+++ b/aiida_quantumespresso/parsers/parse_raw/base.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 """A basic parser for the common format of QE."""
-from aiida_quantumespresso.parsers import get_parser_info
 
 __all__ = ('parse_output_base', 'parse_output_error', 'convert_qe_time_to_sec', 'convert_qe2aiida_structure')
 
@@ -21,7 +20,7 @@ def parse_output_base(filecontent, codename=None, message_map=None):
         raise RuntimeError(f'invalid format `message_map`: should be dictionary with two keys {keys}')
 
     logs = get_logging_container()
-    parsed_data = get_parser_info(parser_info_template='aiida-quantumespresso parser simple v{}')
+    parsed_data = {}
 
     lines = filecontent if isinstance(filecontent, list) else filecontent.split('\n')
 
--- a/aiida_quantumespresso/parsers/parse_raw/cp.py
+++ b/aiida_quantumespresso/parsers/parse_raw/cp.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 from xml.dom.minidom import parseString
-from aiida_quantumespresso.parsers import QEOutputParsingError, get_parser_info
+from aiida_quantumespresso.parsers import QEOutputParsingError
 from aiida_quantumespresso.parsers.parse_xml.pw.legacy import (
     read_xml_card, parse_xml_child_integer, xml_card_header, parse_xml_child_bool, parse_xml_child_str,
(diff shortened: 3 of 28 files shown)


sphuber (commented):
Thanks @mbercx . Changes look good, I was just wondering if we really need this key `parser_info` at all. First thing I though was to change the content to the entry point name of the parser, which then also makes it machine readable. But then I thought that this is not necessary as this value is already stored as an attribute of the `CalcJobNode` itself, so we would merely be duplicating it in the `output_parameters` output node. I think we can just get rid of the parser info key entirely.

mbercx:
> I think we can just get rid of the parser info key entirely.

I was also considering this, and I agree! In this case the `get_parser_info` function only returns a dictionary with the `'parser_warnings'` key, which has an empty list value. Should we still keep the function so it's easier to add other parser info in the future, or remove it and simply initialize the `parser_info` as `{'parser_warnings': []}` wherever it is called?

sphuber:
To be honest, if you look at how the `parser_info` is used, you can simply just get rid of it entirely and simply declare the variable `parser_warnings = []` directly. As far as I can see, this is at the end of the parser simply added as an attribute. No need to keep it as a key inside dictionary `parser_info` if that is the only member.

mbercx:
I've noticed that the `parser_warnings` key isn't used for all calculation types. For example, for `pw.x`, I can't seem to find a single append to the `parser_warnings` list. Should I still keep the `parser_warnings` in the `output_parameters` in this case for consistency?

More generally, why do we have a list of `parser_warnings` in the `output_parameters`? Can't we just warn the user using `logs.warnings.append`? E.g. can't we change these lines:

https://github.com/aiidateam/aiida-quantumespresso/blob/703936541525165328f84b52fde846e726340f6f/aiida_quantumespresso/parsers/parse_raw/neb.py#L54-L55

by 

```Python
        if not finished_run:  # I try to parse it as much as possible
            logs.warning.append('Error while parsing the output file')
```
of course, after creating a logging container at the start of the method with:

```Python
logs = get_logging_container()
```

sphuber:
The reason is historical. The original implementation of the parsers used `parser_warnings` in the `output_parameters` as the and the only way of communicating problems detected in the parsing of the output. This was before I added the concept of exit codes to `aiida-core` as the preferred way of communicating problems that should be machine readable. I kept a system for adding human readable warnings, but directed them through the logging system, as that is where human-readable messages are supposed to go. I implemented this for the `PwParser`, and since, some other parsers adopted this as well. To simplify its usage, the `get_logging_container` utility was created. We haven't gone through the exercise of updating *all* parsers to use this new procedure. Would be great if you are willing to do that in this PR :+1:

mbercx:
> We haven't gone through the exercise of updating _all_ parsers to use this new procedure. Would be great if you are willing to do that in this PR.

I'm happy to work on cleaning up the parsers, but going through some of these warnings, I find a lot more than just `parser_warnings`. Some of them also seem to errors, i.e. in the `Parser.parse()` method they are picked up and call the `exit()` method of the parser with a certain exit code. For example:

https://github.com/aiidateam/aiida-quantumespresso/blob/6d4c70683b1b29487164589a2de922a2db4a388e/aiida_quantumespresso/parsers/parse_raw/neb.py#L45-L48

https://github.com/aiidateam/aiida-quantumespresso/blob/6d4c70683b1b29487164589a2de922a2db4a388e/aiida_quantumespresso/parsers/neb.py#L70-L71

There is also the `job_successful` variable that probably should also be removed. It seems that these parsers could use a good scrubbing that is beyond the scope of this PR, and would make it rather massive/messy? 

I'd rather split this work up for the different parsers, and familiarize myself with the output of these calculations when I work on this, so I understand better what is going wrong, i.e. understand the meaning of all the exit codes and output parameters.

For this PR, I'll just remove all the `parser_warnings` and replace them by proper usage of the logging container.

sphuber:
I actually agree that doing what I proposed is beyond the scope of this PR. I wouldn't even start to introduce the logging container in the other parsers. Let's here just simply get rid of the `get_parser_info` util function and directly declare the `parser_warnings = []` variable in the parsers where it applies. In another PR you can then actually change the behavior of the parsers to use the new logging concept

mbercx:
> I actually agree that doing what I proposed is beyond the scope of this PR. I wouldn't even start to introduce the logging container in the other parsers. Let's here just simply get rid of the `get_parser_info` util function and directly declare the `parser_warnings = []` variable in the parsers where it applies. In another PR you can then actually change the behavior of the parsers to use the new logging concept

Yeah, after working on it a bit, I realize doing this partially also will just create more of a mess. I'll do as you suggest and open some Issues for cleaning up/use logging for the various parsers.

sphuber:
I think you are done with the changes here, correct @mbercx ? If so I will review it

sphuber (approved):
Looks good to me, thanks @mbercx

PR #599, opened 2020-10-26
`ForceConstantsData`: add `filename` and `**kwargs` to constructor

### LINKED ISSUE #598: `ForceConstantsData`: add `filename` argument to constructor
This is now required by the `SinglefileData` subclass and will start to raise in `aiida-core==2.0.0`

sphuber:
Fixes #598 

The constructor signature of the `SinglefileData` class, which is the
base class of the `ForceConstantsData`, was updated sometime ago in
`aiida-core` to include the `filename` and `**kwargs` arguments. These
are now added, because in `aiida-core==2.0.0` having a different
constructor will start to raise.


mbercx (approved):
I haven't tested this, but it looks good to me.

PR #603, opened 2020-10-28
`PwBaseWorkChain`: deprecate the `pseudo_family` input

### LINKED ISSUE #601: Deprecate use of `pseudo_family` in `PwBaseWorkChain`

sphuber:
Fixes #601 

The use of a `pseudo_family`, which has to be passed in as a `Str` node
to represent the name of a `Group`, is not robust with respect to the
provenance. Since groups are mutable, rerunning a workchain that used
the `pseudo_family` to determine what pseudos to run, can lead to a
rerun using different pseudos. This input exists for historical reasons
because it made the usage of the workflows easier for users, since they
wouldn't have to build the mapping of pseudos themselves, but with the
advent of `aiida-pseudo` and the input generators, this is no longer an
issue and it is preferable to directly pass the pseudos into the
`pseudos` input namespace.

### DIFF
--- a/aiida_quantumespresso/workflows/pw/base.py
+++ b/aiida_quantumespresso/workflows/pw/base.py
@@ -15,6 +15,14 @@ from aiida_quantumespresso.utils.resources import cmdline_remove_npools, create_
 PwCalculation = CalculationFactory('quantumespresso.pw')
 
 
+def validate_pseudo_family(value, _):
+    """Validate the `pseudo_family` input."""
+    if value is not None:
+        import warnings
+        from aiida.common.warnings import AiidaDeprecationWarning
+        warnings.warn('`pseudo_family` is deprecated, use `pw.pseudos` instead.', AiidaDeprecationWarning)
+
+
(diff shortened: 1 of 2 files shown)


aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -159,6 +159,11 @@ def validate_pseudos(self):
         pseudos = self.ctx.inputs.get('pseudos', None)
         pseudo_family = self.inputs.get('pseudo_family', None)
 
+        if pseudo_family:
+            import warnings
+            from aiida.common.warnings import AiidaDeprecationWarning
```
mbercx, line 164: Similar to our discussion in #569, shouldn't we create a validator instead of raising a warning, which will just wind up in the daemon logs?

sphuber: very good point

mbercx (approved):
Thanks @sphuber! LGTM! 🚀

PR #604, opened 2020-10-28
`PwBaseWorkChain`: restart from scratch when convergence not reached

### LINKED ISSUE #602: `PwBaseWorkChain`: restart from scratch in the `ERROR_ELECTRONIC_CONVERGENCE_NOT_REACHED` handler
The docstring says it will and the same handler for the relax/vc-relax variant also does so. It should be consistent.

sphuber:
Fixes #602 

This was already being done by the handler for relax/vc-relax
calculations where electronic convergence was not reached, but for
normal scf/nscf calculations, the restart was done from the last
calculation. It is not yet clear which approach will guarantee a higher
convergency rate, as in some cases restarting from scratch is better,
whereas in others it is detrimental, but at least the approaches should
be the same. The handler for `ERROR_ELECTRONIC_CONVERGENCE_NOT_REACHED`
is therefore updated to also restart from scratch, as the docstring
already said it would do.


mbercx (approved):
Thanks @sphuber, LGTM! 🚀

PR #605, opened 2020-11-12
Add atomic_proj.xml to retrieve list of ProjwfcCalculation

qiaojunfeng:
`atomic_proj.xml` is also very useful for postprocessing.

### DIFF
--- a/aiida_quantumespresso/calculations/projwfc.py
+++ b/aiida_quantumespresso/calculations/projwfc.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 """`CalcJob` implementation for the projwfc.x code of Quantum ESPRESSO."""
+import os
 from aiida.orm import RemoteData, FolderData, Dict, XyData
 from aiida_quantumespresso.calculations.namelists import NamelistsCalculation
 
@@ -25,7 +26,14 @@ class ProjwfcCalculation(NamelistsCalculation):
         ('PROJWFC', 'plotboxes', False),
     ]
     _default_parser = 'quantumespresso.projwfc'
-    _internal_retrieve_list = [NamelistsCalculation._PREFIX + '.pdos*']
+    # yapf: disable
+    _internal_retrieve_list = [
+        NamelistsCalculation._PREFIX + '.pdos*',
+        os.path.join(
+            NamelistsCalculation._OUTPUT_SUBFOLDER,
+            f'{NamelistsCalculation._PREFIX}.save',
+            'atomic_proj.xml')
+    ]
 
     @classmethod
     def define(cls, spec):


sphuber:
Thanks @qiaojunfeng . Can you give me an estimate of how big these files are? How often would one want to use these for post processing? I am trying to determine whether this should be retrieved by default because if they are big and not used that often, maybe it is better to let people retrieve it whenever they need it. In the develop branch of `aiida-core` I merged a feature that allows you to retrieve specific files per calcjob through the `additional_retrieve_list` metadata option.

qiaojunfeng:
> Thanks @qiaojunfeng . Can you give me an estimate of how big these files are? How often would one want to use these for post processing? I am trying to determine whether this should be retrieved by default because if they are big and not used that often, maybe it is better to let people retrieve it whenever they need it. 

This XML contains the matrix of shape `num_kpt * num_bands * num_atomic_orbitals`, usually the size is similar to stdout (which also contains a matrix of this shape, but `matrix_in_stdout = matrix_in_xml ** 2`, so the XML has more information). Yes for a very large structure (hundreds of bands) the XML might reach ~100MB, so I am a bit hesitant to retrieve it by default now.

> In the develop branch of `aiida-core` I merged a feature that allows you to retrieve specific files per calcjob through the `additional_retrieve_list` metadata option.

Thanks, indeed `additional_retrieve_list` works as well, I can just use `additional_retrieve_list` in the wannier workflow. I add it to default retrieve list because there are cases when I want to download the XML file but it has been deleted from /scratch, so I have to redo the calculation. 
I guess we can ask @giovannipizzi for his opinion on this.

greschd:
Are we currently parsing the "matrix in stdout"? If so, we could consider putting the XML in `temporary_retrieve_list` and parsing _that_ instead, deprecating the output of "matrix in stdout" in favor of "matrix in XML" which contains more info?

qiaojunfeng:
> Are we currently parsing the "matrix in stdout"? If so, we could consider putting the XML in `temporary_retrieve_list` and parsing _that_ instead, deprecating the output of "matrix in stdout" in favor of "matrix in XML" which contains more info?

Yes currently it is parsing stdout
https://github.com/aiidateam/aiida-quantumespresso/blob/2c6d02f69cf273cb5f78f6f9f39839d6397793ae/aiida_quantumespresso/parsers/projwfc.py#L350-L356

The XML contains the coefficient `<g|psi>` where g is atomic orbital, psi is Bloch function, so it's a complex number, while the stdout is `|<g|psi>|**2`, a real number, so stdout is more chemically intuitive, e.g. `psi = 0.3 p_x + 0.3 p_y + 0.4 p_z`. So usually we still want the stdout_matrix, but when we want the true coefficient, we need the XML.

greschd:
Ah, so the matrix is a factor 2 larger due to the complex / real difference I guess? Yeah then it probably makes sense to keep the current behavior by default.

With the change in `aiida-core`, does that mean we don't need any changes here?

qiaojunfeng:
> Ah, so the matrix is a factor 2 larger due to the complex / real difference I guess? Yeah then it probably makes sense to keep the current behavior by default.

Yes considering the complex/real there will be another factor 2.

> 
> With the change in `aiida-core`, does that mean we don't need any changes here?

sphuber:
@qiaojunfeng I think your last message may be incomplete, it seems you didn't respond to the second quote. Do we still need this PR?

qiaojunfeng:
> @qiaojunfeng I think your last message may be incomplete, it seems you didn't respond to the second quote. Do we still need this PR?

Ah sorry, OK let's keep the current ProjwfcCalculation as it is, and I will use `additional_retrieve_list` to get the XML, so I close the PR, thanks!

giovannipizzi:
Maybe the simplest is indeed for now to retrieve it "by hand" and parse it in the workflow with the additional_retrieve_list, since the file is big.
The original idea is that this file contains valuable information, that will be lost in the parsed output that only contains the squares, but probably it's OK for now.

PR #609, opened 2020-11-17
Dependencies: update the `aiida-pseudo` requirements

sphuber:
The new required version of `aiida-pseudo` allows multiple cutoff sets
to be defined on a pseudopotential family at different levels of
stringency. It also changes the CLI option to pass a pseudopotential
family to `-F`.

The `UpfData` class was also updated to now require the `z_valence`
value to be parsed in addition to the element, so the test fixtures are
updated to include that in the generated pseudo nodes.

The tests are adapted and a new fixture `sssp` is added, which for the
whole session creates an instance of `SsspFamily`. Since this is a
costly operation, it should only be done once and all tests should
reuse it. This means that tests can no longer blindly use the fixture
`clear_database_before_tests`, which anyway was not useful in terms of
efficiency.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,6 +4,7 @@
 import collections
 import os
 import shutil
+import tempfile
 
 import pytest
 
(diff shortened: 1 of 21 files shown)


tests/conftest.py
```diff
@@ -48,12 +49,57 @@ def fixture_code(fixture_localhost):
...
+        for values in elements.values():
+
+            element = values['symbol']
+            upf = generate_upf_data(element)
+            filename = os.path.join(dirpath, f'{element}.upf')
+
+            with open(filename, 'w+b') as handle:
+                with upf.open(mode='rb') as source:
+                    handle.write(source.read())
+                    handle.flush()
```
mbercx, line 91: Just out of curiosity: why do we `flush` the file handle here?

mbercx (approved):
Thanks @sphuber! All goods good to me, just left a question with regard to file handle flushing.

tests/conftest.py
sphuber: just to make sure it is written to disk, but I think it is not actually necessary here, since by the time the file will be used, it is outside of this context manager, the closing of which should guarantee the flushing as well.

PR #613, opened 2020-11-18
`PwParser`: ignore stress threshold if `CELL.cell_dofree != 'all'`

### LINKED ISSUE #612: `PwParser`: ignore stress convergence threshold if `CELL.cell_dofree != 'all'`
If the `cell_dofree` is anything other than the default `all` where no degrees of freedom of the cell are constrained, the stress convergence threshold needs to be ignored.

sphuber:
Fixes #612 

If the `cell_dofree` is anything other than the default `all` where no
degrees of freedom of the cell are constrained, the stress convergence
threshold needs to be ignored. The reason is that often the stress and
therefore pressure are far from a minimum because the cell if articially
constrained.


giovannipizzi (commented):
Thanks, all good - I wouldn't call the variable fixed_volume, though - there are many more options for cell_dofree - e.g. there is `volume` where actually volume changes, but the angles are kept fixed (actually, in this case one could still check the threshold). Or `'x'` where
 only the x component of axis 1 (v1_x) is moved etc.

So maybe I would call it e.g. `contstrained_cell_minimisation` or something like this

sphuber:
@giovannipizzi I changed it to `constrained_cell`

mbercx (approved):
Thanks @sphuber, looks good to me!

PR #616, opened 2020-11-18
`PwRelaxWorkChain`: do not attach output structure for `RelaxType.NONE`

### LINKED ISSUE #615: `PwRelaxWorkChain`: make the `output_structure` optional for `RelaxType.NONE`
In this mode the workchain should essentially just compute the forces and stresses and not change the cell or atomic positions at all.

sphuber:
Fixes #615 

This relax type was mostly added for the common workflows project where
they want to leverage the common relax workflows to also simply provide
a single point calculation to just compute the total energy and forces
for a given structure. In this case, the structure won't change and so
attaching it as an output makes no sense.


mbercx (approved):
Looks good, thanks @sphuber!

PR #619, opened 2020-11-20
Re-enable formatting after `define` methods

greschd:
The `calculations/__init__.py` contained a `# yapf: disable` statement, but no `# yapf: enable`. My understanding is that single `# yapf: disable` are "allowed" only at  the end of a specific line that shouldn't be formatted, otherwise it will apply for the entire rest of the file (see [the examples in their README](https://github.com/google/yapf#why-does-yapf-destroy-my-awesome-formatting)).

Here we add a `# yapf: enable` statement after the things that shouldn't be formatted, and apply the resulting formatting changes.

**EDIT:** Turns out `calculations/__init__.py` isn't the only affected file, updated the PR to add matching `# yapf: enable` everywhere.


sphuber (approved):
Thanks @greschd . Love me some auto formatting

PR #622, opened 2020-11-23
`PwCalculation`: Add parsing of up/dw Fermi energy

### LINKED ISSUE #621: parser for pw.x cannot get two Fermi energies (for spin up&down)
Dear aiida-quantumespresso developers,

I'm doing spin-polarized simulation, and in the output file (aiida.out), there are two fermi energies:

`the spin up/dw Fermi energies are    11.1196    7.8565 ev`

But when I check the node.res data, I can only find the parameter `fermi_energy_units`, but no fermi energy.

I'm guessing the parser can only get 1 value for Fermi energy? I may be wrong.

Best,
ZDH

mbercx:
Fixes #621 

From:
https://lists.quantum-espresso.org/pipermail/users/2011-April/020089.html

"If you force the system to have a certain magnetization
(i.e. setting tot_magnetization) you are actually asking to occupy the
"up" spin orbital more then the "down" spin ones, i.e. you have more up
electrons then down electrons. This is achieved by filling the up orbitals
up to a higher energy than the down orbitals, by setting two distinct
Fermi energies."

When running with tot_magnetization set, pw.x outputs both the spin up
and down Fermi energy. Currently, the `PwCalculation` parser only looks
for the Fermi output in case of a single Fermi. Here we add parsing of
the spin up/dw Fermi energies to the pw.x parser.

#### Questions/TODO

- [x] Do we store the up/dw Fermi energies in a list as a value of the _same_ key (`'fermi_energy'`), or would it be better to distinguish it from the single Fermi energy case with a different key (e.g. `'up_dw_fermi_energies'`)?
- [ ] Are there other parsers that need to accommodate for this?
- [ ] Is the `fermi_energy` from the `output_parameters` used in other parts of the code that we need to update in this PR?


sphuber:
Honestly, I would be tempted to not change the type of the `fermi_energy` attribute, but instead create new ones `fermi_energy_up` and `fermi_energy_down`. In my experience, having a node or attribute change type based on some condition makes code that relies on it more complex and unwieldy. In addition, there is backwards compatibility that we probably would be breaking.

tests/parsers/fixtures/pw/tot_magnetization/data-file.xml
```diff
@@ -0,0 +1,843 @@
...
+        <info name="inv. 180 deg rotation - cart. axis [0,1,1]" class="not found">lattice_symmetry</info>
+        <rotation rank="2" dims="3 3" order="F">
+          1.000000000000000e0 0.000000000000000e0 0.000000000000000e0
+          0.000000000000000e0 0.000000000000000e0 -1.000000000000000e0
+          0.000000000000000e0 -1.000000000000000e0 0.000000000000000e0
+          </rotation>
+      </symmetry>
+      <symmetry>
+        <info name="180 deg rotation - cart. axis [1,-1,0]" class="not found">lattice_symmetry</info>
+        <rotation rank="2" dims="3 3" order="F">
```
sphuber, line 475: You should be able to remove all symmetries and also all but one of the KS eigen states. This will save a lot of lines

aiida_quantumespresso/parsers/parse_raw/pw.py
```diff
@@ -744,6 +744,15 @@ def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None)
...
 
+            elif 'the spin up/dw Fermi energies are' in line:
+                try:
+                    value = line.split('are')[1].split('ev')[0].split()
+                    value = [float(e) for e in value]
+                    trajectory_data.setdefault('fermi_energy', []).extend(value)
+                    parsed_data['fermi_energy' + units_suffix] = default_energy_units
+                except Exception:
+                    logs.warning.append('Error while parsing spin up/dw Fermi energies from the output file.')
+
```
sphuber, line 755: Can this info not be parsed from the XML? We should start to move away from parsing from stdout

tests/parsers/test_pw.py
```diff
@@ -379,6 +379,31 @@ def test_pw_npools_too_high(fixture_localhost, generate_calc_job_node, generate_
...
+    node = generate_calc_job_node(entry_point_calc_job, fixture_localhost, name, generate_inputs())
+    parser = generate_parser(entry_point_parser)
+    results, calcfunction = parser.parse_from_node(node, store_provenance=False)
+
+    assert calcfunction.is_finished, calcfunction.exception
+    assert calcfunction.is_finished_ok, calcfunction.exit_status
+    assert 'output_parameters' in results
+    data_regression.check({
+        'output_parameters': results['output_parameters'].get_dict(),
+    })
```
sphuber, line 404: Why not just check for the presence of the new keys? It _is_ possible that this change could break other code, but given that it adds a single elif clause, that is unlikely. Don't think it is necessary to check the entire output parameters node

sphuber (changes requested):
Thanks @mbercx . Have some suggestions

tests/parsers/fixtures/pw/tot_magnetization/data-file.xml
mbercx: Note that I also had to adjust the `nks` element.

aiida_quantumespresso/parsers/parse_raw/pw.py
mbercx: Yes, changed it! 👍

tests/parsers/test_pw.py
mbercx: Makes sense! It's been adjusted.

mbercx:
Thanks for the review @sphuber! I've moved the parsing to the XML file and trimmed the testing a bit. Let me know if you have further comments.

mbercx:
@hezhengda: as the one who originally reported this, do you still have any comments before this is reviewed and merged? In particular, do you agree with having two outputs in the `output_parameters`, with keys `fermi_energy_up` and `fermi_energy_down`?

sphuber (approved):
@mbercx this would be fine for. I think it is best that the original `fermi_energy` key is not changed, so I like that we simply add two new keys. I guess there may be code out there that may rely on the `fermi_energy` attribute being present, but then again even without this fix, that key wouldn't have been present in this calculation, so I think that problem would already have existed. Still would be good to check with the original feature requester since I have never needed this functionality and so am not sure if it satisfies the needs

mbercx:
> Still would be good to check with the original feature requester since I have never needed this functionality and so am not sure if it satisfies the needs

I pinged @hezhengda to see if he still had any comments over two weeks ago, but haven't heard anything. 🤷  If I don't hear anything by this evening I'll go ahead and merge.

hezhengda:
Sorry, I have been off for a while. Looks great! Thanks for the modification.

Best,
Zhengda
> > Still would be good to check with the original feature requester since I have never needed this functionality and so am not sure if it satisfies the needs
> 
> 
> 
> I pinged @hezhengda to see if he still had any comments over two weeks ago, but haven't heard anything. 🤷  If I don't hear anything by this evening I'll go ahead and merge.

mbercx:
> Sorry, I have been off for a while. Looks great! Thanks for the modification.

No worries! Thanks for the feedback. 🙃

PR #624, opened 2020-11-26
Fix projwfc parser for >=1000 projections

### LINKED ISSUE #623: Projwfc parser fails for >=1000 projections
When running a `ProjwfcCalculation`, I encountered the following error:

```
+-> REPORT at 2020-11-24 13:25:02.007190+00:00
 | [249875|ProjwfcCalculation|on_except]: Traceback (most recent call last):
 |   File "/home/a-dogres/.virtualenvs/aiida-py38/lib/python3.8/site-packages/plumpy/process_states.py", line 225, in execute
 |   File "/data/aiida/source/aiida-core/aiida/engine/processes/calcjobs/calcjob.py", line 288, in parse
 |     exit_code = execmanager.parse_results(self, retrieved_temporary_folder)
 |   File "/data/aiida/source/aiida-core/aiida/engine/daemon/execmanager.py", line 439, in parse_results
 |   File "/data/aiida/source/aiida-quantumespresso/aiida_quantumespresso/parsers/projwfc.py", line 342, in parse
 |     new_nodes_list = self._parse_bands_and_projections(out_info_dict)
 |   File "/data/aiida/source/aiida-quantumespresso/aiida_quantumespresso/parsers/projwfc.py", line 461, in _parse_bands_and_projections
 |     bands_data, projection_data = spin_dependent_subparser(out_info_dict)
 |   File "/data/aiida/source/aiida-quantumespresso/aiida_quantumespresso/parsers/projwfc.py", line 152, in spin_dependent_subparser
 |     bands[i % od['k_states']][j % od['num_bands']] = val
 | ValueError: could not convert string to float: 'eV'
``` 

This seems to be the code raising the error:

https://github.com/aiidateam/aiida-quantumespresso/blob/4fabecc1baff99227afcaad99d5b250930c67821/aiida_quantumespresso/parsers/projwfc.py#L144-L151

I believe the issue is that the calculation has more than 1000 projections, and the corresponding line in the output file changes from 

```
==== e( 999) =     5.83614 eV ==== 
```
to 
```
==== e(1000) =     5.84558 eV ==== 
```
which changes the index after `.split()`. Should be a relatively straightforward fix, for example we could split by `) =` or ` = `.

The only thing I'm a bit confused by is the comments in the code (added in #319): According to those, what I have here is "pre ~6.3" format, but this is run with QE 6.6.
@eimrek @sphuber do you remember / know the details of the format change? Maybe the format was changed back?

If there's a possibility the pre-6.3 format was different still, we could add the new variant as another try/except to be safe.

greschd:
Fixes #623.

Instead of splitting by whitespace alone, we now split at the equals sign with surrounding whitespaces.

In the tests, we use a bit of a hack to test for this; replacing a '1' with '0001' to force the old behavior to break without having to add >=1000 projections.


sphuber (approved):
Thanks @greschd

PR #626, opened 2020-12-02
added qes_030920.xsd from https://github.com/QEF/qeschemas

rikigigi:
…/PW_CPV/previous_schemas/qes_030920.xsd


sphuber:
Thanks @rikigigi . This version is produced with what version of QE? And did they change the date format on purpose? Now the filename uses `030920` which I take to mean September 3rd 2020? So far they have been using the inverse: `200420` i.e. April 20th 2020. Will the version string in produced output XMLs also follow a different date format now?

rikigigi:
This is for the upcoming 6.7 version, the order of the digit is an an accident. But it is not possible to change

pietrodelugas:
yes sorry @sphuber and @rikigigi  I used the wrong order, I didn't want to change it afterwards because I already changed the link. 
I thinks that using the new qes_200420.xsd is compatible  with qes_030920.xsd, so we can hide this away. 
It goes without saying that all new schemas in 2021 will be realeased on the 21st of the month ....

rikigigi:
note that diffing the schemas `diff qes_200420.xsd qes_191206.xsd` we get, at the bottom:
```
<       <element type="boolean" name="do_magnetization" minOccurs="0"/>
---
>       <element type="boolean" name="do_magnetization"/>
```
So I added a "get" in the parsing function

pietrodelugas:
do_magnetizatio is only non non collinear calculation.  It was  non initialized and meaningless in the other cases.

mbercx (approved):
Thanks @rikigigi! Sorry for keeping this PR open for so long. I did a little test run for QE v6.7 and I think this is good to go, unless there is a reason why @sphuber held off the merge?

I did notice that the `xml_format` element still  refers tot the previous version:

```xml
<qes:espresso xsi:schemaLocation="http://www.quantum-espresso.org/ns/qes/qes-1.0 http://www.quantum-espresso.org/ns/qes/qes_030920.xsd" Units="Hartree atomic units" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:qes="http://www.quantum-espresso.org/ns/qes/qes-1.0">
  <!--All quantities are in Hartree atomic units unless otherwise specified-->
  <general_info>
    <xml_format NAME="QEXSD" VERSION="20.04.20">QEXSD_20.04.20</xml_format>
    <creator NAME="PWSCF" VERSION="6.7MaX">XML file generated by PWSCF</creator>
    <created DATE="13Feb2021" TIME=" 0:16: 8">This run was terminated on:   0:16: 8  13 Feb 2021</created>
  </general_info>
```

Could this be a bug, or?

sphuber:
I kept it open because the version for which the schema is written wasn't released yet. Has 6.7 been officially released now? And does the schema that comes with it correspond exactly to what is added in this PR?

> I did notice that the xml_format element still refers tot the previous version

This is what I mentioned in my original response to this PR and Pietro has said that it was a mistake that would be fixed. Would be good to check if there isn't an updated version. Note that it shouldn't be a real problem as that tag is not used to match the schema to the output file, but rather the attribute in the top-level tag `xsi:schemaLocation="http://www.quantum-espresso.org/ns/qes/qes-1.0 http://www.quantum-espresso.org/ns/qes/qes_030920.xsd"` is parsed. The final filename in the last URL is what is used as the schema file in the repo which is `qes_030920.xsd` in this case. Note that as mentioned before as well, this also has an inconsistency as it uses a different date format compared to previous schemas.

mbercx:
> Has 6.7 been officially released now?

I didn't see an official release item on the news page of the Quantum ESPRESSO website, but there is a [release pack](https://github.com/QEF/q-e/releases/tag/qe-6.7.0) available. I got a request from someone to provide support for QE v6.7, so wanted to go ahead and add the schema. 

> This is what I mentioned in my original response to this PR

Ah, I only noticed the comment about the format, my bad.

mbercx:
@rikigigi and @pietrodelugas: just for clarity: is the current XSD schema in this PR part of an official release that we should support, or will the "fixed" schema replace this one?

rikigigi:
Yes @mbercx it is part of a relased QE https://gitlab.com/QEF/q-e/-/releases/qe-6.7MaX-Release

sphuber:
Thanks @rikigigi and @pietrodelugas . We will release this with v3.3.2

PR #630, opened 2020-12-08
Dependencies: update to `aiida-pseudo==0.4.0`

sphuber:
This new minor version comes with some significant changes:

 * All format specific pseudopotential family subclasses have been
   removed, for example, `UpfFamily` no longer exists. The format of
   the pseudos in a family are now stored as an extra and can be
   retrieved through the `pseudo_type` property. The command line
   option param type `PseudoFamilyType` was updated to reflect these
   changes.

 * The recommended cutoffs for families are now always stored in
   electronvolt. They therefore have to be converted explicitly to
   Rydberg before they are passed to the calculations.

### DIFF
--- a/aiida_quantumespresso/workflows/pw/base.py
+++ b/aiida_quantumespresso/workflows/pw/base.py
@@ -17,7 +17,6 @@ from aiida_quantumespresso.utils.resources import cmdline_remove_npools, create_
 from ..protocols.utils import ProtocolMixin
 
 PwCalculation = CalculationFactory('quantumespresso.pw')
-UpfFamily = GroupFactory('pseudo.family.upf')
 SsspFamily = GroupFactory('pseudo.family.sssp')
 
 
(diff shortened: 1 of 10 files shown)


sphuber:
@mbercx the new version of `aiida-pseudo` has not been released yet so that is why the tests are failing, but once it has been (should be relatively soon) I will rerun the tests.

sphuber:
@mbercx (`aiida-pseudo==0.4.0`](https://pypi.org/project/aiida-pseudo/) has now been released, so this is good to go.

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -168,18 +168,17 @@ def get_builder_from_protocol(
         nkinds = len(structure.kinds)
 
         try:
-            types = (UpfFamily, SsspFamily)
-            pseudo_family = orm.QueryBuilder().append(types, filters={'label': pseudo_family}).one()[0]
+            pseudo_family = orm.QueryBuilder().append(SsspFamily, filters={'label': pseudo_family}).one()[0]
         except exceptions.NotExistent as exception:
             raise ValueError(f'required pseudo family `{pseudo_family}` is not installed') from exception
```
mbercx, line 173: Maybe it would be good to give the new user a little hand here, and point them towards installing the default SSSP family with `aiida-pseudo`.

mbercx (approved):
Thanks @sphuber, looks good to me! 👌

Just added a minor comment re helping the user a bit when using `get_builder_from_protocol` without having the required SSSP family installed. Will already approve, so feel free to merge at your leisure!

aiida_quantumespresso/workflows/pw/base.py
sphuber: Done

PR #631, opened 2020-12-11
`PwRelaxWorkChain`: add tests for `relax_type` in generator overrides

sphuber:
The `get_builder_from_protocol` method was not yet tested with respect
to defining a `relax_type` in the `overrides` and make sure that the
correct value is set on the builder that is returned.

### DIFF
--- a/tests/workflows/protocols/pw/test_relax.py
+++ b/tests/workflows/protocols/pw/test_relax.py
@@ -4,7 +4,7 @@ import pytest
 
 from aiida.engine import ProcessBuilder
 
-from aiida_quantumespresso.common.types import ElectronicType, SpinType
+from aiida_quantumespresso.common.types import ElectronicType, RelaxType, SpinType
 from aiida_quantumespresso.workflows.pw.relax import PwRelaxWorkChain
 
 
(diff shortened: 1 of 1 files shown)


tests/workflows/protocols/pw/test_relax.py
```diff
@@ -63,3 +63,12 @@ def test_spin_type(fixture_code, generate_structure):
         parameters = namespace['pw']['parameters'].get_dict()
         assert parameters['SYSTEM']['nspin'] == 2
         assert parameters['SYSTEM']['starting_magnetization'] == {'Si': 0.1}
+
+
+@pytest.mark.parametrize('relax_type', RelaxType)
```
mbercx, line 68: Just so I understand: since `RelaxType` is an `Enum`, and hence an iterable, `pytest.mark.parametrize` will test each of the `RelaxType`'s defined in the class?

mbercx (approved):
All good, thanks @sphuber! Just left a little question to make sure I understand the use of an `Enum` with `@pytest.mark.parametrize`.

tests/workflows/protocols/pw/test_relax.py
sphuber: [Uhuh](https://media.giphy.com/media/l41m35Z4QtUJgVEdy/giphy.gif)

PR #632, opened 2020-12-11
`Protocol`: Switch to initial_magnetic_moments argument

mbercx:
Replace the `starting_magnetization` input argument of the
`get_builder_from_protocol` method of the `PwBaseWorkChain` by
`initial_magmom`.

Allowing the user to define the initial magnetic moments in Bohr
magnetons instead of the `starting_magnetization` has several advantages:

* It's easier to think of an initial guess for the magnetic moment than
the `starting_magnetization` based on e.g. the oxidition state of the
element.
* The magnetization output of pw.x also is expressed in Bohr magnetons.
* Since deriving the `starting_magnetization` from the magnetic moments
requires the `z_valence`, it requires knowledge about which
pseudopotentials are being used for the calculation. This makes it a
more difficult quantity to work with than the magnetic moments when
using the `get_builder_from_protocol` method in other projects.

### DIFF
--- a/aiida_quantumespresso/workflows/protocols/utils.py
+++ b/aiida_quantumespresso/workflows/protocols/utils.py
@@ -109,13 +109,26 @@ def get_magnetization_parameters() -> dict:
         return yaml.safe_load(handle)
 
 
-def get_starting_magnetization(structure, pseudo_family):
+def get_starting_magnetization(structure, pseudo_family, initial_magnetic_moments=None):
     """Return the dictionary with starting magnetization for each kind in the structure.
 
     :param structure: the structure.
     :param pseudo_family: pseudopotential family.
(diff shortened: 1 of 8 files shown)


sphuber (commented):
All good. I would just suggest to call it `initial_magnetic_moments` since we also use `magnetic_moments` in the `TrajectoryData` output nodes. It is more consistent and am not a fan of the `magmom` shorthand to be honest. What do you think?

mbercx:
Sure, I agree. I just felt it was a little longish for a variable name. 🙃 Let me update the code.

Another thing I was considering was adding the `initial_magnetic_moments` argument to the `get_starting_magnetization` function, and then just passing it to the call in the `PwBaseWorkChain.get_protocol_from_builder()` method, removing lines 193-200. Feels like that would be cleaner, what do you think?

sphuber (approved):
absolute blinder

PR #633, opened 2020-12-14
use verbosity instead of iverbosity in PH and EPW calculation

### LINKED ISSUE #606: `iverbosity` is not working in ph.x 6.6 anymore
running `ph.x` 6.6 raising the following error:
```
     Error in routine  read_namelists (1):
      bad line in namelist &inputph: "/" (error could be in the previous line)
``` 
change iverbosity to `verbosity = 'default'` fix this issue.

-----------
aiida-quantumespresso version is 3.2.1

unkcpz:
Fix #606 

The `iverbosity` is obsolete and found causing `ph.x` failed in qe-6.6.


sphuber:
Do you know when `iverbosity` was deprecated and removed, and especially in what version of QE `verbosity` was added? This is important because we need to know if this change would essentially be dropping support for certain QE versions. Also, I just checked that we also still use `iverbosity` in the `PwCalculation` and `CpCalculation` plugins, but I no longer find that option in the documentation of QE v6.3 or later.

unkcpz:
> Do you know when iverbosity was deprecated and removed, and especially in what version of QE verbosity was added? 

I did a quick search and can't find the details. I guess the answer will in the source code anyway, I will make more investigate.

atztogo:
I hit this issue at qe-6.7.
I found information on PH:
https://github.com/QEF/q-e/blob/626d07a1cff06525d08d931a12672b28717f9932/Doc/release-notes#L869

sphuber:
> did a quick search and can't find the details. I guess the answer will in the source code anyway, I will make more investigate.

I ran through the source code and it seems it was changed in commit `a7ef3557fb946e6c226aaa4c96b1fabed0e12346` of the QE source code, which was released with v5.2. I think that version is old enough that we can simply replace the value here and no longer support the old flag

sphuber (approved):
Thanks @unkcpz

atztogo:
Thanks @unkcpz and @sphuber.

PR #634, opened 2020-12-14
`PwRelaxWorkChain`: Move `RelaxType` input to `get_builder_from_protocol`

mbercx:
Instead of having the Relaxtype as an input of the PwRelaxWorkChain, it
would be better to properly separate church and state by making it an
input of the `get_builder_from_protocol` method. Based on the chosen
`Relaxtype`, this will properly populate the builder of the
`PwRelaxWorkChain`.

Second, currently for `RelaxType.NONE`, the `PwRelaxWorkChain` fixes
the atoms using the `if_pos` inputs explained explained here:

https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm1094

and then run a `'relax'` type calculation. This leads to issues with
all calculations raising `ERROR_IONIC_CONVERGENCE_NOT_REACHED` 
exit codes from the parser. 

Instead, here we set `parameters['CONTROL']['calculation'] = 'scf'` for
`RelaxType.NONE` in the `get_builder_from_protocol` method, since 
this "relaxation" type really corresponds to a static calculation.

Besides this, we also:

* Make sure the `PwRelaxWorkChain` only runs 1 `PwBaseWorkChain` for
  `RelaxType`'s `NONE`, `ATOMS` and `SHAPE`. I.e. meta convergence is
  turned off.
* Make sure we do not run a "final_scf" when `'calculation'` is set to `'scf'`
  for the `PwBaseWorkChain` that does the relaxation.
* Make sure we don't run into an error in the `results` step of the outline
  when running with `'calculation'` set to `'scf'`.

#### TODO

- [x] Ask @sphuber if we _really_ need to be able to disguise static calculations 
  as `PwRelaxWorkChain`s.
- [x] Give this work chain logic another hard look and see if it can't be simplified.
- [x] Add more tests for these `RelaxType`s and the work chain logic.
- [x] Test all of the `RelaxType`s and make sure we don't run into more errors.

### DIFF
--- a/aiida_quantumespresso/workflows/pw/relax.py
+++ b/aiida_quantumespresso/workflows/pw/relax.py
@@ -143,12 +143,36 @@ class PwRelaxWorkChain(ProtocolMixin, WorkChain):
         self.ctx.is_converged = False
         self.ctx.iteration = 0
 
+        # Determine the correct `RelaxType` and add it to the context
+        if 'relaxation_scheme' in self.inputs:
+            if self.inputs.relaxation_scheme.value == 'relax':
+                self.ctx.relax_type = RelaxType.ATOMS
+            elif self.inputs.relaxation_scheme.value == 'vc-relax':
+                self.ctx.relax_type = RelaxType.ATOMS_CELL
+            else:
+                raise ValueError('unsupported value for the `relaxation_scheme` input.')
+        else:
+            self.ctx.relax_type = RelaxType(self.inputs.relax_type)
+
+        # Set the meta_convergence and add it to the context
+        self.ctx.meta_convergence = self.inputs.meta_convergence.value
+        if self.inputs.meta_convergence and self.ctx.relax_type in [RelaxType.NONE, RelaxType.ATOMS, RelaxType.SHAPE]:
+            self.report(
+                f'Since there is no change in volume for `{self.ctx.relax_type}`, meta convergence is turned off.'
+            )
+            self.ctx.meta_convergence = False
+
+        # Add the final scf inputs to the context if a final scf should be run
         if self.inputs.final_scf and 'base_final_scf' in self.inputs:
             raise ValueError('cannot specify `final_scf=True` and `base_final_scf` at the same time.')
         elif self.inputs.final_scf:
             self.ctx.final_scf_inputs = AttributeDict(self.exposed_inputs(PwBaseWorkChain, namespace='base'))
         elif 'base_final_scf' in self.inputs:
             self.ctx.final_scf_inputs = AttributeDict(self.exposed_inputs(PwBaseWorkChain, namespace='base_final_scf'))
+
+        if self.ctx.relax_type is RelaxType.NONE:
+            self.report('Work chain will not run final SCF for `RelaxType.NONE`.')
+            self.ctx.pop('final_scf_inputs')
 
     def should_run_relax(self):
         """Return whether a relaxation workchain should be run.
@@ -164,10 +188,7 @@ class PwRelaxWorkChain(ProtocolMixin, WorkChain):
         If the maximum number of meta convergence iterations has been exceeded and convergence has not been reached, the
         structure cannot be considered to be relaxed and the final scf should not be run.
         """
-        return (
-            RelaxType(self.inputs.relax_type) is not RelaxType.NONE and self.ctx.is_converged and
-            'final_scf_inputs' in self.ctx
-        )
+        return self.ctx.is_converged and 'final_scf_inputs' in self.ctx
 
     def run_relax(self):
         """Run the `PwBaseWorkChain` to run a relax `PwCalculation`."""
@@ -181,35 +202,25 @@ class PwRelaxWorkChain(ProtocolMixin, WorkChain):
         inputs.pw.parameters.setdefault('CONTROL', {})
         inputs.pw.parameters['CONTROL']['restart_mode'] = 'from_scratch'
 
-        if 'relaxation_scheme' in self.inputs:
-            if self.inputs.relaxation_scheme.value == 'relax':
-                relax_type = RelaxType.ATOMS
-            elif self.inputs.relaxation_scheme.value == 'vc-relax':
-                relax_type = RelaxType.ATOMS_CELL
-            else:
-                raise ValueError('unsupported value for the `relaxation_scheme` input.')
-        else:
-            relax_type = RelaxType(self.inputs.relax_type)
-
-        if relax_type in [RelaxType.VOLUME, RelaxType.SHAPE, RelaxType.CELL]:
+        if self.ctx.relax_type in [RelaxType.VOLUME, RelaxType.SHAPE, RelaxType.CELL]:
             inputs.pw.settings = self._fix_atomic_positions(inputs.pw.structure, inputs.pw.get('settings', None))
 
-        if relax_type == RelaxType.NONE:
+        if self.ctx.relax_type is RelaxType.NONE:
             inputs.pw.parameters['CONTROL']['calculation'] = 'scf'
             inputs.pw.parameters.pop('CELL', None)
-        elif relax_type == RelaxType.ATOMS:
+        elif self.ctx.relax_type is RelaxType.ATOMS:
             inputs.pw.parameters['CONTROL']['calculation'] = 'relax'
             inputs.pw.parameters.pop('CELL', None)
         else:
             inputs.pw.parameters['CONTROL']['calculation'] = 'vc-relax'
 
-        if relax_type in [RelaxType.VOLUME, RelaxType.ATOMS_VOLUME]:
+        if self.ctx.relax_type in [RelaxType.VOLUME, RelaxType.ATOMS_VOLUME]:
             inputs.pw.parameters['CELL']['cell_dofree'] = 'volume'
 
-        if relax_type in [RelaxType.SHAPE, RelaxType.ATOMS_SHAPE]:
+        if self.ctx.relax_type in [RelaxType.SHAPE, RelaxType.ATOMS_SHAPE]:
             inputs.pw.parameters['CELL']['cell_dofree'] = 'shape'
 
-        if relax_type in [RelaxType.CELL, RelaxType.ATOMS_CELL]:
+        if self.ctx.relax_type in [RelaxType.CELL, RelaxType.ATOMS_CELL]:
             inputs.pw.parameters['CELL']['cell_dofree'] = 'all'
 
         # If one of the nested `PwBaseWorkChains` changed the number of bands, apply it here
@@ -248,7 +259,7 @@ class PwRelaxWorkChain(ProtocolMixin, WorkChain):
             structure = workchain.outputs.output_structure
         except exceptions.NotExistent:
             # If relax_type is RelaxType.NONE, this is expected, so we are done
-            if RelaxType(self.inputs.relax_type) is RelaxType.NONE:
+            if self.ctx.relax_type is RelaxType.NONE:
                 self.ctx.is_converged = True
                 return
 
@@ -267,11 +278,8 @@ class PwRelaxWorkChain(ProtocolMixin, WorkChain):
         if not prev_cell_volume:
             self.ctx.current_cell_volume = curr_cell_volume
 
-            # If meta convergence is switched off, or the relax type doesn't change the volume we are done
-            if (
-                not self.inputs.meta_convergence.value or
-                RelaxType(self.inputs.relax_type) in [RelaxType.ATOMS, RelaxType.SHAPE]
-            ):
+            # If meta convergence is switched off we are done
+            if not self.ctx.meta_convergence:
                 self.ctx.is_converged = True
             return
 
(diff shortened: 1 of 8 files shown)


aiida_quantumespresso/workflows/pw/relax.py
```diff
@@ -164,7 +164,10 @@ def should_run_final_scf(self):
         If the maximum number of meta convergence iterations has been exceeded and convergence has not been reached, the
         structure cannot be considered to be relaxed and the final scf should not be run.
         """
-        return self.ctx.is_converged and 'final_scf_inputs' in self.ctx
+        return (
+            RelaxType(self.inputs.relax_type) is not RelaxType.NONE and self.ctx.is_converged and
```
sphuber, line 168: Instead of converting `self.inputs.relax_type` into a `RelaxType` instance, we should just do it once in `setup` and assign it to the context.  While we are at it, we can move this code
```

        if 'relaxation_scheme' in self.inputs:
            if self.inputs.relaxation_scheme.value == 'relax':
                relax_type = RelaxType.ATOMS
            elif self.inputs.relaxation_scheme.value == 'vc-relax':
                relax_type = RelaxType.ATOMS_CELL
            else:
                raise ValueError('unsupported value for the `relaxation_scheme` input.')
        else:
            relax_type = RelaxType(self.inputs.relax_type)
```
from `run_relax` to setup and just assign `relax_type` to `self.ctx.relax_type`. Then everywhere else in the code you can just use `self.ctx.relax_type` and compare directly to `RelaxType`.

aiida_quantumespresso/workflows/pw/relax.py
```diff
@@ -256,8 +267,11 @@ def inspect_relax(self):
         if not prev_cell_volume:
             self.ctx.current_cell_volume = curr_cell_volume
 
-            # If meta convergence is switched off we are done
-            if not self.inputs.meta_convergence.value:
+            # If meta convergence is switched off, or the relax type doesn't change the volume we are done
+            if (
+                not self.inputs.meta_convergence.value or
+                RelaxType(self.inputs.relax_type) in [RelaxType.ATOMS, RelaxType.SHAPE]
+            ):
```
sphuber, line 274: Wonder if we should do the same here with `meta_convergence` as with `relax_type`. In `setup` we assign `self.ctx.meta_convergence` a boolean which is then used elsewhere in the logic. In principle it just assumes the value of the inputs but it is overridden based on the relax type. In this way we handle with this logic in one place and then can also have a report message in case the relax type overrides the meta_convergence input, so people are not super surprised when meta convergence is ignored

sphuber (changes requested):
Thanks @mbercx . Have a couple of suggestions

mbercx:
Alright, @sphuber, I made the changes and already took a _soft_ look at some of the logic. However, I'm left wondering why we treat the `RelaxType` different from the `SpinType` and `ElectronicType`. Most of the `RelaxType` logic can be moved into the `get_builder_from_protocol` method:

* Setting the correct `pw.parameters['CONTROL']['calculation']` and `pw.parameters['CELL']['cell_dofree']`.
* Setting the `meta_convergence` and `final_scf` inputs.
* The `RelaxType.NONE` logic can also be included by instead looking at the `pw.parameters['CONTROL']['calculation']`. If this is set to `'scf'`, then we won't _have_ an `output_structure` for the one and only `PwBaseWorkChain`, but we can deal with this in the logic of the work chain.

What do you think? Is there anything I've missed? 

I think right now it's a little confusing that the `RelaxType` has to be passed as a `Str` to the inputs of the `PwRelaxWorkChain`, but the `SpinType` and `ElectronicType` have to be passed as an argument to the `get_builder_from_protocol` method.

aiida_quantumespresso/workflows/pw/relax.py
```diff
@@ -143,13 +143,37 @@ def setup(self):
...
+
+        # Add the final scf inputs to the context if a final scf should be run
         if self.inputs.final_scf and 'base_final_scf' in self.inputs:
             raise ValueError('cannot specify `final_scf=True` and `base_final_scf` at the same time.')
         elif self.inputs.final_scf:
             self.ctx.final_scf_inputs = AttributeDict(self.exposed_inputs(PwBaseWorkChain, namespace='base'))
         elif 'base_final_scf' in self.inputs:
             self.ctx.final_scf_inputs = AttributeDict(self.exposed_inputs(PwBaseWorkChain, namespace='base_final_scf'))
 
+        if self.ctx.relax_type is RelaxType.NONE:
```
sphuber, line 173: This should also include `RelaxType.ATOMS` right? As long as the cell doesn't change, we good.

aiida_quantumespresso/workflows/pw/relax.py
```diff
@@ -143,13 +143,37 @@ def setup(self):
...
+            self.report('Work chain will not run final SCF for `RelaxType.NONE`.')
```
sphuber, line 174:
```suggestion
            self.report('final SCF will not be run because the cell will not change with relax type `{self.ctx.relax_type}`.')
```

aiida_quantumespresso/workflows/pw/relax.py
```diff
@@ -334,20 +342,16 @@ def results(self):
...
+        # Get the latest relax workchain and pass the outputs
+        final_relax_workchain = self.ctx.workchains[-1]
+
+        if self.ctx.relax_type is not RelaxType.NONE:
+            self.out('output_structure', final_relax_workchain.outputs.output_structure)
+
+        if 'final_scf_inputs' in self.ctx:
+            self.out_many(self.exposed_outputs(self.ctx.workchain_scf, PwBaseWorkChain))
+        else:
+            self.out_many(self.exposed_outputs(final_relax_workchain, PwBaseWorkChain))
```
sphuber, line 354: I think this logic _should_ in principle work, but would it not be safer to just use
```
try:
    self.out_many(self.exposed_outputs(self.ctx.workchain_scf, PwBaseWorkChain))
except AttributeError:
    self.out_many(self.exposed_outputs(final_relax_workchain, PwBaseWorkChain))
```

sphuber (changes requested):
Two minor things

mbercx:
Alright, @sphuber! As per our discussion, I've properly separated church and state by moving the `RelaxType` from an input of the `PwRelaxWorkChain` to an input argument of the `PwRelaxWorkChain.get_builder_from_protocol` method. I've done some testing for the different `RelaxType`s, and everything seems to be working as expected:

* Input parameters and settings look good, and are now tested in `workflows/protocols/pw/test_relax.py`.
* The work chain only uses meta convergence when the volume can change i.e. as long as `calculation` isn't `scf` or `relax`, and `cell_dofree` isn't `shape`.
* If `calculation` is set to `scf`, the work chain doesn't run the final scf.

Let me know if I should reorganise the commits somewhat. Should we perhaps split this in two PR's? i.e.:
* running with `calculation == 'scf'` for `RelaxType.NONE`
* Move the `RelaxType` to `get_builder_from_protocol`.

aiida_quantumespresso/workflows/pw/relax.py
```diff
@@ -170,41 +214,15 @@ def run_relax(self):
...
-            inputs.pw.parameters['CELL']['cell_dofree'] = 'volume'
-
-        if relax_type in [RelaxType.SHAPE, RelaxType.ATOMS_SHAPE]:
-            inputs.pw.parameters['CELL']['cell_dofree'] = 'shape'
-
-        if relax_type in [RelaxType.CELL, RelaxType.ATOMS_CELL]:
-            inputs.pw.parameters['CELL']['cell_dofree'] = 'all'
+        # If the relaxation uses 'vc-relax make sure there is a 'CELL' namelist
+        if inputs.pw.parameters['CONTROL']['calculation'] == 'vc-relax':
+            inputs.pw.parameters.setdefault('CELL', {})
```
sphuber, line 225: I think this will now fail because you removed the line `inputs.pw.parameters = inputs.pw.parameters.get_dict()` and so `inputs.pw.parameters` is still a stored `Dict` node and therefore immutable. I don't think we need this anymore do we? Is there any code after this that relies on the `CELL` key being there?

tests/workflows/protocols/pw/test_relax.py
```diff
@@ -65,10 +65,50 @@ def test_spin_type(fixture_code, generate_structure):
...
+    """Test ``PwRelaxWorkChain.get_builder_from_protocol`` with ``spin_type`` keyword."""
     code = fixture_code('quantumespresso.pw')
     structure = generate_structure()
-    builder = PwRelaxWorkChain.get_builder_from_protocol(code, structure, overrides={'relax_type': relax_type.value})
-    assert builder.relax_type.value == relax_type.value
+
+    builder = PwRelaxWorkChain.get_builder_from_protocol(code, structure, relax_type=RelaxType.NONE)
+    assert builder.base['pw']['parameters']['CONTROL']['calculation'] == 'scf'
+    with pytest.raises(KeyError):
+        builder.base['pw']['parameters']['CELL']  # pylint: disable=pointless-statement
```
sphuber, line 76: I think the following type of assertions may be more intuitive and cleaner written as
```suggestion
    assert 'CELL' not in builder.base['pw']['parameters']
```

aiida_quantumespresso/workflows/pw/relax.py
sphuber: Ah, I see now that the unpacking is done in `setup`. Maybe it would be cleaner to move this check there as well if it is still necessary

sphuber (changes requested):
Thanks @mbercx . Looking good, just two minor comments.

aiida_quantumespresso/workflows/pw/relax.py
mbercx: Should we then move _all_ lines that set the parameter inputs to `setup`? I.e. also the three lines above these and those defined in `run_final_scf`?

tests/workflows/protocols/pw/test_relax.py
mbercx: Haha, of course, that's a little less convoluted. 😅 I'll fix it!

aiida_quantumespresso/workflows/pw/relax.py
mbercx: Indeed, it's cleaner if all the input parameter specification is moved to `setup`, I think.

mbercx:
Uh oh! Copy Pasta! 😅

tests/workflows/protocols/pw/test_relax.py
mbercx: Actually, seems I made a mistake in the tests (e.g. for `ATOMS_VOLUME` the parameters should have a `CELL` input), but the code still raised a `KeyError` because `builder.base['pw']['parameters']` is a `Dict`.

aiida_quantumespresso/workflows/pw/relax.py
```diff
@@ -143,13 +157,56 @@ def setup(self):
...
+            if self.inputs.relaxation_scheme.value in ('relax', 'vc-relax'):
+                self.ctx.relax_inputs.pw.parameters['CONTROL']['calculation'] = self.inputs.relaxation_scheme.value
+            else:
+                raise ValueError('unsupported value for the `relaxation_scheme` input.')
+
+        # Set the meta_convergence and add it to the context
+        self.ctx.meta_convergence = self.inputs.meta_convergence.value
+        volume_cannot_change = (
+            self.ctx.relax_inputs.pw.parameters['CONTROL']['calculation'] in ('scf', 'relax') or
+            self.ctx.relax_inputs.pw.parameters['CELL']['cell_dofree'] == 'shape'
```
sphuber, line 177: I think this has still the capacity to fail now, no? I don't think it is guaranteed that `['CONTROL']['calculation']` or `['CELL']['cell_dofree']` exists. For the former, we should probably add a validator in the spec because the user now *has* to indicate the calculation mode (before we did it through the `relax_type` input). The second one is sometimes not defined and that is ok, so we should do something like `self.ctx.relax_inputs.pw.parameters.get('CELL', {}).get('cell_dofree', None) == 'shape'`

mbercx: Right, of course. 🤦 

One issue with the validator is that the user might still use the `relaxation_scheme` input, and I don't think I can check if this input is provided in the validator of the `base` input? So the user might not provide the `'calculation'`, in the `base` input, but the `relaxation_scheme` instead.

sphuber: you can put the validator on the top-level namespace for now with a comment that once `relaxation_scheme` is removed, it can be moved to the `parameters` port. In the top-level you will have access to all inputs

mbercx: I've added the validator which checks if the `base` input has `['pw']['parameters']['CONTROL']['calculation']` specified. It might be a little confusing to a user that still relies on `relaxation_scheme` and doesn't specify the `calculation` input, but I maybe this is ok? Or is there an elegant solution to this?

sphuber: yeah, as my previous comment said, you can move the validator to the top level inputs which has access to all
```
def validate_inputs(inputs, ctx):
    """Validator for top level namespace."""
        parameters = inputs['base']['pw']['parameters'].get_dict()

        if 'relaxation_scheme' not in inputs and 'calculation' not in parameters.get('CONTROL', {}):
            return 'The parameters in `base.pw.parameters` do not specify the required key `CONTROL.calculation`.'

def define(cls, spec):
    spec.input()
    ....
    spec.inputs.validator = validate_inputs
```

Note also that you should not raise an exception from a validator but simply return a string with the error message

mbercx: Ah, it seems Github hadn't updated the comments when I replied to yours, sorry about that!

sphuber (approved):

PR #638, opened 2021-01-07
Protocols: Allow for PseudoDojoFamily pseudos

mbercx:
Currently the `get_builder_from_protocol()` method only queries for
pseudopotential families of type `SsspFamily`. Here we also include
the `PseudoDojoFamily` class in the query, so these pseudo families can
also be used for the protocols.


mbercx:
@sphuber currently the `PseudoDojoFamily` isn't used in any of our defined protocols, so these changes aren't that useful unless the user changes the protocol in the `.yaml` files (which I'm doing for the common workflows). However, in the future it might be nice if the user can provide a protocol via a `.yaml` file instead of only the ones we have hard-coded. What do you think?

mbercx:
Another thing to note here is that our current code for the protocol does not allow us to use different cutoff stringencies, which is more of an issue for the `PseudoDojoFamily`, as it ships with 3 different stringencies.

sphuber (approved):

PR #642, opened 2021-01-26
Protocols: Fix ionic convergence thresholds for fast protocol

mbercx:
The ionic convergence thresholds for the `fast` protocol were more
strict than the `moderate` and `precise` protocols. Here we loosen the
ionic convergence thresholds for the `fast` protocols to something more
suitable for test calculations.

### DIFF
--- a/aiida_quantumespresso/workflows/protocols/pw/base.yaml
+++ b/aiida_quantumespresso/workflows/protocols/pw/base.yaml
@@ -47,8 +47,8 @@ protocols:
         kpoints_distance: 0.50
         meta_parameters:
             conv_thr_per_atom: 0.4e-9
-            etot_conv_thr_per_atom: 1.e-6
+            etot_conv_thr_per_atom: 1.e-4
         pw:
             parameters:
                 CONTROL:
-                    forc_conv_thr: 1.e-5
+                    forc_conv_thr: 1.e-3


mbercx:
@sphuber I've loosened the ionic convergence thresholds to one order of magnitude those of the `moderate` protocol. Since the `fast` protocol is only designed for test calculations, this should be ok?

greschd:
This may or may not be relevant here, but in my experience for relax type calculations less strict thresholds can sometimes _increase_ the total runtime because it becomes much harder to converge.

mbercx:
> This may or may not be relevant here, but in my experience for relax type calculations less strict thresholds can sometimes _increase_ the total runtime because it becomes much harder to converge.

Really? That seems somewhat counter-intuitive. Do you mean less strict thresholds on the _electronic_ convergence leading to issues with the _ionic_ convergence?

greschd:
> Do you mean less strict thresholds on the _electronic_ convergence leading to issues with the _ionic_ convergence?

Yes, sorry if that was unclear.

mbercx:
> Yes. Sorry, that was a bit unclear.

No worries, I just wasn't sure because in this PR I only loosen the thresholds on the ionic convergence. The `fast` protocol does have looser convergence conditions on the _electronic_ convergence as well compared to `moderate`, but I think they are still fairly strict. The default `conv_thr` in QE is 1e-6, so with the meta-parameter `conv_thr_per_atom` set to 0.4e-9, we'd need quite a few atoms before having a higher convergence threshold than the default. 😅

greschd:
Aah, I always get `conv_thr` and `etot_conv_thr` mixed up :sweat_smile: Yeah I agree this change doesn't pose a problem then.

sphuber (commented):
Uhm, I think the `precise` protocol is also suffering from my idiocy. It has _looser_ thresholds than the `moderate` protocol. These should be stricter, right?

mbercx:
> Uhm, I think the `precise` protocol is also suffering from my idiocy. It has _looser_ thresholds than the `moderate` protocol. These should be stricter, right?

Eh, the `conv_thr` values of `precise` are always half that of `moderate`, no? And, smaller values are more strict, or am I just getting confused here? 😅

sphuber:
> > Uhm, I think the `precise` protocol is also suffering from my idiocy. It has _looser_ thresholds than the `moderate` protocol. These should be stricter, right?
> 
> Eh, the `conv_thr` values of `precise` are always half that of `moderate`, no? And, smaller values are more strict, or am I just getting confused here? sweat_smile

Join us next time for Episode V: "The idiocy strikes back!"

sphuber (approved):

PR #650, opened 2021-02-18
Protocols: Fix mutation of overrides

### LINKED ISSUE #649: Protocols: overrides are mutated in `get_builder_from_protocol`
When passing overrides to the `get_builder_from_protocol` method of e.g. the `PwRelaxWorkChain`, they are mutated:

```Python
In [1]: PwRelaxWorkChain = WorkflowFactory('quantumespresso.pw.relax')

In [2]: code = load_code(53)
   ...: structure = load_node(54)

In [3]: overrides = {
   ...:     'base': {
   ...:         'pseudo_family': 'PseudoDojo/0.4/PBE/SR/standard/upf'
   ...:     },
   ...: }

In [4]: builder = PwRelaxWorkChain.get_builder_from_protocol(
   ...:     code=code,
   ...:     structure=structure,
   ...:     overrides=overrides
   ...: )

In [5]: overrides
Out[5]: 
{'base': {'pw': {'parameters': {'CONTROL': {'calculation': 'vc-relax',
     'forc_conv_thr': 0.0001,
     'tprnfor': True,
     'tstress': True,
     'etot_conv_thr': 2e-05},
    'SYSTEM': {'nosym': False,
     'occupations': 'smearing',
     'smearing': 'cold',
     'degauss': 0.01,
     'ecutwfc': 36.0,
     'ecutrho': 288.0},
    'ELECTRONS': {'electron_maxstep': 80,
     'mixing_beta': 0.4,
     'conv_thr': 4e-10},
    'CELL': {'press_conv_thr': 0.5, 'cell_dofree': 'all'}}},
  'pseudo_family': 'PseudoDojo/0.4/PBE/SR/standard/upf'}}
```

This is not the intended behaviour, and might be dangerous in some scripts.

mbercx:
Fixes #649 

Currently the `overrides` input dictionary provided to the
`get_builder_from_protocol` method can be mutated to also include the
defaults from the protocol. Here we make a copy of the `right`
dictionary in the `recursive_merge` function, to prevent the key/value
pairs of the `left` dictionary to be added to the `right` one. This
prevents the mutation of the `overrides` argument.

### DIFF
--- a/aiida_quantumespresso/workflows/protocols/utils.py
+++ b/aiida_quantumespresso/workflows/protocols/utils.py
@@ -70,6 +70,7 @@ def recursive_merge(left, right):
     """
     import collections
 
+    # Note that a deepcopy is not necessary, since this function is called recusively.
     right = right.copy()
 
     for key, value in left.items():
--- a/aiida_quantumespresso/workflows/protocols/utils.py
+++ b/aiida_quantumespresso/workflows/protocols/utils.py
@@ -70,6 +70,8 @@ def recursive_merge(left, right):
     """
     import collections
 
+    right = right.copy()
+
     for key, value in left.items():
         if key in right:
             if isinstance(value, collections.abc.Mapping) and isinstance(right[key], collections.abc.Mapping):
--- a/tests/workflows/protocols/test_utils.py
+++ b/tests/workflows/protocols/test_utils.py
@@ -1,0 +1,51 @@
+# -*- coding: utf-8 -*-
+"""Tests for the utility functions for the protocols."""
+
+def test_recursive_merge():
+    """Test the recursive merge function."""
+    from aiida_quantumespresso.workflows.protocols.utils import recursive_merge
+
+    left_dict = {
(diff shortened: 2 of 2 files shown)


aiida_quantumespresso/workflows/protocols/utils.py
```diff
@@ -70,6 +70,8 @@ def recursive_merge(left, right):
     """
     import collections
 
+    right = right.copy()
```
sphuber, line 73: I was going to say that we needed a `deepcopy` here to also copy nested dictionaries, but since this function is called recursively for nested dictionaries, copy should work. Maybe we can add a comment of this here.

sphuber (changes requested):
Thanks @mbercx . Change looks good, just one suggestion and maybe add a test for this, including a nested dictionary to make sure the shallow copying is sufficient because of the recursive nature

sphuber (approved):
Meesterlijk :+1:

PR #652, opened 2021-02-18
Protocols: Add usage of metadata/paral. overrides

### LINKED ISSUE #651: Protocols: `metadata` and `parallelization` overrides not used
Currently, the `metadata` and `parallelization` settings in the `overrides` supplied to the `get_builder_from_protocol()` method are not used (properly). In case of the `metadata`, these are ignored because of this line:

https://github.com/aiidateam/aiida-quantumespresso/blob/ae410f63279d3eb2695a8a33c12ed939a93b176e/aiida_quantumespresso/workflows/pw/base.py#L202

When the user provides the `metadata` correctly, in the `pw` namespace, they are not used. The defaults are still loaded because the `.yaml` file is incorrect.

The `parallelization` overrides are not used at all in the `get_builder_from_protocol()` method.

mbercx:
Fixes #651 

The `metadata` and `parallelization` inputs of the `PwCalculation`,
stored in the `pw` namespace of the `PwBaseWorkChain`, are not
(properly) used in the `get_builder_from_protocol()` method. Here we
make the following changes to correct this issue:

* Move the `metadata` inputs under `pw` in the base.yaml file.
* Correctly obtain the `metadata` from the inputs in the
`get_builder_from_protocol()` method of the `PwBaseWorkChain`.
* Check if there is `parallelization` input in the `pw` namespace of the
inputs, and if so add it to the builder in the
`get_builder_from_protocol()` method of the `PwBaseWorkChain`.

### DIFF
--- a/aiida_quantumespresso/workflows/pw/base.py
+++ b/aiida_quantumespresso/workflows/pw/base.py
@@ -204,7 +204,9 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         builder.pw['pseudos'] = pseudo_family.get_pseudos(structure=structure)  # pylint: disable=no-member
         builder.pw['structure'] = structure  # pylint: disable=no-member
         builder.pw['parameters'] = orm.Dict(dict=parameters)  # pylint: disable=no-member
-        builder.pw['metadata'] = inputs['metadata']  # pylint: disable=no-member
+        builder.pw['metadata'] = inputs['pw']['metadata']  # pylint: disable=no-member
+        if 'parallelization' in inputs['pw']:
+            builder.pw['parallelization'] = orm.Dict(dict=inputs['pw']['parallelization'])
         builder.clean_workdir = orm.Bool(inputs['clean_workdir'])
         builder.kpoints_distance = orm.Float(inputs['kpoints_distance'])
         builder.kpoints_force_parity = orm.Bool(inputs['kpoints_force_parity'])
--- a/aiida_quantumespresso/workflows/pw/base.py
+++ b/aiida_quantumespresso/workflows/pw/base.py
@@ -200,16 +200,18 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
             parameters['SYSTEM']['nspin'] = 2
             parameters['SYSTEM']['starting_magnetization'] = starting_magnetization
 
-        builder.pw['code'] = code  # pylint: disable=no-member
-        builder.pw['pseudos'] = pseudo_family.get_pseudos(structure=structure)  # pylint: disable=no-member
-        builder.pw['structure'] = structure  # pylint: disable=no-member
-        builder.pw['parameters'] = orm.Dict(dict=parameters)  # pylint: disable=no-member
-        builder.pw['metadata'] = inputs['pw']['metadata']  # pylint: disable=no-member
+        # pylint: disable=no-member
+        builder.pw['code'] = code
+        builder.pw['pseudos'] = pseudo_family.get_pseudos(structure=structure)
+        builder.pw['structure'] = structure
+        builder.pw['parameters'] = orm.Dict(dict=parameters)
+        builder.pw['metadata'] = inputs['pw']['metadata']
         if 'parallelization' in inputs['pw']:
             builder.pw['parallelization'] = orm.Dict(dict=inputs['pw']['parallelization'])
         builder.clean_workdir = orm.Bool(inputs['clean_workdir'])
         builder.kpoints_distance = orm.Float(inputs['kpoints_distance'])
         builder.kpoints_force_parity = orm.Bool(inputs['kpoints_force_parity'])
+        # pylint: enable=no-member
 
         return builder
 
--- a/tests/workflows/protocols/pw/test_base.py
+++ b/tests/workflows/protocols/pw/test_base.py
@@ -93,3 +93,37 @@ def test_initial_magnetic_moments(fixture_code, generate_structure):
 
     assert parameters['SYSTEM']['nspin'] == 2
     assert parameters['SYSTEM']['starting_magnetization'] == {'Si': 0.25}
+
+
+def test_metadata_overrides(fixture_code, generate_structure):
+    """Test that pw metadata is correctly passed through overrides."""
+    code = fixture_code('quantumespresso.pw')
+    structure = generate_structure()
+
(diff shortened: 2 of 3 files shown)


aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -195,14 +195,18 @@ def get_builder_from_protocol(
...
-        builder.pw['parameters'] = orm.Dict(dict=parameters)  # pylint: disable=no-member
-        builder.pw['metadata'] = inputs['metadata']  # pylint: disable=no-member
+        # pylint: disable=no-member
+        builder.pw['code'] = code
+        builder.pw['pseudos'] = pseudo_family.get_pseudos(structure=structure)
+        builder.pw['structure'] = structure
+        builder.pw['parameters'] = orm.Dict(dict=parameters)
+        builder.pw['metadata'] = inputs['pw']['metadata']
+        if 'parallelization' in inputs['pw']:
+            builder.pw['parallelization'] = orm.Dict(dict=inputs['pw']['parallelization'])
```
sphuber, line 205: can't the parallelization already be an instance of `Dict`?

mbercx: Usually I add the overrides as one big regular ol' dictionary, and as far as I know the loaded defaults are also just in a dictionary format? 

Second, I was actually also sort of hoping that the `Dict` constructor would be idempotent, but alas:

```Python
In [3]: d = Dict(dict={'a': 1})

In [4]: d2 = Dict(dict=d)
AttributeError                            Traceback (most recent call last)
<ipython-input-4-b2188aaaaf40> in <module>
----> 1 d2 = Dict(dict=d)

~/envs/aiida-dev/code/aiida-core/aiida/orm/nodes/data/dict.py in __init__(self, **kwargs)
     60         super().__init__(**kwargs)
     61         if dictionary:
---> 62             self.set_dict(dictionary)
     63 
     64     def __getitem__(self, key):

~/envs/aiida-dev/code/aiida-core/aiida/orm/nodes/data/dict.py in set_dict(self, dictionary)
     81             # Clear existing attributes and set the new dictionary
     82             self.clear_attributes()
---> 83             self.update_dict(dictionary)
     84         except exceptions.ModificationNotAllowed:  # pylint: disable=try-except-raise
     85             # I reraise here to avoid to go in the generic 'except' below that would raise the same exception again

~/envs/aiida-dev/code/aiida-core/aiida/orm/nodes/data/dict.py in update_dict(self, dictionary)
     98         :param dictionary: a dictionary with the keys to substitute
     99         """
--> 100         for key, value in dictionary.items():
    101             self.set_attribute(key, value)
    102 

AttributeError: 'Dict' object has no attribute 'items'
```

I think we should be working on making the `Dict` more like `dict`, so this is no longer an issue.

sphuber: That might not be as straighforward as you imagine as far as the expected behavior is concerned. For example, consider the behavior of `dict`:
```
In [1]: d = dict(a=1)

In [2]: id(d)
Out[2]: 140302121245760

In [3]: e = dict(d)

In [4]: id(e)
Out[4]: 140302119318464
```
if you apply this behavior to `Dict` and then use it in the way you have, you would be creating a new `Dict` node. I would argue that this is not what you want here. If it is already a `Dict` node, you want to use exactly that node that was an input and not create a new one. You create a new one only when you really have to

mbercx: Fair point. 👍  I was thinking the initial `Dict` node might not have been stored in the database, but actually, it could very well be; and this might be exactly the reason to support providing `Dict` nodes as an input in the `overrides`.

The same point could be made for other inputs, however. Consider for example the `clean_workdir` input for the `PwBaseWorkChain`:

```Python
In [1]: PwBaseWorkChain = WorkflowFactory('quantumespresso.pw.base')

In [2]: code = load_code(53)
   ...: structure = load_node(54)
   ...: 
   ...: clean_workdir = Bool(False)
   ...: clean_workdir.store()
Out[2]: <Bool: uuid: 26c7022a-6e63-49ab-9421-52bf016da014 (pk: 2384) value: False>

In [3]: overrides = {
   ...:     'clean_workdir': clean_workdir
   ...: }

In [4]: builder = PwBaseWorkChain.get_builder_from_protocol(
   ...:     code=code,
   ...:     structure=structure,
   ...:     overrides=overrides
   ...: )

In [5]: builder.clean_workdir
Out[5]: <Bool: uuid: f07c9464-37d6-459c-ba67-d560ce58b4cc (unstored) value: False>
```

So we'd have to update the logic in the `get_builder_from_protocol()` method everywhere, or simply only allow `isinstance(overrides, dict)`. Considering the fact that there might be a use case for re-using nodes, I'd suggest the former, but I will do this in a separate PR with a proper explanation.

However, I still think we should promote `Dict` to a proper `BaseType`, and make sure the behaviour is as close to `dict` as possible. But I suggest we move that discussion to an issue I had opened that discusses similar changes to `List` and `Dict`:

https://github.com/aiidateam/aiida-core/issues/4495

I'll add a note re this discussion there.

mbercx:
@sphuber although I think the comment in your review is valid, I will open a separate PR to deal with the possibility of `override`s being e.g. `Dict` nodes for all inputs constructed in the `get_builder_from_protocol` methods of all work chains. If you have no further comments, perhaps this PR can be merged so it can be added to the next patch release?

sphuber (commented):
Ok to merge as is @mbercx . Would it be possible to add a quick test that the metadata and parallelization are now respected by the `get_builder_from _protocol`?

mbercx:
2 addition tests, coming right up!

tests/workflows/protocols/pw/test_base.py
```diff
@@ -93,3 +93,37 @@ def test_initial_magnetic_moments(fixture_code, generate_structure):
...
     assert parameters['SYSTEM']['nspin'] == 2
     assert parameters['SYSTEM']['starting_magnetization'] == {'Si': 0.25}
+
+
+def test_metadata_overrides(fixture_code, generate_structure):
+    """Test that pw metadata is correctly passed through overrides."""
+    code = fixture_code('quantumespresso.pw')
+    structure = generate_structure()
+
+    overrides = {'pw': {'metadata': {'options': {'resources': {'num_machines': 2}, 'max_wallclock_seconds': 3600}}}}
```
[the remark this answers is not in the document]
mbercx: ludicrous numbers, coming right up!

sphuber (approved):

Issue #655, opened 2021-03-03
Increase `quantumespresso.pp` versatility (two step process)


ramirezfranciscof:
I would like to be able to use PP to generate the information for the charge density of individual kpoint-band pairs (using `plotnum=7`, [see description](https://www.quantum-espresso.org/Doc/INPUT_PP.html#idm24)) and then have PP add up a combination of those into a single charge density output (using `filepp(i)` and `weight(i)`, [see description](https://www.quantum-espresso.org/Doc/INPUT_PP.html#idm103)).

If I try to specify both settings of `&INPUTPP` and `&PLOT` at the same time, QE complains:
```
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
     task #         0
     from chdens  : error #         1
     can't mix nfile > 1 with multiple output files
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
```

It is possible to do this in succession (first run the `&INPUTPP` with `&PLOT` empty and then run the `&PLOT` with `&INPUTPP` empty) but if I try to do this with `quantumespresso.pp` I run into a couple of limitations:

1. `quantumespresso.pp` complains if I don't set up the `INPUTPP.plot_num` (I need this blank for the second step of only doing `&PLOT`).
2. Every step needs to be a different execution of PP and so `quantumespresso.pp` will create a separate folder for it. I can set a `parent_folder` and it will copy the content of the `out` folder, but the outputs of the `&INPUTPP` step are outside and won't be copied in this way.
3. `quantumespresso.pp` also currently requires `PLOT.iflag`, and so it will `&PLOT`-process, retrieve, and parse all the outputs of the first `&INPUTPP` step, which is what I'm trying to avoid.

Am I missing any current way to do this? If not, could we flexibilize a little the previous 3 limitations? I'm sure there are other use cases where one needs to run these two steps separatedly (I understand PP is meant to support this two step approach, as it says in the [description](https://www.quantum-espresso.org/Doc/INPUT_PP.html#idm3)). I could try to do the modifications myself and PR the changes if I get some guidelines on how to best do this.

ramirezfranciscof:
[...]

PR #656, opened 2021-03-03
Proof of concept

ramirezfranciscof:
This is just a proof of concept to illustrate changes that would be necessary to implement the request #655. It seems to at least work in my use case.


aiida_quantumespresso/calculations/pp.py
```diff
@@ -13,28 +13,30 @@
 def validate_parameters(value, ctx=None):  # pylint: disable=unused-argument
     """Validate 'parameters' dict."""
     parameters = value.get_dict()
+    missing_namelists = 0
```
ramirezfranciscof, line 16: For step1 only you don't need `parameters['PLOT']['iflag']`, and for step2 only you don't need `parameters['INPUTPP']['plot_num']`, but at least one of those two must be present. I keep count of how many are missing with this variable.

aiida_quantumespresso/calculations/pp.py
```diff
@@ -137,6 +139,9 @@ def prepare_for_submission(self, folder):  # pylint: disable=too-many-branches,t
         else:
             settings = {}
 
+        skip_plot = 'PLOT' not in parameters
+        copy_files = 'INPUTPP' not in parameters
```
ramirezfranciscof, line 143: This is the criteria to decide what is happening: which namelist is missing from the parameters provided. I used these names since they are more related to the implications, but you can assume these logicals mean `step1_only` and `step2_only` respectively.

aiida_quantumespresso/calculations/pp.py
```diff
@@ -197,6 +209,17 @@ def prepare_for_submission(self, folder):  # pylint: disable=too-many-branches,t
                 parent_calc_folder.computer.uuid,
                 os.path.join(parent_calc_folder.get_remote_path(), self._PSEUDO_SUBFOLDER), self._PSEUDO_SUBFOLDER
             ))
+
+            if copy_files:
+                remote_symlink_list.append((
+                    parent_calc_folder.computer.uuid,
+                    os.path.join(parent_calc_folder.get_remote_path(), f'aiida.in'), 'aiida.ref'
+                ))
```
ramirezfranciscof, line 217: This I need to keep a reference of the `plot_num` used in the `INPUTPP` of the first step because the parser uses this value to decide the units of the `ArrayData` output node. Since it will be retrieved and is small, I could use the `remote_copy_list` to be extra sure but retrieving symlinked files works as well.

aiida_quantumespresso/calculations/pp.py
```diff
@@ -197,6 +209,17 @@ def prepare_for_submission(self, folder):  # pylint: disable=too-many-branches,t
...
+                remote_symlink_list.append((
+                    parent_calc_folder.computer.uuid,
+                    os.path.join(parent_calc_folder.get_remote_path(), f'{self._FILPLOT}_*'), '.'
+                ))
```
ramirezfranciscof, line 221: I don't want to duplicate these files as they are tipically heavy, so just simlink them.

aiida_quantumespresso/calculations/pp.py
```diff
@@ -216,9 +239,10 @@ def prepare_for_submission(self, folder):  # pylint: disable=too-many-branches,t
         calcinfo.codes_info = [codeinfo]
         calcinfo.local_copy_list = local_copy_list
         calcinfo.remote_copy_list = remote_copy_list
+        calcinfo.remote_symlink_list = remote_symlink_list
 
         # Retrieve by default the output file
-        calcinfo.retrieve_list = [self.inputs.metadata.options.output_filename]
+        calcinfo.retrieve_list = [self.inputs.metadata.options.output_filename, 'aiida.ref']
```
ramirezfranciscof, line 245: I retrieve this because I may need to parse it. I could also add it to `retrieve_temporary_list`...

aiida_quantumespresso/parsers/pp.py
```diff
@@ -92,6 +92,16 @@ def parse(self, **kwargs):
...
+        if 'plot_num' not in self.output_parameters:
+            try:
+                stdref_raw = retrieved.get_object_content('aiida.ref')
+                for dataline in stdref_raw.splitlines():
+                    if 'plot_num' in dataline:
+                        self.output_parameters['plot_num'] = int(dataline.split('=')[1])
+            except Exception:
+                self.logger.error(traceback.format_exc())
+                return self.exit_codes.ERROR_UNEXPECTED_PARSER_EXCEPTION
+
```
ramirezfranciscof, line 104: This is why I need the `aiida.ref` file, if I'm doing just the second step, `plot_num` does not appear in the output and thus won't be parsed in the section above this one.

The whole PR in general, but specially this parsing of the `plot_num` is very "tied together with wire", as we say (precariously done, with many specifics such as hardcoding the `aiida.ref` filename, etc), just to illustrate what would need to be modified. Ideas about how to properly do this are welcome.

aiida_quantumespresso/parsers/pp.py
```diff
@@ -145,8 +158,9 @@ def get_key_from_filename(filename):
         # If we don't have any parsed files, we exit. Note that this will not catch the case where there should be more
         # than one file, but the engine did not retrieve all of them. Since often we anyway don't know how many files
         # should be retrieved there really is no way to check this explicitly.
-        if not data_parsed:
-            return self.exit_codes.ERROR_OUTPUT_DATAFILE_MISSING.format(filename=filename_prefix)
+        #if not data_parsed:
+        #    return self.exit_codes.ERROR_OUTPUT_DATAFILE_MISSING.format(filename=filename_prefix)
+        # It is now also possible that nothing will be parsed
```
ramirezfranciscof, line 163: If doing just step1 then may not retrieve any output. I could do more elaborate checks such as only checking this if the `PLOT` namelist was provided, etc. But since it anyways says that generically detecting missing files is too complicated, I'm not sure if this is even worth it.

aiida_quantumespresso/parsers/pp.py
```diff
@@ -102,9 +112,12 @@ def parse(self, **kwargs):
...
         # `PpCalculation` plugin which only ever sets 0 - 4, but we check in order for the code not to except.
-        iflag = self.node.inputs.parameters.get_attribute('PLOT')['iflag']
-        if iflag not in range(5):
-            return self.exit_codes.ERROR_UNSUPPORTED_DATAFILE_FORMAT
+        # NOTE: The previous is no longer valid, needs to be corrected
+        plot_namelist = self.node.inputs.parameters.get_attribute('PLOT', None)
+        if plot_namelist is not None:
+            iflag = plot_namelist['iflag']
+            if iflag not in range(5):
+                return self.exit_codes.ERROR_UNSUPPORTED_DATAFILE_FORMAT
```
ramirezfranciscof, line 120: Now the `PLOT` namelist might be missing, but if it is there then it should have the complying `iflag`.

aiida_quantumespresso/calculations/pp.py
```diff
@@ -153,17 +158,23 @@ def prepare_for_submission(self, folder):  # pylint: disable=too-many-branches,t
...
-            4: 0,  # Polar on a sphere -> # Gnuplot, 1D
-        }
-        parameters['PLOT']['output_format'] = dimension_to_output_format[parameters['PLOT']['iflag']]
-
-        namelists_toprint = self._default_namelists
+        if skip_plot:
+            parameters.pop('PLOT')
+            namelists_toprint = list(self._default_namelists)
+            if 'PLOT' in namelists_toprint:
+                namelists_toprint.remove('PLOT')
```
ramirezfranciscof, line 165: If I'm just doing step1, then I just remove the `PLOT` from the parameters (it is there because there were some defaults loaded in after checking the user didn't set anything themselves) and from the list of `namelists_toprint`.

I need to remove it from `namelists_toprint` because the program iterates there to know what namelists to look inside the parameters and print them in the input file (it pops them out of the parameters).

I need to remove it from `parameters` because it will be checked after setting the input for namelists that have not been poped out and will except.

ramirezfranciscof (commented):
A couple of explanations of why of the changes made.

PR #658, opened 2021-03-04
`RelaxType`: Rename `ATOMS` to `POSITIONS`

mbercx:
Renames the `RelaxType.ATOMS` to `RelaxType.POSITIONS`, since it is the
positions of the atoms that are being relaxed. Similarly,
`ATOMS_VOLUME`, `ATOMS_SHAPE` and `ATOMS_CELL` are renamed to
`POSITIONS_VOLUME`, `POSITIONS_SHAPE` and `POSITIONS_CELL`,
respectively.

This change is also necessary to accommodate a similar change in
`aiida-common-workflows`, see:

https://github.com/aiidateam/aiida-common-workflows/pull/146

### DIFF
--- a/aiida_quantumespresso/common/types.py
+++ b/aiida_quantumespresso/common/types.py
@@ -15,13 +15,13 @@ class RelaxType(enum.Enum):
     """Enumeration of known relax types."""
 
     NONE = 'none'  # All degrees of freedom are fixed, essentially performs single point SCF calculation
-    ATOMS = 'atoms'  # Only the atomic positions are relaxed, cell is fixed
+    POSITIONS = 'positions'  # Only the atomic positions are relaxed, cell is fixed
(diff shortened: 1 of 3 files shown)


sphuber (approved):
Thanks a lot @mbercx . If you have the time, could you please test that with this fix the plugin works with the `develop` branch of `aiida-common-workflows`? If so, then it would be great if we can release `v3.3.2`.

mbercx:
> Thanks a lot @mbercx . If you have the time, could you please test that with this fix the plugin works with the `develop` branch of `aiida-common-workflows`? If so, then it would be great if we can release `v3.3.2`.

I already tested it ;). It works fine, but I'm also looking into adding some more structure files to avoid having to manually create the structures that require more kinds, see https://github.com/aiidateam/aiida-common-workflows/issues/154. 

How about I prepare `v3.3.2` release, if you review #622, so we can add two commits to the release? After all, "Eentje is geentje, hé!" ;)

PR #660, opened 2021-03-08
`PwRelaxWorkChain`: remove compatibility for volume optimization

sphuber:
Quantum ESPRESSO only supports optimizing the volume for simple cubic
systems. Since for now by default we accept any structure and set
`ibrav=0`, we simply do not support the relax types `POSITIONS_VOLUME`
and `VOLUME`. Once Quantum ESPRESSO supports volume optimization for any
symmetry or we add some validation to determine simple cubic structures,
we are simply going to remove support in the `get_builder_from_protocol`.


mbercx (approved):
Looks good, thanks @sphuber!

PR #663, opened 2021-03-11
Maintenance: address deprecation warnings

sphuber:
Note that there is one deprecation warning remaining that comes from the `pytest-regressions` dependency. They have [already addressed it](https://github.com/ESSS/pytest-regressions/commit/ffad2c7fd1d110f420f4e3ca3d39d90cae18a972) but we just need to wait for it to be released. When that happens I will update this PR and then it can be merged.


mbercx (approved):
Thanks @sphuber! Not sure why I haven't approved this one already, feels like I just forgot to click the "Submit review" button. 😅

sphuber:
Turns out they also released new version of `pytest-regressions` few weeks ago, so I included that as well

PR #666, opened 2021-03-12
Add support for Python 3.9

### LINKED ISSUE #664: Add support for Python 3.9

sphuber:
Fixes #664 

Also up the requirement for `aiida-core` to `~=1.5` since that is the
version that also added Python 3.9 support.


greschd:
> Also up the requirement for aiida-core to ~=1.5 since that is the
version that also added Python 3.9 support.

I'm not sure this is necessary; shouldn't the `pip` resolver take care of this at install time when on 3.9? 

Were there actual breaking changes, or does `1.4` essentially work also on `3.9`?

sphuber:
No you are right, it shouldn't really matter. I just noticed that the compatibility table in the `README.md` incorrectly shows `aiida-core~=1.5` for the v3.* series of `aiida-quantumespresso` even though in the `setup.json` it is set to `aiida_core[atomic_tools]~=1.4,>=1.4.4`. I guess I could just as well fix the `README.md` and leave the requirement, just with the lower boundary on the patch version it is a bit ugly. I don't remember exactly why this was necessary. The commit that upped it was to up the dependency for `aiida-pseudo` which might actually require it.

sphuber:
Corrected it @greschd

.github/workflows/develop.yml
```diff
@@ -14,7 +14,7 @@ jobs:
 
         strategy:
             matrix:
-                python-version: [3.7, 3.8]
```
mbercx, line 17: I think adding Python 3.6 here will break the nightly build again, since `aiida-core` no longer supports Python v3.6 (due to the pymatgen update of its numpy dependency).

sphuber: Good one, removed it again from the develop workflow

mbercx (approved):
All good, thanks @sphuber!

PR #671, opened 2021-04-13
Dependencies: update to `aiida-pseudo==0.6.0`

### LINKED ISSUE #670: Add support for `aiida-pseudo` v0.6.0
The upcoming `aiida-pseudo` release will include functionality for specifying units for the recommended cutoffs, as well as an overhaul of the methods for setting and obtaining the cutoffs. This will require changes in the `get_builder_from_protocol()` method of the `PwBaseWorkChain` (and possibly elsewhere?).

One question is if we want to keep on supporting pseudo families that were installed with previous versions of `aiida-pseudo` (e.g. by checking if any units are defined for the family). In the past several users have had issues because they installed a pseudo family with an outdated version of `aiida-pseudo`, then _thought_ they were all good after only updating the package.

mbercx:
Fixes #670 

This new minor version introduces two important changes:

*  Add support for units for the recommended cutoffs. This means that we
can now request the recommended cutoffs in a specified unit. We add
support for this feature here.
* Changes the `set_cutoffs` method of the `RecommendedCutoffMixin` class
so it can set the recommended cutoffs for a single stringency instead
all at the same time. This only has influence on the `sssp` fixture for
the tests in `conftest.py`.

### DIFF
--- a/aiida_quantumespresso/cli/calculations/cp.py
+++ b/aiida_quantumespresso/cli/calculations/cp.py
@@ -23,11 +23,10 @@ def launch_calculation(code, structure, pseudo_family, max_num_machines, max_wal
     """Run a CpCalculation."""
     from aiida.orm import Dict
     from aiida.plugins import CalculationFactory
-    from qe_tools import CONSTANTS
 
     from aiida_quantumespresso.utils.resources import get_default_options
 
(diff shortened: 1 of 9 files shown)


mbercx:
Tests failing because `aiida-pseudo==0.6.0` hasn't been released yet. ;)

sphuber (approved):
To the moon :rocket:

PR #674, opened 2021-04-14
CI: use `postgres-12` since update to Ubuntu Focal 20.04

sphuber:
The CI configuration uses `ubuntu-latest` which recently became Ubuntu
Focal Fossa (20.04) which no longer has `postgres-10` available, but
instead provides `postgres-12`.


mbercx (approved):
🚀

PR #676, opened 2021-04-18
`BasePwCpInputGenerator`: Improve parameters input validation

### LINKED ISSUE #675: `BasePwCpInputGenerator`: Improve `parameters` input validation
### Problem description

Currently, the `parameters` input for the `BasePwCpInputGenerator` is:

1. Not validated as a validator, but in the `prepare_for_submission`  method, raising `InputValidationError`s when there are issues.
2. Lacking a lot of validation that causes e.g. `pw.x` to crash.

[1] Causes the calculation to `Except`, e.g. when using the `PwBaseWorkChain.get_builder_from_protocol()` method:

```Python
PwBaseWorkChain = WorkflowFactory('quantumespresso.pw.base')

    structure=orm.load_node(3),
    code=orm.load_code(1)
)
builder.pw.parameters['CONTROL']['calculation'] = 'GME stock'
```
submitting this `builder` leads to an `Excepted` work chain with the following error:

```Python
aiida.common.exceptions.InputValidationError: Unknown 'calculation' value in CONTROL namelist GME stock. Otherwise, specify the list of namelists using the NAMELISTS inside the 'settings' input node
```

[2] Usually starts running just fine, but `pw.x` will crash, leading to `exit_status` `305` for the work chain, e.g.:

```Python
PwBaseWorkChain = WorkflowFactory('quantumespresso.pw.base')

    structure=orm.load_node(3),
    code=orm.load_code(1)
)
builder.pw.parameters['CONTROL']['wf_collect'] = 'at the moon'
```
submitting this `builder` leads to `exit_status` `305`, which requires the user to use `verdi calcjob outputcat` to figure out what's going wrong by looking at the outputs:

```console
     Error in routine  read_namelists (1):
      bad line in namelist &control: "/" (error could be in the previous line)
```
and even then the user still has to look at the input files to figure out what went wrong.

### Suggested improvements

Use validation based on `jsonschema` along with a `validate_parameters` static method assigned as a validator to the `parameters` input. Move validation that cannot be validated with the schema from the `prepare_for_submission` or `_generate_PWCPinputdata` methods to the `validate_parameters` method, and remove validation that has become unnecessary with the `jsonschema` validation.

### Open questions

- [ ] Should the `validator` be allowed to change the input? An example here is the following fix executed in the `_generate_PWCPinputdata` method:

  ```Python
          input_params = _uppercase_dict(parameters.get_dict(), dict_name='parameters')
          input_params = {k: _lowercase_dict(v, dict_name=k) for k, v in input_params.items()}
  ```
  I'd be inclined to move this to the `validate_parameters` method.

- [ ] Currently, the only `jsonschema` validation we have is for the `PdosWorkChain`. Here, the `jsonschema` is stored in the module itself, defined as a dictionary. I'd be inclined to define the schemas in `yaml` format and place them all in a `schemas` folder as a subfolder of the `calculations` folder, with the following structure:
  ```console
  └── schemas
      ├── parameters.yaml
      └── pw
          └── parameters.yaml
      └── cp
          └── parameters.yaml
  ...
  ```
  Where the top level `parameters.yaml` file contains parameters for the `BasePwCpInputGenerator` class, i.e. that are shared by the `pw.x` and `cp.x` calculations.

- [ ] Are there such schemas already available for Quantum ESPRESSO? Although it would not be an insurmountable task to set these up ourselves, if each released version of QE would have schemas for the inputs, that'd make life quite a bit easier. 😅 
- [ ] How to make the validation compatible with multiple QE versions, and how far back this support should extend.

### Other context

There is quite a bit of validation in the `calculation.helpers.pw_input_helper()` method. A couple of points here:

* If we make the `validate_parameters` a `staticmethod` of the `BasePwCpInputGenerator` and `PwCalculation` classes, the user can validate the inputs parameters via (following the above examples):

  ```Python
  PwCalculation.validate_parameters(builder.pw.parameters, _)
  ```

  For example [1], this would lead to:

  ```Python
  ValidationError: 'GME stock' is not one of ['scf', 'nscf', 'bands', 'relax', 'md', 'vc-relax', 'vc-md']

  Failed validating 'enum' in schema['properties']['CONTROL']['properties']['calculation']:
      {'default': 'scf',
       'enum': ['scf', 'nscf', 'bands', 'relax', 'md', 'vc-relax', 'vc-md'],
       'type': 'string'}

  On instance['CONTROL']['calculation']:
      'GME stock'
  ```
  This would be easier to find than the `pw_input_helper` method.
* Since the `pw_input_helper` method also considers the `structure` input, not all validation that is done here can be moved to the `validate_parameters` method.

mbercx:
Fixes #675 

First working example of how we can add validation of the input
parameters of the Quantum ESPRESSO calculation jobs by converting the
INPUT-{}.XML files that are generated for the QE documentation into JSON
schemas that are then used in the validator of the `parameters` input.

The `convert_to_json_schema()` function in the `calculations.schemas.utils.py`
module is used to generate the `calculations.schemas.pw.parameters.yaml`
file, based on the `calculations.helpers.INPUT_PW-6.4.xml` file. This
yaml file is then loaded in the validator for the `parameters` input and
used to check if the parameters adhere to the prescriptions of the JSON
schema.

### DIFF
--- a/aiida_quantumespresso/calculations/__init__.py
+++ b/aiida_quantumespresso/calculations/__init__.py
@@ -115,7 +115,8 @@ class BasePwCpInputGenerator(CalcJob):
         spec.input('structure', valid_type=orm.StructureData,
             help='The input structure.')
         spec.input('parameters', valid_type=orm.Dict,
-            help='The input parameters that are to be used to construct the input file.')
+            help='The input parameters that are to be used to construct the input file.',
+            validator=cls.validate_parameters)
         spec.input('settings', valid_type=orm.Dict, required=False,
             help='Optional parameters to affect the way the calculation job and the parsing are performed.')
         spec.input('parent_folder', valid_type=orm.RemoteData, required=False,
@@ -151,6 +152,11 @@ class BasePwCpInputGenerator(CalcJob):
             invalid_values = [val for val in value_dict.values() if not isinstance(val, numbers.Integral)]
             if invalid_values:
                 return f'Parallelization values must be integers; got invalid values {invalid_values}.'
+
+    @staticmethod
+    def validate_parameters(cls, value, _):
+        """Validate the input parameters of the pw.x calculation."""
+        pass
 
     def prepare_for_submission(self, folder):
         """Create the input files from the input nodes passed to this instance of the `CalcJob`.
@@ -442,11 +448,8 @@ class BasePwCpInputGenerator(CalcJob):
             pseudo = pseudos[kind.name]
             if kind.is_alloy or kind.has_vacancies:
                 raise exceptions.InputValidationError(
-                    "Kind '{}' is an alloy or has "
-                    'vacancies. This is not allowed for pw.x input structures.'
-                    ''.format(kind.name)
+                    f'Kind `{kind.name}` is an alloy or has vacancies. This is not allowed for `pw.x` input structures.'
                 )
-
             try:
                 # If it is the same pseudopotential file, use the same filename
                 filename = pseudo_filenames[pseudo.pk]
--- a/aiida_quantumespresso/calculations/pw.py
+++ b/aiida_quantumespresso/calculations/pw.py
@@ -1,10 +1,14 @@
 # -*- coding: utf-8 -*-
 """`CalcJob` implementation for the pw.x code of Quantum ESPRESSO."""
 import os
+import yaml
+import jsonschema
 
+from pathlib import Path
 from aiida import orm
 from aiida.common.lang import classproperty
 from aiida.plugins import factories
+from ..workflows.protocols.utils import recursive_merge
 
 from aiida_quantumespresso.calculations import BasePwCpInputGenerator
 
@@ -146,6 +150,15 @@ class PwCalculation(BasePwCpInputGenerator):
         spec.exit_code(541, 'ERROR_SYMMETRY_NON_ORTHOGONAL_OPERATION',
             message='The variable cell optimization broke the symmetry of the k-points.')
         # yapf: enable
+
+    @classmethod
+    def validate_parameters(cls, value, _):
+        """Validate the input parameters of the pw.x calculation."""
+
+        with (Path(__file__).resolve().parent / 'schemas' / 'pw' / 'parameters.yaml').open() as handle:
+            parameters_schema = yaml.safe_load(handle)
+
+        jsonschema.validate(value.get_dict(), parameters_schema)
 
     @classproperty
     def filename_input_hubbard_parameters(cls):
--- a/aiida_quantumespresso/calculations/schemas/pw/parameters.yaml
+++ b/aiida_quantumespresso/calculations/schemas/pw/parameters.yaml
@@ -1,0 +1,546 @@
+$schema: http://json-schema.org/draft/2019-09/schema#
+additionalProperties: false
+properties:
+  CELL:
+    additionalProperties: false
+    properties:
+      cell_dofree:
+        enum:
+        - all
+        - ibrav
+        - x
+        - y
+        - z
+        - xy
+        - xz
+        - yz
+        - xyz
+        - shape
+        - volume
+        - 2Dxy
+        - 2Dshape
+        - epitaxial_ab
+        - epitaxial_ac
+        - epitaxial_bc
+        type: string
+      cell_dynamics:
+        enum:
+        - none
+        - sd
+        - damp-pr
+        - damp-w
+        - bfgs
+        - none
+        - pr
+        - w
+        type: string
+      cell_factor:
+        type: number
+      press:
+        type: number
+      press_conv_thr:
+        type: number
+      wmass:
+        type: number
+    type: object
+  CONTROL:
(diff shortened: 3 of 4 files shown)


mbercx:
@sphuber this is a first working example of converting an `INPUT_PW-{}.xml` file into a JSON schema in `.yaml` format that can then be used to validate the input parameters. Here's the code that I used to generate the `.yaml` file:

```Python
import xml, yaml

from aiida_quantumespresso.calculations.schemas.utils import convert_to_json_schema

xml_path = 'INPUT_PW-6.4.xml'

with open(xml_path, 'r') as file:
    json_schema = convert_to_json_schema(xml.dom.minidom.parse(file))
    
with open('parameters.yaml', 'w') as file:
    file.write(yaml.safe_dump(json_schema, default_flow_style=False))
```

Currently the validation checks:

* If the input `namelist` and `tag` are in the list of those of the JSON schema (`additionalProperties: false`).
* If the input is of the correct `type`.
* In case there are only a limited number of options: if the input is in the list of options (`enum`).

Example [1] from #675, where:
```Python
builder.pw.parameters['CONTROL']['calculation'] = 'GME stock'
```
Now gives:
```Python
ValidationError: 'GME stock' is not one of ['scf', 'nscf', 'bands', 'relax', 'md', 'vc-relax', 'vc-md']

Failed validating 'enum' in schema['properties']['CONTROL']['properties']['calculation']:
    {'enum': ['scf', 'nscf', 'bands', 'relax', 'md', 'vc-relax', 'vc-md'],
     'type': 'string'}

On instance['CONTROL']['calculation']:
    'GME stock'
```

Example [2] from #675, where:
```Python
builder.pw.parameters['CONTROL']['wf_collect'] = 'at the moon'
```
Now gives:
```Python
ValidationError: 'at the moon' is not of type 'boolean'

Failed validating 'type' in schema['properties']['CONTROL']['properties']['wf_collect']:
    {'type': 'boolean'}

On instance['CONTROL']['wf_collect']:
    'at the moon'
```

mbercx:
And deeper down the validation rabbit hole I go...

To do some basic testing of which schema accepts which parameters, I've added the `generate_params_from_schema()` function in the `calculations.schemas.utils.py` module. This is a rather rudimentary approach at generating an example dictionary that fills in every input for the schema with an acceptable value. So, while these parameters most likely don't make sense, they are "complete", and hence good for testing the `jsonschema` validation when `additionalProperties: false`.

Unfortunately, it turned out that e.g. taking the schema that corresponds to the latest version (6.7) doesn't validate the generated parameters by previous schemas (since inputs/options were removed) or vice versa (since inputs/options were added). So, I wanted to look into merging the JSON schemas into one that has all the inputs/options, that can be used as a default.

This lead me to the [`genson`](https://github.com/wolverdude/GenSON) package, which can be used for exactly that purpose. By using the `SchemaBuilder`, one can simply add multiple schemas and `genson` will try to merge them into one that accepts data that conforms to each schema. However, the `'enum'` keyword was not yet supported, so I [took a stab at adding it](https://github.com/wolverdude/GenSON/pull/57), and it seems to be working nicely for my use case.

Merging the `pw.x` JSON schemas for major versions (5 and 6) works fine (i.e. there are no conflicts), but not between schemas that correspond to different major versions.  This seems to be in part related to the fact that the `5.X` versions of the XML input files for `pw.x` do not have the option (`opt`) variables defined, so the generated JSON schemas are limited to checking the `type`, since the `enum`s are generated from the `opt`. This is similar for _all_ `cp.x` XML input files, i.e. in QE v6.X the XML files also do not have options defined. The advantage here is that for the `cp.x` schemas I _can_ generate one schema that successfully validates generated parameters from _all_ schemas.

So far I've just added a few schemas, and chosen the all-encompassing schema for `cp.x`, and the one generated for major version 6.X for `pw.x` as the defaults used in the `validate_parameters` method. This seems to be working nicely, however someone trying to run v5.X for `pw.x` _could_ run into validation errors. 

A couple more points to hash out/work on:
- [ ] Although I'm partial to only providing support to v6.0, I think it makes more sense in the end to move towards including `version` information in the `Code` (set as an extra, for now). This will not only allow us to make the validation more strict (and correct, in the case of running v5.X), but also come in handy when we want to indicate that the `Code` is SIRIUS-enabled (e.g. `version == 'sirius-6.5'`. This will need to be documented properly, though, or maybe enforced and added to the code setup step. Can we maybe demand more information based on the calculation plugin during the code setup?
- [ ] Remove obsolete code from the `BasePwCpInputGenerator._generate_PWCPinputdata()` method. An example:
https://github.com/aiidateam/aiida-quantumespresso/blob/33cb11e3c9a1f382fac69c15d74a2b4f6d331d4c/aiida_quantumespresso/calculations/__init__.py#L381-L385
  This code served no purpose anymore, since errors in the case with be caught by the validation.
- [ ] The documentation also indicates which inputs are `REQUIRED`, so I can also add this to the JSON schema.
- [ ] Add input validation that relies on other inputs to a top-level input validator. This might also include the JSON schema validation in case we need to check the version of the input `Code`.

Issue #653, opened 2021-02-18
Protocols: Consider way of "popping" when `overrides` is `None`


mbercx:
Currently, we call quite a lot of `pop` methods in the `get_builder_from_protocol()` methods to remove inputs which are not necessary for the underlying work chain, for example:

https://github.com/aiidateam/aiida-quantumespresso/blob/128ec5967e0193571ff2052abc68f5c894a8c3cd/aiida_quantumespresso/workflows/pw/bands.py#L135-L143

I was thinking it would be good to be able to simply add this removal of inputs to the `.yaml` files. I.e. replace:

https://github.com/aiidateam/aiida-quantumespresso/blob/128ec5967e0193571ff2052abc68f5c894a8c3cd/aiida_quantumespresso/workflows/protocols/pw/bands.yaml#L1-L4

with

```yaml
default_inputs:
    bands_kpoints_distance: 0.025
    clean_workdir: True
    nbands_factor: 3.0
    relax:
        structure: null
        clean_workdir: null
        base_final_scf: null
    scf:
        pw:
            structure: null
<...>
```

I think this would be cleaner since you can see the inputs that are added/updated/removed in one place: the `.yaml` file. I especially noticed this when working on the `PdosWorkChain` protocol (see [this PR](https://github.com/chrisjsewell/aiida-quantumespresso/pull/1) in case you're just too curious and already want to have a look and maybe review it, @sphuber 😁 ). There are several inputs which need to be removed from the default values for the `nscf` `PwBaseWorkChain`, and I think it would cleaner and more readable to do this in the `PwBaseWorkChain`.

There are other advantages to adding the possibility of being able to remove inputs by setting the `overrides` value to `None`. @flavianojs was trying to switch to using the `kpoints` input (i.e. providing a `KpointsData`) _instead of_ the `kpoints_distance` input. This is currently not possible by providing the correct `overrides`. Of course, it's possible to adjust this after obtaining the builder, but I think it should be possible to change all inputs with the `overrides`. This will also make it easier in the future for people to define their own protocols  - which can touch every input - by writing `.yaml` files.

[...]

PR #678, opened 2021-04-20
`ProtocolMixin`: increase flexibility and usability for other packages

### LINKED ISSUE #677: `ProtocolMixin`: don't use `__file__` to obtain path to protocol `.yaml` files
Currently, the `ProtocolMixin` class uses `__file__` to obtain the path to the `.yaml` files that define the protocols:

https://github.com/aiidateam/aiida-quantumespresso/blob/33cb11e3c9a1f382fac69c15d74a2b4f6d331d4c/aiida_quantumespresso/workflows/protocols/utils.py#L105-L106

While this is perfectly sensible for the work chains in the `aiida-quantumespresso` package, once another package wants to rely on the `ProtocolMixin` class to add protocols for their work chains, it no longer is a viable strategy, since these protocols will be in the corresponding directory of that package. Instead, we should try and use the path of the module where the work chain class is defined.

mbercx:
Fixes #677 

The load_protocol_file method in the protocol utils.py
module uses the path of the module file to obtain the path to the .yaml
file that described the protocols. However, this fails in case the
`ProtocolMixin` class is used by other packages, since the protocol files
of the package's work chains will be stored in that package.

Moreover, the user can only adapt the protocol by specifying
`overrides` as a dictionary, or adapting the builder afterwards.

Here we adapt the `ProtocolMixin` class as follows:

* Add a `get_protocol_filepath()` class method that needs to be
implemented by the work chain classes that use the `ProtocolMixin`.
* Move the `load_protocol_file` method into the `ProtocolMixin` class as
a hidden method that relies on the `get_protocol_filepath` method.
* The `get_protocol_inputs` method is made more flexible by accepting
a `pathlib.Path` for the `overrides` input. This allows the user to
define protocol overrides in a `.yaml` file.

The `Path` to the hardcoded protocols that ship with the package are now
also obtained using `importlib_resources`.

### DIFF
--- a/aiida_quantumespresso/workflows/pdos.py
+++ b/aiida_quantumespresso/workflows/pdos.py
@@ -308,6 +308,13 @@ class PdosWorkChain(ProtocolMixin, WorkChain):
         spec.expose_outputs(ProjwfcCalculation, namespace='projwfc')
 
     @classmethod
+    def get_protocol_filepath(cls):
+        """Return ``pathlib.Path`` to the ``.yaml`` file that defines the protocols."""
+        from importlib_resources import files
+        from . import protocols
+        return files(protocols) / 'pdos.yaml'
+
+    @classmethod
     def get_builder_from_protocol(
         cls, pw_code, dos_code, projwfc_code, structure, protocol=None, overrides=None, **kwargs
     ):
@@ -324,11 +331,10 @@ class PdosWorkChain(ProtocolMixin, WorkChain):
         :return: a process builder instance with all inputs defined ready for launch.
         """
 
-        args = (pw_code, structure, protocol)
 
         inputs = cls.get_protocol_inputs(protocol, overrides)
-        builder = cls.get_builder()
 
+        args = (pw_code, structure, protocol)
         scf = PwBaseWorkChain.get_builder_from_protocol(*args, overrides=inputs.get('scf', None), **kwargs)
         scf['pw'].pop('structure', None)
         scf.pop('clean_workdir', None)
@@ -338,6 +344,7 @@ class PdosWorkChain(ProtocolMixin, WorkChain):
         nscf['pw']['parameters']['SYSTEM'].pop('degauss', None)
         nscf.pop('clean_workdir', None)
 
+        builder = cls.get_builder()
         builder.structure = structure
         builder.clean_workdir = orm.Bool(inputs['clean_workdir'])
         builder.scf = scf
--- a/aiida_quantumespresso/workflows/protocols/utils.py
+++ b/aiida_quantumespresso/workflows/protocols/utils.py
@@ -1,7 +1,5 @@
 # -*- coding: utf-8 -*-
 """Utilities to manipulate the workflow input protocols."""
-import functools
-import os
 import pathlib
 import yaml
 
@@ -10,13 +8,18 @@ class ProtocolMixin:
     """Utility class for processes to build input mappings for a given protocol based on a YAML configuration file."""
 
     @classmethod
+    def get_protocol_filepath(cls):
+        """Return the ``pathlib.Path`` to the ``.yaml`` file that defines the protocols."""
+        raise NotImplementedError
+
+    @classmethod
     def get_default_protocol(cls):
         """Return the default protocol for a given workflow class.
 
         :param cls: the workflow class.
         :return: the default protocol.
         """
-        return load_protocol_file(cls)['default_protocol']
+        return cls._load_protocol_file()['default_protocol']
 
     @classmethod
     def get_available_protocols(cls):
@@ -26,7 +29,7 @@ class ProtocolMixin:
         :return: dictionary of available protocols, where each key is a protocol and value is another dictionary that
             contains at least the key `description` and optionally other keys with supplementary information.
         """
-        data = load_protocol_file(cls)
+        data = cls._load_protocol_file()
         return {protocol: {'description': values['description']} for protocol, values in data['protocols'].items()}
 
     @classmethod
@@ -39,7 +42,7 @@ class ProtocolMixin:
             maintain the exact same nesting structure as the input port namespace of the corresponding workflow class.
         :return: mapping of inputs to be used for the workflow class.
         """
-        data = load_protocol_file(cls)
+        data = cls._load_protocol_file()
         protocol = protocol or data['default_protocol']
 
         try:
@@ -48,7 +51,6 @@ class ProtocolMixin:
             raise ValueError(
                 f'`{protocol}` is not a valid protocol. Call ``get_available_protocols`` to show available protocols.'
             ) from exception
-
         inputs = recursive_merge(data['default_inputs'], protocol_inputs)
         inputs.pop('description')
 
@@ -56,6 +58,12 @@ class ProtocolMixin:
             return recursive_merge(inputs, overrides)
 
         return inputs
+
+    @classmethod
+    def _load_protocol_file(cls):
+        """Return the contents of the protocol file for workflow class."""
+        with cls.get_protocol_filepath().open() as file:
+            return yaml.safe_load(file)
 
 
 def recursive_merge(left, right):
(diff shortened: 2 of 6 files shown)


chrisjsewell:
Or you could use [importlib.resources](https://importlib-resources.readthedocs.io) 😉

mbercx:
> Or you could use [importlib.resources](https://importlib-resources.readthedocs.io) 😉

Nice, wasn't aware of this module 👌. Will have a closer look tomorrow.

sphuber:
I am wondering if we want to really make this utility more reusable by other plugins, why not make it more flexible still. The changes are still making quite some assumptions about where protocol files are supposed to be stored. I remember that we were also thinking of passing a completely custom protocol file on call of `get_builder_from_protocol` so shouldn't we already move to an interface where the path is literally passed in, or can be at least?

sphuber:
There is also the question whether a custom protocol file that is passed completely replaces existing protocols, or simply adds to it, i.e. the final protocol is the union of the default and custom files taken together. I think that would maybe give the highest flexibility. In that case, I would also not repurpose the `protocol` argument as a way to specify the path, but simply add an optional argument `protocol_filepath`. If specified, the overlap of that file and the standard file is taken where the former takes precedence when both protocols specify the same input. I would then also keep the same format for the custom protocol file as the standard one, except the `default_protocol` and `default_inputs` keys are optional. One could therefore for example create a file:
```
protocols:
    moderate:
         kpoints_distance: 0.20
```

mbercx:
I'm not sure if I fully agree. If the user wants to override _some_ settings of an already existing protocol, the `overrides` input argument can be used. This also adds another input argument, making the method more complex to use. I suggest the following inputs for determining the protocol:

* `protocol: Union[str, pathlib.Path, None] = None` - Specify the protocol to use. This is either a string identifier (e.g. 'moderate', 'fast', 'precise') or a ``Path`` to a file that defines the protocol. If not specified, the default protocol will be used.
* `overrides: Union[dict, pathlib.Path, None] = None` - optional inputs to override the defaults of the protocol. Can be either a ``dict`` or a ``Path`` to a ``.yaml`` file that contains the overrides.

I think these two inputs cover all use cases, e.g. by repurposing the `overrides` argument to also accept a ``Path``, your suggested use case is covered, and it's also immediately clear that the inputs defined in this `yaml` file will override the defaults specified in the `protocol` input.

mbercx:
After working on this a bit, I've come to the conclusion that allowing the `protocol` input to also accept a `Path` adds a lot of complexity to the implementation for relatively little gain. The problem arises when trying to do this for e.g. the `PwRelaxWorkChain`. The hardcoded protocol file only contains the defaults for the top-level inputs, and overrides for the lower-level `PwBaseWorkChain`s. So either this input protocol file would have to specify which protocol to use for the lower-level work chains along with the overrides, or contain all the defaults for the underlying work chains. And then we'd have to deal with all these cases in the `ProtocolMixin` class and the `get_builder_from_protocol` methods. 

We _could_ add another `protocol_filepath ` input that has the same structure as the hardcoded file, and then load that one instead. But, then the user would have to specify both this file and the `protocol` input. If this file would just override the default values, having the ability to pass a `Path` for the `overrides` does pretty much the same thing, without the added complexity of starting with:

```
protocols:
    moderate:
```
and then having to select the `moderate` protocol for this to take effect.

A couple of notes:
* As I mention in https://github.com/aiidateam/aiida-quantumespresso/issues/653, the `overrides` don't let us remove input tags, only change them. We could fix this by letting the user set `null` to remove an input, but this is still a little of a roundabout way of defining your own protocol with a `.yaml` file.
* Initially I thought about keeping the logic for constructing the file path to the `.yaml` files based on the module and putting it in the `ProtocolMixin.get_protocol_filepath` method, so we wouldn't have to implement it for our work chains here. But then I figured that it might be helpful for others who want to implement protocols to simply see the `get_protocol_filepath` method (and be able to copy it). Note that removing this method and simply adding a `protocol_filepath` input to `get_protocol_inputs` won't do, because then we'd also have to add this input to `get_default_protocol` and `get_available_protocols`.

sphuber:
Ok, let's try to separate the various requests in action here. If I understand the current code in this PR, it mostly makes the `ProtocolMixin` more flexible and actually usable by other `WorkChain` class by allowing to define where the protocol file is defined. That is already great and should be merged either way I would say.

Then there is the question how the existing protocol files that ship with the package, can be (partially) modified at runtime. I think the use case here is that a user of the `aiida-quantumespresso` plugin would like to run `PwBaseWorkChain.get_builder_from_protocol` but use a slightly modified protocol. I am getting a bit confused on what the difference is between the protocols and the overrides. I also see the potential problem that if we have some way of dynamically overriding the protocol of a workchain, that should allow to also override the settings of all lower lying workchains. I don't think it will be feasible for the interface to allow to specify a custom protocol file for each sub workchain that is involved, so instead a single protocol file should do the trick. But then again, I feel like we already have this no? The protocol file of the PwRelax _can_ and _does_ already specify values that override values of the PwBase protocol. Ultimately, it just follows the same structure of the entire spec. With the custom protocol file it should just be "overlaid" onto the base one and wherever they overlap, the custom takes precedence. Won't this be possible and address everything? (The difference between protocol and overrides left aside for now)

mbercx:
> Then there is the question how the existing protocol files that ship with the package, can be (partially) modified at runtime. I think the use case here is that a user of the `aiida-quantumespresso` plugin would like to run `PwBaseWorkChain.get_builder_from_protocol` but use a slightly modified protocol. I am getting a bit confused on what the difference is between the protocols and the overrides.

To me, what you are describing as _slightly modifying_ is exactly the overrides.

> I also see the potential problem that if we have some way of dynamically overriding the protocol of a workchain, that should allow to also override the settings of all lower lying workchains. I don't think it will be feasible for the interface to allow to specify a custom protocol file for each sub workchain that is involved, so instead a single protocol file should do the trick. But then again, I feel like we already have this no? The protocol file of the PwRelax _can_ and _does_ already specify values that override values of the PwBase protocol.

Yes, but our protocol files are a little mixed. On the one hand it specifies _defaults_ for the _top-level_ inputs, and then _overrides_ for the _lower-level_ inputs. Moreover, it is even more complex, since each protocol file also defines two more protocols that again override the values that are defined by the `default_protocol`. Now, I do agree with this design, since it is more maintainable (i.e. changing a default `mixing_beta` for the `pw.x` calculation only has to be changed in one file). But it does mean that we can't have the user supply 1 _protocol_ file that _replaces_ (not _overrides_) e.g. the `moderate` protocol without making this file quite complex (i.e. define the protocols for the work chain and _all_ its lower-level work chains). 

> Ultimately, it just follows the same structure of the entire spec. With the custom protocol file it should just be "overlaid" onto the base one and wherever they overlap, the custom takes precedence. Won't this be possible and address everything? (The difference between protocol and overrides left aside for now)

Again, this is exactly what the `overrides` do, no? With this PR the user can specify these `overrides` as a `pathlib.Path` to a `.yaml` file.

---

Let me try and crystallise the difference between a protocol and the overrides somewhat more:

* A `protocol` defines a set of default inputs for the work chains.
* The `overrides` defines a set of inputs to change for a given protocol.
 
The `'moderate'` protocol is a _pure_ protocol. The `'fast'` and `'precise'` protocols are only defined as `overrides` of the `'moderate'` protocol. With the current implementation the user can define a new protocol in the sense of the `'fast'` or `'precise'` protocol, i.e. it can pass a `dict` or `.yaml` file that defines overrides for the `'moderate'` protocol.

What the user _can't_ do yet, is easily define a _pure_ protocol (which I tried to do by passing a `Path` to `protocol`). Now, almost everything can be achieved with `overrides` (except removing inputs, see #653), but it's a little convoluted that the user always has to define their protocol in relation to the default one, instead of just being able to write their own _pure_ protocol.

When I first looked at how the protocol files were set up, I was a little confused. I was expecting 1 `.yaml` file for each protocol, that defined the defaults for all the work chains. However, this double overriding strategy made sense to me to avoid duplication and consequently maintenance issues. However, it does make it tricky for the user to just e.g. copy a protocol file and define their own _pure_ protocol that they can run for higher-level work chains.

Sorry for the long-winded explanation. I hope I wasn't being redundant. 😅

mbercx:
@sphuber the changes have been separated in two commits as discussed, and the PR is ready for review!

sphuber:
noticed that you said "allow `cutoffs` to be passed" in the commit message instead of `overrides` and couldn't help myself :grimacing:  Don't worry though, I adapted it and the future of humanity is saved

sphuber (approved):
All good (except for the fumble in one of the commit message, but you have been forgiven) thanks @mbercx

mbercx:
hero

mbercx:
> noticed that you said "allow cutoffs to be passed" in the commit message

Working on `aiida-pseudo` is getting to me. 😅

Issue #528, opened 2020-06-12
Fix resource validation in all workchains


dpshoema:
I was running through the tutorials and all my local submitted jobs hang at

` PwCalculation    ⨯ Excepted       Waiting for transport task: upload`

Running the test script (aiida-quantum-espresso calculation launch ...) helped give a more specific error

`TypeError: At least two among num_machines, num_mpiprocs_per_machine or tot_num_mpiprocs must be specified`

So updating the script to 

`builder.metadata.options.resources = {'num_machines': 1,'num_mpiprocs_per_machine': 1}`

Fixed the problem. It was actually quite hard to pin down since I never saw the resources error when submitting to the daemon. Not sure if this is only my problem or more common. If the latter, does the example script need to be updated?

sphuber:
Thanks for the report. The documentation, including the examples is unfortunately outdated here and there due to the big updates since `aiida-core==1.0.0`. I haven't had the time yet to correct it. The problem you faced should already be made easier with a more recent version of AiiDA. Since then, the resources are validated before actually submitting it. So you would see the error that you saw when running through the `aiida-quantumespresso` CLI. Unfortunately, the question of the resources is not as easy, as it depends on the scheduler what constitutes "correct resources". There is an open issue: https://github.com/aiidateam/aiida-core/issues/3887

BeZie:
@sphuber 
You are probably already aware of it but the current versions of the base workflows (e.g. ph) are checking for num_machines, but now some scheduler must not have num_machines set, e.g. the SGE scheduler.  Not sure if I should open a new issue or keep it here.

sphuber:
Thanks for the heads-up @BeZie . I have fixed the problem in `aiida-core` which was released in v1.3 last week. I will try to find the time soon to update the workflows here as well.

mbercx:
Pretty sure this was fixed in AiiDA core some time ago, along with the work in https://github.com/aiidateam/aiida-quantumespresso/pull/683. Feel free to re-open in case I'm mistaken.

PR #683, opened 2021-05-05
`PwBaseWorkChain`: fix bug in `validate_resources` validator

### LINKED ISSUE #682: `PwBaseWorkChain` validation of resources is incorrect
It assumes that `num_machines` and `max_wallclock_seconds` are set, however, this only applies to certain schedulers. For example for the `SgeScheduler` you _cannot_ specify `num_machines` so the workchain is fundamentally incompatible with those schedulers. The solution is simple: just remove the validator. The reason is that this validator was added at a time when the resources for the scheduler were not yet validated in `aiida-core`. But now they are (should check starting from what version) so now we can just rely on this, and even better, the validation will happen on instantiation and not when the workchain starts running.

sphuber:
Fixes #682 

The `validate_resources` outline step would check that the
`metadata.options` namespace included at least the input
`resources.num_machines` and `max_wallclock_seconds` as long as the
automatic parallelization feature was not enabled. This was added when
the workchain was first developed, because the scheduler resources were
not validated on the base `CalcJob` in `aiida-core`.

However, since `aiida-core==1.0.0` the scheduler resources are validated
on the `CalcJob` base class. This means that a `PwBaseWorkChain` will
now not even be instantiated if the resources for the `PwCalculation`
are insufficient, so it is better to leave the validation up to the base
class.

What is worse, the logic in `validate_resources` is actually incorrect
as it assumed that the resources always need to include `num_machines`.
However, this is not the case as it is actually `Scheduler`
implementation specific, and for example for the `SgeScheduler` this
field is even prohibited and will raise when specified. This made that
the `PwBaseWorkChain` was fundamentally incompatible with those
schedulers before this fix.

The `PhBaseWorkChain` suffered from the same bug and has also been
fixed.

### DIFF
--- a/aiida_quantumespresso/workflows/ph/base.py
+++ b/aiida_quantumespresso/workflows/ph/base.py
@@ -30,7 +30,6 @@ class PhBaseWorkChain(BaseRestartWorkChain):
         spec.outline(
             cls.setup,
             cls.validate_parameters,
-            cls.validate_resources,
             while_(cls.should_run_process)(
                 cls.prepare_process,
                 cls.run_process,
@@ -40,7 +39,8 @@ class PhBaseWorkChain(BaseRestartWorkChain):
         )
         spec.expose_outputs(PwCalculation, exclude=('retrieved_folder',))
         spec.exit_code(204, 'ERROR_INVALID_INPUT_RESOURCES_UNDERSPECIFIED',
-            message='The `metadata.options` did not specify both `resources.num_machines` and `max_wallclock_seconds`.')
+            message='The `metadata.options` did not specify both `resources.num_machines` and `max_wallclock_seconds`. '
+                    'This exit status has been deprecated as the check it corresponded to was incorrect.')
         spec.exit_code(300, 'ERROR_UNRECOVERABLE_FAILURE',
             message='The calculation failed with an unrecoverable error.')
         # yapf: enable
@@ -66,21 +66,7 @@ class PhBaseWorkChain(BaseRestartWorkChain):
         if self.inputs.only_initialization.value:
             self.ctx.inputs.settings['ONLY_INITIALIZATION'] = True
 
-    def validate_resources(self):
-        """Validate the inputs related to the resources.
-
-        The `metadata.options` should at least contain the options `resources` and `max_wallclock_seconds`, where
-        `resources` should define the `num_machines`.
-        """
-        num_machines = self.ctx.inputs.metadata.options.get('resources', {}).get('num_machines', None)
-        max_wallclock_seconds = self.ctx.inputs.metadata.options.get('max_wallclock_seconds', None)
-
-        if num_machines is None or max_wallclock_seconds is None:
-            return self.exit_codes.ERROR_INVALID_INPUT_RESOURCES_UNDERSPECIFIED
-
-        self.set_max_seconds(max_wallclock_seconds)
-
-    def set_max_seconds(self, max_wallclock_seconds):
+    def set_max_seconds(self, max_wallclock_seconds: None):
         """Set the `max_seconds` to a fraction of `max_wallclock_seconds` option to prevent out-of-walltime problems.
 
         :param max_wallclock_seconds: the maximum wallclock time that will be set in the scheduler settings.
@@ -95,6 +81,11 @@ class PhBaseWorkChain(BaseRestartWorkChain):
         If a `restart_calc` has been set in the context, its `remote_folder` will be used as the `parent_folder` input
         for the next calculation and the `restart_mode` is set to `restart`.
         """
+        max_wallclock_seconds = self.ctx.inputs.metadata.options.get('max_wallclock_seconds', None)
+
+        if max_wallclock_seconds is not None and 'max_seconds' not in self.ctx.inputs.parameters['INPUTPH']:
+            self.set_max_seconds(max_wallclock_seconds)
+
         if self.ctx.restart_calc:
             self.ctx.inputs.parameters['INPUTPH']['recover'] = True
             self.ctx.inputs.parent_folder = self.ctx.restart_calc.outputs.remote_folder
--- a/aiida_quantumespresso/workflows/pw/base.py
+++ b/aiida_quantumespresso/workflows/pw/base.py
@@ -76,7 +76,6 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
             cls.validate_parameters,
             cls.validate_kpoints,
             cls.validate_pseudos,
-            cls.validate_resources,
             if_(cls.should_run_init)(
                 cls.validate_init_inputs,
                 cls.run_init,
@@ -99,9 +98,11 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         spec.exit_code(202, 'ERROR_INVALID_INPUT_KPOINTS',
             message='Neither the `kpoints` nor the `kpoints_distance` input was specified.')
         spec.exit_code(203, 'ERROR_INVALID_INPUT_RESOURCES',
-            message='Neither the `options` nor `automatic_parallelization` input was specified.')
+            message='Neither the `options` nor `automatic_parallelization` input was specified. '
+                    'This exit status has been deprecated as the check it corresponded to was incorrect.')
         spec.exit_code(204, 'ERROR_INVALID_INPUT_RESOURCES_UNDERSPECIFIED',
-            message='The `metadata.options` did not specify both `resources.num_machines` and `max_wallclock_seconds`.')
+            message='The `metadata.options` did not specify both `resources.num_machines` and `max_wallclock_seconds`. '
+                    'This exit status has been deprecated as the check it corresponded to was incorrect.')
         spec.exit_code(210, 'ERROR_INVALID_INPUT_AUTOMATIC_PARALLELIZATION_MISSING_KEY',
             message='Required key for `automatic_parallelization` was not specified.')
         spec.exit_code(211, 'ERROR_INVALID_INPUT_AUTOMATIC_PARALLELIZATION_UNRECOGNIZED_KEY',
@@ -282,26 +283,6 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
             self.report(f'{exception}')
             return self.exit_codes.ERROR_INVALID_INPUT_PSEUDO_POTENTIALS
 
-    def validate_resources(self):
-        """Validate the inputs related to the resources.
-
-        One can omit the normally required `options.resources` input for the `PwCalculation`, as long as the input
-        `automatic_parallelization` is specified. If this is not the case, the `metadata.options` should at least
-        contain the options `resources` and `max_wallclock_seconds`, where `resources` should define the `num_machines`.
-        """
-        if 'automatic_parallelization' not in self.inputs and 'options' not in self.ctx.inputs.metadata:
-            return self.exit_codes.ERROR_INVALID_INPUT_RESOURCES
-
-        # If automatic parallelization is not enabled, we better make sure that the options satisfy minimum requirements
-        if 'automatic_parallelization' not in self.inputs:
-            num_machines = self.ctx.inputs.metadata.options.get('resources', {}).get('num_machines', None)
-            max_wallclock_seconds = self.ctx.inputs.metadata.options.get('max_wallclock_seconds', None)
-
-            if num_machines is None or max_wallclock_seconds is None:
-                return self.exit_codes.ERROR_INVALID_INPUT_RESOURCES_UNDERSPECIFIED
-
-            self.set_max_seconds(max_wallclock_seconds)
-
     def set_max_seconds(self, max_wallclock_seconds):
         """Set the `max_seconds` to a fraction of `max_wallclock_seconds` option to prevent out-of-walltime problems.
 
@@ -419,6 +400,11 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         for the next calculation and the `restart_mode` is set to `restart`. Otherwise, no `parent_folder` is used and
         `restart_mode` is set to `from_scratch`.
         """
+        max_wallclock_seconds = self.ctx.inputs.metadata.options.get('max_wallclock_seconds', None)
+
+        if max_wallclock_seconds is not None and 'max_seconds' not in self.ctx.inputs.parameters['CONTROL']:
+            self.set_max_seconds(max_wallclock_seconds)
+
         if self.ctx.restart_calc:
             self.ctx.inputs.parameters['CONTROL']['restart_mode'] = 'restart'
             self.ctx.inputs.parent_folder = self.ctx.restart_calc.outputs.remote_folder
--- a/tests/workflows/ph/test_base.py
+++ b/tests/workflows/ph/test_base.py
@@ -1,36 +1,11 @@
 # -*- coding: utf-8 -*-
 # pylint: disable=no-member,redefined-outer-name
 """Tests for the `PhBaseWorkChain` class."""
-import pytest
-
-from plumpy import ProcessState
-
 from aiida.common import AttributeDict
 from aiida.engine import ProcessHandlerReport
 
 from aiida_quantumespresso.calculations.ph import PhCalculation
 from aiida_quantumespresso.workflows.ph.base import PhBaseWorkChain
-
-
-@pytest.fixture
-def generate_workchain_ph(generate_workchain, generate_inputs_ph, generate_calc_job_node):
-    """Generate an instance of a `PhBaseWorkChain`."""
-
-    def _generate_workchain_ph(exit_code=None):
-        entry_point = 'quantumespresso.ph.base'
-        process = generate_workchain(entry_point, {'ph': generate_inputs_ph()})
-
-        if exit_code is not None:
-            node = generate_calc_job_node()
-            node.set_process_state(ProcessState.FINISHED)
-            node.set_exit_status(exit_code.status)
-
-            process.ctx.iteration = 1
-            process.ctx.children = [node]
-
-        return process
-
-    return _generate_workchain_ph
 
 
 def test_setup(generate_workchain_ph):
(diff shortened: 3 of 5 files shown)


mbercx:
Thanks @sphuber, will review later, but just wanted to mention that I recognise the issue with the failing test:


```Python
self = frozendict({'q2r': {'code': <Code: Remote code 'test.quantumespresso.q2r' on localhost-test, pk: 170, uuid: 96e8b36c-9...ored)>, 'metadata': {'options': {'resources': {'num_machines': 1}, 'max_wallclock_seconds': 1800, 'withmpi': False}}}})
args = ('_initialised', True), kwargs = {}

    def _immutable(self, *args, **kwargs):
        r"""
        Function for not implemented method since the object is immutable
        """
    
>       raise AttributeError(f"'{self.__class__.__name__}' object is read-only")
E       AttributeError: 'AttributesFrozendict' object is read-only
```

This issue was also raised by @asle85 on slack. She noticed that her `frozendict` version had been upgraded to `2.0.2`, while it was `1.2` before. Not sure what is causing the upgrade of this dependency, but reverting back to `1.2` fixed the problem for her. Will do a bit more digging.

sphuber:
The problem is that for Python 3.6, `plumpy==0.15.1` gets installed as that is the last version to support that Python version. That `plumpy` version still uses the `frozendict` library to implement its `FrozenDict` class, which is mostly used as subclass for the `AttributesFrozenDict` which in turns is used to wrap the inputs of a process in before returning it in `self.inputs`. This is to make sure they cannot be mutated during the process running code, such as in a workchain, which would be bad. Anyway, in `frozendict==1.2` the `__setattr__` method was callable and this is actually called in `plumpy` here:

https://github.com/aiidateam/plumpy/blob/7be3b8efaef7350ee32aae4be966db32f95156ec/plumpy/utils.py#L110

However, in `frozendict==2.0` they added this line:
https://github.com/Marco-Sulla/python-frozendict/blob/e2b59b190ee77c3d7bdb1f2cd772c0b531694a4c/frozendict/core.py#L160

So now that `self._initialized = True` call in the constructor of `AttributesFrozenDict` causes the exception. Really the only solution is to release a `v0.15.2` for plumpy that restricts the version of `frozendict<2.0`. Otherwise all new installations with Python 3.6 will be broken meaning all versions `aiida-core<1.6` will be broken. I am not sure if we want to have that. I will open an issue on `plumpy`.

sphuber:
All clear now capt'n! @mbercx

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -100,7 +99,8 @@ def define(cls, spec):
         spec.exit_code(203, 'ERROR_INVALID_INPUT_RESOURCES',
             message='Neither the `options` nor `automatic_parallelization` input was specified.')
         spec.exit_code(204, 'ERROR_INVALID_INPUT_RESOURCES_UNDERSPECIFIED',
-            message='The `metadata.options` did not specify both `resources.num_machines` and `max_wallclock_seconds`.')
+            message='The `metadata.options` did not specify both `resources.num_machines` and `max_wallclock_seconds`. '
+                    'This exit status has been deprecated as the check it corresponded too was incorrect.')
```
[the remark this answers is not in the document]
mbercx: Also, shouldn't this exit code also be deprecated for the `PhBaseWorkChain`?

aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -95,6 +80,8 @@ def prepare_process(self):
         If a `restart_calc` has been set in the context, its `remote_folder` will be used as the `parent_folder` input
         for the next calculation and the `restart_mode` is set to `restart`.
         """
+        self.set_max_seconds(self.ctx.inputs.metadata.options['max_wallclock_seconds'])
```
mbercx, line 83: Won't this fail now in case the user hasn't specified the `max_wallclock_seconds`?

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -272,26 +272,6 @@ def validate_pseudos(self):
...
 
-    def validate_resources(self):
-        """Validate the inputs related to the resources.
-
-        One can omit the normally required `options.resources` input for the `PwCalculation`, as long as the input
-        `automatic_parallelization` is specified. If this is not the case, the `metadata.options` should at least
-        contain the options `resources` and `max_wallclock_seconds`, where `resources` should define the `num_machines`.
-        """
-        if 'automatic_parallelization' not in self.inputs and 'options' not in self.ctx.inputs.metadata:
-            return self.exit_codes.ERROR_INVALID_INPUT_RESOURCES
```
mbercx, line 283: Isn't this exit code deprecated now as well?

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -409,6 +389,8 @@ def prepare_process(self):
         for the next calculation and the `restart_mode` is set to `restart`. Otherwise, no `parent_folder` is used and
         `restart_mode` is set to `from_scratch`.
         """
+        self.set_max_seconds(self.ctx.inputs.metadata.options['max_wallclock_seconds'])
```
mbercx, line 392: Same issue as for `PhBaseWorkChain`.

aiida_quantumespresso/workflows/ph/base.py
mbercx: Just tested this for the `PwBaseWorkChain`:
```Python
$ verdi process report 792
2021-05-06 07:31:54 [217 | REPORT]: [792|PwBaseWorkChain|on_except]: Traceback (most recent call last):
  File "/home/mbercx/.virtualenvs/aiida-dev/lib/python3.8/site-packages/plumpy/process_states.py", line 230, in execute
    result = self.run_fn(*self.args, **self.kwargs)
  File "/home/mbercx/envs/aiida-dev/code/aiida-core/aiida/engine/processes/workchains/workchain.py", line 214, in _do_step
    finished, stepper_result = self._stepper.step()
  File "/home/mbercx/.virtualenvs/aiida-dev/lib/python3.8/site-packages/plumpy/workchains.py", line 299, in step
    finished, result = self._child_stepper.step()
  File "/home/mbercx/.virtualenvs/aiida-dev/lib/python3.8/site-packages/plumpy/workchains.py", line 532, in step
    finished, result = self._child_stepper.step()
  File "/home/mbercx/.virtualenvs/aiida-dev/lib/python3.8/site-packages/plumpy/workchains.py", line 299, in step
    finished, result = self._child_stepper.step()
  File "/home/mbercx/.virtualenvs/aiida-dev/lib/python3.8/site-packages/plumpy/workchains.py", line 250, in step
    return True, self._fn(self._workchain)
  File "/home/mbercx/envs/aiida-dev/code/aiida-quantumespresso/aiida_quantumespresso/workflows/pw/base.py", line 392, in prepare_process
    self.set_max_seconds(self.ctx.inputs.metadata.options['max_wallclock_seconds'])
KeyError: 'max_wallclock_seconds'
```

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -272,26 +272,6 @@ def validate_pseudos(self):
...
-
-        One can omit the normally required `options.resources` input for the `PwCalculation`, as long as the input
-        `automatic_parallelization` is specified. If this is not the case, the `metadata.options` should at least
-        contain the options `resources` and `max_wallclock_seconds`, where `resources` should define the `num_machines`.
-        """
-        if 'automatic_parallelization' not in self.inputs and 'options' not in self.ctx.inputs.metadata:
-            return self.exit_codes.ERROR_INVALID_INPUT_RESOURCES
-
-        # If automatic parallelization is not enabled, we better make sure that the options satisfy minimum requirements
-        if 'automatic_parallelization' not in self.inputs:
```
mbercx, line 286: Looking at the `automatic_parallelization` some more, won't the fact that `num_machines` can't be defined for some schedulers cause problems here? E.g. this line:

https://github.com/aiidateam/aiida-quantumespresso/blob/33cb11e3c9a1f382fac69c15d74a2b4f6d331d4c/aiida_quantumespresso/workflows/pw/base.py#L347

In fact, after reading https://github.com/aiidateam/aiida-quantumespresso/commit/70d97b1cf40e3a6a4d9fe8bf0fccba09a3d4d102, does this feature still work for recent  QE versions? Maybe it would be a good idea to fix this in this PR.

... Just kidding on that last part. 😉

mbercx (changes requested):
Thanks @sphuber! Gave the PR a pass and left some comments. 

Now that we are touching this piece of the code, I'm wondering if in this PR we should perhaps also fix the fact that in case the user sets `max_seconds` in the input parameters, this is silently ignored by both `PwBaseWorkChain` and `PhBaseWorkChain`, since:

https://github.com/aiidateam/aiida-quantumespresso/blob/33cb11e3c9a1f382fac69c15d74a2b4f6d331d4c/aiida_quantumespresso/workflows/pw/base.py#L295-L302

Wouldn't it be better to _only_ set `max_seconds` based on `delta_factor_max_seconds` in case the user _hasn't_ specified `max_seconds` in the input parameters, i.e. only call the `set_max_seconds` method if `'max_seconds' not in self.ctx.inputs.parameters['CONTROL']`?

aiida_quantumespresso/workflows/ph/base.py
sphuber: Good point. I could have sworn that the `max_wallclock_seconds` option was required but the `CalcJob` spec says otherwise. Maybe all scheduler implementations require it though?

aiida_quantumespresso/workflows/pw/base.py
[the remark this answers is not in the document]
sphuber: Honestly, I don't know. This is unknown terrain as for us it hasn't come up before. I did think about this way back when introducing the concept of exit codes and decided to store the actual exit message on the node. So each node has the `exit_message` attribute which contains the message of the exit code as it was defined at runtime, in addition to the `exit_status` which stores the corresponding integer. So even if we change the implementation, existing nodes will still reflect the original message through things like `verdi process status/list` etc. Still, the integer remains the same, so I am not sure if we should repurpose it, because if you then perform a query on it, you might get processes that failed for a different reason but it seems to be the same one (based off of just the status and ignoring the message).

So in short, I am not sure. I don't think we should repurpose the exit code, so better to leave it in because if we remove it, at some point we forget it existed and might use the integer again. Then again, without a deprecation pathway they might start to accumulate over time and get in the way. Open to ideas and suggestions

aiida_quantumespresso/workflows/ph/base.py
sphuber: Just checked and it is not required, so this should indeed be fixed.

aiida_quantumespresso/workflows/pw/base.py
sphuber: > .. Just kidding on that last part. wink

Haha I was going to say, you are giving me a "cookie of my own dough" ;)

There is already an open issue for the automatic parallelization because I doubt it has been working at all for a long time.  There is another problem that I realized. Originally, since the `metadata.options.resources` were not validated on the base `CalcJob` you could submit with them missing and so the `PwBaseWorkChain` could set them instead if automatic parallelization was selected. This is why this validate resource function was even there. Now though, the `PwCalculation` requires it so even if the user _were_ to want to use automatic parallelization, they would still have to define acceptable resources, even though they may be overwritten later. Not the worst problem, since there is a workaround, but also not the nicest. We could replace the validator of the `CalcJob` class but that would be tricky because the resource validation is done in the top level namespace validator and we only want to get rid of part that, meaning we would have to copy paste all the rest of the validator, which would be a nightmare to maintain.

Anyway, I am not to worried about the AP feature here because it was always experimental and don't think it is critical as long as the normal functionality for the base works.

aiida_quantumespresso/workflows/pw/base.py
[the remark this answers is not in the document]
mbercx: Hmm, good points, thanks for the explanation. Just off the top of my head I would consider adding a `deprecated_exit_status` class variable to the process classes where you can keep track of deprecated exit codes, so the class can raise in case a new exit code is defined on the spec with an already deprecated `exit_status`. This would definitely be shorter than keeping them all in the `define` method.

Maybe it would be best to open a discussion on the `aiida-core` discussions board?

mbercx:
While browsing through the issues, I also found #528. Does this PR fix that one as well?

aiida_quantumespresso/workflows/pw/base.py
mbercx: Alright, we'll have a closer look at the automatic parallelisation some other time then. 😀

aiida_quantumespresso/workflows/pw/base.py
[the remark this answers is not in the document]
sphuber: > Maybe it would be best to open a discussion on the aiida-core discussions board?

yes, let's

sphuber:
> While browsing through the issues, I also found #528. Does this PR fix that one as well?

The original issue should have been fixed indirectly by newer `aiida-core` versions that include resource validation in the `CalcJob` spec. I think the direct problem is that the `computer` configured in the tutorial does not define a default number MPI procs and the script didn't specify explicitly, hence the error. The comment by the other user would actually be fixed by this PR and is the issue described in #682  so I think we can close both issues.

aiida_quantumespresso/workflows/pw/base.py
sphuber: yes

sphuber:
@mbercx thanks for the review. Have addressed everything except the question of `max_seconds` always being overridden. Would be happy to add it here but what do we need additional validation? What if it is equal or larger than the `max_wallclock_seconds`? Do we just set whatever the user asks?

mbercx:
> @mbercx thanks for the review. Have addressed everything except the question of `max_seconds` always being overridden. Would be happy to add it here but what do we need additional validation? What if it is equal or larger than the `max_wallclock_seconds`? Do we just set whatever the user asks?

You _could_ add extra validation, but perhaps this isn't necessary. If the user specifies both `max_wallclock_seconds` and `max_seconds` and sets `max_seconds > max_wallclock_seconds`, isn't that sort of on them? 😅 

It's strange that the `max_wallclock_seconds` isn't a required input though... What happens if you try to run without it?

sphuber:
> You could add extra validation, but perhaps this isn't necessary.

Of course I could, but that was exactly the question: how do we want the validation to behave. Maybe we indeed just say that we only set `max_seconds` if not defined explicitly and if it is, then it is up to the user to make sure it makes sense. Setting it too low or too high should lead to the calculation prematurely exiting or the value even being ignored. Either way, not really our problem. Fine I will add the simplest version here then :)

> It's strange that the `max_wallclock_seconds` isn't a required input though... What happens if you try to run without it?

It probably isn't required on the `CalcJob` because it would most likely be again scheduler dependent. I doubt that it would work for a standard SLURM submission, as that would almost certainly require a max walltime, but for the `DirectScheduler` I am not so sure. I actually think that for that scheduler it doesn't even work if you _do_ specify it and the job runs forever.

sphuber:
@mbercx  [All done Sire!](https://www.youtube.com/watch?v=eO6OaVnbqaY)

aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -95,6 +81,11 @@ def prepare_process(self):
         If a `restart_calc` has been set in the context, its `remote_folder` will be used as the `parent_folder` input
         for the next calculation and the `restart_mode` is set to `restart`.
         """
+        max_seconds = self.ctx.inputs.metadata.options.get('max_wallclock_seconds', None)
+
+        if max_seconds is not None and 'max_seconds' not in self.inputs.parameters['INPUTPH']:
```
mbercx, line 86:
```suggestion
        if max_seconds is not None and 'max_seconds' not in self.ctx.inputs.parameters['INPUTPH']:
```

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -419,6 +400,11 @@ def prepare_process(self):
         for the next calculation and the `restart_mode` is set to `restart`. Otherwise, no `parent_folder` is used and
         `restart_mode` is set to `from_scratch`.
         """
+        max_seconds = self.ctx.inputs.metadata.options.get('max_wallclock_seconds', None)
+
+        if max_seconds is not None and 'max_seconds' not in self.inputs.parameters['CONTROL']:
```
mbercx, line 405:
```suggestion
        if max_seconds is not None and 'max_seconds' not in self.ctx.inputs.parameters['CONTROL']:
```

mbercx (changes requested):
It's a good thing at least one of us test runs your implementations. 😏

sphuber:
> It's a good thing at least one of us test runs your implementations. smirk

That's the thing, you are so diligent at double-checking my shitty code, that it is making me even more sloppy. So really this is all your fault

mbercx:
> That's the thing, you are so diligent at double-checking my shitty code, that it is making me even more sloppy. So really this is all your fault

Haha, that comment turned my remark around on me so fast it gave me whiplash. 😅

sphuber:
> Haha, that comment turned my remark around on me so fast it gave me whiplash. sweat_smile



She's all fixed up boss! With tests and all

tests/workflows/pw/test_base.py
```diff
@@ -137,3 +137,32 @@ def test_sanity_check_no_bands(generate_workchain_pw):
 
     calculation = process.ctx.children[-1]
     assert process.sanity_check_insufficient_bands(calculation) is None
+
+
+def test_set_max_seconds(generate_workchain_pw):
+    """Test that `max_seconds` get sets in the parameters based on `max_wallclock_seconds` unless already set."""
+    inputs = generate_workchain_pw(return_inputs=True)
```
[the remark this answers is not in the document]
sphuber: because the `generate_inputs_pw` does not return all the minimally required inputs needed for the workchain. If you look at `generate_workchain_pw` it needs to add the inputs to a namespace, and pop an input. I don't want to have to add these lines in every test.

tests/workflows/ph/test_base.py
```diff
@@ -84,3 +59,32 @@ def test_handle_convergence_not_achieved(generate_workchain_ph):
 
     result = process.inspect_process()
     assert result.status == 0
+
+
+def test_set_max_seconds(generate_workchain_ph):
+    """Test that `max_seconds` get sets in the parameters based on `max_wallclock_seconds` unless already set."""
+    inputs = generate_workchain_ph(return_inputs=True)
```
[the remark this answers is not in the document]
sphuber: Problem described for the comment on `pw` is less pronounced for `ph` but I want to keep it consistent and the input definition for the `PhBasWorkChain` may become more complicated in the future

aiida_quantumespresso/workflows/pw/base.py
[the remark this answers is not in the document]
mbercx: https://github.com/aiidateam/aiida-core/discussions/4933

PR #684, opened 2021-05-06
Protocols: Add support for `CutoffsPseudoPotentialFamily`

mbercx:
Fixes #684 

In the `aiida-pseudo` release v1.6.1, a generic pseudo family class was added
which has the features of the `RecommendedCutoffsMixin` class. This
allows the user to install pseudo families other than the supported SSSP
and pseudo-dojo configurations and set recommended cutoffs for specified
stringencies.

Here we add support for the `CutoffsPseudoPotentialFamily` class to the
`get_builder_from_protocol` method of the `PwBaseWorkChain`. Since these
pseudo families aren't configured with recommended cutoffs by default,
we catch any ValueErrors raised by the `get_recommended_cutoffs` method
and add reraise them with a message that clearly indicates that
the failure is related to obtaining the recommended cutoffs.

### DIFF
--- a/setup.json
+++ b/setup.json
@@ -92,7 +92,7 @@
     },
     "install_requires": [
         "aiida_core[atomic_tools]~=1.4,>=1.4.4",
-        "aiida-pseudo~=0.6,>=0.6.1",
+        "aiida-pseudo~=0.6.1",
         "jsonschema",
         "packaging",
         "qe-tools~=2.0rc1",
--- a/setup.json
+++ b/setup.json
@@ -92,7 +92,7 @@
     },
     "install_requires": [
         "aiida_core[atomic_tools]~=1.4,>=1.4.4",
-        "aiida-pseudo~=0.6.0",
+        "aiida-pseudo~=0.6,>=0.6.1",
         "jsonschema",
         "packaging",
         "qe-tools~=2.0rc1",
--- a/aiida_quantumespresso/workflows/pw/base.py
+++ b/aiida_quantumespresso/workflows/pw/base.py
@@ -19,6 +19,7 @@ from ..protocols.utils import ProtocolMixin
 PwCalculation = CalculationFactory('quantumespresso.pw')
 SsspFamily = GroupFactory('pseudo.family.sssp')
 PseudoDojoFamily = GroupFactory('pseudo.family.pseudo_dojo')
+CutoffsPseudoPotentialFamily = GroupFactory('pseudo.family.cutoffs')
 
 
 def validate_pseudo_family(value, _):
@@ -167,7 +168,7 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         natoms = len(structure.sites)
 
         try:
-            pseudo_set = (PseudoDojoFamily, SsspFamily)
+            pseudo_set = (PseudoDojoFamily, SsspFamily, CutoffsPseudoPotentialFamily)
             pseudo_family = orm.QueryBuilder().append(pseudo_set, filters={'label': pseudo_family}).one()[0]
         except exceptions.NotExistent as exception:
             raise ValueError(
@@ -175,7 +176,12 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
                 'install it.'
             ) from exception
 
-        cutoff_wfc, cutoff_rho = pseudo_family.get_recommended_cutoffs(structure=structure, unit='Ry')
+        try:
+            cutoff_wfc, cutoff_rho = pseudo_family.get_recommended_cutoffs(structure=structure, unit='Ry')
+        except ValueError as exception:
+            raise ValueError(
+                f'failed to obtain recommended cutoffs for pseudo family `{pseudo_family}`: {exception}'
+            ) from exception
 
         parameters = inputs['pw']['parameters']
         parameters['CONTROL']['etot_conv_thr'] = natoms * meta_parameters['etot_conv_thr_per_atom']


mbercx:
@sphuber I figured a re-raise of the `ValueError` would make the final error message more clear to the user, what do you think?

```Python
KeyError                                  Traceback (most recent call last)
~/envs/aiida-dev/code/aiida-core/aiida/orm/implementation/entities.py in get_extra(self, key)
    322         try:
--> 323             return self._dbmodel.extras[key]
    324         except KeyError as exception:

KeyError: '_default_stringency'

The above exception was the direct cause of the following exception:

AttributeError                            Traceback (most recent call last)
~/.virtualenvs/aiida-dev/lib/python3.8/site-packages/aiida_pseudo/groups/mixins/cutoffs.py in get_default_stringency(self)
    114         try:
--> 115             return self.get_extra(self._key_default_stringency)
    116         except AttributeError as exception:

~/envs/aiida-dev/code/aiida-core/aiida/orm/entities.py in get_extra(self, key, default)
    503         try:
--> 504             extra = self.backend_entity.get_extra(key)
    505         except AttributeError:

~/envs/aiida-dev/code/aiida-core/aiida/orm/implementation/entities.py in get_extra(self, key)
    324         except KeyError as exception:
--> 325             raise AttributeError(f'extra `{exception}` does not exist') from exception
    326 

AttributeError: extra `'_default_stringency'` does not exist

The above exception was the direct cause of the following exception:

ValueError                                Traceback (most recent call last)
~/envs/aiida-dev/code/aiida-quantumespresso/aiida_quantumespresso/workflows/pw/base.py in get_builder_from_protocol(cls, code, structure, protocol, overrides, electronic_type, spin_type, initial_magnetic_moments, **_)
    179         try:
--> 180             cutoff_wfc, cutoff_rho = pseudo_family.get_recommended_cutoffs(structure=structure, unit='Ry')
    181         except ValueError as exception:

~/.virtualenvs/aiida-dev/lib/python3.8/site-packages/aiida_pseudo/groups/mixins/cutoffs.py in get_recommended_cutoffs(self, elements, structure, stringency, unit)
    297         cutoffs_rho = []
--> 298         cutoffs = self.get_cutoffs(stringency)
    299 

~/.virtualenvs/aiida-dev/lib/python3.8/site-packages/aiida_pseudo/groups/mixins/cutoffs.py in get_cutoffs(self, stringency)
    189         """
--> 190         stringency = stringency or self.get_default_stringency()
    191 

~/.virtualenvs/aiida-dev/lib/python3.8/site-packages/aiida_pseudo/groups/mixins/cutoffs.py in get_default_stringency(self)
    116         except AttributeError as exception:
--> 117             raise ValueError('no default stringency has been defined.') from exception
    118 

ValueError: no default stringency has been defined.

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
<ipython-input-8-481d6f6b1740> in <module>
----> 1 builder = PwBaseWorkChain.get_builder_from_protocol(
      2     code=orm.load_code(1),
      3     structure=orm.load_node(590),
      4     overrides={'pseudo_family': 'SSSP/test/PBE/efficiency'}
      5 )

~/envs/aiida-dev/code/aiida-quantumespresso/aiida_quantumespresso/workflows/pw/base.py in get_builder_from_protocol(cls, code, structure, protocol, overrides, electronic_type, spin_type, initial_magnetic_moments, **_)
    180             cutoff_wfc, cutoff_rho = pseudo_family.get_recommended_cutoffs(structure=structure, unit='Ry')
    181         except ValueError as exception:
--> 182             raise ValueError(f'failed to obtain recommended cutoffs for pseudo family `{pseudo_family}`: {exception}.')
    183 
    184         parameters = inputs['pw']['parameters']

ValueError: failed to obtain recommended cutoffs for pseudo family `CutoffsPseudoPotentialFamily<SSSP/test/PBE/efficiency>`: no default stringency has been defined.
```

setup.json
```diff
@@ -92,7 +92,7 @@
     },
     "install_requires": [
         "aiida_core[atomic_tools]~=1.4,>=1.4.4",
-        "aiida-pseudo~=0.6.0",
+        "aiida-pseudo~=0.6,>=0.6.1",
```
sphuber, line 95: these are not really the same, now it will also allow `aiida-pseudo==0.7` and higher. whereas before it would only install the patches of 0.6 series. Instead you could do, 
```suggestion
        "aiida-pseudo~=0.6.1",
```
which should achieve what you intended I think

sphuber (changes requested):
Thanks @mbercx looks good, just a point on the requirement. I think that the `from exception` is fine, even though i think that most users actually get confused by more stack trace. In my experience, what they look at is the actual final message, at most. But mostly exceptions are scary and seem like everything is broken, as opposed to some data is just missing or not well configured. But for me would be fine to keep it

setup.json
mbercx: Hmm, I see, thanks! Will this also upgrade `aiida-pseudo` from `0.6.0` to `0.6.1` when installing the new `aiida-quantumespresso` release?

mbercx:
> Thanks @mbercx looks good, just a point on the requirement. I think that the `from exception` is fine, even though i think that most users actually get confused by more stack trace. In my experience, what they look at is the actual final message, at most. 

I agree, that's why I wanted to catch and re-raise, so the final message clearly indicates the problem with the pseudo family.

setup.json
sphuber: yes, I `~=0.6.1` literally is identical to `>= 0.6.1, == 0.6.*`. See [PEP 440](https://www.python.org/dev/peps/pep-0440/#compatible-release)

sphuber (approved):

PR #686, opened 2021-05-07
Fix/ph parser symmetry labels

mkotiuga:
Adding symmetry label parsing from the std out of a ph.x run. The symmetry labels are printed after the mode frequencies for each computed q-point. 

A dictionary for each q-point (determined by the q-points listed in the header of the std out) has been added to the parsed 'output_parameters' as well as a key-value pair containing the number of q-points found and parsed in the std out, 'num_q_found'. As  PhCalculation can run out of wall time (thus not completing all the q-points) or be a restart (thus missing the data from the q-points calculated in a previous run), the dictionaries for each q_point is created with the key "q_point_i" containing the coordinated of the q_point from the header of std out. If the results of this q-point are in this std out they are parsed and added to 'q_point_i". 

For a PhBaseRestartWorkChain, which can call multiple iterations of PhCalculation, the results of the parsed q_points_i need to be merged. This is done by calcfunction added to PhBaseRestartWorkChain and is called after a check that all the instances of PhCalculation have the same number of q-points to calculate and that this is the number of q_points found and parsed across all the called instances of PhCalculation (i.e. the sum of num_q_found for all called  PhCalculation == number_of_qpoints". The merged results are output as "merged_output_parameters". As the data output to DYN_MAT is cumulative upon multiple restarts of a PhCalculation, the final merge is done by taking the output of the last called PhCalculation and adding in the q_points_i dictionaries from the other called instances of PhCalculation. The values for 'num_q_found', 'number_of_irr_representations_for_each_q', 'wall_time', 'wall_time_seconds' also need to be merged (the rest of the data can be simply copied from the last instance of PhCalculation). 

Possible extensions:
- 'number_of_irr_representations_for_each_q' could be added for each q-point to the respective q_point_i dict
- Add the parsed q_point_i data with the dynamical_matrix_i data. This should be done in the calcfunction  
- The 'merged_output_parameters' could replace the current 'output_parameters' which are just forwarded from the last call of PhCalculation by PhBaseRestartWorkChain

Issues with current implentation  - No error 'proper' handling.

### DIFF
--- a/aiida_quantumespresso/parsers/parse_raw/ph.py
+++ b/aiida_quantumespresso/parsers/parse_raw/ph.py
@@ -204,6 +204,20 @@ def parse_ph_text_output(lines, logs):
                     logs.warning.append('Number q-points found several times with different values')
                 else:
                     parsed_data['number_of_qpoints'] = num_qpoints
+                    for iq in range(num_qpoints): #parse the qpoints - mkotiuga
+                        q_ind = int(lines[count+iq+2].split()[0])
+                        qpt_str = lines[count+iq+2].split()[1:4]
+                        qpt = [float(i) for i in qpt_str]
+                        if f'q_point_{q_ind}' in list(parsed_data.keys()):
+                            if parsed_data[f'q_point_{q_ind}']['q_point'] == qpt:
+                                pass
+                            else:
+                                logs.warning.append('Unexpected q-point found in the header')
+                                #TODO exception
+                        else:
+                            parsed_data[f'q_point_{q_ind}'] = {'q_point':qpt}
+                            #parse the qpoints done- mkotiuga
+
             except Exception:
                 logs.warning.append('Error while parsing number of q points.')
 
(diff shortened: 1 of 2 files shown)


mbercx:
Thanks a lot, @mkotiuga! I have little experience running the `ph.x` code I'm afraid. I hope to remedy this at the QE school soon, so if @sphuber or @giovannipizzi haven't reviewed by then, I'll definitely take a look asap.

One general comment regarding parsing is that we're trying to move towards parsing as much as possible of the outputs from the `data-file-schema.xml` using the XSD schemas provided by QE. Do you know if this output is also present in this file?

mkotiuga:
Thanks @mbercx! I think the short answer is no. Without changes to the output of ph.x this info is only in the stdout. 

Looking at one of my own runs as an example: the xml output of ph is a number of files (not liek pw.x which puts everything into data-file-schema.xml). These files have the status of the run (control_ph.xml, status_run.xml), pieces of the dynamical matrix (dynmat.*.*.xml), the tensor quantities computed (tensors.xml) and the displacement patterns associated with each q-point (patterns.*.xml). The symmetry labels I am parsing are present in "patterns.*.xml" but not the eigenenergy (i.e. frequency) they are associated with - I have only found this in the stdout. 

This could be fixed at the ph.x level and then aiida could just parse the xml. A related aside, aiida is parsing the frequencies from the DYN_MAT files, not the xml files, and as far as I have looked so far, I have not seen the frequencies in the xml files either. So if there are changes at the level of ph.x, some other implementations could be: 

1) "most xml consistent and most work": either the frequencies and symmetry labels can be saved in xml and aiida parses only xml (a lot of the aiida ph parses would be redone); or,

2) "least work on ph and aiida sides, not using xml": The symmetry of modes could be printed by ph.x in the DYN_MAT files along their frequency (the piece I have written would then be replaced by a small tweak to the dynamical matrices parsing in aiida)

mbercx:
Thanks for the response, @mkotiuga! I'll look into opening an issue on the Quantum ESPRESSO GitLab repo. In general we are pushing to have more output into XML formatted files, but it will probably take some time for this to be implemented. I'm not sure how the _dynmat_ and _patterns_ xml files work, but for the `data-file-schema.xml` we have XSD files that are version dependent, so we can easily keep supporting the outputs of different versions of QE without things breaking when there are changes or writing complex parsing code that deals with these differences.

mbercx:
Notes from a short meeting with @mkotiuga:

1. In my view it makes more sense to just have one `output_parameters` instead of having _also_ `merged_output_parameters`, since the latter is just a superset of the former.
2. Would it make sense to just track the symmetry labels in the context instead of merging them at the end with the `create_merged_outputs` step?

sphuber:
superseded by #818

PR #690, opened 2021-05-10
Release `v3.4.2`

mbercx:


CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
...
+### Improvements
+- `ProtocolMixin`: increase flexibility and usability for other packages [[#678]](https://github.com/aiidateam/aiida-quantumespresso/pull/678)
+- Protocols: Add support for `CutoffsPseudoPotentialFamily` [[#684]](https://github.com/aiidateam/aiida-quantumespresso/pull/684)
+
+### Bug Fixes
+- Protocols: Add usage of metadata and parallelisation `overrides` [[#652]](https://github.com/aiidateam/aiida-quantumespresso/pull/652)
+- `PwBaseWorkChain`: fix bug in `validate_resources` validator [[#683]](https://github.com/aiidateam/aiida-quantumespresso/pull/683)
+
+### Deprecation
+- `PwBandStructureWorkChain`: Deprecate and fix [[#688]](https://github.com/aiidateam/aiida-quantumespresso/pull/688)
```
sphuber, line 15: Maybe we could have added here that we suggest to use the `PwBandsWorkChain` instead, especially with the `get_builder_from_protocol` functionality.

sphuber (commented):
Just a minor suggestion for the changelog, but not necessary. If you temporarily lift the setting that new commits invalidate reviews (in branch settings) you can update it and push and then merge

mbercx:
Hmm, same issue as with the nightly build. Will have a look.

sphuber:
I know what is probably happening. With the stream, the seeker position is not being properly reset after it is consumed. Which means that whoever tries to read it second will read an empty string and so the md5 is different.

mbercx:
Can confirm that this fixes the tests! 👌  Will you open PR or shall I?

sphuber:
> Can confirm that this fixes the tests! ok_hand Will you open PR or shall I?

Well that is good in a way, but I am now actually confused as to why it does. I cannot reproduce this locally and I don't see who is using the stream before it is passed to `set_file` and we already _do_ reset it after the `set_file` super call. And why wasn't this caught by the tests then?

sphuber:
Got it. The problem is not with `aiida-pseudo` but with `aiida-quantumespresso`. The `generate_upf_data` fixture passes `handle` as fileobject to the `UpfData` constructor`, but that fails to reset the seeker position. It _flushed_ the content to disk, which is pointless since we are passing in the stream, but the stream will be at its end. So simply running `handle.seek(0)` before calling `UpfData(handle)` will fix the problem.

mbercx:
Indeed, my bad here as well, my example above wasn't entirely correct. I actually just added this quickly to a code snippet I already had in a notebook where I tested the `get_or_create` method:

```
with pathlib.Path('W_ONCV_PBE-1.1.upf').open('rb') as file:
    upf3 = UpfData.get_or_create(source=file)
    upf4 = UpfData(file=file)
```

Obviously this fails because the byte stream is already exhausted. 🤦

sphuber:
Is it possible that you cherry-picked commit `a969a48 `? Because the relevant PR hadn't been merged yet. This won't work I'm afraid, since we squash merge and so that commit got a different hash on `develop`, namely `79798c8e86c06608b7008d38308a11e95609cf35`. You should remove the cherry-picked commit and then rebase on `develop` again. Then we can merge this

mbercx:
> Is it possible that you cherry-picked commit [a969a48](https://github.com/aiidateam/aiida-quantumespresso/commit/a969a485c793a855d7901e2f594b945350b98ee1) ? Because the relevant PR hadn't been merged yet. This won't work I'm afraid, since we squash merge and so that commit got a different hash on `develop`, namely [79798c8](https://github.com/aiidateam/aiida-quantumespresso/commit/79798c8e86c06608b7008d38308a11e95609cf35). You should remove the cherry-picked commit and then rebase on `develop` again. Then we can merge this

I see, thanks. I didn't think you'd squash merge since there was only one commit in the PR, so I rebased on that commit. Will fix 👍

sphuber:
> I see, thanks. I didn't think you'd squash merge since there was only one commit in the PR, so I rebased on that commit. Will fix +1

We still need to squash though, because for some reason the Github interface doesn't allow fast forward merges without merge commit. So either we need to accept the merge commit (definite :-1: ) or we squash merge (rebase merge also creates new commits so that doesn't help either). I could have done the merge on the CLI with a fast-forward and then it would have worked but hadn't noticed that you had already cherry-picked.

sphuber (approved):
Choo-choo full-steam ahead :steam_locomotive:

mbercx:
> > I see, thanks. I didn't think you'd squash merge since there was only one commit in the PR, so I rebased on that commit. Will fix +1
> 
> We still need to squash though, because for some reason the Github interface doesn't allow fast forward merges without merge commit. So either we need to accept the merge commit (definite 👎 ) or we squash merge (rebase merge also creates new commits so that doesn't help either). I could have done the merge on the CLI with a fast-forward and then it would have worked but hadn't noticed that you had already cherry-picked.

Right, I see! definite no-no on the merge commit, will be more patient in the future. ;)

PR #693, opened 2021-05-19
Fixtures: fix bug in the `generate_upf_data` fixture

sphuber:
The fixture is constructing a `UpfData` by writing the content of a
pseudo in UPF format to a temporary file but then passing the stream to
the constructor. However, the stream was just written to and so the
pointer will be at the stream which will cause the constructor of the
node to read an empty string of bytes.

Since we are using a stream in memory, we don't actually need a file on
disk at all, so the temporary file is replaced by a simple `io.BytesIO`
stream which solves the problem.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,6 +2,7 @@
 # pylint: disable=redefined-outer-name,too-many-statements
 """Initialise a text database and profile for pytest."""
 import collections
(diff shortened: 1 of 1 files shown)


mbercx (approved):
Muchas gracias, @sphuber!

PR #695, opened 2021-05-20
New error handling for SCF convergence edge cases

### LINKED ISSUE #694: Allowing electronic non-convergence if `electron_maxstep = 0` in `PwCalculation`
To start, a quote from Paolo Giannozzi:

> QA Engineer walks into a bar. Orders a beer. Orders 0 beers. Orders 999999999 beers. Orders a lizard. Orders -1 beers. Orders a sfdeljknesv.

For my research, I've ended up hacking `pw.x` so that it handles the following input better:
```
&CONTROL
    calculation = 'scf'
/
&ELECTRONS
    electron_maxstep = 0
    scf_must_converge = .false.
/
...
```

In current versions, this input will lead to the Fermi level and occupations not being written out to `data-file-schema.xml`, but in my hacked version it does write these.

Soo.... I can use `dos.x` for Fermi-centered DOS 😄 .

However, `aiida-quantumespresso`, ever vigilant, still finds the pesky `convergence NOT achieved` in the output and the PWscf parser returns a `401`, which causes the `PwBaseWorkChain` and eventually the `PdosWorkChain` to pull the fire alarm.

I propose to put a check in the PWscf parser so that if `electron_maxstep == 0` or `not scf_must_converge`, the `401` exit status is not returned.

I'm not sure if this would cause any unintended side effects (nausea, vomiting, diarrhea, etc.), so I'd be happy to hear any feedback before I do a tiny PR.

zooks97:
Fixes #694

### DIFF
--- a/aiida_quantumespresso/calculations/pw.py
+++ b/aiida_quantumespresso/calculations/pw.py
@@ -145,6 +145,11 @@ class PwCalculation(BasePwCpInputGenerator):
             message='The electronic minimization cycle did not reach self-consistency.')
         spec.exit_code(541, 'ERROR_SYMMETRY_NON_ORTHOGONAL_OPERATION',
             message='The variable cell optimization broke the symmetry of the k-points.')
+
+        # Strong warnings about calculation results, but something tells us that you're ok with that
+        spec.exit_code(710, 'WARNING_ELECTRONIC_CONVERGENCE_NOT_REACHED',
+            message='The electronic minimization cycle did not reach self-consistency, but `scf_must_converge` '
+                    'is `False` and/or `electron_maxstep` is 0.')
         # yapf: enable
 
     @classproperty
--- a/aiida_quantumespresso/workflows/pw/base.py
+++ b/aiida_quantumespresso/workflows/pw/base.py
@@ -115,6 +115,9 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
             message='The initialization calculation failed.')
         spec.exit_code(501, 'ERROR_IONIC_CONVERGENCE_REACHED_EXCEPT_IN_FINAL_SCF',
             message='Then ionic minimization cycle converged but the thresholds are exceeded in the final SCF.')
+        spec.exit_code(710, 'WARNING_ELECTRONIC_CONVERGENCE_NOT_REACHED',
+            message='The electronic minimization cycle did not reach self-consistency, but `scf_must_converge` '
+                    'is `False` and/or `electron_maxstep` is 0.')
         # yapf: enable
 
     @classmethod
@@ -588,3 +591,15 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         action = f'reduced beta mixing from {mixing_beta} to {mixing_beta_new} and restarting from the last calculation'
         self.report_error_handled(calculation, action)
         return ProcessHandlerReport(True)
+
+    @process_handler(priority=420, exit_codes=[
+        PwCalculation.exit_codes.WARNING_ELECTRONIC_CONVERGENCE_NOT_REACHED,
+    ])
+    def handle_electronic_convergence_warning(self, calculation):
+        """Handle `WARNING_ELECTRONIC_CONVERGENCE_NOT_REACHED': consider finished."""
+        self.ctx.is_finished = True
+        self.ctx.restart_calc = calculation
+        action = 'electronic convergence not reached but inputs say this is ok: consider finished.'
+        self.report_error_handled(calculation, action)
+        self.results()  # Call the results method to attach the output nodes
+        return ProcessHandlerReport(True, self.exit_codes.WARNING_ELECTRONIC_CONVERGENCE_NOT_REACHED)
--- a/tests/workflows/pw/test_base.py
+++ b/tests/workflows/pw/test_base.py
@@ -130,6 +130,34 @@ def test_handle_relax_recoverable_ionic_convergence_error(generate_workchain_pw,
     assert result.status == 0
 
 
+def test_handle_electronic_convergence_warning(generate_workchain_pw, generate_structure):
+    """Test `PwBaseWorkChain.handle_electronic_convergence_warning`."""
+    from aiida.common.links import LinkType
+
+    inputs = generate_workchain_pw(return_inputs=True)
+    inputs['pw']['parameters']['scf_maxstep'] = 0
+    inputs['pw']['parameters']['scf_must_converge'] = False
+
+    process = generate_workchain_pw(
+        exit_code=PwBaseWorkChain.exit_codes.WARNING_ELECTRONIC_CONVERGENCE_NOT_REACHED, inputs=inputs
+    )
+    process.setup()
+
+    calculation = process.ctx.children[-1]
+    structure = generate_structure().store()
+    structure.add_incoming(calculation, link_label='output_structure', link_type=LinkType.CREATE)
+
+    result = process.handle_electronic_convergence_warning(calculation)
+    assert isinstance(result, ProcessHandlerReport)
+    assert result.do_break
+    assert result.exit_code == PwBaseWorkChain.exit_codes.WARNING_ELECTRONIC_CONVERGENCE_NOT_REACHED
+    assert process.node.get_outgoing().all() is not None
+    assert process.ctx.restart_calc is calculation
+
+    result = process.inspect_process()
+    assert result == PwBaseWorkChain.exit_codes.WARNING_ELECTRONIC_CONVERGENCE_NOT_REACHED
+
+
 def test_sanity_check_no_bands(generate_workchain_pw):
     """Test that `sanity_check_insufficient_bands` does not except if there is no `output_band`, which is optional."""
     process = generate_workchain_pw(exit_code=ExitCode(0))
--- a/aiida_quantumespresso/parsers/pw.py
+++ b/aiida_quantumespresso/parsers/pw.py
@@ -150,7 +150,14 @@ class PwParser(Parser):
             return
 
         if 'ERROR_ELECTRONIC_CONVERGENCE_NOT_REACHED' in logs['error']:
-            return self.exit_codes.ERROR_ELECTRONIC_CONVERGENCE_NOT_REACHED
+            scf_must_converge = self.node.inputs.parameters.get_attribute('ELECTRONS',
+                                                                          {}).get('scf_must_converge', True)
+            electron_maxstep = self.node.inputs.parameters.get_attribute('ELECTRONS', {}).get('electron_maxstep', 1)
+
+            if electron_maxstep == 0 or not scf_must_converge:
+                return self.exit_codes.WARNING_ELECTRONIC_CONVERGENCE_NOT_REACHED
+            else:
+                return self.exit_codes.ERROR_ELECTRONIC_CONVERGENCE_NOT_REACHED
 
     def validate_dynamics(self, trajectory, parameters, logs):
         """Analyze problems that are specific to `dynamics` type calculations: i.e. `md` and `vc-md`."""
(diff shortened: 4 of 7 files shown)


aiida_quantumespresso/calculations/pw.py
```diff
@@ -145,6 +145,11 @@ def define(cls, spec):
             message='The electronic minimization cycle did not reach self-consistency.')
         spec.exit_code(541, 'ERROR_SYMMETRY_NON_ORTHOGONAL_OPERATION',
             message='The variable cell optimization broke the symmetry of the k-points.')
+
+        # Strong warnings about calculation results, but something tells us that you're ok with that
+        spec.exit_code(610, 'WARNING_ELECTRONIC_CONVERGENCE_NOT_REACHED',
+            message='The electronic minimization cycle did not reach self-consistency, but `scf_must_converge` '
+                    'is `False` and/or `scf_maxstep` is 0.')
```
sphuber, line 152:
```suggestion
                    'is `False` and/or `electron_maxstep` is 0.')
```

aiida_quantumespresso/calculations/pw.py
```diff
@@ -145,6 +145,11 @@ def define(cls, spec):
             message='The electronic minimization cycle did not reach self-consistency.')
         spec.exit_code(541, 'ERROR_SYMMETRY_NON_ORTHOGONAL_OPERATION',
             message='The variable cell optimization broke the symmetry of the k-points.')
+
+        # Strong warnings about calculation results, but something tells us that you're ok with that
+        spec.exit_code(610, 'WARNING_ELECTRONIC_CONVERGENCE_NOT_REACHED',
```
sphuber, line 150: See comment on the issue
```suggestion
        spec.exit_code(710, 'WARNING_ELECTRONIC_CONVERGENCE_NOT_REACHED',
```

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -113,6 +113,9 @@ def define(cls, spec):
             message='The initialization calculation failed.')
         spec.exit_code(501, 'ERROR_IONIC_CONVERGENCE_REACHED_EXCEPT_IN_FINAL_SCF',
             message='Then ionic minimization cycle converged but the thresholds are exceeded in the final SCF.')
+        spec.exit_code(502, 'WARNING_ELECTRONIC_CONVERGENCE_NOT_REACHED',
```
sphuber, line 116:
```suggestion
        spec.exit_code(710, 'WARNING_ELECTRONIC_CONVERGENCE_NOT_REACHED',
```

sphuber: Wondering if we should maybe try and maintain the same exit codes as much as possible. Not sure if this will be possible but I can imagine that it might be confusing if exit codes keep switching

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -113,6 +113,9 @@ def define(cls, spec):
...
+            message='The electronic minimization cycle did not reach self-consistency, but `scf_must_converge` '
+                    'is `False` and/or `scf_maxstep` is 0.')
```
sphuber, line 118:
```suggestion
                    'is `False` and/or `electron_maxstep` is 0.')
```

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -585,3 +588,15 @@ def handle_electronic_convergence_not_achieved(self, calculation):
         action = f'reduced beta mixing from {mixing_beta} to {mixing_beta_new} and restarting from the last calculation'
         self.report_error_handled(calculation, action)
         return ProcessHandlerReport(True)
+
+    @process_handler(priority=610, exit_codes=[
```
sphuber, line 592: note that the priority has nothing to do with the exit code. By making it `610` this handler will now actually be ran before all others. I don't have a waterproof argument for where it should be otherwise, but I would put it close, probably just before the handler of the actual `ERROR_ELECTRONIC_CONVERGENCE_NOT_REACHED`, which has priority 410, so maybe we give this one 420.

sphuber (changes requested):
Thanks @zooks97 . Couple of minor changes and then we would need a test. All the infrastructure to write one is already in place. Just have a look in `tests/workflows/pw/test_base.py`. You can see examples there: essentially you create a workchain instance  with specific inputs, call a few outline steps to initialize the state and then you call the new handler, checking afterwards that the report and context are correctly changed.

zooks97:
I've attempted to write a simple test for the new exit code, but I'm getting a bunch of md5 mismatch errors when running pytest (maybe my setup is bad?). Would like to have feedback on if the test is reasonable (maybe should test all combinations of the two pw inputs).

aiida_quantumespresso/workflows/pw/base.py
mbercx: I actually had a question about this: does the `PwCalculation` ever have multiple exit codes? Just by glancing at the `PwParser`, doesn't the parsing stop once e.g.: 

https://github.com/aiidateam/aiida-quantumespresso/blob/f4528aff3165dbb8c598e8ef5ab91e81d9587b60/aiida_quantumespresso/parsers/pw.py#L105-L107

so even _if_ other exit codes could be added that are dealt with in the `PwBaseWorkChain` at a higher priority, will this happen? If not, does the priority really matter if the `PwCalculation` only has one exit code? Maybe I'm misunderstanding how this all works. I was looking into it when I had the ambition of reviewing this PR, but I went too deep down the rabbit hole and forgot what I was doing at some point. 😅

sphuber: > I actually had a question about this: does the `PwCalculation` ever have multiple exit codes? 

Any `Process` can only ever have a single exit code, that is a limitation imposed by `aiida-core`.

> so even _if_ other exit codes could be added that are dealt with in the `PwBaseWorkChain` at a higher priority, will this happen? If not, does the priority really matter if the `PwCalculation` only has one exit code?

It does matter, because process handlers return instances of `ProcessHandlerReport` which have an attribute called `do_break`. If that is set to `False` (which is the default) the `BaseRestartWorkChain` will continue calling other process handlers. So depending on the return values of process handlers the order in which they are executed *can* matter.

tests/workflows/pw/test_base.py
```diff
@@ -130,6 +130,34 @@ def test_handle_relax_recoverable_ionic_convergence_error(generate_workchain_pw,
     assert result.status == 0
 
 
+def test_handle_electronic_convergence_warning(generate_inputs_pw, generate_workchain_pw, generate_structure):
+    """Test `PwBaseWorkChain.handle_electronic_convergence_warning`."""
+    from aiida.common.links import LinkType
+
+    inputs = generate_inputs_pw()
```
sphuber, line 137: This is why the test fails. This fixture generates inputs for the `PwCalculation` not the `PwBaseWorkChain`. Use the following instead:
```
inputs = generate_workchain_pw(return_inputs=True)
```
that generates inputs for the workchain and returns the inputs instead of creating the instance, giving you the opportunity to update them as you like. Then call `generate_workchain_pw` again.

sphuber (commented):
Thanks @zooks97 .  Test looks fine in principle there is just a mistake which causes it to fail. That isn't anything to do with md5s though, but I reckon that is something you have locally. Most likely this is due to the SSSP pseudos. Are you running against a clean test profile? Can you post the exact error and stack trace?

aiida_quantumespresso/workflows/pw/base.py
mbercx: Alright, thanks for the explanation! Will be looking closer at this in the coming month when reviewing the tutorial material on the topic.

tests/workflows/pw/test_base.py
zooks97: Great, thanks!

The error does come from `aiida-pseudo`:
```
ERROR tests/workflows/pw/test_base.py::test_handle_electronic_convergence_warning - ValueError: md5 does not match that of stored file: 62f7fc0614e61f39d2530dbfe4ec53b6 != d41d8cd98f00b204e9800998ecf8427e
```

```
____________________________________________________________________________________________________________________________ ERROR at setup of test_handle_electronic_convergence_warning ____________________________________________________________________________________________________________________________

aiida_profile = <aiida.manage.tests.TestManager object at 0x7f0047c85640>, generate_upf_data = <function generate_upf_data.<locals>._generate_upf_data at 0x7f00429fab80>

    @pytest.fixture(scope='session', autouse=True)
    def sssp(aiida_profile, generate_upf_data):
        """Create an SSSP pseudo potential family from scratch."""
        from aiida.common.constants import elements
        from aiida.plugins import GroupFactory
    
        aiida_profile.reset_db()
    
        SsspFamily = GroupFactory('pseudo.family.sssp')
    
        cutoffs = {}
        stringency = 'standard'
    
        with tempfile.TemporaryDirectory() as dirpath:
            for values in elements.values():
    
                element = values['symbol']
>               upf = generate_upf_data(element)

tests/conftest.py:128: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/conftest.py:295: in _generate_upf_data
    return UpfData(file=handle)
../aiida-core/aiida/orm/nodes/data/singlefile.py:48: in __init__
    self.set_file(file)
../aiida-pseudo/aiida_pseudo/data/pseudo/upf.py:86: in set_file
    super().set_file(source, filename, **kwargs)
../aiida-pseudo/aiida_pseudo/data/pseudo/pseudo.py:134: in set_file
    self.md5 = md5_from_filelike(source)
../aiida-pseudo/aiida_pseudo/data/pseudo/pseudo.py:186: in md5
    self.validate_md5(value)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <UpfData: uuid: 457be4f1-e194-417f-9727-9b210445ea43 (unstored)>, md5 = '62f7fc0614e61f39d2530dbfe4ec53b6'

    def validate_md5(self, md5: str):
        """Validate that the md5 checksum matches that of the currently stored file.
    
        :param value: the md5 checksum.
        :raises ValueError: if the md5 does not match that of the currently stored file.
        """
        with self.open(mode='rb') as handle:
            md5_file = md5_from_filelike(handle)
            if md5 != md5_file:
>               raise ValueError(f'md5 does not match that of stored file: {md5} != {md5_file}')
E               ValueError: md5 does not match that of stored file: 62f7fc0614e61f39d2530dbfe4ec53b6 != d41d8cd98f00b204e9800998ecf8427e

../aiida-pseudo/aiida_pseudo/data/pseudo/pseudo.py:111: ValueError
```

sphuber: Ah, this was a bug in our fixtures in `conftest.py`. I already fixed it, just rebase on latest `develop` and you should be good. Good lesson to always update your working branches ;)

sphuber:
Thanks @zooks97 but you and `git` are not a match made in heaven :sweat_smile: I don't know how you do it, but you managed to get lines into the PR that are already in develop. I will fix her up

sphuber (approved):

sphuber:
Fixed the branch and also added a test for the `PwParser` whose functionality was also changed. Thanks for the PR @zooks97

zooks97:
Thanks!

PR #699, opened 2021-05-25
Read PHONON version number from stdout

atztogo:
PR #633 is critical. To merge PR #633, @sphuber suggested to check version number, but PHONON parameter output doesn't contain the version information.

This PR is made to read the version of PHONON. The code is taken from `parse_output_base` method. Here I follow `parse_output_base`'s naming of the key `'code_version'`, but this may be inconsistent with `'creator_version'` in PW, though I don't know what exactly means 'creator'.

### DIFF
--- a/aiida_quantumespresso/parsers/parse_raw/ph.py
+++ b/aiida_quantumespresso/parsers/parse_raw/ph.py
@@ -195,7 +195,10 @@ def parse_ph_text_output(lines, logs):
 
         detect_important_message(logs, line)
 
-        if 'q-points for this run' in line:
+        if 'PHONON' in line and 'starts on' in line:
+            parsed_data['code_version'] = line.split('PHONON')[1].split('starts on')[0].strip()
+
+        elif 'q-points for this run' in line:
             try:
                 num_qpoints = int(line.split('/')[1].split('q-points')[0])
                 if (


unkcpz (changes requested):
Thanks! @atztogo 
The implementation is good, but you should run `tests/parsers/test_ph.py` with `--force-regen` to count this new attribute.

As for 'creator_version' I believe it is the same as the code_version, while 'creator_version' only parsed in pw with XML format. Let's see what @sphuber says.

atztogo:
Thanks @unkcpz. I could not succeed to run the test. I simply don't understand the testing manner in aiida-quantumespresso. I looked into the documentation, but didn't find a good introduction (https://aiida-quantumespresso.readthedocs.io/en/latest/devel_guide/parsertests.html#running-tests). I would like you to include something like this PR to your PR #633 by hand (i.e., copy and paste etc) if you are fine. Then I will close this PR.

sphuber:
Thanks @atztogo . I am updating the PR of @unkcpz with the necessary info and then merging it. We can keep this open and I can add a quick test for you. No need to close it.

sphuber:
I came back to this PR, but I cannot actually find in #633 where I would have called for parsing the version? It is not actually being used in the code anywhere for checks. So I think we can close this. Thanks nonetheless @atztogo !

Issue #702, opened 2021-06-04
Adding more namelists for environ?


sudarshanv01:
I believe that currently only the &ENVIRON namelist is written out to the environ.in file (_ENVIRON_INPUT_FILE_NAME). There are a few other namelists which would be useful to have when running [Environ](https://environ.readthedocs.io/en/latest/) calculations. Two (BOUNDARY and ELECTROSTATIC) of them can be easily added by doing much the same thing that was done for the ENVIRON tag. 

https://github.com/aiidateam/aiida-quantumespresso/blob/f4528aff3165dbb8c598e8ef5ab91e81d9587b60/aiida_quantumespresso/calculations/__init__.py#L238-L251

PR #703, opened 2021-06-05
Adding ENVIRON namelists

sudarshanv01:
This PR is in reference to [this](https://github.com/aiidateam/aiida-quantumespresso/issues/702#issue-911764109) issue. Added a couple of namelists (boundary and electrostatics) specific to [Environ](http://www.quantum-environ.org/installation.html) which can be added to the settings dictionary. It is written the same way as the ENVIRON namelist was written - but only called if an ENVIRON key is in the settings dictionary. 

 Also added an example to test the changes with some typical keys. This example tries to recreate the same calculation as in Example04 with the [examples](https://gitlab.com/olivieroandreussi/Environ.git) folder of Environ.

### DIFF
--- a/aiida_quantumespresso/calculations/__init__.py
+++ b/aiida_quantumespresso/calculations/__init__.py
@@ -238,6 +238,11 @@ class BasePwCpInputGenerator(CalcJob):
         if environ_namelist is not None:
             if not isinstance(environ_namelist, dict):
                 raise exceptions.InputValidationError('ENVIRON namelist should be specified as a dictionary')
+
+            ## Check for BOUNDARY and ELECTROSTATIC Keys as well
+            boundary_namelist = settings.pop('BOUNDARY', None)
+            electrostatic_namelist = settings.pop('ELECTROSTATIC', None)
+
             # We first add the environ flag to the command-line options (if not already present)
             try:
                 if '-environ' not in settings['CMDLINE']:
@@ -254,6 +259,18 @@ class BasePwCpInputGenerator(CalcJob):
                 for key, value in sorted(environ_namelist.items()):
                     handle.write(convert_input_to_namelist_entry(key, value, mapping=mapping_species))
                 handle.write('/\n')
+
+                if boundary_namelist is not None:
+                    handle.write('&BOUNDARY\n')
+                    for key, value in sorted(boundary_namelist.items()):
+                        handle.write(convert_input_to_namelist_entry(key, value, mapping=mapping_species))
+                    handle.write('/\n')
+
+                if electrostatic_namelist is not None:
+                    handle.write('&ELECTROSTATIC\n')
+                    for key, value in sorted(electrostatic_namelist.items()):
+                        handle.write(convert_input_to_namelist_entry(key, value, mapping=mapping_species))
+                    handle.write('/\n')
 
         # Check for the deprecated 'ALSO_BANDS' setting and if present fire a deprecation log message
         also_bands = settings.pop('ALSO_BANDS', None)
--- a/docs/source/user_guide/get_started/examples/example_environ_calculation.py
+++ b/docs/source/user_guide/get_started/examples/example_environ_calculation.py
@@ -1,0 +1,143 @@
+
+from aiida import orm
+from aiida.engine import run, submit
+import numpy as np
+from ase import Atoms
+from ase import units
+
+"""
+Calculation that recreates the environ calculation of
+CO on Pt(111) with a unit charge
+
+More details about the solvation models and potential corrections
+can be found here: 
+
+O. Andreussi, I. Dabo and N. Marzari, J. Chem. Phys. 136,064102 (2012).
+I. Dabo et al. Phys.Rev. B 77, 115139 (2008)
+
+This system is taken directly from Example04 in the 
+examples folder from Environ
+"""
+
+def calculator():
+    """
+    This is the SCF calculator, which is identical
+    and unchanged from what one would do in the case 
+    of a standard SCF calculation - nothing environ
+    related. Remember to add in the tot_charge here
+    """
+    param_dict =  {
+    "CONTROL":{
+        'calculation':'scf',
+        'tprnfor':True,
+                },
+    "SYSTEM": {
+        "ecutwfc": 35,
+        "ecutrho": 280,
+        "occupations":'smearing',
+        'smearing':'cold',
+        'degauss':0.03,
+        'tot_charge':1,
+                },
+    "ELECTRONS": {
+        "conv_thr": 5e-9,
+        'electron_maxstep': 200,
+        'mixing_beta': 0.2,
+        'mixing_ndim': 15,
+        'diagonalization': 'david',
+                },
+    }
+
+    return param_dict
+
+def environ_calculator():
+    """
+    Environ parameters
+    """
+    param_dict = {
+        'ENVIRON':{
+            'verbose': 0,
+            'environ_thr': 1.0,
+            'environ_type': 'input',
+            'env_static_permittivity': 1,
+            'env_surface_tension': 0.0,
+            'env_pressure': 0.0,
+            'env_electrostatic': True,
+        },
+        'BOUNDARY':{
+            'solvent_mode':'full',
+        },
+        'ELECTROSTATIC':{
+            'pbc_correction':'parabolic',
+            'pbc_dim': 2, 
+            'pbc_axis': 3, 
+            'tol': 5e-13,
+        }
+    }
+    return param_dict
+
+def runner(code, structure):
+
+    ## Base calculation for testing
+    PwBaseWorkChain = WorkflowFactory('quantumespresso.pw.base')
+    builder = PwBaseWorkChain.get_builder()
+    builder.pw.structure = structure
+
+    builder.metadata.label = 'Single point environ'
+    builder.metadata.description = 'Testing calculation with environ for Pt(111) + CO'
+
+    KpointsData = DataFactory('array.kpoints')
+    kpoints = KpointsData()
+    kpoints.set_kpoints_mesh([1, 1, 1]) # Not physical just a test
+    builder.kpoints = kpoints
+
+    family = load_group('SSSP/1.1/PBE/efficiency')
+    builder.pw.pseudos = family.get_pseudos(structure=structure)
+
+    calculation = calculator()
+    environ_calculation = environ_calculator()
+    ## these are the main cube files that could potentially be parsed
+    ## if the verbosity is set to 2 or higher 
+    # environ_calculation['additional_retrieve_list'] = ['epsilon.cube', \
+    #     'vreference.cube', 'velectrostatic.cube', 'vsoftcavity.cube', \
+    #     'electrons.cube', 'charges.cube', 'smeared_ions.cube']
+
+    builder.pw.parameters = orm.Dict(dict=calculation)
+    builder.pw.settings = orm.Dict(dict=environ_calculation)
+    builder.pw.metadata.options.resources = {'num_machines': 1}
+    builder.pw.metadata.options.max_wallclock_seconds = 5 * 60
+    builder.pw.code = code
+
+    calculation = submit(builder)
+
+
+if __name__ == '__main__':
+    code = load_code('pw_6-7@juwels')
+
+    StructureData = DataFactory('structure')
+    ## these are the original coordinates for the Pt-CO system
+    positions = [
+       [5.335084148,   4.646723426,  12.901029877],
+       [5.335009643,   4.619623254,  15.079854269],
+       [8.061327071,   0.098057998,   8.992142901],
+       [2.608989366,   0.098058283,   8.992140585],
+       [0.000036609,   4.720846294,   8.968756935],
+       [5.335159557,   4.721612729,   9.380196435],
+       [0.000041121,   7.802951963,   4.604626508],
+       [5.335161233,   7.697749113,   4.753489408],
+       [2.697860636,   3.152173889,   4.688412329],
+       [7.972463687,   3.152174491,   4.688415209],
+    ]
+
+    ## setting up the system with ASE 
+    ## notice the units that are being used
+    atoms = Atoms('COPt8')
+    atoms.set_positions(np.array(positions)*units.Bohr)
+    atoms.set_pbc([True, True, True])
+    a = 10.6881 * units.Bohr
+    b = 0.866025 * a * units.Bohr
+    c = 3.95422 * a * units.Bohr
+    atoms.set_cell([a, b, c])
+    structure = StructureData(ase=atoms)
+
+    runner(code=code, structure=structure)


mbercx:
Thanks for your contribution @sudarshanv01! Looks great at a first glance, will do a more thorough review later next week when I have time to test this. Two more things:

* Although the example you provide is already great, could you maybe write a short piece of documentation on how to use ENVIRON? I think you can just add it as an extra section to the page on [running `pw.x` calculations](https://aiida-quantumespresso.readthedocs.io/en/latest/user_guide/get_started/examples/pw_tutorial.html), i.e. [this file](https://github.com/aiidateam/aiida-quantumespresso/blob/develop/docs/source/user_guide/get_started/examples/pw_tutorial.rst) in the `docs`. As @giovannipizzi mentioned, this isn't currently documented anywhere, so would be good if we could add this. Note that the documentation still needs a lot of work, that'll be my main focus after the tutorial has finished. :)
* I'm not sure how familiar you are with writing tests, but I think it would be good to add a test for these namelists to the tests for the `PwCalculation`, see [this file](https://github.com/aiidateam/aiida-quantumespresso/blob/develop/tests/calculations/test_pw.py). Let me know if you need help with this, and I'd also be happy to add a test myself once I'm done reviewing the code.

As a sidenote: does ENVIRON produce any new output in the stdout? Or does it only produce extra files with AiiDA then has to retrieve?

sudarshanv01:
Absolutely - will write a section on ENVIRON and will add tests!

That is a good point, about the stdout. There are a few quantities that could be potentially interesting to parse from stdout - I imagine that would go into the parser - I can take a look at adding those in as well!

mbercx:
> Absolutely - will write a section on ENVIRON and will add tests!

Great, thanks a lot! Let me know if you need any help. 🙃 

> That is a good point, about the stdout. There are a few quantities that could be potentially interesting to parse from stdout - I imagine that would go into the parser - I can take a look at adding those in as well!

Just a note: In general, it's better _not_ to parse from the stdout _in case_ there is an alternative that is less likely to break. For example in the `pw.x` parsing, we try to get as much information from the `data-file-schema.xml`, since we als have the XSD schemas for the different versions of QE, so this will not break when changes are made by the QE developers, since we can simply load the schema of the correct version. If there is something like this for ENVIRON, it would be better than parsing information from the stdout, since any change in formatting can break the parser.

tests/calculations/test_pw.py
```diff
@@ -250,3 +250,26 @@ def test_pw_parallelization_duplicate_cmdline_flag(fixture_sandbox, generate_cal
...
+            'BOUNDARY': {
+                'solvent_mode': 'electronic',
+                'electrolyte_mode': 'electronic',
+            },
+            'ELECTROSTATIC': {
+                'pbc_correction': 'parabolic',
+            }
+        },
+    )
+    generate_calc_job(fixture_sandbox, entry_point_name, inputs)
```
sphuber, line 275: This test is not actually yet testing that the namelists are (properly) generates. The `generate_calc_job` just creates the `CalcJob` and calls the `prepare_for_submission` method. You still need to check that the input file that was generated is correct. To see how to do this, you can look in this same file at the test `test_pw_default` all the way at the top:
```
    with fixture_sandbox.open('aiida.in') as handle:
        input_written = handle.read()
        file_regression.check(input_written, encoding='utf-8', extension='.in')
```
The `file_regression` fixture will compare the input file that was written with a reference file. If that file doesn't yet exist (i.e. before you have ever run this test) it will automatically create it. If you rerun the tests a second time, it will work.

sphuber (changes requested):
Thanks @sudarshanv01 . The changes look ok. There is just a change needed for the added test to actually make it test. And since you are also changing the parser, I would appreciate it if you could add a test for this as well. Have a look at the other `pw.x` parser tests to see how it is done and feel free to ask me for help if you have any questions or run into trouble

aiida_quantumespresso/parsers/parse_raw/pw.py
```diff
@@ -743,6 +743,14 @@ def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None)
...
                 except Exception:
                     logs.warning.append('Error while parsing Fermi energy from the output file.')
+            elif 'Gaussian-smeared nuclei' in line:
+                try:
+                    value_potential_shift = float(line.split()[-2])
+                    unit_potential_shift = line.split()[-1]
+                    parsed_data['environ_potential_shift'] = value_potential_shift
+                    parsed_data['environ_unit_potential_shift'] = unit_potential_shift
+                except Exception:
+                    pass
```
giovannipizzi, line 753: Add a log here as well?

giovannipizzi (commented):
Just one small comment

tests/calculations/test_pw.py
sudarshanv01: Hi @sphuber - thanks a lot for the comments! I have a few questions about how these tests work. I have added in `file_regression` in the most recent commit, but I have not really set up a corresponding input file that it should compare to, and I missing how it automatically creates it. 

Regarding the parser, I modeled the Environ test after the `test_pw_default` test and added in some checks after running an Environ calculation. A very basic question - how does this work in practice, is a calculation actually run - in which case it might break because Environ is a plugin that doesn't ship with the code I believe.

aiida_quantumespresso/parsers/parse_raw/pw.py
sudarshanv01: Thanks Giovanni! I have added it to the log in the most recent commit - thanks!

https://github.com/sudarshanv01/aiida-quantumespresso/blob/894b95a9e44fda4875b0a4f3fcd580bd40242e29/aiida_quantumespresso/parsers/parse_raw/pw.py#L753

tests/calculations/test_pw.py
sphuber: If you just add the `file_regression` call in the test, the comparison will be automatically created once you run the tests. If you then run the tests a second time, the comparison file is there and the test should therefore pass. If you ever change anything in the test that would imply that the comparison file has to be updated, you don't have to do this manually either. If you run the tests and they fail because the comparison file is outdated, you can update it by rerunning the tests with the flag `--force-regen` and it will update all comparison files that are used and whose test fails.

> A very basic question - how does this work in practice, is a calculation actually run - in which case it might break because Environ is a plugin that doesn't ship with the code I believe.

No, it doesn't run the calculation. I set up the framework for these parser tests that it simply mocks an already completed calculation with outputs taking from the test directory. Each test has its own folder that corresponds to the name of the test. So all you have to do is add a folder for your test with the output files of a run you did manually with the environ module enabled.

sudarshanv01: Thanks a lot! So I think I understand now how the tests for the calculations work (and have updated the latest commit with the right files). The same thing seems to work for the parser, but I am not sure I am testing this properly. All inputs to Environ would be passed through the settings, so I am not sure if this information is included in the yml that is generated. 
Thanks!

docs/source/user_guide/get_started/examples/environ_calculation.py
```diff
@@ -0,0 +1,146 @@
...
+
+O. Andreussi, I. Dabo and N. Marzari, J. Chem. Phys. 136,064102 (2012).
+I. Dabo et al. Phys.Rev. B 77, 115139 (2008)
+
+This system is taken directly from Example04 in the
+examples folder from Environ
+"""
+
+
+def calculator():
```
sphuber, line 23: Why did you make this and the `environ_calculator` a function if all it does is return a static dictionary? I think it might be simpler if to literally just include the dictionary in the main method where you assign it directly to the building

docs/source/user_guide/get_started/examples/environ_calculation.py
```diff
@@ -0,0 +1,146 @@
...
+
+def runner(code, structure):
+
+    ## Base calculation for testing
+    PwBaseWorkChain = WorkflowFactory('quantumespresso.pw.base')
+    builder = PwBaseWorkChain.get_builder()
+    builder.pw.structure = structure
+
+    builder.metadata.label = 'Single point environ'
+    builder.metadata.description = 'Testing calculation with environ for Pt(111) + CO'
```
sphuber, line 90: This hardcodes the structure even though the structure is mutable through the argument. Not a huge deal since it is an example, but still a bit weird.

tests/parsers/test_pw.py
```diff
@@ -907,3 +907,35 @@ def test_pw_vcrelax_failed_not_converged_nstep(
...
+
+def test_environ(
+    fixture_localhost,
+    generate_calc_job_node,
+    generate_parser,
+    generate_inputs,
+    data_regression,
+):
+    """Test a simple Environ calculation."""
+    name = 'default'
```
sphuber, line 920: You are almost there with creating a parser test. This `name` variable here, actually corresponds to a folder in the `tests/parsers/fixtures/pw` folder. You will find the `default` folder there. It actually contains output files that were generated by a `PwCalculation`, i.e. the `aiida.out` and `data-file.xml`. Through the `generate_calc_job_node` function called on line 924, a `CalcJobNode` is mocked with a `retrieved` output node that contains these files. This is replicating what happens when you actually run a `PwCalculation` and the node is passed to the parser.

What you then have to do is create a new folder in that fixtures base folder, let's call it `default_environ` and put the `aiida.out` and `data-file.xml` in there that you take from an actual `PwCalculation` run that included the `ENVIRON` namecard. Only then will you actually test the parser for output files that were produced by an environ run. In the current state, you are simply testing exactly the same as the `test_default` test, which tests the outputs for a simple scf calculation with `pw.x`. I hope this is clear. If not, feel to ask for more clarification.

sphuber (changes requested):
Thanks @sudarshanv01 . Great start with the tests, but as is, they are not actually testing the new functionality. I have given more info in the code how to change this.

mbercx:
@sudarshanv01 just a quick ping on this: will you still have time to update the PR based on @sphuber's suggestions? If not, that's fine, I'd happily take over to fix the tests so we can get this baby merged.

tests/calculations/test_pw.py
```diff
@@ -250,3 +250,33 @@ def test_pw_parallelization_duplicate_cmdline_flag(fixture_sandbox, generate_cal
...
+
+    with fixture_sandbox.open('aiida.in') as handle:
+        input_written = handle.read()
+
+    # Checks on the files written to the sandbox folder as raw input
+    assert sorted(fixture_sandbox.get_content_list()) == sorted(['aiida.in', 'pseudo', 'out', 'environ.in'])
+    file_regression.check(input_written, encoding='utf-8', extension='.in')
```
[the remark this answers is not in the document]
sudarshanv01: Hi @sphuber - apologies that this reply is so late, this pr seems to have slipped through somehow. Just made some changes. I have now added two tests now as you suggested, one for the `aiida.in` file called `test_environ_pw_namelists` and the second for `environ.in` files called `test_environ_namelists`.

aiida_quantumespresso/parsers/parse_raw/pw.py
sudarshanv01: Yup, I guess it is specific to ENVIRON calculations, right now it says `Not an Environ calculation with an open boundary condition`, but this might show up for all calculations, which might be a bit annoying? Is there a better way to have a log message here without it popping up every time - and only for ENVIRON calculations?

docs/source/user_guide/get_started/examples/environ_calculation.py
sudarshanv01: Yup good point, has been corrected.

docs/source/user_guide/get_started/examples/environ_calculation.py
sudarshanv01: Yup, also corrected.

tests/parsers/test_pw.py
sudarshanv01: Yup all makes sense - thanks! Let me just describe what I did here. I added a new structure with `structure_id=platinum` to the `conftest.py` in order to test the different environ outputs as they won't work for the two example cases for Si or water (I need a 2D slab to check for the right flags). I then ran an `pw` calculation with QE v.6.7 and stored the `aiida.out` and the `data-file.xml` file in `fixtures/default_environ`. So far so good, except the auto-generated `test_environ.yml` file doesn't seem to have the right specifications (it reports Si and not Pt). Moreover the `k-points` do not seem to be parsed properly (commented out currently). I can also bypass this and run everything for Si, but my thinking is that it would be nice to have the test mirror that of the Environ package.

aiida_quantumespresso/parsers/parse_raw/pw.py
sphuber: Sure, just add a conditional to the `elif 'Gaussian-smeared nuclei' in line` to only run when it is an ENVIRON calculation. How to determine whether we actually are parsing an ENVIRON question depends. Probably you could parse the stdout for some flag, but this is really fragile. If the XML doesn't contain some switch that we can rely on (which would be the best and most robust solution) you can use the input parameters. If I understand correctly, if you want to run an ENVIRON calculation, you have to add a particular key to the `parameters` input node. You can check for that at the beginning of the parser. Something like
```python
is_environ = 'ENVIRON' in self.inputs.parameters.get_dict()
```

and then the parsing line would become
```python
elif is_environ and 'Gaussian-smeared nuclei' in line:
```

docs/source/user_guide/get_started/examples/pw_tutorial.rst
```diff
@@ -623,6 +623,32 @@ More options are available, and can be explored by expanding
...
+        'ELECTROSTATIC':{
+            'pbc_correction':'parabolic',
+            'pbc_dim': 2,
+            'pbc_axis': 3,
+        }
+    }
+
+This dictionary can be passed on to the builder as ::
+
+    builder.pw.settings = orm.Dict(dict=environ_param_dict)
```
sphuber, line 651:
```suggestion
    builder.settings = orm.Dict(dict=environ_param_dict)
```

tests/conftest.py
```diff
@@ -314,8 +314,21 @@ def _generate_structure(structure_id='silicon'):
...
+            structure.append_atom(position=[5.335084148, 4.646723426, 12.901029877], symbols='C', name='C')
+            structure.append_atom(position=[5.335009643, 4.619623254, 15.079854269], symbols='O', name='O')
+            structure.append_atom(position=[8.061327071, 0.098057998, 8.992142901], symbols='Pt', name='Pt')
+            structure.append_atom(position=[2.608989366, 0.098058283, 8.992140585], symbols='Pt', name='Pt')
+            structure.append_atom(position=[0.000036609, 4.720846294, 8.968756935], symbols='Pt', name='Pt')
+            structure.append_atom(position=[5.335159557, 4.721612729, 9.380196435], symbols='Pt', name='Pt')
+            structure.append_atom(position=[0.000041121, 7.802951963, 4.604626508], symbols='Pt', name='Pt')
+            structure.append_atom(position=[5.335161233, 7.697749113, 4.753489408], symbols='Pt', name='Pt')
+            structure.append_atom(position=[2.697860636, 3.152173889, 4.688412329], symbols='Pt', name='Pt')
+            structure.append_atom(position=[7.972463687, 3.152174491, 4.688415209], symbols='Pt', name='Pt')
```
sphuber, line 329: Do you really need a different structure for the tests? Usually you are just mocking a structure and the test shouldn't really care about what actual structure is used.

tests/calculations/test_pw.py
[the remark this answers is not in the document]
sphuber: Why do you need two different tests for this? The tests are identical, you are just checking a different file in each. You can merge these tests and check both files in one with something like this:
```python
file_regression.check(input_written, encoding='utf-8', extension='.aiida.in')
file_regression.check(input_written, encoding='utf-8', extension='.environ.in')
```

aiida_quantumespresso/parsers/parse_raw/pw.py
sudarshanv01: Sure, absolutely - just put this change in. In the current form I would need to pass an extra argument of settings to the `parse_raw` function. I could also just send `is_environ` as an argument, but maybe it is more complete to pass the `settings` dictionary, just like `parameters`?

tests/calculations/test_pw.py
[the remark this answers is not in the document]
sudarshanv01: Yup I do not need two tests for this as you mentioned. Has been combined into one as you said!

tests/conftest.py
sudarshanv01: Very fair - as long as there is some vacuum somewhere. Maybe the structure of water could be used instead, would that work?

aiida_quantumespresso/parsers/parse_raw/pw.py
```diff
@@ -267,7 +267,7 @@ def detect_important_message(logs, line):
             logs.warning.append(message)
 
 
-def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None):
+def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None, settings={}):
```
sphuber, line 270: Note that using mutable objects (as a `dict` is) for function defaults is dangerous. The reason is that the default can actually change if it is mutated. Rather, you should define `settings=None` and then in the function body you call `settings = settings or {}` which will assign the empty dict if it is `None`.

aiida_quantumespresso/parsers/pw.py
```diff
@@ -295,7 +295,7 @@ def parse_xml(self, dir_with_bands=None, parser_options=None):
 
         return parsed_data, logs
 
-    def parse_stdout(self, parameters, parser_options=None, parsed_xml=None):
+    def parse_stdout(self, parameters, parser_options=None, parsed_xml=None, settings={}):
```
sphuber, line 298: See other comment about having mutable defaults

tests/conftest.py
sphuber: What part of the code will fail if there is no vacuum?

tests/conftest.py
```diff
@@ -314,8 +314,21 @@ def _generate_structure(structure_id='silicon'):
...
         else:
-            raise KeyError('Unknown structure_id=\'{}\''.format(structure_id))
+            raise KeyError('Unknown structure_id=\'{}\''.format(structure_id))  # pylint: disable=consider-using-f-string
```
sphuber, line 331:
```suggestion
            raise KeyError(f'Unknown structure_id=\'{structure_id}\''
```

tests/conftest.py
```diff
@@ -234,7 +234,7 @@ def _generate_calc_job_node(
             from qe_tools.exceptions import ParsingError
             from aiida_quantumespresso.tools.pwinputparser import PwInputFile
             try:
-                with open(filepath_input, 'r') as input_file:
+                with open(filepath_input, 'r') as input_file:  # pylint: disable=unspecified-encoding
```
sphuber, line 237:
```suggestion
                with open(filepath_input, 'r', encoding='utf-8') as input_file:
```

tests/parsers/test_pw.py
```diff
@@ -907,3 +907,36 @@ def test_pw_vcrelax_failed_not_converged_nstep(
...
+        entry_point_calc_job, fixture_localhost, name,
+        generate_inputs(settings=environ_settings, structure_id='platinum')
+    )
+    parser = generate_parser(entry_point_parser)
+    results, calcfunction = parser.parse_from_node(node, store_provenance=False)
+
+    assert calcfunction.is_finished, calcfunction.exception
+    assert calcfunction.is_finished_ok, calcfunction.exit_message
+    assert not orm.Log.objects.get_logs_for(node), [log.message for log in orm.Log.objects.get_logs_for(node)]
+    # assert 'output_kpoints' in results
```
sphuber, line 934:
```suggestion
```

tests/parsers/test_pw.py
```diff
@@ -907,3 +907,36 @@ def test_pw_vcrelax_failed_not_converged_nstep(
...
+    assert 'output_parameters' in results
+    assert 'output_trajectory' in results
+
+    data_regression.check({
+        # 'output_kpoints': results['output_kpoints'].attributes,
```
sphuber, line 939:
```suggestion
```

tests/calculations/test_pw.py
```diff
@@ -250,3 +250,37 @@ def test_pw_parallelization_duplicate_cmdline_flag(fixture_sandbox, generate_cal
...
+    generate_calc_job(fixture_sandbox, entry_point_name, inputs)
+
+    with fixture_sandbox.open('aiida.in') as handle:
+        input_written = handle.read()
+    with fixture_sandbox.open('environ.in') as handle:
+        environ_written = handle.read()
+
+    # Checks on the files written to the sandbox folder as raw input
+    assert sorted(fixture_sandbox.get_content_list()) == sorted(['aiida.in', 'pseudo', 'out', 'environ.in'])
+    # Check the aiida.in files are written correctly
```
sphuber, line 284:
```suggestion
```

tests/calculations/test_pw.py
sphuber: The tests now look fine :+1:

sphuber (changes requested):
Thanks @sudarshanv01 . The last changes need some final improvements. See the comments

tests/calculations/test_pw.py
sudarshanv01: Great, thanks!

aiida_quantumespresso/parsers/parse_raw/pw.py
sudarshanv01: I see, that makes sense, good to know - thanks!

tests/conftest.py
sudarshanv01: I don't think it would necessarily fail, I just thought it would be nice to test the pbc corrections schemes, as we have them in the parser now, though this is relatively minor. If we are allowed to use a different structure from the one that was used for the DFT calculation, then that would be checked anyway.

mbercx:
@matthew-truscott just pinging you so you are aware of this proposed change by @sudarshanv01. Would the `EnvPwCalculation` still be needed once this PR is merged?

matthew-truscott:
Thanks for the mention, and good to see more support for Environ with AiiDA. `EnvPwCalculation` in its current state does not provide much more functionality (there is an update to the github mirror scheduled soon where we introduce datatypes to handle specific Environ charge objects that can be added to the input file as cards, and parsing options for the debug file that is generated by Environ). We also plan on adding exit codes and logging messages that reflect various environ specifc issues that can arise in a PW self consistent calculation. 

Environ itself is getting some pretty significant updates that will warrant a completely separate calcfunction, in particular, Environ will act less like a plugin on top of QE and more like a standalone code that leverages the QE drivers, among other things.

PR #707, opened 2021-07-21
Fix psycopg2 dependency problem (temporary)

ramirezfranciscof:
A recent patch release of `psycopg2 v2.8` has broken (at least) the continuous
integration tests suite. This was fixed in `aiida-core v1.6.x` but not (yet?)
backported to previous versions that are currently supported by this plugin.
This pin can be removed once a higher version of `aiida-core` is supported
or once `psycopg2` has released a fix for these problems.

### DIFF
--- a/setup.json
+++ b/setup.json
@@ -93,6 +93,7 @@
         ]
     },
     "install_requires": [
+        "psycopg2-binary<2.9",
         "aiida_core[atomic_tools]~=1.4,>=1.4.4",
         "aiida-pseudo~=0.6.1",
         "jsonschema",


setup.json
```diff
@@ -93,6 +93,7 @@
         ]
     },
     "install_requires": [
+        "psycopg2-binary~=2.8.3",
```
sphuber, line 96: Since all we really care about here is just an upper limit it is better to do just that. There is no need to also set a lower limit and can only introduce further problems.
```suggestion
        "psycopg2-binary<2.9",
```

sphuber (changes requested):
Thanks @ramirezfranciscof . Please provide a detailed commit message why this is introduced and when it can and should be removed.

setup.json
ramirezfranciscof: But the thing is that from what I understand, the problem was actually with a _patch_ release of psycopg. See [here](https://github.com/aiidateam/aiida-core/pull/4989/files#diff-9efd195f4e9bfb79ccd456a1d8370fafcc4bcb0b00ea3799222667d2ae818533L31-R31), we did the following change:

```
- psycopg2-binary>=2.8.3,~=2.8
+ psycopg2-binary~=2.8.3
```

I don't know why we required >=2.8.3, but I understand `~=2.8` means that the problematic version was one of 2.8 patches, no?

ramirezfranciscof:
> Please provide a detailed commit message why this is introduced and when it can and should be removed.

I edited the OP to add the message, I can also open an issue to leave as a reminder (this is what we did in `aiida-core`).

setup.json
sphuber: Nope it was due to the new 2.9 release. The `~2.8` is identical to `==2.*,>=2.8`, i.e., all versions of `2.x` but at least `2.8`. [See here for details](https://www.python.org/dev/peps/pep-0440/#compatible-release). So `~-2.8.3` indirectly fixes the problem because it implicitly says `==2.8.*` which is equivalent to `<2.9`.

ramirezfranciscof:
@sphuber ah, shoot, I misremembered how the symbols of the dependency specifications worked, sorry. I now changed the `~=2.8.3` to the preferred `<2.9` and added the description directly in the commit message.

sphuber (approved):

PR #708, opened 2021-07-22
Allow for user to set `SYSTEM/nbnd` for final scf calculation in relax workchain.

Crivella:
This PR is meant to increase the versatility of the `PwRelaxWorkChain`.

As of now, the workchain overrides the `SYSTEM/nbnd` input given using the `base_final_scf` namelist using the number of bands read from the output of the relax calculation.
In case the user wants to use more bands (e.g.: relax without SOC and scf with SOC, which now causes the calculation to fail for "too few bands"), as of now this cannot be done.
This change checks whether the user has set the `SYSTEM/nbnd` manually in the `base_final_scf.pw.parameters` input dictionary, and uses the one from the relax calculation only if not present.


aiida_quantumespresso/workflows/pw/relax.py
```diff
@@ -338,7 +338,8 @@ def run_final_scf(self):
         inputs = self.ctx.final_scf_inputs
         inputs.pw.structure = self.ctx.current_structure
 
-        if self.ctx.current_number_of_bands is not None:
+        bnd = inputs.pw.parameters.get('SYSTEM', {}).get('nbnd', None)
```
mbercx, line 341:
```suggestion
        inputs_nbnd = inputs.pw.parameters.get('SYSTEM', {}).get('nbnd', None)
```

aiida_quantumespresso/workflows/pw/relax.py
```diff
@@ -338,7 +338,8 @@ def run_final_scf(self):
...
+        if self.ctx.current_number_of_bands is not None and bnd is None:
```
mbercx, line 342:
```suggestion
        if self.ctx.current_number_of_bands is not None and inputs_nbnd is None:
```

mbercx (changes requested):
Thanks @Crivella! I just have a minor suggestion to be a bit more explicit in the variable name.

I'd be inclined to agree with this change. In case the user provides a specific `nbnd` for the `final_scf`, I think it's fair to assume they know what they are doing. Silently overriding this input in the logic of the work chain seems like a bad idea. Unless @sphuber objects, I'll go ahead and merge this after the variable name is changed.

PR #709, opened 2021-07-22
Tests: fix immigrate tests after change in `aiida-core`

sphuber:
The nightly build started failing because a recent change in
`aiida-core`. The `CalcJob` now has a new namespace in the metadata
options called `stash`. This means that the options of any process builder
will now include this namespace even if nothing is specified.

This change was causing the immigrate tests to fail since the inputs
options do not have this namespace and so the comparison was failing.


mbercx (approved):
Looks good, thanks @sphuber! I will start taking up my responsibilities as `aiida-quantumespresso` maintainer properly again soon, I promise. 😅

PR #715, opened 2021-08-18
Bugfix when a directory is present in the parent_calc_folder

sponce24:
This should fix the problem of local copy when a directory is present in the parent_calc_folder


sphuber:
Thanks @sponce24 . This is indeed a problem, but the solution provided here doesn't solve the entire problem. It should recursively traverse the parent folder in case it contains multi-nested directories. I'm closing this in favor of #719 which does include this recursive logic.

PR #719, opened 2021-08-27
Modify nameslist for use with  a local repository 

mkotiuga:
Modification of nameslist such that  directories within the local repository are copied with the correct file structure.

### DIFF
--- a/aiida_quantumespresso/calculations/namelists.py
+++ b/aiida_quantumespresso/calculations/namelists.py
@@ -7,6 +7,7 @@ import os
 
 from aiida.common import datastructures, exceptions
 from aiida.orm import Dict, FolderData, RemoteData, SinglefileData
+from aiida.repository.common import FileType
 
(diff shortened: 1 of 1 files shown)


sphuber:
Thanks @mkotiuga . This change seems more correct compared to that of #715 because this properly recurses into the directory in case it contains nested directories. Since multiple people have had this issue, it would be good to add a test. I can help with this if you don't know how to write one or where to add it. I will also open an issue (if it doesn't already exist) on `aiida-core` because it should really just be possible to specify a `FolderData` and have `aiida-core` simply copy over all its contents recursively

sphuber:
As I thought, there already is an open issue: https://github.com/aiidateam/aiida-core/issues/3869. I will try to get this added to `aiida-core==2.0`.

mkotiuga:
Thanks @sphuber  - I would be great to have some help setting up the test (naturally I have some calculations where I have used this feature)

sphuber:
I added native support for this use case in `aiida-core==2.0` which is now released. I am closing this in favor of the simpler implementation in #805

PR #722, opened 2021-08-30
`PwBaseWorkChain`: improve restart from `parent_folder`

### LINKED ISSUE #721: `PwBaseWorkChain`: Allow restarts from `RemoteData` that do not have a `creator`
Currently, the `PwBaseWorkChain` expects the `RemoteData` provided to the `pw.parent_folder` input to have a creator:

https://github.com/aiidateam/aiida-quantumespresso/blob/c49def7a238617439feb75afd0d0cf9fbb681613/aiida_quantumespresso/workflows/pw/base.py#L248-L249

This is then used to set up a `restart_calc` in the context, similar to how actual restarts are done in the work chain. However, there can be cases where the user wants to provide a `RemoteData` that does _not_ have a `PwCalculation` as a creator. Maybe this is a `RemoteData` that is constructed manually or is the result of a `calcfunction`.

Moreover, later in the `prepare_process` step, the input for the `restart_mode` tag is overwritten in case a `restart_calc` is found in the context:

https://github.com/aiidateam/aiida-quantumespresso/blob/c49def7a238617439feb75afd0d0cf9fbb681613/aiida_quantumespresso/workflows/pw/base.py#L418-L420

This means that when providing the `parent_folder` input, `restart_mode` is always set to `'restart'`, overriding the input provided by the user. This means that restarting from _only_ a charge density with `startingpot` is not possible.

mbercx:
Fixes #721 

Remove some of the logic in the `PwBaseWorkChain` regarding restarting
from a previous calculation using a `RemoteData` provided to the
`pw.parent_folder` input.

The current logic expected the `RemoteData` to have a `PwCalculation`
creator, which is not always the case. Moreover, the `restart_mode`
chosen by the user is overriden, which means that e.g. restarting from
_only_ the charge density with `startingpot` is not possible.


mbercx:
@sphuber this was quickly field tested and seems to work fine. Not sure if there is an important reason why the `parent_folder` input is converted into the `restart_calc` stored in the `ctx` that I'm missing?

It's def possible that the user doesn't provide the correct parameters to actually make use of the `parent_folder` though. Maybe we should add some validation here?

(Also possible tests will fail, will fix when we agree)

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -418,9 +415,6 @@ def prepare_process(self):
         if self.ctx.restart_calc:
             self.ctx.inputs.parameters['CONTROL']['restart_mode'] = 'restart'
             self.ctx.inputs.parent_folder = self.ctx.restart_calc.outputs.remote_folder
-        else:
-            self.ctx.inputs.parameters['CONTROL']['restart_mode'] = 'from_scratch'
-            self.ctx.inputs.pop('parent_folder', None)
```
sphuber, line 423: I think we should keep this login. The idea is that if there is not restart calculation, we are not restarting, and so we should probably set `from_scratch`. What would happen now if a user provides a `parent_folder` (so wanting do a restart) but sets the incorrect `restart_mode` in the parameters? I am not sure what the behavior will be and maybe QE will internally ignore it and do the right thing, but the idea was to make sure the correct parameter was set. Or maybe it is not even set at all. It would be good to know what happens when the `restart_mode` is inconsistent with what we are requesting. Depending on that, we might want to add this logic to a port validator instead that warns the user in case of inconsistent inputs. But ideally we still do this automatically, if it is acceptable to always override user input in certain cases.

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -245,9 +245,6 @@ def validate_parameters(self):
         self.ctx.inputs.parameters = self.ctx.inputs.parameters.get_dict()
         self.ctx.inputs.settings = self.ctx.inputs.settings.get_dict() if 'settings' in self.ctx.inputs else {}
 
-        if 'parent_folder' in self.ctx.inputs:
-            self.ctx.restart_calc = self.ctx.inputs.parent_folder.creator
-
```
sphuber, line 250: The only place where `self.ctx.restart_calc` is checked is indeed in `prepare_process`. So it could potentially be fine to remove this, if indeed we do not have to explicitly set the `restart_mode` but as described in the other comment, in some cases we might, when the parameters are inconsistent. Alternatively, we could change the `restart_calc` to a boolean instead of a `CalcJobNode` since really we are merely using it as a flag and don't actually need any info from the creator calcjob.

aiida_quantumespresso/workflows/pw/base.py
mbercx: I think a validator is indeed the way to go here. If a user wants to restart from _only_ the charge density, the provided input for `startingpot` is ignored if `restart_mode` is set to `'restart'`. So we shouldn't automatically set this value and override the user provided inputs.

aiida_quantumespresso/workflows/pw/base.py
mbercx: Hmm, do we even need the `restart_calc` then? Can't we just update the `self.ctx.inputs` whenever a calculation must restart? That way we can also differentiate between different modes of restarting more directly. For example:

1. For the `handle_out_of_walltime` process handler, we can just set the `restart_mode` to `'from_scratch'` in case the structure has changed, and make sure to `pop` any `startingpot` or `startingwfc` inputs.
2. If we want a full restart, we simply set `restart_mode` to `'restart'`.
3. If we want a restart from the charge density, as you would for example want in case the number of bands is changed by `sanity_check_insufficient_bands`, we set the `restart_mode` to `'from_scratch'` and `startingpot` to `'file'`.

Actually, maybe a clean solution would be to have a `restart_mode` setting in the context (perhaps defined as an enum?) that lists the different ways one can restart, and then in the `prepare_process` step this mode is used to set the correct inputs?

mbercx: Also see https://github.com/aiidateam/aiida-quantumespresso/issues/691, which is def relevant here.

sphuber: Yeah, the main point is to centralize the logic for dealing with restarts instead of having to deal with this each time in each handler explicitly. If an enum can cover all cases, that would be great, because that is easy to set. However, this might not be serializable and so may not be possible to be stored in the context.

mbercx:
@sphuber thanks for the feedback! I've willingly introduced some scope creep in this PR to also deal with some of the issues in #691 (not sure if all though, and there are also some comments there that should be raised on the QE GitLab, so I wouldn't say that issue is fixed by this PR at this time.).

I've mainly done two things:

1. Introduced a top-level validator to the `PwCalculation` to check if _any_ correct restart parameters are set when the `parent_folder` input is provided.
2. Removed the `restart_calc` logic in favor of a `set_restart_type` method (thanks to @ramirezfranciscof for the suggestion of making this a method) that takes a `RestartType` enum and the previous calculation to properly set the input parameters and `parent_folder` based on the chosen restart method for each process handler. We've used the `RestartType` enum to indicate the different "restart modes" in order to be consistent with the other `Type`s we have defined and to avoid any confusion with the `restart_mode` input tag of QE.

Finally, some more notes/questions regarding the choice of the restart type:

* `sanity_check_insufficient_bands`: Updated the process handler to restart from the charge density after increasing the number of bands. I see no reason _not_ to restart from the charge density here? It should already be quite close to the correct result.
* `handle_out_of_walltime`: I'm not sure I understand why we cannot do a full restart or _at least_ restart from the charge density in case the walltime has run out and the structure has changed. How would doing a full restart differ from just having more walltime here?
* `handle_relax_recoverable_ionic_convergence_error`: Here I can understand that perhaps a restart from scratch might help with kicking the calculation out of the ionic convergence issue?
* `handle_relax_recoverable_electronic_convergence_error` and `handle_electronic_convergence_not_achieved`: Not sure how to best restart here. Is there a reason not to e.g. restart from the charge density for the ionic case and do a full restart instead in case it's just an SCF?

Also pinging @ramirezfranciscof and @qiaojunfeng for comments and field testing! :)

(will fix tests once the review has processed past the design stage)

bastonero:
I think indeed having a `restart_type` would be nicer, which by default could be `restart_mode='restart'`.
This latter indeed was designed explicitly for restarting calculation which would have taken too much time. So, I think that its proper use is just in the case of 'out of wall time'. In other cases, having this restart mode can be very dramatic. In fact if you want to have an initial density, and just restart from that, this "full" restart will also take the structure from the output folder, which is NOT good. 
Depending on the problem, one should use the "full" restart or the `startingport='file'`+`from_scratch`. In fact, sometimes you already have a well-converged ground-state and you may want to use it, e.g. to speed up calculation with atom displacements for phonons. In this example, if you do a "full" restart, you end up on also reading the structure from the output folder, thus discarding the displaced input structure.
As a consequence, some handler will benefit (or even might be corrected!) from the type of restart. Regarding the out of wall time, one indeed would like to remove the `starting*='file'` and restart fully. 
Regarding the `handle_electronic_convergence_not_achieved`: I also found myself changing slightly the handler. Not sure if one restarts from the previous density how it will affect the scf loop. For sure restarting from scratch is the safest one.

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -302,6 +298,33 @@ def set_max_seconds(self, max_wallclock_seconds):
         max_seconds = max_wallclock_seconds * max_seconds_factor
         self.ctx.inputs.parameters['CONTROL']['max_seconds'] = max_seconds
 
+    def set_restart_type(self, restart_type, calculation):
+        """Set the restart type for the calculation."""
+
+        if restart_type == RestartType.FROM_SCRATCH:
+            self.ctx.inputs.parameters['CONTROL']['restart_mode'] = 'from_scratch'
+            self.ctx.inputs.parameters['ELECTRONS'].pop('startingpot', None)
```
sphuber, line 306: Is the `ELECTRONS` key guaranteed to be set? The `CONTROL` key is explicitly created at the start of the workchain in `validate_parameters`. Maybe we should do the same for `ELECTRONS` (and `SYSTEM` as well while we are at it?) We could then also remove all other occurrences during the workchain of `parameters.setdefault(CARD, {})` and `parameters.get(CARD, {})` as we are guaranteed that it will exist.

sphuber (commented):
Thanks @mbercx . I like the approach a lot, there are just a few minor issues still I think

aiida_quantumespresso/calculations/pw.py
```diff
@@ -152,6 +154,26 @@ def define(cls, spec):
...
+        - Check that the restart input parameters are set when a ``parent_folder`` is provided.
+        """
+        if 'parent_folder' in value:
+
+            parameters = value['parameters'].get_dict()
+            if (
+                parameters.get('CONTROL', {}).get('restart_mode', None) != 'restart' and
+                parameters.get('ELECTRONS', {}).get('startingpot', None) != 'file' and
+                parameters.get('ELECTRONS', {}).get('startingwfc', None) != 'file'
+            ):
```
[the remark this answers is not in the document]
ramirezfranciscof: Just briefly commenting that we may want to also include `parameters['CONTROL']['calculation'] not in {'nscf', 'bands'}` as one of the possibilities. From the [PW input web](https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm883): 

> startingpot = 'file'
> 
> (...) For nscf and bands calculation this is the default
> and the only sensible possibility.

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -233,7 +233,6 @@ def setup(self):
         internal loop.
         """
         super().setup()
-        self.ctx.restart_calc = None
```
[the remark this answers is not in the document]
mbercx: I'm not sure if I agree with this approach. I think it's fine to override inputs when restarting with a handler, but we should try and avoid overriding the initial inputs provided by the user, instead relying on their common sense and the validator. For the first iteration, I don't think we should call `set_restart_type`. For restarts, perhaps we should _always_ call `set_restart_type`, so there is no confusion or incorrect restart?

I do agree with switching to the `parent_folder` as an input instead of `calculation`, though, let me already make that change.

aiida_quantumespresso/calculations/pw.py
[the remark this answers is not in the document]
mbercx: @sphuber the default for `restart_mode` is `'from_scratch'`, which is the required setting when using `startingpot` or `startingwfc`. So only setting `ELECTRONS.startingpot == 'file'` works perfectly fine, I think.

However, @ramirezfranciscof rightly points out that not setting _any_ inputs when running `nscf` or `bands` calculations is also fine, so we shouldn't warn in these cases. Maybe the validation here should be to make sure that the `parent_folder` is actually there, and a non-sensible value for the restart inputs is not set?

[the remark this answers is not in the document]
mbercx: Would it be sensible to enforce that a `parent_folder` is provided for a `nscf` or `bands` calculation?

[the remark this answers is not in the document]
mbercx: Similarly, since running with `ELECTRONS.startingpot == 'file'` is the only sensible choice for these calculations, should we override the inputs and warn that we have done so in case they are set incorrectly?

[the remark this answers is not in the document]
ramirezfranciscof: > I'm a little apprehensive about doing this though. But maybe just raising warnings here isn't enough?

One idea that I had been thinking about lately was to add some sort of input keyword for workchains like `force` (`failsafe = False`?) that allows the user to turn some errors into warnings so that they can fully mess with the workchain taking responsability of the results. Not sure if this is the place to discuss this as a more general approach, but your comment reminded me of this.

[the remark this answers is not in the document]
sphuber: > Would it be sensible to enforce that a parent_folder is provided for a nscf or bands calculation?

I think this is necessary no? Does it ever make sense to run either of these without a `parent_folder`?

> should we override the inputs and warn that we have done so in case they are set incorrectly? I'm a little apprehensive about doing this though. But maybe just raising warnings here isn't enough?

I would say that if we _know_ that for certain calculation types certain inputs _have_ to be set, than we should do that. If there are settings that are likely incorrect, but could still make sense in certain edge cases, then a warning might be better.

I think there is still a problem with the conditional though. If `calculation == 'scf'` _and_ there is a `parent_folder` you _only_ warn if _all_ three parameters are incorrectly set (you are using `and`). But shouldn't this be `or` statements? If _any_ of the parameters are incorrect, it should warn?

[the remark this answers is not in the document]
mbercx: > I would say that if we know that for certain calculation types certain inputs have to be set, than we should do that. If there are settings that are likely incorrect, but could still make sense in certain edge cases, then a warning might be better.

Again, I would be in favor of just raising errors for non-sensible inputs instead of just overriding them. If the user somehow has set wrong inputs, this might indicate a problem in their workflow they would want to be aware of, instead of having it silently fixed in the work chain. In this light, I may change the warnings for `nscf` and `nbands` to actually raise errors. If there _is_ some edge case that we missed, we can always revert this later with a warning that clearly explains this use case.

> I think there is still a problem with the conditional though. If calculation == 'scf' and there is a parent_folder you only warn if all three parameters are incorrectly set (you are using and). But shouldn't this be or statements? If any of the parameters are incorrect, it should warn?

I think the logic is inverted (sort of): In case `parent_folder` is set, _at least_ one of `restart_mode == 'restart'`, `startingpot == 'file'` or `startingwfc == 'file'` must be set for an `scf` calculation. Perhaps it would be clearer to change:

```Python
if (
    parameters.get('CONTROL', {}).get('CONTROL', None) != 'restart' and
    parameters.get('ELECTRONS', {}).get('startingpot', None) != 'file' and
    parameters.get('ELECTRONS', {}).get('startingwfc', None) != 'file'
):
```

into 

```Python
if not any([
    parameters.get('CONTROL', {}).get('CONTROL', None) == 'restart' and
    parameters.get('ELECTRONS', {}).get('startingpot', None) == 'file' and
    parameters.get('ELECTRONS', {}).get('startingwfc', None) == 'file'
]):
```

We should consider if there are other reasons to provide a `parent_folder`, though. 🤔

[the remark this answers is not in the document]
mbercx: > One idea that I had been thinking about lately was to add some sort of input keyword for workchains like force (failsafe = False?) that allows the user to turn some errors into warnings so that they can fully mess with the workchain taking responsability of the results. Not sure if this is the place to discuss this as a more general approach, but your comment reminded me of this.

@sphuber is there a way to disable _all_ validation? As long as we carefully stick to using validators to make sure inputs are sensible (and not just override them), we can disable validation and the user can go where there be dragons. ^^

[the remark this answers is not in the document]
sphuber: No, currently there is no way to do this dynamically through the inputs. Might be worth a requested feature discussion on `aiida-core`, but it may have only niche use

mbercx:
@sphuber I think this is ready for another round of review. To summarize the logic of these changes a bit again:

1. Instead of setting the input tags for the restart when a `parent_folder` is provided, we trust the user to set the restart settings she/he desires.
2. A validator is added instead, that checks if the inputs are sensible based on the `calculation`. If the inputs are simply incorrect (e.g. no `parent_folder` for an "nscf" calculation), this validator will raise. If a `parent_folder` is provided but no restart inputs are set, a warning is raised.
3. We added a `set_restart_type` method to the work chain, which sets the correct input tags based on the provided `RestartType` and `parent_folder` (if not restarting from scratch). This is exclusively used in the `process_handler`s of the `PwBaseWorkChain`.
4. The following restart types are used for each process handler:
  * `sanity_check_insufficient_bands`: `FROM_CHARGE_DENSITY`
  * `handle_unrecoverable_failure`: *no restart*
  * `handle_known_unrecoverable_failure`: *no restart*
  * `handle_out_of_walltime`: `FULL` if the structure hasn't changed, `FROM_SCRATCH` otherwise
  * `handle_vcrelax_converged_except_final_scf`: *no restart*
  * `handle_relax_recoverable_ionic_convergence_error`: `FROM_SCRATCH`
  * `handle_relax_recoverable_electronic_convergence_error`: `FROM_SCRATCH`
  * `handle_electronic_convergence_not_achieved`: `FULL`
  * `handle_electronic_convergence_warning`: *no restart*

I would still consider making the following changes to [4]:
* `handle_out_of_walltime`: restart `FROM_CHARGE_DENSITY` in case structure has changed.
* `handle_relax_recoverable_ionic_convergence_error`: restart `FROM_CHARGE_DENSITY`

mbercx:
@qiaojunfeng or @ramirezfranciscof maybe you'll have a chance to review this some time this week? If we all think the logic is solid, I can fix the tests and we can merge this for LUMI.

aiida_quantumespresso/calculations/pw.py
```diff
@@ -152,6 +155,38 @@ def define(cls, spec):
...
+        # Check that the restart input parameters are set correctly
+        if calculation_type in ('nscf', 'bands'):
+            if 'parent_folder' not in value:
+                return f'`parent_folder` not provided for `{calculation_type}` calculation.'
+            if parameters.get('ELECTRONS', {}).get('startingpot', 'file') != 'file':
+                return f'`startingpot` should be set to `file` for a `{calculation_type}` calculation.'
+            if parameters.get('CONTROL', {}).get('CONTROL', None) == 'restart':
+                return f'`restart_mode` should be set to `from_scratch` for a `{calculation_type}` calculation.'
+        else:
+            if 'parent_folder' in value:
```
sphuber, line 179:
```suggestion
        elif 'parent_folder' in value:
```

aiida_quantumespresso/common/types.py
```diff
@@ -31,3 +31,12 @@ class SpinType(enum.Enum):
...
     SPIN_ORBIT = 'spin_orbit'
+
+
+class RestartType(enum.Enum):
+    """Enumeration of ways to restart a calculation in Quantum ESPRESSO."""
+
+    FROM_SCRATCH = 'from_scratch'
+    FULL = 'full'
+    FROM_CHARGE_DENSITY = 'from_charge_density'
+    FROM_WAVE_FUNCTIONS = 'from_wave_functions'
```
sphuber, line 42: Feel free to ignore the following, but looks better aesthetically and order feels more natural as to what they mean. Still I recognize the nitpickiness and subjectiveness degree of this suggestion ^^
```suggestion
    FULL = 'full'
    FROM_SCRATCH = 'from_scratch'
    FROM_CHARGE_DENSITY = 'from_charge_density'
    FROM_WAVE_FUNCTIONS = 'from_wave_functions'
```

aiida_quantumespresso/calculations/pw.py
```diff
@@ -152,6 +155,38 @@ def define(cls, spec):
...
+                if not any([
+                    parameters.get('CONTROL', {}).get('CONTROL', None) == 'restart' and
+                    parameters.get('ELECTRONS', {}).get('startingpot', None) == 'file' and
+                    parameters.get('ELECTRONS', {}).get('startingwfc', None) == 'file'
+                ]):
```
sphuber, line 184: Don't think this does what you mean it too. Why make a list if you will get only a single boolean because of the `and` operators? Did you mean to use comma's instead of `and`? Or you just remove the `not any([])` from the conditional.

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -302,6 +300,36 @@ def set_max_seconds(self, max_wallclock_seconds):
         max_seconds = max_wallclock_seconds * max_seconds_factor
         self.ctx.inputs.parameters['CONTROL']['max_seconds'] = max_seconds
 
+    def set_restart_type(self, restart_type, parent_folder=None):
+        """Set the restart type for the next iteration."""
+
+        if parent_folder is None and restart_type != RestartType.FROM_SCRATCH:
+            raise ValueError('When not restarting from scratch, a `parent_folder` must be provided.')
```
sphuber, line 307: we are duplicating code from the `validate_inputs` but I think it is ok in this case. Just something to keep in mind if the validation logic changes. Maybe add a comment here that this is doubling checks from the validator

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -440,7 +456,8 @@ def sanity_check_insufficient_bands(self, calculation):
 
         Verify that the occupation of the last band is below a certain threshold, unless `occupations` was explicitly
         set to `fixed` in the input parameters. If this is violated, the calculation used too few bands and cannot be
-        trusted. The number of bands is increased and the calculation is restarted, starting from the last.
+        trusted. The number of bands is increased and the calculation is restarted, using the
+        charge density from the previous calculation.
```
sphuber, line 460:
```suggestion
        trusted. The number of bands is increased and the calculation is restarted, using the charge density
        from the previous calculation.
```

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -501,14 +522,20 @@ def handle_known_unrecoverable_failure(self, calculation):
         PwCalculation.exit_codes.ERROR_OUT_OF_WALLTIME,
     ])
     def handle_out_of_walltime(self, calculation):
-        """Handle `ERROR_OUT_OF_WALLTIME` exit code: calculation shut down neatly and we can simply restart."""
+        """Handle `ERROR_OUT_OF_WALLTIME` exit code.
+
+        In this case the calculation shut down neatly and we can simply restart. We consider two cases:
+
+        1. If the structure is unchanged, we do a full restart.
+        2. If the structure has changed during the calculation, we restart from scratch.
```
sphuber, line 530: Since this corresponds to a clean exit, wouldn't we be able to reuse the charge density at least, even if the structure was changed? I know you didn't change anything in the logic, but I think there is no reason _not_ to reuse the charge density, is there?

aiida_quantumespresso/calculations/pw.py
```diff
@@ -152,6 +155,38 @@ def define(cls, spec):
...
+            if 'parent_folder' not in value:
+                return f'`parent_folder` not provided for `{calculation_type}` calculation.'
+            if parameters.get('ELECTRONS', {}).get('startingpot', 'file') != 'file':
+                return f'`startingpot` should be set to `file` for a `{calculation_type}` calculation.'
+            if parameters.get('CONTROL', {}).get('CONTROL', None) == 'restart':
+                return f'`restart_mode` should be set to `from_scratch` for a `{calculation_type}` calculation.'
+        else:
+            if 'parent_folder' in value:
+                if not any([
+                    parameters.get('CONTROL', {}).get('CONTROL', None) == 'restart' and
```
sphuber, line 181: Should probably be
```suggestion
                    parameters.get('CONTROL', {}).get('calculation', None) == 'restart' and
```
There are quite a few weird things/bugs in the validator. Would be good to add some unit tests for it. Those are easy to write and cheap to run anyway

aiida_quantumespresso/calculations/pw.py
```diff
@@ -152,6 +155,38 @@ def define(cls, spec):
...
+        parameters = value['parameters'].get_dict()
+        calculation_type = parameters.get('CONTROL', {}).get('calculation', 'scf')
+
+        # Check that the restart input parameters are set correctly
+        if calculation_type in ('nscf', 'bands'):
+            if 'parent_folder' not in value:
+                return f'`parent_folder` not provided for `{calculation_type}` calculation.'
+            if parameters.get('ELECTRONS', {}).get('startingpot', 'file') != 'file':
+                return f'`startingpot` should be set to `file` for a `{calculation_type}` calculation.'
+            if parameters.get('CONTROL', {}).get('CONTROL', None) == 'restart':
```
sphuber, line 176: Should probably be
```suggestion
            if parameters.get('CONTROL', {}).get('calculation', None) == 'restart':
```

aiida_quantumespresso/calculations/pw.py
```diff
@@ -152,6 +155,38 @@ def define(cls, spec):
...
+                return f'`restart_mode` should be set to `from_scratch` for a `{calculation_type}` calculation.'
```
sphuber, line 177: Think this should be
```suggestion
                return f'`restart_mode` should be set to `restart` for a `{calculation_type}` calculation.'
```
or otherwise the check doesn't make any sense

sphuber (changes requested):
Thanks @mbercx . Concept looks good to me, just some minor changes and then the validator contains quite a few problems, so would really like to see some unit tests there. Also, we are essentially coding the default values from QE in there. This may become a problem when they change. Not sure how likely that is and also don't really have an alternative solution for now, but just something we should be aware of.

aiida_quantumespresso/calculations/pw.py
mbercx: Whoops 😅! I wanted to have commas indeed. Basically, if none of these are set for calculation that isn't `nscf` or `bands`, providing a `parent_folder` will have no effect, since QE's defaults are not set to restart in this case.

aiida_quantumespresso/common/types.py
mbercx: image

aiida_quantumespresso/workflows/pw/base.py
mbercx: Hmm, actually, this is not really duplicating code I think, since the `set_restart_type` method is not used to set the restart settings for the initial inputs, only for restarts via `process_handler`s. 

So it can be either considered superfluous or a good check on our use of the method when recovering from errors. ^^

aiida_quantumespresso/workflows/pw/base.py
mbercx: Myyyes, concur with your reasoning, I do. See the final two lines of [this comment](https://github.com/aiidateam/aiida-quantumespresso/pull/722#issuecomment-916835694). I also think that for `handle_relax_recoverable_ionic_convergence_error` we can restart from the charge density.

aiida_quantumespresso/calculations/pw.py
mbercx: Wowza, that seems like some pretty rotten copy 🍝 . I will have to spending extra time in the unit-test-writing corner for these. 😓

aiida_quantumespresso/calculations/pw.py
mbercx: Actually, I think it does?

With the correct key being requested in the CONTROL namespace:

```Python
            if parameters.get('CONTROL', {}).get('restart_mode', None) == 'restart':
                return f'`restart_mode` should be set to `from_scratch` for a `{calculation_type}` calculation.'
```

It will basically check if the `restart_mode` has been set to `'restart'`.  This may sound intuitively correct, but looking at the [pw.x input description](https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm60):

```
restart' :
From previous interrupted run. Use this switch only if you want to
continue, using the same number of processors and parallelization,
an interrupted calculation. Do not use to start a new one, or to
perform a non-scf calculations.  Works only if the calculation was
cleanly stopped using variable max_seconds, or by user request
with an "exit file" (i.e.: create a file "prefix".EXIT, in directory
"outdir"; see variables prefix, outdir). The default for
startingwfc and startingpot is set to 'file'.
```

So in order to properly start from the charge density (as required by `nscf` and `bands` calculations), you have to either not set this tag, or set it explicitly to `'from_scratch'` (the default).

mbercx:
Much obliged for the kind review, @sphuber! Nits have been picked, copy 🍝 has been untangled, and tests have been fixed and added for the validation.

Extra notes on changes: 

* also dealing with the following deprecation:

```
DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
```

* Removed the `validate_parameters` step in the outline and moved the code into the `setup` step`, since it seems to have a better home there, i.e. nothing seems to be validated?

Final thing to do is to change the logic of the error handlers when the structure changes, on [which I think we agree](https://github.com/aiidateam/aiida-quantumespresso/pull/722#discussion_r712578140). Just a confirmation that you agree we should restart from the charge density for both:

* `handle_out_of_walltime`: restart `FROM_CHARGE_DENSITY` in case structure has changed.
* `handle_relax_recoverable_ionic_convergence_error`: restart `FROM_CHARGE_DENSITY` 

And I'll update the logic/tests.

aiida_quantumespresso/calculations/pw.py
ramirezfranciscof: Why not check directly that `parameters.get('CONTROL', {}).get('restart_mode', 'from_scratch') != 'from_scratch'`? It catches more possible problems, no? (i.e. `restart_mode = typo`)

aiida_quantumespresso/workflows/pw/base.py
sphuber: Would the fact that the cell might have changed be a potential problem? Or since the charge density would also have been written on the new grid, that wouldn't matter? Not sure here

aiida_quantumespresso/calculations/pw.py
sphuber: Yeah, you are correct, but I misread because I was expecting a negative conditional like @ramirezfranciscof suggests. That makes more sense and also catches other problems if it is accidentally to set any other value

tests/conftest.py
```diff
@@ -515,18 +516,27 @@ def _generate_inputs_pw():
...
+                    'CONTROL': {
+                        'calculation': 'scf'
+                    },
+                    'SYSTEM': {
+                        'ecutrho': 240.0,
+                        'ecutwfc': 30.0
+                    },
+                    'ELECTRONS': {
+                        'electron_maxstep': 60,
+                    }
```
ramirezfranciscof, line 537: What happened here with the auto-formatter? This is crazy...

mbercx: Haha, right, didn't notice. Let's see if it keeps being difficult.

aiida_quantumespresso/workflows/pw/base.py
mbercx: But the charge density is just a 3 dimenional scalar function right? The issue with the cell changing is that the basis set is no longer complete, but I would say that this gets regenerated anyways and then the input charge density is expressed in this new basis set? 🤔

aiida_quantumespresso/calculations/pw.py
mbercx: Right, makes sense. More readable and functional, love it! 👌

tests/conftest.py
mbercx: Got it to behave! 😁

mbercx:
@ramirezfranciscof @sphuber thanks once more for your reviewing efforts! The only question that now remains is the logic regarding starting from a changed structure. However, this was once implemented by someone, perhaps for a good reason, and since we not _sure_ that starting from the charge density is a good approach, perhaps we should declare this logic change "beyond the scope of this PR" and open an issue for revisiting all the restart types?

ramirezfranciscof (approved):
All looks good to me, I mainly checked the logics related to the restarting part, I took a look at the testing for the process handlers but that part I understand a bit less, but I assume @sphuber 's review of it should suffice.

aiida_quantumespresso/calculations/pw.py
qiaojunfeng: Wait, I am a bit confused, for nscf calculation if I only set `restart_mode = restart` without any `startingpot` and `startingwfc`, it should also work, but this validator will invalidate this usage?

tests/calculations/test_pw.py
```diff
@@ -250,3 +250,67 @@ def test_pw_parallelization_duplicate_cmdline_flag(fixture_sandbox, generate_cal
...
+    parameters = inputs['parameters'].get_dict()
+    parameters['CONTROL']['calculation'] = calculation
+
+    if calculation in ('nscf', 'bands'):
+
+        # No parent_folder -> raise
+        inputs['parameters'] = orm.Dict(dict=parameters)
+        with pytest.raises(ValueError) as exc:
+            generate_calc_job(fixture_sandbox, entry_point_name, inputs)
+        assert f'`parent_folder` not provided for `{calculation}` calculation.' in str(exc.value)
```
sphuber, line 273: If you want to check the exception message, the following is recommended
```suggestion
        with pytest.raises(ValueError, match=r'`parent_folder` not provided for `.*` calculation.'):
            generate_calc_job(fixture_sandbox, entry_point_name, inputs)
```

tests/calculations/test_pw.py
```diff
@@ -250,3 +250,67 @@ def test_pw_parallelization_duplicate_cmdline_flag(fixture_sandbox, generate_cal
...
+):
+    """Test the input validation of restart settings for the ``PwCalculation``."""
+    remote_data = generate_remote_data(computer=fixture_localhost, remote_path='/cat_stevens')
+    entry_point_name = 'quantumespresso.pw'
+
+    inputs = generate_inputs_pw()
+    parameters = inputs['parameters'].get_dict()
+    parameters['CONTROL']['calculation'] = calculation
+
+    if calculation in ('nscf', 'bands'):
```
sphuber, line 267: Since you essentially have two types of test, I would just make two separate parametrized tests:
`test_pw_validate_inputs_calculation_base` and `test_pw_validate_inputs_calculation_nscf`

tests/conftest.py
```diff
@@ -672,6 +694,7 @@ def _generate_workchain_pdos():
         nscf_pw_inputs['parameters']['CONTROL']['calculation'] = 'nscf'
         nscf_pw_inputs['parameters']['SYSTEM']['occupations'] = 'tetrahedra'
         nscf_pw_inputs['parameters']['SYSTEM']['nosym'] = True
+        nscf_pw_inputs['parent_folder'] = generate_remote_data(computer=fixture_localhost, remote_path='/daddy_is_here')
```
sphuber, line 697: everything alright there Hoss?

tests/workflows/pw/test_base.py
```diff
@@ -40,29 +39,59 @@ def test_handle_unrecoverable_failure(generate_workchain_pw):
...
+    assert result.do_break
+
+    result = process.inspect_process()
+    assert result.status == 0
+
+
+def test_handle_electronic_convergence_not_achieved(generate_workchain_pw, fixture_localhost, generate_remote_data):
     """Test `PwBaseWorkChain.handle_electronic_convergence_not_achieved`."""
-    process = generate_workchain_pw(exit_code=PwCalculation.exit_codes.ERROR_ELECTRONIC_CONVERGENCE_NOT_REACHED)
+    remote_data = generate_remote_data(computer=fixture_localhost, remote_path='/come_to_papa')
```
sphuber, line 79: ohhh boy

aiida_quantumespresso/calculations/pw.py
sphuber: If `calculation == 'nscf'` it won't get in this part of the conditional though. And even then, according to the docs, `restart_mode = restart` is non-sensical for `nscf` and one should use `from_scratch`.

sphuber (changes requested):
Sorry to bother again, but since code got added, I have a few additional small comments

tests/workflows/pw/test_base.py
mbercx: At 3 am I though this was quite funny. 😅

ramirezfranciscof: I would try to keep the jokes to a PG rating though, haha.

aiida_quantumespresso/calculations/pw.py
qiaojunfeng: Thanks!
I tried to look into QE code, indeed the `restart_mode = restart` tag is aiming to restart from an unfinished but correctly stopped previous calculation (it could read additional temp files during scf iteration). When the previous calculation successfully finished (and correctly stopped), the `calculation == 'nscf'` will default `startingpot` to `file` and `startingwfc` to `atomic+random`, in this case although the `restart_mode = restart` cannot find the additional temp files during scf iteration, it however does not harm and QE keeps running nscf. 
So my previous nscf are always using `restart_mode = restart`, they work but in theory I should use `restart_mode = from_scratch` together with `startingpot = file`.

mbercx: I see, thanks for the comments @qiaojunfeng! So basically the following validation is in principle not necessary:

```python
            if parameters.get('CONTROL', {}).get('restart_mode', 'from_scratch') != 'from_scratch':
                return f'`restart_mode` should be set to `from_scratch` for a `{calculation_type}` calculation.'
```

Since QE will _only_ restart from the `startingpot` file anyways, even though `restart_mode` is set to `'restart'`? Still, would probably be best to nudge users to just setting the correct input tags according to the documentation.

aiida_quantumespresso/workflows/pw/base.py
sphuber: Should be a scalar yes, but it is defined on a grid that relates to the crystal cell. If that changes but the grid of the charge density doesn't in a comparable way, they are effectively out of sync, no? However, I agree with your statement further down that we maybe not change this behavior now and keep what was there. Then we can address this in a follow-up PR and investigate it properly first

aiida_quantumespresso/calculations/pw.py
sphuber: This is what I wrote in an original comment (which I can't find any longer) that before we wouldn't really raise an error if there were "inconsistent" parameters defined, and the workchain would simply set what is correct, even if those inconsistent parameters would work fine in QE because it would ignore certain conflicting values. I can see how there is something to be said for warning users for inconsistent input, but maybe we should not start raising errors like we currently are doing for `nscf` and `bands` calculations.

qiaojunfeng: > ```python
>             if parameters.get('CONTROL', {}).get('restart_mode', 'from_scratch') != 'from_scratch':
>                 return f'`restart_mode` should be set to `from_scratch` for a `{calculation_type}` calculation.'
> ```

Yes it is not necessary. For `calculation = 'nscf'`, `restart_mode = 'restart'` also works, and remove `restart_mode` can work as well. Maybe if the user has set a `restart_mode = 'restart'` for nscf, we just print a warning instead of an error?

mbercx: > This is what I wrote in an original comment (which I can't find any longer) that before we wouldn't really raise an error if there were "inconsistent" parameters defined, and the workchain would simply set what is correct, even if those inconsistent parameters would work fine in QE because it would ignore certain conflicting values.

I still will object to silently "fixing" inconsistent inputs on the fly out of principle. 🙃 If the user is providing non-sensible inputs, something is wrong with how they are setting up the calculations, and they should be made aware about that.

I'm happy to raise a warning for the final conditional, since the calculation will still run as intended. The first two will not run, if I understand correctly, so these should raise.

sphuber: > I still will object to silently "fixing" inconsistent inputs on the fly out of principle. upside_down_face If the user is providing non-sensible inputs, something is wrong with how they are setting up the calculations, and they should be made aware about that.

Normally I would agree, but Quantum ESPRESSO does the same thing. For that reason, lot's of people will actually specify `restart_mode = restart` for an NSCF calculation, because that is what makes sense intuitively and it works, because QE just silently ignores it. In this case I think it is probably not a good idea if this starts breaking stuff on our side all of a sudden.

> I'm happy to raise a warning for the final conditional, since the calculation will still run as intended. The first two will not run, if I understand correctly, so these should raise.

Seems like the golden middle road. If you add this final change we can merge.

sphuber (dismissed):
Bombs away!

mbercx:
Darn, rebased 5 seconds too late! 😭

sphuber:
> Darn, rebased 5 seconds too late! sob

you mean too early ^^

sphuber:
Uhm, just merged another PR and forgot this was waiting. We can just squash merge this though with running tests again, I just updated the `README.md`.

sphuber:
> If you want to check the result without merging, you can always go to the branch of your PR, right?

I did and it looked fine. But I realize now that I forgot to add empty rows for the Python version of older releases of `aiida-quantumespresso` which I don't think specified explicit python requirements. I was anyway going to add another change to add shields for QE as well.

sphuber:
@mbercx look at [this deliciousness](https://github.com/aiidateam/aiida-quantumespresso/tree/fix/readme-qe-shields)

mbercx:
As I am too shy to merge with admin powers, I once again seek your approval @sphuber. 🙃

sphuber (approved):

aiida_quantumespresso/common/types.py
```diff
@@ -31,3 +31,12 @@ class SpinType(enum.Enum):
...
     SPIN_ORBIT = 'spin_orbit'
+
+
+class RestartType(enum.Enum):
+    """Enumeration of ways to restart a calculation in Quantum ESPRESSO."""
+
+    FULL = 'full'
+    FROM_SCRATCH = 'from_scratch'
+    FROM_CHARGE_DENSITY = 'from_charge_density'
+    FROM_WAVE_FUNCTIONS = 'from_wave_functions'
```
bastonero, line 42: Hi guys, I wanted just to point out that it could be useful to have also `FROM_FILES` (or something like that), so that one can restart from the density and wave functions. I know that 'full' is meant to do so, but it will also read the atomic positions.

mbercx: My first inclination would be to say that restarting from the wave functions, it should be pretty fast for QE to recalculate the charge density. But if I remember correctly, QE doesn't actually do this, instead just plugging in the wave functions but for the potential calculated from the atomic charge density, correct?

bastonero: Yes, should be like this. Two case scenario would be:
- electric enthalpy routine: it uses the wfcs to build the polarization operator (essentially you start with a slightly different hamiltonian)
- maybe in some restart when vc-relaxing?

But I do agree at the end it is not such a difference. I was just wondering, since the new implementation is so cool (great job!!! :D ), it would be quite easy to just add that one more, just in case one needs it. 
Or do you think that the "experienced" user can still always tweak the inputs if that is the case?

mbercx: Thanks for the clarification, @bastonero! Note that the restart types are only used for setting the restarts after an error has been handled using the `set_restart_type` method. So if a user wants to restart from a previous calculation, the correct inputs have to be provided, not the restart type. 

Happy to implement another restart type (not sure about the name, `FROM_FILES` seems a bit too general, maybe `FROM_CHARGE_AND_WFC`?). Do you think there is already a current error handler where this restart type would be used though?

bastonero: Ah ok, I see, thanks! For the moment, actually, I still don't have an example of handler where it would be useful. `FROM_CHARGE_AND_WFC` sounds good anyway.

PR #725, opened 2021-09-10
`Protocols`: Add settings from `overrides`

mbercx:
Any settings passed to the `overrides` argument of the
`get_builder_from_protocol()` method are currently not passed to the
builder that is returned by the method. Here we fix this issue by adding
them from the inputs that are merged from the protocol inputs and the
overrides.


sphuber:
Code is failing at the line you are adding. Probably since `settings` is optional, you cannot assume it is there like you do now. Should add a conditional.

Also, why is the commit attributed to me?

mbercx:
@sphuber yes, I noticed when I looked at the tests, will fix and add a test.

I actually added this on `theossrv5` because I was testing the new environment there, but forgot to realise the authentication there is still on your GitHub account. Will move the development to my work station so I'm not hijacking your GitHub account. 😅

sphuber (approved):

PR #727, opened 2021-09-10
`PwParser`: Fix magnetic moments for QE v6.8

### LINKED ISSUE #726: `PwParser`: Parsing of magnetic moments fails for QE v6.8
After doing a test run with QE v6.8 on eiger with spin-polarisation, it seems that the parsing of the magnetic moments fails:

```
+-> CRITICAL at 2021-09-10 11:54:54.744964+00:00
 |   File "/home/aiida/envs/aiida-lumi/code/aiida-quantumespresso/aiida_quantumespresso/parsers/pw.py", line 324, in parse_stdout
 |     parsed_data, logs = parse_stdout(stdout, parameters, parser_options, parsed_xml)
 |   File "/home/aiida/envs/aiida-lumi/code/aiida-quantumespresso/aiida_quantumespresso/parsers/parse_raw/pw.py", line 640, in parse_stdout
 |     line2 = data_step[count + j]
 | IndexError: list index out of range
+-> ERROR at 2021-09-10 11:54:54.747584+00:00
+-> WARNING at 2021-09-10 11:54:54.749605+00:00
```

Closer inspection reveals that this is most likely due to a change in format of the corresponding output. This is the format for the magnetic moments in v6.8:

```
     Magnetic moment per site  (integrated on atomic sphere of radius R)
     atom   1 (R=0.292)  charge=  7.9920  magn= -0.0000
     atom   2 (R=0.292)  charge=  1.2243  magn=  0.0000
```

Compare that to the following example from v6.7:

```
     Magnetic moment per site:
     atom:    1    charge:    0.4485    magn:    0.0000    constr:    0.0000
     atom:    2    charge:    0.4485    magn:    0.0000    constr:    0.0000
     atom:    3    charge:    4.1095    magn:   -0.0000    constr:    0.0000
     atom:    4    charge:    4.1095    magn:   -0.0000    constr:    0.0000
```

This messes up the current parsing:

https://github.com/aiidateam/aiida-quantumespresso/blob/c49def7a238617439feb75afd0d0cf9fbb681613/aiida_quantumespresso/parsers/parse_raw/pw.py#L641-L643

mbercx:
Fixes #726 

Due to a change in the format of the output of the magnetic moments per
site, the parsing of spin-polarised `pw.x` calculations was broken.

Here we add a few lines of parsing that fixes the issue for v6.8.
Basically, if the line doesn't contain `atom:` with a colon, the parser
now also checks if `atom` is there, and parses the atomic charges and
magnetic moments correctly in this case.


mbercx:
Hmm, seems the parsing still leads to exceptions in case a run is interrupted:

```
+-> REPORT at 2021-09-10 18:40:06.375625+00:00
 | [88175|PwCalculation|on_except]: Traceback (most recent call last):
 |   File "/home/aiida/.virtualenvs/aiida-lumi/lib/python3.8/site-packages/plumpy/process_states.py", line 230, in execute
 |   File "/home/aiida/.virtualenvs/aiida-lumi/lib/python3.8/site-packages/aiida/engine/processes/calcjobs/calcjob.py", line 388, in parse
 |     exit_code_retrieved = self.parse_retrieved_output(retrieved_temporary_folder)
 |   File "/home/aiida/.virtualenvs/aiida-lumi/lib/python3.8/site-packages/aiida/engine/processes/calcjobs/calcjob.py", line 468, in parse_retrieved_output
 |   File "/home/aiida/envs/aiida-lumi/code/aiida-quantumespresso/aiida_quantumespresso/parsers/pw.py", line 65, in parse
 |     trajectory = self.build_output_trajectory(parsed_trajectory, structure)
 |   File "/home/aiida/envs/aiida-lumi/code/aiida-quantumespresso/aiida_quantumespresso/parsers/pw.py", line 425, in build_output_trajectory
 |     trajectory.set_trajectory(
 |   File "/home/aiida/.virtualenvs/aiida-lumi/lib/python3.8/site-packages/aiida/orm/nodes/data/array/trajectory.py", line 137, in set_trajectory
 |     self._internal_validate(stepids, cells, symbols, positions, times, velocities)
 |   File "/home/aiida/.virtualenvs/aiida-lumi/lib/python3.8/site-packages/aiida/orm/nodes/data/array/trajectory.py", line 64, in _internal_validate
 |     raise ValueError('TrajectoryData.cells must have shape (s,3,3), with s=number of steps')
 | ValueError: TrajectoryData.cells must have shape (s,3,3), with s=number of steps
```

mbercx:
It doesn't seem to be related to the magnetic moments though. Strange that the `cells` don't have the correct shape, I don't seem to see a difference in this output in v6.8.

mbercx:
Found the culprit:

```
     Error in routine bfgs (1):
     dE0s is positive which should never happen

     stopping ...
00   0.000000000   8.465436411

ATOMIC_POSITIONS (angstrom)
Nb            1.8368309585        0.0000000000        1.8536657717
Nb            0.0000000000        1.8368309585        6.6117706388
Si            0.0000000000        0.0000000000        0.0000000000
Si            1.8368309585        1.8368309585        0.0000000000
Sb            1.8368309585        0.0000000000        5.4797314139
Sb            0.0000000000        1.8368309585        2.9857049967
```

Seems that when this error appears, the printing of the `CELL_PARAMETERS` is somehow messed up? You can still see the last parameters being printed right after the error. This causes there to be one step missing in the `lattice_vectors_relax`. This seems to be a separate problem though, so I'll raise a separate issue.

mbercx:
See https://github.com/aiidateam/aiida-quantumespresso/issues/728. I think this PR is ready for review. :)

aiida_quantumespresso/parsers/parse_raw/pw.py
```diff
@@ -641,6 +641,9 @@ def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None)
                         if 'atom:' in line2:
                             mag_moments.append(float(line2.split('magn:')[1].split()[0]))
                             charges.append(float(line2.split('charge:')[1].split()[0]))
+                        elif 'atom' in line2:
```
sphuber, line 644: Please add a comment here that this is for >= v6.8

tests/parsers/fixtures/pw/magnetic_moments_v68/aiida.out
```diff
@@ -0,0 +1,835 @@
+
+     Program PWSCF v.6.8 starts on 10Sep2021 at 18:59:30 
+
+     This program is part of the open-source Quantum ESPRESSO suite
```
sphuber, line 4: Think we can snip the kpoint information and the timing at the end of this file to reduce the new lines added somewhat.

tests/parsers/test_pw.py
```diff
@@ -937,3 +937,25 @@ def test_pw_vcrelax_failed_not_converged_nstep(
     assert 'output_parameters' in results
     assert 'output_structure' in results
     assert 'output_trajectory' in results
+
+
+def test_magnetic_moments_v68(
+    fixture_localhost, generate_calc_job_node, generate_parser, generate_inputs, data_regression
+):
+    """Test the parsing of the magnetic moments in QE v6.8."""
```
sphuber, line 945:
```suggestion
    """Test the parsing of the magnetic moments in QE v6.8 from stdout.
    
    The format of the output changed in v6.8 and so needs a special conditional in the parser.
    """
```

sphuber (changes requested):
Thanks @mbercx . Just a few minor things

tests/parsers/test_pw.py
```diff
@@ -937,3 +937,25 @@ def test_pw_vcrelax_failed_not_converged_nstep(
...
+    inputs = generate_inputs(calculation_type='relax', metadata={'options': {'without_xml': True}})
+    node = generate_calc_job_node(entry_point_calc_job, fixture_localhost, name, inputs)
+    parser = generate_parser(entry_point_parser)
+    results, calcfunction = parser.parse_from_node(node, store_provenance=False)
+
+    assert calcfunction.is_finished_ok, calcfunction.exit_message
+    assert 'output_trajectory' in results
+
+    data_regression.check({
+        'output_trajectory': results['output_trajectory'].attributes,
```
qiaojunfeng, line 960: All things good! Just one question out of curiosity: we only check the `output_trajectory.attributes`, but not the parsed actual data of the attributes? sth. like checking `output_trajectory.get_array('atomic_magnetic_moments')`?

qiaojunfeng (approved):

tests/parsers/test_pw.py
mbercx: It's a fair question! I figured that the main point is to see if the parsing fails or not, and as long as the size of the arrays returned is still correct, we're probably in good shape. Also, numpy arrays are a problem for the regression check, it seems:

```
FAILED test_pw.py::test_magnetic_moments_v68 - yaml.representer.RepresenterError: ('cannot represent an object', array([[7.9805, 1.2176]]))
```

Probably because they aren't readily serialisable into YAML format. Of course, we can convert them into lists:

```Python
    data_regression.check({
        'atomic_charges': results['output_trajectory'].get_array('atomic_charges').tolist(),
        'atomic_magnetic_moments': results['output_trajectory'].get_array('atomic_magnetic_moments').tolist(),
    })
``` 

And that works fine. So I'll use that instead, will also add less lines. 😁

sphuber (approved):
Good shit, thanks @mbercx

PR #732, opened 2021-09-16
`PhBaseWorkChain`: spec exposed outputs of incorrect process class

### LINKED ISSUE #723: Fails attach output nodes of PhCalculation with aiida-core v1.6.5
aiida-quantumespresso: c49def7a2
ph.x: PHONON v.6.7MaX
aiida-core: v1.6.5

I encountered an issue about the title. I may be related to the update of `BaseRestartWorkChain` as we encountered in VASP plugin (https://github.com/aiida-vasp/aiida-vasp/pull/520). But I have no idea how to fix it yet. I just want to report it though maybe it's just my mistake. I confirmed that it works with aiida-core v1.6.4.

With aiida-core v1.6.5, `19134  15m ago    PhBaseWorkChain             ⏹ Finished [11]`
```
% verdi process report 19134
2021-09-10 05:55:03 [6159 | REPORT]: [19134|PhBaseWorkChain|run_process]: launching PhCalculation<19143> iteration #1
2021-09-10 05:56:15 [6160 | REPORT]: [19134|PhBaseWorkChain|results]: work chain completed after 1 iterations
2021-09-10 05:56:15 [6161 | REPORT]: [19134|PhBaseWorkChain|results]: required output 'remote_folder' was not an output of PhCalculation<19143>
2021-09-10 05:56:15 [6162 | REPORT]: [19134|PhBaseWorkChain|results]: required output 'retrieved' was not an output of PhCalculation<19143>
2021-09-10 05:56:15 [6163 | REPORT]: [19134|PhBaseWorkChain|results]: required output 'output_parameters' was not an output of PhCalculation<19143>
2021-09-10 05:56:15 [6164 | REPORT]: [19134|PhBaseWorkChain|on_terminated]: remote folders will not be cleaned
```

sphuber:
Fixes #723 

A simple copy paste type had the `PhBaseWorkChain` expose the outputs of
the `PwCalculation` instead of the `PhCalculation` causing no outputs to
be registered at all and the process ending with exit status `11`.


mbercx (approved):
Wow, crazy no one noticed this before. 😅  Best merge this one quickly! 👍

unkcpz:
Although the problem is solved, but I am confused why the logs say 'required output 'remote_folder' was not an output of PhCalculation'. Since the `PhCalculation` have the outputs of `output_parameters`, `remote_folder` and `retrieved`. 

That is to say, the log messages can be more precise when the exposed ports is incompatible.

sphuber:
> Although the problem is solved, but I am confused why the logs say 'required output 'remote_folder' was not an output of PhCalculation'. Since the `PhCalculation` have the outputs of `output_parameters`, `remote_folder` and `retrieved`.
> 
> That is to say, the log messages can be more precise when the exposed ports is incompatible.

The reason is because the `exposed_outputs` will match the process class as well. When passed a `PhCalculation` it will check for exposed output ports that match that process class, but because the exposed ports belonged to `PwCalculation` they didn't match. I am not sure how the message can be much clearer because the class name is mentioned. I don't think the error can _guess_ that there may have been a type by also looking at all other exposed outputs to see if their classes happen to have a name that is close.

unkcpz:
Thanks @sphuber ! I look into the relative part of the code in aiida-core a bit and think we can improve the log by add this incompatible exposed case into `results` method of `BaseRestartWorkChain`, that is where it happened. https://github.com/aiidateam/aiida-core/pull/5149

PR #737, opened 2021-09-21
Docs: add compatibility policy for Quantum ESPRESSO

### LINKED ISSUE #736: ❓ Clearly define what versions of QE we support, and when we plan to drop each
It's not entirely clear to me what versions of Quantum ESPRESSO we are meant to support with the plugin at the moment. I'm sure some older versions (<6.0) are not really supported anymore, which I think is perfectly fine since the last release of 5.X was 7 years ago.

Since the minor versions of QE often contain backwards incompatible changes, it gets more and more complex to support _all_ minor versions in one major version. We've had to add two version-specific fixes in the latest release, and going through the parsing code I find a lot of examples of other version-specific code. Generally, this makes for quite messy code that is hard to maintain.

Just wondering if we already have a policy on this. Maybe we should say that we only support the last X minor releases, or drop support for older releases within X time after the release of the next one?

sphuber:
Fixes #736 

Added the current compatibility to the matrix in the README.md together
with the new explicit policy. Section is added to the documentation that
links the README.md.


README.md
```diff
@@ -8,13 +8,15 @@ This is the official AiiDA plugin for [Quantum ESPRESSO](https://www.quantum-esp
...
+| Plugin | AiiDA | Python | Quantum ESPRESSO |
+|-|-|-|-|
+| `v3.4.0 < v4.0.0` | <img class="svg-badge" title="Compatible with aiida-core >=1.4.4,<2.0.0" src="https://img.shields.io/badge/AiiDA->=1.4.4,<2.0.0-007ec6.svg?logo=[aiida-logo-base64]"> |  [![PyPI pyversions](https://img.shields.io/pypi/pyversions/aiida-quantumespresso.svg)](https://pypi.org/project/aiida-quantumespresso.svg) | 6.0, 6.1, 6.2, 6.3, 6.4, 6.5, 6.6, 6.7 |
```
mbercx, line 13: Much cleaner! 👌

README.md
```diff
@@ -8,13 +8,15 @@ This is the official AiiDA plugin for [Quantum ESPRESSO](https://www.quantum-esp
...
+| `v3.0.0 < v3.1.0` | <img class="svg-badge" title="Compatible with aiida-core >=1.0.0,<1.1.0" src="https://img.shields.io/badge/AiiDA->=1.0.0,<1.1.0-007ec6.svg?logo=[aiida-logo-base64]"> |  [![PyPI pyversions](https://img.shields.io/pypi/pyversions/aiida-quantumespresso/3.0.0.svg)](https://pypi.org/project/aiida-quantumespresso/3.0.0/) | 6.0, 6.1, 6.2, 6.3, 6.4, 6.5, 6.6, 6.7 |
+| `v2.0.0 < v3.0.0` | <img class="svg-badge" title="Compatible with aiida-core >=0.12,<1.0.0" src="https://img.shields.io/badge/AiiDA->=0.12,<1.0.0-007ec6.svg?logo=[aiida-logo-base64]"> |  [![PyPI pyversions](https://img.shields.io/pypi/pyversions/aiida-quantumespresso/2.1.svg)](https://pypi.python.org/pypi/aiida-quantumespresso/2.1.0/) | 6.0, 6.1, 6.2, 6.3, 6.4, 6.5, 6.6, 6.7 |
+
+Starting from `aiida-quantumespresso==4.0`, the last three minor versions of Quantum ESPRESSO are supported up to a maximum of two years.
```
mbercx, line 19: This might be interpreted differently than what we discussed in #736. Perhaps:

```suggestion
Starting from `aiida-quantumespresso==4.0`, the last three minor versions of Quantum ESPRESSO are supported. Older versions are supported up to a maximum of two years.
```

mbercx (approved):
All good, thanks @sphuber! Just one small comment regarding

image

README.md
sphuber: Except ... it doesn't work. Referencing only works with actual links, not variable substitution. Seeing if I can still make it work somehow

sphuber: I will remove this change from this PR and do it in a separate one.

README.md
sphuber: I may have misunderstood what we discussed then. I thought even the last three minor versions would be subject to the 2 years max rule. For example, even if the oldest supported minor version happens to be 10 years old, do we still support it? Maybe it doesn't matter. And with what I propose, I guess we could technically end up with _no_ supported versions, if the most recent version is older than two years :sweat_smile: but in that case there might be bigger problems at hand.

PR #739, opened 2021-09-21
`README.md`: add shields for Quantum ESPRESSO compatibility

sphuber:
Encode the Quantum ESPRESSO logo for the shield as a base64 encoded
string. The image used replaces the white background of the original
logo with #5d5d5d which is the background used by the shield pills. We
cannot use transparency, because that will make the image of the png too
large, since it requires saving a separate transparency channel, and
img.shields.io will return a 400 Bad Request.


sphuber:
@mbercx this should be the final update, so I will let you have a proper look to check all is ok

mbercx (approved):
So purty! 😍

sphuber:
OK, thanks @mbercx . I am done messing with the repo now, pinky promise. So go ahead and update the other PR and rebase-squash away

PR #741, opened 2021-09-21
`PdosWorkChain`: Fix initialization and protocol usage of `ElectronicType`

### LINKED ISSUE #685: Protocols: `overrides` should take precedence over `ElectronicType`
Since in the `PwBaseWorkChain` the `overrides` are merged with the inputs before adapting the inputs based on the `ElectronicType`, the latter will change the final inputs submitted to the builder. This is problematic in case of e.g. the `PdosWorkChain`, where `occupations` should be set to `tetrahedra`. Trying to run the `PdosWorkChain` with `ElectronicType.INSULATOR` will simply fail immediately due to the validation.

It seems to me that the `overrides` should be absolute, i.e. merged with the inputs right before setting up the builder. However, since the `overrides` are also used to pass default protocol settings down the hierarchy, I'd have to double check if this causes problems anywhere.

### LINKED ISSUE #740: `PwCalculation`: drop `parent_folder` validation for `nscf`/`bands`
When adding validation for the restart inputs in #722, we added a check that a `parent_folder` is present in the inputs for `nscf` and `bands` calculations. This makes perfect sense when checking the setup of a _separate_ `PwCalculation`, but when such a calculation is wrapped in a parent work chain that also wraps a `scf` calculation which during the logic of the work chain provides the `parent_folder`, it becomes impossible to submit such a work chain because the input validation of the `nscf` `PwCalculation` fails.

So basically we can't run the`PdosWorkChain` anymore now. 😅  I don't see a way around this, unless there is some magical validation setting that I'm not aware of?

@chrisjsewell seems I managed to break your work chain after all, without even touching it. Do I get a cookie now? 🍪

mbercx:
Fixes #740 
Fixes #685

Fixes two issues when using the `PdosWorkChain`:

1. Remove the validation of the `parent_folder` input for `nscf`
`PwCalculation`s. This made it impossible to use the `PdosWorkChain`
with an initial `scf` run, since the `parent_folder` is only created
inside the logic of the work chain, and hence cannot be provided as an
input upon submission.
2. In the `PdosWorkChain.get_builder_from_protocol()` method, override
the `occupations` setting to make sure it is set to `tetrahedra`.
Although this setting is also passed to the overrides from the `yaml`
file that defines the protocol, if the user sets
`ElectronicType.INSULATOR`, the occupations is set to `fixed` in the
`PwBaseWorkChain.get_builder_from_protocol()` method.

### DIFF
--- a/aiida_quantumespresso/calculations/pw.py
+++ b/aiida_quantumespresso/calculations/pw.py
@@ -160,17 +160,14 @@ class PwCalculation(BasePwCpInputGenerator):
         """Validate the top level namespace.
 
         1. Check that the restart input parameters are set correctly. In case of 'nscf' and 'bands' calculations, this
-        means that ``parent_folder`` is provided, ``startingpot`` is set to 'file' and ``restart_mode`` is
-        'from_scratch'. For other calculations, if the ``parent_folder`` is provided, the restart settings must be set
-        to use some of the outputs.
+        means that ``startingpot`` is set to 'file' and ``restart_mode`` is 'from_scratch'. For other calculations, if
+        the ``parent_folder`` is provided, the restart settings must be set to use some of the outputs.
         """
         parameters = value['parameters'].get_dict()
         calculation_type = parameters.get('CONTROL', {}).get('calculation', 'scf')
 
         # Check that the restart input parameters are set correctly
         if calculation_type in ('nscf', 'bands'):
-            if 'parent_folder' not in value:
-                return f'`parent_folder` not provided for `{calculation_type}` calculation.'
             if parameters.get('ELECTRONS', {}).get('startingpot', 'file') != 'file':
                 return f'`startingpot` should be set to `file` for a `{calculation_type}` calculation.'
             if parameters.get('CONTROL', {}).get('restart_mode', 'from_scratch') != 'from_scratch':
--- a/aiida_quantumespresso/workflows/pdos.py
+++ b/aiida_quantumespresso/workflows/pdos.py
@@ -340,6 +340,7 @@ class PdosWorkChain(ProtocolMixin, WorkChain):
         scf.pop('clean_workdir', None)
         nscf = PwBaseWorkChain.get_builder_from_protocol(*args, overrides=inputs.get('nscf', None), **kwargs)
         nscf['pw'].pop('structure', None)
+        nscf['pw']['parameters']['SYSTEM']['occupations'] = 'tetrahedra'
         nscf['pw']['parameters']['SYSTEM'].pop('smearing', None)
         nscf['pw']['parameters']['SYSTEM'].pop('degauss', None)
         nscf.pop('clean_workdir', None)
--- a/tests/workflows/protocols/test_pdos.py
+++ b/tests/workflows/protocols/test_pdos.py
@@ -1,0 +1,91 @@
+# -*- coding: utf-8 -*-
+"""Tests for the ``PwBandsWorkChain.get_builder_from_protocol`` method."""
+import pytest
+
+from aiida.engine import ProcessBuilder
+from aiida.plugins import WorkflowFactory
+
+from aiida_quantumespresso.common.types import ElectronicType, SpinType
+
+PdosWorkChain = WorkflowFactory('quantumespresso.pdos')
+
+
+def test_get_available_protocols():
+    """Test ``PdosWorkChain.get_available_protocols``."""
+    protocols = PdosWorkChain.get_available_protocols()
+    assert sorted(protocols.keys()) == ['fast', 'moderate', 'precise']
+    all('description' in protocol for protocol in protocols)
+
+
+def test_get_default_protocol():
+    """Test ``PdosWorkChain.get_default_protocol``."""
+    assert PdosWorkChain.get_default_protocol() == 'moderate'
+
+
+def test_default(fixture_code, generate_structure, data_regression, serialize_builder):
+    """Test ``PdosWorkChain.get_builder_from_protocol`` for the default protocol."""
+    pw_code = fixture_code('quantumespresso.pw')
+    dos_code = fixture_code('quantumespresso.dos')
+    projwfc_code = fixture_code('quantumespresso.projwfc')
+    structure = generate_structure()
+    builder = PdosWorkChain.get_builder_from_protocol(
+        pw_code=pw_code, dos_code=dos_code, projwfc_code=projwfc_code, structure=structure
+    )
+
+    assert isinstance(builder, ProcessBuilder)
+    data_regression.check(serialize_builder(builder))
+
+
+def test_electronic_type(fixture_code, generate_structure):
+    """Test ``PdosWorkChain.get_builder_from_protocol`` with ``electronic_type`` keyword."""
+    pw_code = fixture_code('quantumespresso.pw')
+    dos_code = fixture_code('quantumespresso.dos')
+    projwfc_code = fixture_code('quantumespresso.projwfc')
+    structure = generate_structure()
+
+    with pytest.raises(NotImplementedError):
+        builder = PdosWorkChain.get_builder_from_protocol(
+            pw_code=pw_code,
+            dos_code=dos_code,
+            projwfc_code=projwfc_code,
+            structure=structure,
+            electronic_type=ElectronicType.AUTOMATIC
+        )
+
+    builder = PdosWorkChain.get_builder_from_protocol(
+        pw_code=pw_code,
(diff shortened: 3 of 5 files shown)


tests/workflows/protocols/test_pdos.py
```diff
@@ -0,0 +1,91 @@
...
+from aiida_quantumespresso.common.types import ElectronicType, SpinType
+
+PdosWorkChain = WorkflowFactory('quantumespresso.pdos')
+
+
+def test_get_available_protocols():
+    """Test ``PdosWorkChain.get_available_protocols``."""
+    protocols = PdosWorkChain.get_available_protocols()
+    assert sorted(protocols.keys()) == ['fast', 'moderate', 'precise']
+    all('description' in protocol for protocol in protocols)
```
qiaojunfeng, line 17: Should have an `assert` before `all('description' ...`?

qiaojunfeng (changes requested):

tests/workflows/protocols/test_pdos.py
qiaojunfeng, line 17:
```suggestion
    assert all('description' in protocol for protocol in protocols)
```

mbercx:
@qiaojunfeng thanks for the review! 🙏  Seems that the code I copy 🍝 'ed from the other tests was missing an assert indeed, as well as `.values()`. Fixed that up!

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -340,6 +340,7 @@ def get_builder_from_protocol(
         scf.pop('clean_workdir', None)
         nscf = PwBaseWorkChain.get_builder_from_protocol(*args, overrides=inputs.get('nscf', None), **kwargs)
         nscf['pw'].pop('structure', None)
+        nscf['pw']['parameters']['SYSTEM']['occupations'] = 'tetrahedra'
```
sphuber, line 343: what if the user explicitly specified a value for the `occupations` in the `nscf` step in the `overrides`? Even though this may be "incorrect", didn't we agree that the `overrides` should have the final say?

aiida_quantumespresso/calculations/pw.py
```diff
@@ -160,17 +160,14 @@ def validate_inputs(value, _):
...
+        means that ``startingpot`` is set to 'file' and ``restart_mode`` is 'from_scratch'. For other calculations, if
+        the ``parent_folder`` is provided, the restart settings must be set to use some of the outputs.
         """
         parameters = value['parameters'].get_dict()
         calculation_type = parameters.get('CONTROL', {}).get('calculation', 'scf')
 
         # Check that the restart input parameters are set correctly
         if calculation_type in ('nscf', 'bands'):
-            if 'parent_folder' not in value:
-                return f'`parent_folder` not provided for `{calculation_type}` calculation.'
```
sphuber, line 173: where is this going to get checked now? Do we have a check somewhere in the outline steps or is it just going to horribly except when it tries to call `self.inputs.parent_folder` somewhere? If that is the case, it should be caught and transformed in an understandable exception message. I also think we should maybe open an issue describing this use case to see if we can come up with a solution that allows to still have the check, but have it disabled in wrapping work chains that guarantee that they will be specifying it themselves.

tests/workflows/protocols/test_pdos.py
```diff
@@ -0,0 +1,91 @@
+# -*- coding: utf-8 -*-
+"""Tests for the ``PwBandsWorkChain.get_builder_from_protocol`` method."""
```
sphuber, line 2:
```suggestion
"""Tests for the ``PdosWorkChain.get_builder_from_protocol`` method."""
```

tests/workflows/protocols/test_pdos.py
```diff
@@ -0,0 +1,91 @@
...
+    assert isinstance(builder, ProcessBuilder)
+    data_regression.check(serialize_builder(builder))
+
+
+def test_electronic_type(fixture_code, generate_structure):
+    """Test ``PdosWorkChain.get_builder_from_protocol`` with ``electronic_type`` keyword."""
+    pw_code = fixture_code('quantumespresso.pw')
+    dos_code = fixture_code('quantumespresso.dos')
+    projwfc_code = fixture_code('quantumespresso.projwfc')
+    structure = generate_structure()
```
sphuber, line 44: You use this multiple times. Would be good to make a fixture `get_generator_inputs` that returns this as a dictionary. Then you just need to update one input each time (`electronic_type` in this case) and then you `**` it to `get_builder_from_protocol`.

sphuber (changes requested):

aiida_quantumespresso/workflows/pdos.py
mbercx: Fair point! I suppose we could _only_ set `occupations` to `tetrahedra` if the setting is not in the `overrides`, and let the `validate_nscf` validation fail in case it's not set appropriately. 

Note that the only reason we have to do this here is because the `overrides` _don't_ have the final say in the `PwBaseWorkChain`, as `SYSTEM.occupations` gets adjusted by the `ElectronicType` _after_ the overrides are already merged with the protocol inputs.

aiida_quantumespresso/calculations/pw.py
mbercx: I'd have to double check, but I think that for both the `PwBandsWorkChain` and `PdosWorkChain`, this validation is unnecessary. The PwBandsWorkChain _always_ runs an SCF first, and the `PdosWorkChain` has its own validation for the `parent_folder` input (i.e. either an SCF is run, or the `parent_folder` must be provided for the `nscf`. So the only issue is that for a bare `nscf`/`bands` `PwCalculation`, or one wrapped in the `PwBaseWorkChain`, the QE will fail since it doesn't have the required files in the folder.

I suppose we could add an `exclude_validation` argument to the `expose_inputs` method (similar to how we use `exclude` to exclude inputs). The user could pass a tuple that is very similar to the `exclude` one (i.e. using the `.` notation for inputs nested in `PortNameSpaces`. One question would be how to indicate _all_ validation, or only the top-level validation should be skipped. Let's indeed open an issue to discuss this.

For this PR: I think that we can simply remove this validation for now, and say that validating that an `nscf` or `bands` calculation has the `parent_folder` input is the responsibility of the work chains that wrap them, as it was before #722.

aiida_quantumespresso/workflows/pdos.py
mbercx: So I've tried to address this by specifically looking at the `overrides` in the `PdosWorkChain.get_builder_from_protocol()` method, but it all feels a little hacky to me. Now I'm specifically adding extra code to preserve an override that will simply cause the work chain to except anyways.

The problem lies with the fact that the `overrides` are not considered to be absolute in the lower level `PwBaseWorkChain`, which is why I need to add this extra line in the first place. But then the overrides are also not considered to be absolute when setting `RelaxType` in the `PwRelaxWorkChain`. So I think this is just a symptom of a larger problem that we discuss in https://github.com/aiidateam/aiida-quantumespresso/issues/745.

Since this PR is blocking https://github.com/aiidateam/aiida-quantumespresso/pull/747 and an improved solution for https://github.com/aiidateam/aiida-quantumespresso/pull/714, I'd be inclined to accept the fact that the incorrect overrides are ignored for now, until we come up with a solution for all the work chains.

mbercx: @sphuber I've bit the bullet and adapted the `PwBaseWorkChain.get_builder_from_protocol()` method so it follow the principle we decided on in #745, i.e. that the `overrides` are absolute. This avoids having to set the `SYSTEM.occupations` to `tetrahedra` in the `PdosWorkChain.get_builder_from_protocol()` method, since now the "protocol overrides" specified in the YAML file of the `PdosWorkChain` protocol is now applied correctly even if the `ElectronicType` is set to `Insulator`.

Additionally, this allowed me to remove the logic we added in #731, since now by default _all_ parameters provided in the overrides are recursively merged so they override process inputs generated by the protocol inputs. This means that the `overrides` _do_ take precedence over any supplied `initial_magnetic_moments` (which we can consider a protocol input). Note the change in the corresponding test.

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -152,7 +152,7 @@ def get_builder_from_protocol(
             an initial guess for the magnetic moment is automatically set in case neither is provided.
```
qiaojunfeng, line 152: So now the `overrides` takes precedence over the `initial_magnetic_moments`, so we need to update the doc

PR #744, opened 2021-09-22
`PwBaseWorkChain`: handle diagonalization issues

### LINKED ISSUE #742: `PwBaseWorkChain`: error handler for diagonalization issues
While running calculations (especially `nscf`, I've noticed), one often runs into the following errors:

1. Too many bands not converged
```
from c_bands : error # 1
too many bands are not converged
```

2. Cholesky
```
     Error in routine  cdiaghg (161): 
      problems computing cholesky
```

[2] is already parsed and raises a specific error. [1] should also be added to the list of errors for the `PwCalculation`. For both issues, I've tried switching `ELECTRONS.diagonalization` to `'cg'`, and this was successful in fixing my `nscf` calculations ( 2 cases of [1], 12 cases of [2]). 

@pietrodelugas would you say that this is a good approach to deal with these issues in general, or did I just get lucky? 🙃 

Will be doing a lot more `PwBandsWorkChain` and `PdosWorkChain`s soon, and will try adding this error handler to see how many failures it can deal with. If it performs well, I'll open a PR.

mbercx:
Fixes #742 

Add parsing for an error related to the diagonalization, and add a
corresponding exit code to the `PwCalculation`.

Add an error handler to deal with both this exit code as well as the
`ERROR_COMPUTING_CHOLESKY` exit code. In both cases, switching to using
conjugate gradient for the diagonalization seems to be a quite
successful approach for recovering from these failures. If the
calculation was already running conjugate gradient, the
`PwBaseWorkChain` fails with a known unrecoverable failure.

### DIFF
--- a/aiida_quantumespresso/workflows/pw/base.py
+++ b/aiida_quantumespresso/workflows/pw/base.py
@@ -529,7 +529,6 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         density. In case the run already used conjugate gradient diagonalization, abort.
         """
         if self.ctx.inputs.parameters['ELECTRONS'].get('diagonalization', None) == 'cg':
-            print('ELLO')
             action = 'found diagonalization issues but already running with conjugate gradient algorithm, aborting...'
             self.report_error_handled(calculation, action)
             return ProcessHandlerReport(True, self.exit_codes.ERROR_KNOWN_UNRECOVERABLE_FAILURE)
@@ -537,7 +536,6 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         self.ctx.inputs.parameters['ELECTRONS']['diagonalization'] = 'cg'
         action = 'found diagonalization issues, switching to conjugate gradient diagonalization.'
         self.report_error_handled(calculation, action)
-        print('WAVE')
         return ProcessHandlerReport(True)
 
     @process_handler(priority=580, exit_codes=[
--- a/aiida_quantumespresso/workflows/pw/base.py
+++ b/aiida_quantumespresso/workflows/pw/base.py
@@ -506,9 +506,7 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
             self.report_error_handled(calculation, 'unrecoverable error, aborting...')
             return ProcessHandlerReport(True, self.exit_codes.ERROR_UNRECOVERABLE_FAILURE)
 
-    @process_handler(priority=590, exit_codes=[
-        PwCalculation.exit_codes.ERROR_COMPUTING_CHOLESKY,
-    ])
+    @process_handler(priority=590, exit_codes=[])
     def handle_known_unrecoverable_failure(self, calculation):
         """Handle calculations with an exit status that correspond to a known failure mode that are unrecoverable.
 
@@ -516,6 +514,29 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         """
         self.report_error_handled(calculation, 'known unrecoverable failure detected, aborting...')
         return ProcessHandlerReport(True, self.exit_codes.ERROR_KNOWN_UNRECOVERABLE_FAILURE)
+
+    @process_handler(
+        priority=585,
+        exit_codes=[
+            PwCalculation.exit_codes.ERROR_COMPUTING_CHOLESKY,
+            PwCalculation.exit_codes.ERROR_DIAGONALIZATION_TOO_MANY_BANDS_NOT_CONVERGED,
+        ]
+    )
+    def handle_diagonalization_errors(self, calculation):
+        """Handle known issues related to the diagonalization.
+
+        Switch to ``diagonalization = 'cg'`` if not already running with this setting, and restart from the charge
+        density. In case the run already used conjugate gradient diagonalization, abort.
+        """
+        if self.ctx.inputs.parameters['ELECTRONS'].get('diagonalization', None) == 'cg':
+            action = 'found diagonalization issues but already running with conjugate gradient algorithm, aborting...'
+            self.report_error_handled(calculation, action)
+            return ProcessHandlerReport(True, self.exit_codes.ERROR_KNOWN_UNRECOVERABLE_FAILURE)
+
+        self.ctx.inputs.parameters['ELECTRONS']['diagonalization'] = 'cg'
+        action = 'found diagonalization issues, switching to conjugate gradient diagonalization.'
+        self.report_error_handled(calculation, action)
+        return ProcessHandlerReport(True)
 
     @process_handler(priority=580, exit_codes=[
         PwCalculation.exit_codes.ERROR_OUT_OF_WALLTIME,
--- a/aiida_quantumespresso/workflows/pw/base.py
+++ b/aiida_quantumespresso/workflows/pw/base.py
@@ -529,6 +529,7 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         density. In case the run already used conjugate gradient diagonalization, abort.
         """
         if self.ctx.inputs.parameters['ELECTRONS'].get('diagonalization', None) == 'cg':
+            print('ELLO')
             action = 'found diagonalization issues but already running with conjugate gradient algorithm, aborting...'
             self.report_error_handled(calculation, action)
             return ProcessHandlerReport(True, self.exit_codes.ERROR_KNOWN_UNRECOVERABLE_FAILURE)
@@ -536,6 +537,7 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         self.ctx.inputs.parameters['ELECTRONS']['diagonalization'] = 'cg'
         action = 'found diagonalization issues, switching to conjugate gradient diagonalization.'
         self.report_error_handled(calculation, action)
+        print('WAVE')
         return ProcessHandlerReport(True)
 
     @process_handler(priority=580, exit_codes=[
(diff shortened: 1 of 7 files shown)


sphuber (commented):
Looks good @mbercx , thanks. Would just at least add a test for the `PwParser` to check the error gets picked up correctly. Bonus points if you add a test for the workchain handler itself, but I think that is less critical

mbercx:
I've been testing this in the field, and it seems to work quite nicely:

1. Cholesky issues (exit code `462`)
```
PwBaseWorkChain<213713> Finished [0] [5:results]
    ├── create_kpoints_from_distance<213820> Finished [0]
    ├── PwCalculation<213888> Finished [462]
    └── PwCalculation<215363> Finished [0]
PwBaseWorkChain<213899> Finished [0] [5:results]
    ├── create_kpoints_from_distance<213900> Finished [0]
    ├── PwCalculation<213909> Finished [462]
    └── PwCalculation<215765> Finished [0]
...
```

So far I've encountered this error 18 times in my test runs, and _every_ time switching to conjugate gradient seems to fix the problem.

2. "Too many bands not converged" (exit code `463`)
```
PwBaseWorkChain<214832> Finished [0] [5:results]
    ├── create_kpoints_from_distance<215196> Finished [0]
    ├── PwCalculation<215224> Finished [463]
    └── PwCalculation<217771> Finished [0]
PwBaseWorkChain<215379> Finished [0] [5:results]
    ├── create_kpoints_from_distance<215888> Finished [0]
    ├── PwCalculation<215926> Finished [463]
    └── PwCalculation<218431> Finished [0]
...
```

Here it also works most of the time, but there are still some failures. One run already started with `'cg'`, and yet the calculation failed:

```
PwBaseWorkChain<219924> Finished [310] [4:while_(should_run_process)(2:inspect_process)]
    └── PwCalculation<219935> Finished [463]
```

```
$ verdi process report 219924
2021-09-22 17:01:03 [45694 | REPORT]: [219924|PwBaseWorkChain|run_process]: launching PwCalculation<219935> iteration #1
2021-09-22 17:03:55 [45838 | REPORT]: [219924|PwBaseWorkChain|report_error_handled]: PwCalculation<219935> failed with exit status 463: Too many bands failed to converge during the diagonalization.
2021-09-22 17:03:55 [45839 | REPORT]: [219924|PwBaseWorkChain|report_error_handled]: Action taken: found diagonalization issues but already running with conjugate gradient algorithm, aborting...
2021-09-22 17:03:55 [45840 | REPORT]: [219924|PwBaseWorkChain|inspect_process]: PwCalculation<219935> failed but a handler detected an unrecoverable problem, aborting
2021-09-22 17:03:55 [45841 | REPORT]: [219924|PwBaseWorkChain|on_terminated]: remote folders will not be cleaned
```

mbercx:
> Looks good @mbercx , thanks. Would just at least add a test for the `PwParser` to check the error gets picked up correctly. Bonus points if you add a test for the workchain handler itself, but I think that is less critical

Indeed, still have to fix the tests, on it! Will see how complex it is to get those bonus points. 😎

ramirezfranciscof (commented):
I probably shouldn't be approving this since I never done an error handler before (so I'm more learning how it works by looking at this rather than actually checking if it works), but the general logic looks fine to me.

mbercx:
> I probably shouldn't be approving this since I never done an error handler before (so I'm more learning how it works by looking at this rather than actually checking if it works), but the general logic looks fine to me.

You can always give it a spin for one of your problematic systems if you'd like. ^^ I've tested it in the field, though, and it seems to work fine! Just going to compare with `ParO` instead of `cg` since that should be much faster [according to Pietro](https://github.com/aiidateam/aiida-quantumespresso/issues/742#issuecomment-925141664).

sphuber:
> Just going to compare with `ParO` instead of `cg` since that should be much faster [according to Pietro](https://github.com/aiidateam/aiida-quantumespresso/issues/742#issuecomment-925141664).

Note that [ParO wasn't added until QE v6.6](https://github.com/QEF/q-e/releases/tag/qe-6.6), so depending on our new version support protocol we may not yet start to rely on it existing. Making it conditional would be tricky because there is no official way to determine what code version will be used.

mbercx:
> > Just going to compare with `ParO` instead of `cg` since that should be much faster [according to Pietro](https://github.com/aiidateam/aiida-quantumespresso/issues/742#issuecomment-925141664).
> 
> Note that [ParO wasn't added until QE v6.6](https://github.com/QEF/q-e/releases/tag/qe-6.6), so depending on our new version support protocol we may not yet start to rely on it existing. Making it conditional would be tricky because there is no official way to determine what code version will be used.

Good point! Thanks for looking into this. Re the version of the code: we should really move to adding this to the `Code` somehow. Can we perhaps automatically add the code version to the extras of the `Code` at the start of the parsing?

mbercx:
@sphuber @ramirezfranciscof since we currently still support all v6.X QE versions, and have no way to determine the version of the code run before parsing, I think we can go ahead and merge this with `'cg'`. It's better to be able to handle this error with a slow method than not handle it at all.

Will open another issue to discuss how we can deal with code version-dependent logic and add a note there that we should update this error handler when we figure this out.

mbercx:
@sphuber I got the bonus points, but had to comment out the `test_handle_known_unrecoverable_failure` test since we now have no exit codes that need to get handled by this error handler. Not sure if this is fine?

tests/workflows/pw/test_base.py
```diff
@@ -97,15 +97,41 @@ def test_handle_electronic_convergence_not_achieved(generate_workchain_pw, fixtu
...
+#     process = generate_workchain_pw(exit_code=PwCalculation.exit_codes.<INSERT EXIT CODE HERE>)
+#     process.setup()
+
+#     result = process.handle_known_unrecoverable_failure(process.ctx.children[-1])
+#     assert isinstance(result, ProcessHandlerReport)
+#     assert result.do_break
+#     assert result.exit_code == PwBaseWorkChain.exit_codes.ERROR_KNOWN_UNRECOVERABLE_FAILURE
+
+#     result = process.inspect_process()
+#     assert result == PwBaseWorkChain.exit_codes.ERROR_KNOWN_UNRECOVERABLE_FAILURE
```
sphuber, line 112:
```suggestion
pytest.mark.skip('Reactivate once we have an unrecoverable failure once again.')
 def test_handle_known_unrecoverable_failure(generate_workchain_pw):
     """Test `PwBaseWorkChain.handle_known_unrecoverable_failure`."""
     process = generate_workchain_pw(exit_code=PwCalculation.exit_codes.<INSERT EXIT CODE HERE>)
     process.setup()

     result = process.handle_known_unrecoverable_failure(process.ctx.children[-1])
     assert isinstance(result, ProcessHandlerReport)
     assert result.do_break
     assert result.exit_code == PwBaseWorkChain.exit_codes.ERROR_KNOWN_UNRECOVERABLE_FAILURE

     result = process.inspect_process()
     assert result == PwBaseWorkChain.exit_codes.ERROR_KNOWN_UNRECOVERABLE_FAILURE
```

tests/workflows/pw/test_base.py
```diff
@@ -97,15 +97,41 @@ def test_handle_electronic_convergence_not_achieved(generate_workchain_pw, fixtu
...
-    result = process.handle_known_unrecoverable_failure(process.ctx.children[-1])
+    process.ctx.inputs.parameters['ELECTRONS']['diagonalization'] = 'david'
+
+    result = process.handle_diagonalization_errors(process.ctx.children[-1])
     assert isinstance(result, ProcessHandlerReport)
+    assert process.ctx.inputs.parameters['ELECTRONS']['diagonalization'] == 'cg'
+    assert result.do_break
+
+    result = process.handle_diagonalization_errors(process.ctx.children[-1])
     assert result.do_break
```
sphuber, line 134: wait, why is this called a second time?

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -517,6 +515,31 @@ def handle_known_unrecoverable_failure(self, calculation):
...
+        ]
+    )
+    def handle_diagonalization_errors(self, calculation):
+        """Handle known issues related to the diagonalization.
+
+        Switch to ``diagonalization = 'cg'`` if not already running with this setting, and restart from the charge
+        density. In case the run already used conjugate gradient diagonalization, abort.
+        """
+        if self.ctx.inputs.parameters['ELECTRONS'].get('diagonalization', None) == 'cg':
+            print('ELLO')
```
sphuber, line 532:
```suggestion
```

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -517,6 +515,31 @@ def handle_known_unrecoverable_failure(self, calculation):
...
+            action = 'found diagonalization issues but already running with conjugate gradient algorithm, aborting...'
+            self.report_error_handled(calculation, action)
+            return ProcessHandlerReport(True, self.exit_codes.ERROR_KNOWN_UNRECOVERABLE_FAILURE)
+
+        self.ctx.inputs.parameters['ELECTRONS']['diagonalization'] = 'cg'
+        action = 'found diagonalization issues, switching to conjugate gradient diagonalization.'
+        self.report_error_handled(calculation, action)
+        print('WAVE')
```
sphuber, line 540:
```suggestion
```

sphuber (changes requested):
Thanks @mbercx . Couple of small things but should be good to merge after that

tests/workflows/pw/test_base.py
sphuber: I see, but that means we still have an `ERROR_KNOWN_UNRECOVERABLE_FAILURE` pathway and you don't have to actually skip the test that checks that path, as you do now. Anyway, this path is checked in the test discussed in this comment, so it is ok to leave for now.

sphuber (approved):

PR #747, opened 2021-10-03
`projwfc.x`: parse from XML instead of parent calc

### LINKED ISSUE #299: ProjwfcParser uses input nodes up to 3 links back, and prevents easy testing
`ProjwfcParser` wants to access not only the `parent_folder` that we are passing as input, but its creator `CalcJobNode` (a pw.x calculation) and the input `structure` and output `output_parameters` of that calculation (i.e. up to 3 steps away the current `CalcJobNode`). This is not only a bit weird, but prevents easy testing of the Parser class in isolation because it requires mocking of all these nodes.

* We should decide whether we can access all the information the parser needs just from the `parent_folder` input.
* Whatever the decision, we should add a parser test.

mbercx:
Fixes #299 

The current `ProjwfcParser` uses several in and output nodes from the
parent calculation. This increased the complexity of the tests for this
parser, and made running `opengrid.x` in between the `pw.x` and
`projwfc.x` run impossible without adding these in and output nodes to
the calculation job of `opengrid.x`.

Here we switch to parsing the XML instead of relying on the parent
calculation. The `data-file-schema.xml` of the parent calculation is
retrieved and parsed, providing the required information for the
subsequent parsing of the `projwfc.x` output. All the parsing tests are
updated to include the XML output file and remove the in/output links
for the parent calculation.

The `convert_qe_to_kpoints` function is added to convert the k-points
data in the XML to a `KpointsData` node.


mbercx:
This PR is still blocked since it builds on #741 (I used the `PdosWorkChain` to generate the parser test files ^^).

Here I simply switch to parsing the XML, but leave the 🍝 -code of the parser. I'll open a second PR soon with the new (and hopefully more readable) code. I split these up for a couple of reasons:

1. Make this PR easier to review since it's time-critical.
2. Update the tests for using the XML with the old parsing code. I shouldn't have to adapt them when I switch to my code, so the tests should still pass without adapting them, demonstrating that the parsing is unchanged by my refactoring.
3. Although I'm sure the parsing is the same for QE v6.6 (the version I ran my tests with), I'm also quite confident that e.g. QE v6.3 will no longer parse properly, since I've removed some extra logic required to support it. Since starting from `aiida-quantumespresso==4.0.0` we'll only support QE v6.5+, the PR with the refactored code can only be merged when we're getting ready to release 4.0.

mbercx:
@sphuber one thing I was considering is that we duplicate a lot of files in the repository by retrieving the XML file. I suppose it would be better to instead add the XML file to the `retrieve_temporary_list`.

Additionally, I was wondering if we should move the PDOS files (e.g. `aiida.pdos_atm#1(Fe)_wfc#6(p_j1.5)`) to the temporary retrieve list as well. Adding all these to the repository adds quite a bit of files, since you get one of these for each orbital of each atom. Even for the simple 2 atom case of Fe, this results in 16 extra files in the repository when running spin-orbit calculations. I don't see the use case for keeping these in the repository save for testing, since all of this data is stored in the `ProjectionData` output under the link `projections`.

mbercx:
In the end I decided to _only_ move the XML file to the temporary retrieve list here, since technically no longer retrieving the PDOS files is backwards incompatible. I've added this to #749 instead.

**Note**: Please don't pay too much attention to coding style for this PR. I've reworked the code _a lot_ in #749, so leave coding style comments for that one. I too was very annoyed by the use of `out_info_dict` to just pass _everything_ between the various methods.

mbercx:
@qiaojunfeng I've adapted the branch so it no longer builds on top of #741, so it's ready for review. I guess the main question is if this allows you to simplify the definition of the `opengrid.x` calculation job in https://github.com/aiidateam/aiida-quantumespresso/pull/714.

qiaojunfeng (changes requested):
Thanks Marnik! Only few minor issues.

aiida_quantumespresso/calculations/projwfc.py
```diff
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 """`CalcJob` implementation for the projwfc.x code of Quantum ESPRESSO."""
+import os
```
sphuber, line 3: pathlib?

aiida_quantumespresso/calculations/projwfc.py
```diff
@@ -25,7 +27,17 @@ class ProjwfcCalculation(NamelistsCalculation):
...
+    xml_filename = 'data-file-schema.xml'
+
+    _internal_retrieve_list = [
+        NamelistsCalculation._PREFIX + '.pdos*',
+    ]
+    _retrieve_temporary_list = [
+        os.path.join(
+            NamelistsCalculation._default_parent_output_folder, f'{NamelistsCalculation._PREFIX}.save', xml_filename
+        )
+    ]
```
sphuber, line 40: Any particular reason to put have this in the temporary retrieved? Is the full data parsed and stored in output nodes?

aiida_quantumespresso/parsers/projwfc.py
```diff
@@ -168,16 +171,11 @@ def spin_dependent_subparser(out_info_dict):
         raise QEOutputParsingError('the standard out file does not comply with the official documentation.')
 
     bands_data = BandsData()
-    # Attempts to retrieve the kpoints from the parent calc
-    parent_calc = out_info_dict['parent_calc']
+    kpoints = od['kpoints']
```
sphuber, line 174: Huh? Where is this `od` variable coming from? I imagine this is the `out_info_dict` but I can't see where the name change is happening. Anyway, not sure I am a fan of `od`. Not very descriptive name

sphuber (changes requested):

aiida_quantumespresso/parsers/projwfc.py
qiaojunfeng: It's here 😄 https://github.com/aiidateam/aiida-quantumespresso/blob/c28d19c370cec6cecf878cb8e93b376899f4d77a/aiida_quantumespresso/parsers/projwfc.py#L130

mbercx: @sphuber I know, the whole code here isn't the most readable, but as I lament in [my comment above](https://github.com/aiidateam/aiida-quantumespresso/pull/747#issuecomment-932989255), I've refactored the code quite a bit in https://github.com/aiidateam/aiida-quantumespresso/pull/749. However, that PR doesn't support all 6.X versions, since I removed some of the logic that deals with very old versions in cleaning up the code. It's also not backwards compatible since I think we shouldn't add all the PDOS files to the repository, since all the info they contain are in the `ProjectionData` output. So I'd leave that for v4.0.0, from which we'll only support QE v6.5+.

aiida_quantumespresso/calculations/projwfc.py
```diff
@@ -45,10 +57,18 @@ def define(cls, spec):
         spec.output('projections', valid_type=ProjectionData, required=False)
         spec.output('bands', valid_type=BandsData, required=False)
         spec.default_output_node = 'output_parameters'
+        spec.exit_code(303, 'ERROR_OUTPUT_XML_MISSING',
+            message='The retrieved folder did not contain the required required XML file.')
```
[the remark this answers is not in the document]
mbercx: Well spotted! I just copy pasted this 😁

aiida_quantumespresso/parsers/projwfc.py
```diff
@@ -347,6 +361,33 @@ def parse(self, **kwargs):
...
+        xml_filepath = Path(retrieved_temporary_folder) / self.node.process_class.xml_filename
+
+        if not xml_filepath.exists():
+            self.exit(self.exit_codes.ERROR_OUTPUT_XML_MISSING)
+
+        try:
+            with xml_filepath.open('r') as handle:
+                parsed_xml, logs = parse_xml(handle, None)
+        except IOError:
+            self.exit(self.exit_codes.ERROR_OUTPUT_XML_READ)
```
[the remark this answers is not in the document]
mbercx: Hmm, I think once `self.exit()` is called, the parsing stops? In the `PwParser` the exit codes are stored in the attributes, and then later in the parsing some logic is applied to determine which exit code should be returned. However, here we _need_ the XML information to parse the outputs, so in case the parsing fails, I think we can just exit with the corresponding exit code.

aiida_quantumespresso/calculations/projwfc.py
mbercx: yessir!

aiida_quantumespresso/parsers/projwfc.py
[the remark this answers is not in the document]
mbercx: Right, of course, it needs to be returned in the `parse` method. I actually looked at these lines when I saw your comment:

https://github.com/aiidateam/aiida-quantumespresso/blob/27c93e5848829219eda7ab6ee5256bfbbd960f42/aiida_quantumespresso/parsers/pw.py#L109-L118

And somehow failed to notice the `return` statements 🤦 . Thanks for your dilligence @qiaojunfeng! Will fix.

PR #425, opened 2019-10-19
Add raw parser of Lowdin charges from projwfc output file

chrisjsewell:
This is intended to be merged after #424, then used to add an output node to `ProjwfcCalculation` containing this data.

PR #749, opened 2021-10-03
`ProjwfcParser`: refactor code and temp retrieve PDOS files

mbercx:
Massive refactoring of the `ProjwfcParser` to make the code less entangled
and more readable. 

TODO:

- [ ] Integrate changed from https://github.com/aiidateam/aiida-quantumespresso/pull/425 in this PR and make sure @chrisjsewell gets his attribution.


mbercx:
Note: the Python 3.6 tests fail because `numpy.ArrayLike` requires a numpy version that no longer supports Python3.6. However, since the EOL of Python 3.6 is in a couple months and the PR is meant to go into v4.0.0, I think we can just leave it as is and drop Python 3.6 support before we merge it.

More importantly, the tests succeed for 581bd57  and Python >= 3.7, without touching the tests. I've also run both the old parser and refactored one for the 4 different spin settings, and compared the content of the orbitals and PDOS arrays. So I'm quite confident that the parsing is unaltered.

mbercx:
@sphuber this PR is both blocked by #747 and on hold because it contains backwards incompatible changes:

- Drop support for pre-v6.5 QE versions.
- Only temporarily retrieve the PDOS files to lighten the repository.

mbercx:
@Minotakm @t-reents @AndresOrtegaGuerrero I still have to add the Lowdin parsing, but other than that the PR is ready for review. A line-by-line code review might be quite a bit of work, I'm mainly asking you to test the new parser class and see if it still has any issues. E.g. I think you mentioned the parsing for some cases (spinorbit?) was incorrect.

Minotakm:
Thanks @mbercx  I will give it a spin and see how it goes.

src/aiida_quantumespresso/parsers/projwfc.py
```diff
@@ -2,458 +2,423 @@
...
+            return keys
+
+        # Read the `pdos_tot` file
+        try:
+            pdostot_filepath = next(retrieved_temporary_folder.glob('*pdos_tot*'))
+            with pdostot_filepath.open('r') as pdostot_file:
+                # Columns: Energy(eV), Ldos, Pdos
+                pdostot_array = np.atleast_2d(np.genfromtxt(pdostot_file))
+                energy = pdostot_array[:, 0]
+                dos = pdostot_array[:, 1]
```
t-reents, line 304: In a spin-polarized calculation (`nspin ==2 `, we only parse the spin-up column at the moment). The suggestion is following the logic of the `dos.x` Parser: 

https://github.com/aiidateam/aiida-quantumespresso/blob/c84ef21dbea1109814e9780d1d18e7303fa2fc25/src/aiida_quantumespresso/parsers/dos.py#L51-L69

```suggestion
                pdostot_array = np.atleast_2d(np.genfromtxt(pdostot_file))
                
                dos = []
                dos_names = []
                energy = pdostot_array[:, 0]
                if nspin ==2:
                    dos += [pdostot_array[:, 1]]
                    dos_names += ['dos_spin_up']
                    dos += [pdostot_array[:, 2]]
                    dos_names += ['dos_spin_down']
                else:
                    dos += [pdostot_array[:, 1]]
                    dos_names += ['dos']
```

src/aiida_quantumespresso/parsers/projwfc.py
```diff
@@ -2,458 +2,423 @@
...
+                pdos_array = np.concatenate([pdos_array[:, 1::2], pdos_array[:, 0::2]], axis=1)
+        else:
+            # Here all the 'up' orbitals for each l number come first, so the PDOS columns must be sorted accordingly
+            pdos_list = []
+            for wfc_pdos_array in [pdos_atm_array_dict[name] for name in pdos_file_names]:
+                # Reorder the columns so the 'up' spin columns _for this l number_ come first
+                pdos_list.append(np.concatenate([wfc_pdos_array[:, 0::2], wfc_pdos_array[:, 1::2]], axis=1))
+            pdos_array = np.concatenate(pdos_list, axis=1)
+
+        return energy, dos, pdos_array
```
t-reents, line 343:
```suggestion
        return energy, dos, dos_names, pdos_array
```

src/aiida_quantumespresso/parsers/projwfc.py
```diff
@@ -2,458 +2,423 @@
...
+
+        1. Sort the filenames by the atom # and wfc #, in that order.
+        2. Concatenate all the PDOS files into one large array.
+        3. Reorder the columns depending on the ``npsin`` value and if spin-orbit is used.
+
+        :param retrieved_temporary_folder: temporary folder of retrieved files that is deleted after parsing.
+        :param nspin: nspin value of the parent calculation.
+        :param spinorbit: True if the calculation used spin-orbit coupling.
+
+        :return: tuple of three arrays containing the energy grid, the total DOS and the PDOS
```
t-reents, line 279:
```suggestion
        :return: tuple of three arrays containing the energy grid, the total DOS, the labels of the DOS 
        (spin-up and spin-down) and the PDOS
```

src/aiida_quantumespresso/parsers/projwfc.py
```diff
@@ -2,458 +2,423 @@
...
 
-        # check and read all of the individual pdos_atm files
-        pdos_atm_filenames = fnmatch.filter(out_filenames, '*pdos_atm*')
-        pdos_atm_array_dict = {}
-        for name in pdos_atm_filenames:
-            with self.retrieved.base.repository.open(name, 'r') as pdosatm_file:
-                pdos_atm_array_dict[name] = np.atleast_2d(np.genfromtxt(pdosatm_file))
+        orbitals = self._parse_orbitals(header, structure, non_collinear, spinorbit)
+        bands, projections = self._parse_bands_and_projections(kpoint_blocks, len(orbitals))
+        energy, dos, pdos_array = self._parse_pdos_files(retrieved_temporary_folder, nspin, spinorbit, logs)
```
t-reents, line 66:
```suggestion
        energy, dos, dos_names, pdos_array = self._parse_pdos_files(retrieved_temporary_folder, nspin, spinorbit, logs)
```

src/aiida_quantumespresso/parsers/projwfc.py
```diff
@@ -2,458 +2,423 @@
...
+        for linkname, node in output_node_dict.items():
             self.out(linkname, node)
 
-        Dos_out = XyData()
-        Dos_out.set_x(energy, 'Energy', 'eV')
-        Dos_out.set_y(dos, 'Dos', 'states/eV')
-        self.out('Dos', Dos_out)
+        dos_out = XyData()
+        dos_out.set_x(energy, 'Energy', 'eV')
+        dos_out.set_y(dos, 'Dos', 'states/eV')
```
t-reents, line 76:
```suggestion
        dos_out.set_y(dos, dos_names, ['states/eV'] * len(dos_names))
```

t-reents (changes requested):
Thanks @mbercx, I went through it and it looks nice! Definitely easier to understand what is going on, compared to the last version.

As you mentioned, there was still an issue, namely the proper parsing of the total DOS #1041. I added the fix as suggestions (let me know if I should open a PR to your branch instead) and quickly tested them.  @AndresOrtegaGuerrero, please have a look as well.

mbercx:
@t-reents thanks a lot for the review! Feel free to add a commit here with your changes. If possible it would be great if:

* We add a commit that tests this issue, probably via a `pytest` data regression. This will adapt the YAML files corresponding to that regression test. The tests will still pass after the file is updated, but hopefully it's clear that the DOS parsing is incorrect.
* We add a second commit that fixes this parsing problem, clearly showing the change in the data regression YAML files.

This PR will then have 4 commits to merge in the end. I think that is the most clear for future developers.

mbercx:
> Thanks @mbercx I will give it a spin and see how it goes.

Haha, just noticed the beautiful pun here. Stellar work @Minotakm 😂

AndresOrtegaGuerrero:
thank you @mbercx I will test it once the new commits are merged then :)

src/aiida_quantumespresso/parsers/projwfc.py
```diff
@@ -63,19 +63,16 @@ def parse(self, **kwargs):
 
         orbitals = self._parse_orbitals(header, structure, non_collinear, spinorbit)
         bands, projections = self._parse_bands_and_projections(kpoint_blocks, len(orbitals))
-        energy, dos, pdos_array = self._parse_pdos_files(retrieved_temporary_folder, nspin, spinorbit, logs)
+        energy, dos_node, pdos_array = self._parse_pdos_files(retrieved_temporary_folder, nspin, spinorbit, logs)
+
+        self.out('Dos', dos_node)
```
mbercx, line 68: I'd be inclined to change this output name to `dos` (all lowercase), but that is technically backwards-incompatible. However, for the spin-polarised case, by fixing the bug we are also adding an extra array to the `XyData` output, so we are already breaking backwards incompatibility in a sense? Of course changing the output link name would also break backwards incompatibility for the non-spin-polarised case.

Thoughts @t-reents @sphuber @Minotakm @AndresOrtegaGuerrero?

src/aiida_quantumespresso/parsers/projwfc.py
```diff
@@ -300,11 +297,24 @@ def natural_sort_key(sort_key, _nsre=re.compile('([0-9]+)')):
...
+
+        # For spin-polarised calculations the total DOS is split up in spin up and down
+        if nspin == 2:
+            dos_node.set_y(
+                (pdostot_array[:, 1], pdostot_array[:, 2]),
+                ('dos_up', 'dos_down'),
+                ('states/eV', 'states/eV')
+            )
+        else:
+            dos_node.set_y(pdostot_array[:, 1], 'Dos', 'states/eV')
```
mbercx, line 317: Similar comment as above: I'd use lowercase `dos` for the array name here.

src/aiida_quantumespresso/parsers/projwfc.py
AndresOrtegaGuerrero: How much is the impact on previous calculations , not much right ?

mbercx: I mean the QEapp will definitely break, since it will look for the `Dos` output, not the `dos` one, I assume. ^^

AndresOrtegaGuerrero: we can open an issue to add logic for backwards-compatibility

mbercx: In the end, I think I'll do the inverse: I'll try and preserve backwards compatibility as much as possible, and open an issue to track breaking changes we want to make. I'd also really like to have `PwCalculation.outputs.structure` instead of `PwCalculation.outputs.output_structure` for example. ^^

Minotakm (approved):
LGTM! :rocket:

PR #751, opened 2021-10-07
`parse_output_error`: fix parser for QE stdout errors

### LINKED ISSUE #750: `parse_output_base`: Move detection of 'JOB DONE' to the end
Currently in `parse_output_base()`, the detection of `JOB DONE` string in QE stdout is at the beginning,
https://github.com/aiidateam/aiida-quantumespresso/blob/27c93e5848829219eda7ab6ee5256bfbbd960f42/aiida_quantumespresso/parsers/parse_raw/base.py#L27-L31
This will hide the detection of any QE reported error e.g.
 ```
     Error in routine set_para_diag (1):
     Too few bands for required ndiag
```
These kinds of errors are handled by the following lines at the end of `parse_output_base()`
https://github.com/aiidateam/aiida-quantumespresso/blob/27c93e5848829219eda7ab6ee5256bfbbd960f42/aiida_quantumespresso/parsers/parse_raw/base.py#L55-L57

Can we move the `JOB DONE` lines to the end so the errors won't be hidden by the too generic `ERROR_OUTPUT_STDOUT_INCOMPLETE`?

### PR DESCRIPTION
**Pull request by qiaojunfeng**
The previous code returned a wrong `line_number_end`, thus all the QE
stdout errors were discarded. Moreover, the code converted lines into a
`set` to remove repetitions, but this led to random ordering of error
messages in the final `logs.error` list, causing difficulties in pytest
data_regression. Now the repeated messages are explicitly discarded so
the error message ordering in QE stdout is still preserved. Finally, a
test is added to ensure errors are correctly recorded in `logs.error`.

Fix #750

### DIFF
--- a/aiida_quantumespresso/parsers/parse_raw/base.py
+++ b/aiida_quantumespresso/parsers/parse_raw/base.py
@@ -74,29 +74,36 @@ def parse_output_error(lines, line_number_start, logs, message_map=None):
             if marker in line:
                 if message is None:
                     message = line
-                logs.error.append(message)
+                if message not in logs.error:
+                    logs.error.append(message)
 
         for marker, message in message_map['warning'].items():
             if marker in line:
                 if message is None:
                     message = line
-                logs.warning.append(message)
+                if message not in logs.warning:
+                    logs.warning.append(message)
 
     # First determine the line that closes the error block which is also marked by ``%%%%%%%`` in the line
-    for line_number, line in enumerate(lines[line_number_start + 1:]):
+    # Skip the starting line '%%%%%%%%%%%%'
+    line_number_start += 1
+    for line_number, line in enumerate(lines[line_number_start:]):
         if '%%%%%%%%%%%%' in line:
-            line_number_end = line_number
+            line_number_end = line_number_start + line_number
             break
     else:
         return
 
     # Get the set of unique lines between the error indicators and pass them through the message map, or if not provided
     # simply append the message to the `error` list of the logs container
-    for message in set(lines[line_number_start:line_number_end]):
+    for message in lines[line_number_start:line_number_end]:
+        if message.strip() == '':
+            continue
         if message_map is not None:
             map_message(message, message_map, logs)
         else:
-            logs.error(message)
+            if message not in logs.error:
+                logs.error.append(message)
 
     return
 
--- a/tests/parsers/fixtures/parse_raw/base/aiida.out
+++ b/tests/parsers/fixtures/parse_raw/base/aiida.out
@@ -1,0 +1,130 @@
+
+     Program OPEN_GRID v.6.8 starts on  7Oct2021 at  1:52:44 
+
+     This program is part of the open-source Quantum ESPRESSO suite
+     for quantum simulation of materials; please cite
+         "P. Giannozzi et al., J. Phys.:Condens. Matter 21 395502 (2009);
+         "P. Giannozzi et al., J. Phys.:Condens. Matter 29 465901 (2017);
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by obgeneralao** [comment]
Hi
Is this issue related to the error I got as shown below?

```bash
+-> WARNING at 2021-12-04 14:20:42.160632+00:00
+-> ERROR at 2021-12-04 14:20:42.157234+00:00
+-> CRITICAL at 2021-12-04 14:20:42.152651+00:00
 |   File "/home/jupyter-admin/.local/lib/python3.9/site-packages/aiida_quantumespresso/parsers/pw.py", line 325, in parse_stdout
 |     parsed_data, logs = parse_stdout(stdout, parameters, parser_options, parsed_xml)
 |   File "/home/jupyter-admin/.local/lib/python3.9/site-packages/aiida_quantumespresso/parsers/parse_raw/pw.py", line 396, in parse_stdout
 |     trajectory_data['atomic_species_name'] = [data_lines[i + 1 + j].split()[1] for j in range(nat)]
 |   File "/home/jupyter-admin/.local/lib/python3.9/site-packages/aiida_quantumespresso/parsers/parse_raw/pw.py", line 396, in <listcomp>
 |     trajectory_data['atomic_species_name'] = [data_lines[i + 1 + j].split()[1] for j in range(nat)]
 | IndexError: list index out of range
```

sphuber:
@obgeneralao I doubt it. This seems to be a different problem in the parser where it assumes a certain format in the stdout. Could you share the stdout file as an attachment to this PR? Then we can try to determine the problem and fix it. You can obtain the output using the CLI with `verdi calcjob outputcat <PK>  > stdout.txt` where you replace `<PK>` with the pk of the failed calcjob.

obgeneralao:
Hi,

The ```verdi calcjob outputcat <PK>``` output is attached
[stdout.txt](https://github.com/aiidateam/aiida-quantumespresso/files/7659610/stdout.txt)

sphuber:
Thanks @obgeneralao . There is indeed a problem with your output file, but it is not due to the bug of this PR. Instead, it is because your output file is corrupted because you somehow ran a serial version of QE in parallel. You can see that at the beginning of your file, the header is printed multiple times. Therefore the output of the various processors that are all writing to the same file is being interleaved leading to a corrupt file and so the parser crashes. You should make sure that if you either compile a parallel version of QE, so with MPI support, or you run this executable with `metadata.options.withmpi = False` in the inputs of the `PwCalculation`.

obgeneralao:
Hi,

Thank you very much. I got it fixed by recompiling it with OpenMP.

mbercx:
@qiaojunfeng is it ok if I close this PR, since it's been superseded by the refactoring work done in https://github.com/aiidateam/aiida-quantumespresso/pull/756? I think the test you add here is very similar to the test that is added in https://github.com/aiidateam/aiida-quantumespresso/pull/714? If not, could you migrate it to that PR? Thanks!

qiaojunfeng:
> @qiaojunfeng is it ok if I close this PR, since it's been superseded by the refactoring work done in #756? 

Sure, always welcome a refactor 😁

> I think the test you add here is very similar to the test that is added in #714? If not, could you migrate it to that PR? Thanks!

The test here is only for the `parse_output_error` function, although I was a bit lazy that I reused a opengrid stdout with some errors to test the function 😅. Since this is a test for the `parse_output_error` function which is used everywhere, so I think it's better to test it in #756?

mbercx:
Great, thanks @qiaojunfeng! I think I'll adapt your file to do some testing on the base parsing indeed. Will close this PR.

PS: If you have a change to work on #714, that'd be great, then we can also get that merged :)

Issue #753, opened 2021-11-15
PhBaseWorkChain: not recover from ran out of walltime


unkcpz:
My `PhBaseWorkChain` got exit when reaching the walltime. But there is already the walltime handler there in `PhBaseWorkChain` to handle the walltime issue. 

The process report is:

```
2021-11-04 11:58:36 [4706 | REPORT]: [14787|PhBaseWorkChain|run_process]: launching PhCalculation<14790> iteration #1
2021-11-04 13:04:59 [5070 | REPORT]: [14787|PhBaseWorkChain|report_error_handled]: PhCalculation<14790> failed with exit status 120: The job ran out of walltime.
2021-11-04 13:04:59 [5071 | REPORT]: [14787|PhBaseWorkChain|report_error_handled]: Action taken: unrecoverable error, aborting...
2021-11-04 13:04:59 [5072 | REPORT]: [14787|PhBaseWorkChain|inspect_process]: PhCalculation<14790> failed but a handler detected an unrecoverable problem, aborting
2021-11-04 13:04:59 [5073 | REPORT]: [14787|PhBaseWorkChain|on_terminated]: remote folders will not be cleaned
```

versions:
`aiida-core==1.6.5`
`aiida-quantumespresso` current develop branch at https://github.com/aiidateam/aiida-quantumespresso/commit/0874d956c94bba6f2853f8a4984121330b5c77f8

Please tell me how to give you more information to anchor this issue.

unkcpz:
By search `exit status` of `120` in `aiida-core`. I find that it is in `CalcJob` for walltime detect in scheduler:

```
        spec.exit_code(120, 'ERROR_SCHEDULER_OUT_OF_WALLTIME',
            message='The job ran out of walltime.')
```

sphuber:
Yes, this was a newly added feature to `aiida-core`. Scheduler plugins can now add parsing to the scheduler stdout and stderr and return this exit code if they detect an OOW.  If the job was cleanly stopped, we can simply restart the calculation but that should be checked first. The `PhParser` should be updated to check for this exit code and then do additional parsing. If there was a clean stop, it should probably change the exit code to a `400` otherwise leave it at `120`.

unkcpz:
`PhParser` already try to check the `ERROR_OUT_OF_WALLTIME`. I think the problem is that the parser detect `Maximum CPU time exceeded` in output for this exit status while this line is not in the output text. 
[...]

PR #754, opened 2021-11-19
PhBaseWorkChain: add handle schedulor out of walltime

unkcpz:
fixes https://github.com/aiidateam/aiida-quantumespresso/issues/753

Add a handler for `exit_status==120` when scheduler wall time reached before the calculation shut down neatly.
The new calculation is restart from scratch by setting `['INPUTPH']['recover']=False`.


unkcpz:
I am not so sure about this `recover=false` setting will correctly do the restart from scratch. You may want to have a close look at this PR. Since for the handler `convergence_not_reached` it says restart from scratch but set the `restart_calc` and thus `recover=True` which to me this is not restart from scratch for PH calculation.

unkcpz:
@sphuber Could you have a look at this PR?

aiida_quantumespresso/parsers/ph.py
```diff
@@ -56,6 +56,10 @@ def parse(self, **kwargs):
         self.emit_logs(logs)
         self.out('output_parameters', orm.Dict(dict=parsed_data))
 
+        # scheduler plugin detected a problem
+        if self.node.exit_status is not None:
```
sphuber, line 60: This should be a more specific check. You should compare the status to the actual status of the OOW exit code

aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -109,6 +109,24 @@ def handle_unrecoverable_failure(self, node):
             self.report_error_handled(node, 'unrecoverable error, aborting...')
             return ProcessHandlerReport(True, self.exit_codes.ERROR_UNRECOVERABLE_FAILURE)
 
+    @process_handler(priority=610, exit_codes=PhCalculation.exit_codes.ERROR_SCHEDULER_OUT_OF_WALLTIME)
+    def handle_scheduler_out_of_walltime(self, node):
+        """Handle `ERROR_SCHEDULER_OUT_OF_WALLTIME` exit code: decrease the max_secondes and restart from scratch."""
+
+        # decrease max_seconds a lot in order to make sure the calculation shut down neatly before reaching the
+        # scheduler wall time
```
sphuber, line 117:
```suggestion
        # Decrease `max_seconds` significantly in order to make sure that the calculation has the time to shut down
        # neatly before reaching the scheduler wall time and one can restart from this calculation.
```

aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -109,6 +109,24 @@ def handle_unrecoverable_failure(self, node):
...
+        factor = 0.5
+        max_seconds = self.ctx.inputs.parameters.get('INPUTPH', {}).get('max_seconds', None)
+        max_seconds_new = max_seconds * factor
```
sphuber, line 120: It is possible that you will multiple `None` with `0.5` here. If `max_seconds` is not defined, you should create it yourself from the `max_wallclock_seconds`. I know that currently in the code it is "guaranteed" that it will be set because the workchain itself sets it, but still I think it is more robust to check here explicitly

aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -117,8 +135,8 @@ def handle_out_of_walltime(self, node):
         return ProcessHandlerReport(True)
 
     @process_handler(priority=410, exit_codes=PhCalculation.exit_codes.ERROR_CONVERGENCE_NOT_REACHED)
-    def handle_convergence_not_achieved(self, node):
-        """Handle `ERROR_CONVERGENCE_NOT_REACHED` exit code: decrease the mixing beta and restart from scratch."""
+    def handle_convergence_not_reached(self, node):
```
sphuber, line 138: Why change this method name?

aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -109,6 +109,24 @@ def handle_unrecoverable_failure(self, node):
...
+        """Handle `ERROR_SCHEDULER_OUT_OF_WALLTIME` exit code: decrease the max_secondes and restart from scratch."""
+
+        # decrease max_seconds a lot in order to make sure the calculation shut down neatly before reaching the
+        # scheduler wall time
+        factor = 0.5
+        max_seconds = self.ctx.inputs.parameters.get('INPUTPH', {}).get('max_seconds', None)
+        max_seconds_new = max_seconds * factor
+
+        self.ctx.restart_calc = node
+        self.ctx.inputs.parameters.setdefault('INPUTPH', {})['recover'] = False
```
sphuber, line 123: I am not too familiar with `ph.x` so I am not sure if this is the correct way to do it, but it could very well be. Just saying that I cannot really sign off on this as I have no idea. I would advise that we ask someone that is knowledgeable on `ph.x` to validate this strategy.

sphuber (changes requested):

sphuber:
> @sphuber Could you have a look at this PR?

Thanks @unkcpz . Mostly some minor comments. Regarding the restart strategy: as mentioned in a comment, I think we should ask someone that uses `ph.x` a lot, as I don't know.

aiida_quantumespresso/parsers/ph.py
unkcpz: My idea here is that when an exit status is already set it means the scheduler catch a problem, what it does is propagate it up and let it be handled by the workchain handler. 
Maybe I misunderstand the documentation here https://aiida.readthedocs.io/projects/aiida-core/en/latest/topics/calculations/usage.html?highlight=scheduler%20handler#scheduler-errors ?

aiida_quantumespresso/workflows/ph/base.py
unkcpz: In fact, there are two changes here :smile: 
First I change the function name from `achieved` to `reached`, I guess this should be the same as the exit status name? Any reason they are not the same previously?
Second, I remove the `from scratch` from the docstring since I think the setting here use the `recover=True` and it is not run from scratch. But as you suggest below, we need to ask someone more familiar with ph.x.

aiida_quantumespresso/parsers/ph.py
sphuber: That is completely true, but there could be a variety of exit statuses being set and we might not all want to treat them the same way. So what I am suggesting is to do this instead:
```python
# If the scheduler detected OOW, simply keep that exit code by not returning anything more specific.
if self.node.exit_status == PhCalculation.exit_codes.ERROR_SCHEDULER_OUT_OF_WALLTIME:
    return
```

The documentation does indeed not set the right example here. I will open a PR to fix it.

aiida_quantumespresso/workflows/ph/base.py
sphuber: > First I change the function name from achieved to reached, I guess this should be the same as the exit status name? Any reason they are not the same previously?

Fair enough, it is just that other base workchain (`PwBaseWorkChain` for example) have the same handler. If we make this change, would be good to do it there as well for consistency.

And I noticed the change in the docstring and didn't comment because I agree with it :)

sphuber:
Thanks for the changes @unkcpz . Everything looks ok now. If we can get a confirmation on the use of `recover = True` I will merge this.

unkcpz:
Thanks @sphuber ! I am asking Samuel about this, I think he has much experience in running ph calculations. Let's wait for a while, and I will keep this pr updated.

unkcpz:
Hi @sphuber , I confirm @sponce24 that the `parent_folder` should always be set which in our case the `self.ctx.restart_calc` always set to the previous calculation node to get `remote_folder`, either from finished pw run or previous failed ph run. 
Then set `recover=.false.` will control whether to recalculate the irr representations already saved.
So for the 'pure' restart from scratch, set the `self.ctx.restart_calc` and set `recover=.false.`.

sphuber (approved):

sphuber:
Thanks a lot @unkcpz

Issue #748, opened 2021-10-03
♻️ REFACTOR: Revisit/clean up each parser for v4.0.0


mbercx:
From v4.0.0, we have a clear policy on which QE versions the plugin will support. This would be a good time to go through each of the parsers and clean up any code that is designed to support old output formats. And, well, after having gone through the `ProjwfcParser`, clean up the code itself.

Below is a checkbox list with the various codes we have to look at. Once a PR is merged that cleans up the parser, these can be ticked!

- [ ] `CpParser`
- [ ] `DosParser`
- [ ] `MatDynParser`
- [ ] `NebParser`
- [ ] `PhParser`
- [ ] `PpParser`
- [ ] `ProjwfcParser`
- [ ] `PwParser`
- [ ] `Pw2gwParser`
- [ ] `Pw2wannier90Parser`
- [ ] `Q2rParser`

Also related to https://github.com/aiidateam/aiida-quantumespresso/issues/600

PR #756, opened 2021-11-23
♻️  REFACTOR: move basic parsing into `BaseParser` class

### LINKED ISSUE #750: `parse_output_base`: Move detection of 'JOB DONE' to the end

### LINKED ISSUE #797: Error during conversion of walltime during parsing of `pw.x` calculation
Reported by @Ezio-droid in https://github.com/aiidateam/aiida-core/discussions/5452

---

Hi all,

I am following the tutorial given [on following link](https://aiida-tutorials.readthedocs.io/en/tutorial-qe-short/source/sections/qe.html#importing-a-structure-and-inspecting-it). And while I was seeing the process report for the job I got following errors. Can someone suggest the reason for these errors?

```console
verdi process report 116
*** 116: None
*** (empty scheduler output file)
*** (empty scheduler errors file)
*** 3 LOG MESSAGES:
+-> CRITICAL at 2022-03-16 00:32:15.000564+00:00
| File "/home/tanay/miniconda3/envs/aiida/lib/python3.9/site-packages/aiida_quantumespresso/parsers/parse_raw/pw.py", line 434, in parse_stdout
| parsed_data['wall_time_seconds'] = convert_qe_time_to_sec(time)
| File "/home/tanay/miniconda3/envs/aiida/lib/python3.9/site-packages/aiida_quantumespresso/parsers/parse_raw/base.py", line 137, in convert_qe_time_to_sec
| num_seconds = (float(seconds) + float(minutes) * 60. + float(hours) * 3600. + float(days) * 86400.)
| ValueError: could not convert string to float: 'PWSCF : 2.46'
|
| During handling of the above exception, another exception occurred:
|
| File "/home/tanay/miniconda3/envs/aiida/lib/python3.9/site-packages/aiida_quantumespresso/parsers/pw.py", line 325, in parse_stdout
| parsed_data, logs = parse_stdout(stdout, parameters, parser_options, parsed_xml)
| File "/home/tanay/miniconda3/envs/aiida/lib/python3.9/site-packages/aiida_quantumespresso/parsers/parse_raw/pw.py", line 436, in parse_stdout
| raise QEOutputParsingError('Unable to convert wall_time in seconds.')
| aiida_quantumespresso.parsers.QEOutputParsingError: Unable to convert wall_time in seconds.
+-> ERROR at 2022-03-16 00:32:15.006631+00:00
+-> WARNING at 2022-03-16 00:32:15.009005+00:00
```

mbercx:
Fixes #750 
Fixes #797

Large refactoring of the basic parsing features to deduplicate code and rely on regular expressions via `re` instead of iterations over lines. The main changes are:

* The `Parser` class is renamed to `BaseParser` class to avoid confusion with the `aiida.core` `Parser` class.
* The `BaseParser.parse_stdout_from_retrieved` method can be used by all subclasses to obtain the `stdout`, basic parsed data and logs from retrieving and parsing the `stdout`.
* The error and warning messages are defined as class attributes. Base error messages are defined on the `BaseParser` class, which can be extended on the subclasses via the `class_[error,warning]_map` attributes.
* A `BaseParser.check_base_errors()` method is added, which checks for basic errors that indicate the `stdout` cannot be parsed for more data.
* Shared exit codes are moved to the `NamelistsCalculation` spec from that of the subclasses.
* Several changes in the logic of returning exit codes and emitting the logs. Some of the discussions on these topics have not been fully concluded, and will be revisited when integrating the `BaseParser` into the `PwParser`, see:
  https://github.com/aiidateam/aiida-quantumespresso/issues/953


mbercx:
Some timing information for a very long `aiida.out` file (~140k lines) without errors:

* Old `parse_output_base`:

```
14.4 ms ± 313 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

* New `parse_output_base`:

```
5.14 ms ± 68.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

and a very short (131 lines) file with errors:

* Old `parse_output_base`:

```
11.3 µs ± 231 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

* New `parse_output_base`:

```
48.1 µs ± 815 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

aiida_quantumespresso/parsers/projwfc.py
```diff
@@ -366,6 +344,14 @@ def parse(self, **kwargs):
         Dos_out.set_y(dos, 'Dos', 'states/eV')
         self.out('Dos', Dos_out)
 
+    def parse_stdout(self, stdout, out_info_dict):
+        """Parse the ``stdout`` content of a Quantum ESPRESSO ``projwfc.x`` calculation.
+
+        :param stdout: the stdout content as a string.
+        """
+        out_info_dict['out_file'] = stdout.split('\n')
```
mbercx, line 352: I know using the `out_info_dict` is messy as hell, but that'll be fixed in https://github.com/aiidateam/aiida-quantumespresso/pull/749. :)

mbercx:
@sphuber currently the `PwParser` and `CpParser` don't reuse the base parsing methods yet, but these are rather large and messy parsers I'd like to clean up in separate PRs for v4.0.

aiida_quantumespresso/calculations/projwfc.py
```diff
@@ -61,10 +61,6 @@ def define(cls, spec):
             message='The retrieved temporary folder could not be accessed.')
         spec.exit_code(303, 'ERROR_OUTPUT_XML_MISSING',
             message='The retrieved folder did not contain the required XML file.')
-        spec.exit_code(310, 'ERROR_OUTPUT_STDOUT_READ',
-            message='The stdout output file could not be read.')
-        spec.exit_code(312, 'ERROR_OUTPUT_STDOUT_INCOMPLETE',
-            message='The stdout output file was incomplete probably because the calculation got interrupted.')
         spec.exit_code(320, 'ERROR_OUTPUT_XML_READ',
```
sphuber, line 64: Think these XML exit codes are also kind of generic for codes that have an XML output. Shouldn't `PwCalculation` also have these exit codes? If so might be worth looking into abstracting them as well.

aiida_quantumespresso/calculations/pw2gw.py
```diff
@@ -37,16 +37,10 @@ def define(cls, spec):
         spec.output('eps', valid_type=orm.ArrayData,
             help='The `eps` output node containing 5 arrays `energy`, `epsX`, `epsY`, `epsZ`, `epsTOT`')
 
-        spec.exit_code(302, 'ERROR_OUTPUT_STDOUT_MISSING',
-            message='The retrieved folder did not contain the required stdout output file.')
         spec.exit_code(305, 'ERROR_OUTPUT_FILES',
             message='The eps*.dat output files could not be read or parsed.')
-        spec.exit_code(310, 'ERROR_OUTPUT_STDOUT_READ',
-            message='The stdout output file could not be read.')
         spec.exit_code(311, 'ERROR_OUTPUT_STDOUT_PARSE',
```
sphuber, line 42: This should also be generic and can be moved to base class

aiida_quantumespresso/parsers/base.py
```diff
@@ -3,15 +3,128 @@
...
+from aiida.parsers import Parser
+from aiida.common import AttributeDict
+from aiida.engine import ExitCode
+from aiida_quantumespresso.utils.mapping import get_logging_container
+from aiida_quantumespresso.parsers.parse_raw.base import convert_qe_time_to_sec
+
+__all__ = ('BaseParser',)
+
+
+class BaseParser(Parser):  # pylint: disable=abstract-method
```
sphuber, line 18:
```suggestion
class BaseParser(Parser, metaclass=abc.ABCMeta):
```

aiida_quantumespresso/parsers/base.py
```diff
@@ -3,15 +3,128 @@
...
+        return error_map
+
+    @classmethod
+    def get_warning_map(cls):
+        """The full error map of the parser class."""
+        warning_map = cls.base_warning_map.copy()
+        warning_map.update(cls.class_warning_map)
+        return warning_map
+
+    def _retrieve_parse_stdout(self, **kwargs) -> typing.Tuple[str, dict, AttributeDict]:
```
sphuber, line 46: Wouldn't call this `_retrieve` because it is a bit misleading with the usual meaning of retrieving in AiiDA. The files have actually already been retrieved, you are just reading them from the `retrieved` output node and parsing them. Think `parse_stdout` would be sufficient, but if you really want it more specific, maybe `parse_retrieved_stdout` would work?

aiida_quantumespresso/parsers/base.py
```diff
@@ -3,15 +3,128 @@
...
+        try:
+            with self.retrieved.open(filename_stdout, 'r') as handle:
+                stdout = handle.read()
+        except OSError:
+            logs.error.append('ERROR_OUTPUT_STDOUT_READ')
+            return {}, logs
+
+        parsed_data, stdout_logs = self.parse_stdout(stdout, **kwargs)
+
+        for log_type, log_items in stdout_logs.items():
```
sphuber, line 68: `log_type` here would be like `warning` or `error` etc.? If so, probably `log_level` is better because that is the canonical term for it.

aiida_quantumespresso/parsers/base.py
sphuber: Edit, I see now that you have the `parse_stdout` classmethod already that does the pure parsing of the stdout content. Maybe this should be called `parse_stdout_from_retrieved` ?

tests/parsers/test_dos/test_dos_default.yml
```diff
@@ -10,6 +10,6 @@ dos:
   - states/eV
   - states/eV
 parameters:
-  code_version: v.6.4.1
+  code_version: 6.4.1
```
sphuber, line 13: Should we keep this the same to not break backwards-compatibility?

tests/parsers/test_neb.py
```diff
@@ -58,7 +58,6 @@ def test_neb_default(fixture_localhost, generate_calc_job_node, generate_parser,
 
     assert calcfunction.is_finished, calcfunction.exception
     assert calcfunction.is_finished_ok, calcfunction.exit_message
-    assert not orm.Log.objects.get_logs_for(node)
```
sphuber, line 61: Why remove these tests? This is making sure there were no error/warning logs

aiida_quantumespresso/parsers/base.py
```diff
@@ -3,15 +3,128 @@
...
+    @classmethod
+    def get_error_map(cls):
+        """The full error map of the parser class."""
+        error_map = cls.base_error_map.copy()
+        error_map.update(cls.class_error_map)
+        return error_map
+
+    @classmethod
+    def get_warning_map(cls):
+        """The full error map of the parser class."""
```
sphuber, line 41:
```suggestion
        """The full warning map of the parser class."""
```

sphuber (changes requested):
Thanks a lot @mbercx great work! This really cleans things up nicely, just have some minor requests/suggestions

aiida_quantumespresso/calculations/projwfc.py
mbercx: I agree entirely! But I will leave refactoring the XML parsing to another PR for v4.0, where we can also clean up that part of the code significantly when we no longer support older versions of QE. I've added a note to https://github.com/aiidateam/aiida-quantumespresso/issues/748 on this.

aiida_quantumespresso/calculations/pw2gw.py
mbercx: Actually, this currently isn't returned in any parser. Maybe I should add a try/except with a generic `Exception` around the `parsed_data, stdout_logs = self.parse_stdout(stdout, **kwargs)` call in `BaseParser._retrieve_parse_stdout`?

tests/parsers/test_dos/test_dos_default.yml
mbercx: Hmm, perhaps. I preferred removing the `v.`, because it's cleaner and more consistent with e.g. the plugin version added to the metadata. But it's not a big deal, so probably backwards compatibility is the stronger argument.

tests/parsers/test_neb.py
mbercx: Actually, with the new warning parsing there is a warning here for the line:

```
DEPRECATED: symmetry with ibrav=0, use correct ibrav instead
```

Let me see if I can just get the errors tested here.

mbercx: This should do the trick:

```python
assert not [log for log in orm.Log.objects.get_logs_for(node) if log.levelname == 'ERROR']
```

tests/parsers/test_dos/test_dos_default.yml
mbercx: Looking at this again, I'd be less shy about breaking backwards compatibility regarding this version number. It's probably not being used at all atm, and removing the `v.` is cleaner in my view.

mbercx:
@sphuber she's almost ready, just a bit more cleaning. I would also:

1. Remove the `v.` for the code versions. 
2. Remote the `wall_time`. It's a useless string and there is `wall_time_seconds`.

src/aiida_quantumespresso/parsers/matdyn.py
```diff
@@ -3,27 +3,31 @@
...
+        base_exit_code = self.check_base_errors(logs)
+        if base_exit_code:
+            return self.exit(base_exit_code, logs)
 
-        if 'JOB DONE' not in retrieved.base.repository.get_object_content(filename_stdout):
-            return self.exit(self.exit_codes.ERROR_OUTPUT_STDOUT_INCOMPLETE)
+        self.out('output_parameters', orm.Dict(parsed_data))
 
-        if filename_frequencies not in retrieved.base.repository.list_object_names():
-            return self.exit(self.exit_codes.ERROR_OUTPUT_STDOUT_READ)
```
mbercx, line 26: This is removed on purpose:

1. I'm don't think this _ever_ happens if the stdout is "complete"
2. It should use its own exit code, and I don't want to define one if [1] is true.

src/aiida_quantumespresso/parsers/base.py
```diff
@@ -50,14 +113,96 @@ def emit_logs(self, logging_dictionaries, ignore=None):
...
 
         This is a utility function if one wants to return from the parse method and automically add the exit message
-        associated to the exit code as a log message to the node: e.g. `return self.exit(self.exit_codes.LABEL))`
+        associated to the exit code as a log message to the node: e.g. ``return self._exit(self.exit_codes.LABEL))``
 
         :param exit_code: an `ExitCode`
         :return: the exit code
         """
+        if logs:
+            self.emit_logs(logs)
```
mbercx, line 154: I was wondering what to do here. I liked the idea of always passing the `logs` to the `exit` method, so the logs are always emitted before the parsing stops. But this lead to a few questions:

1. What does the `emit_logs` method _really_ do? It seems to only use `getattr`, how does it really "emit" anything?
2. What's the purpose of this `logger` exactly? 
3. this reminds me, perhaps we should end each `parse` method with an explicit `return` of a zero exit code via the `exit` method? Explicit is better than implicit, and then it passes through this method which emits the logs (now).

mbercx:
@sphuber ok, I refactored my initial refactor quite a bit, so perhaps this needs a full review again, apologies. It was also quite a bit of work to fix all the merge conflicts after rebase, so I would now really push to get this PR in.

A couple of quick notes that I decided upon:

* The code version in the `output_parameters` shouldn't have a `v.` prefix. It's redundant.
* The `wall_time` output is similarly redundant and less useful than `wall_time_seconds`.

Now a more elaborate design discussion on the changes here and TODO's for parsers.

## Keeping the `parse` method readable

Understanding what the parser parses when and which exit codes it returns at what point is difficult with many of the parser classes. We want to make sure that:

1. The "best" exit code is returned, i.e. the one that is most specific or allows for useful error handling.
2. As much data as possible is parsed and added to the outputs of the calculation job.

To improve clarity and make these two requirements possible, I think the `parse` method should:

1. Not contain any _actual_ parsing. Any real parsing should happen in a separate method.
3. Instantiate a single logging container that is passed _through_ these parsing methods.
4. Take care of decisions on when to return which exit code.
5. Output data at the appropriate time.

Note that this PR _does not_ fully implement these ideas for all parser classes. It's already big enough as it is. But it already shows how this would be possible for e.g. the `open_grid.x` parser.

## Other improvements 

* **Exit code registry**: Although this PR already make an effort of making sure exit codes are consistent across the calculation job classes, it's still tricky to keep track of all the exit codes they are used any why they are there or have a certain number. I think we should create an (automated) exit code registry in the documentation where we keep track of all this. Although two calculation jobs can in principle have the same exit code for different issues, we should probably also avoid this.

* **Stop using strings for exit codes**: Now we use strings to keep track of the exit codes. I think this is a bit fragile and doesn't allow for nice features like tab-completion. It would be better to have a different approach, but I'd have to think a bit on how to do this.

* `PhCalculation` and `PpCalculation` aren't subclasses from `NamelistCalculation`, and hence have to redefine the basic exit codes. So do `PwCalculation` and `CpCalculation`. There may be reasons to treat these calculation jobs differently, but we should investigate if we can't further consolidate this calculation job hierarchy.

mbercx:
Because of the size of the PR, and the fact that even though the tests pasts it can definitely break things, perhaps:

* @PNOGillespie @superstar54 maybe you can have a look a the XSpectra parsing?
* @qiaojunfeng can you have a look at the `open_grid.x` parsing changes?

src/aiida_quantumespresso/parsers/base.py
sphuber: > What does the `emit_logs` method _really_ do? It seems to only use `getattr`, how does it really "emit" anything?

Believe it or not, it emits log messages 😄 It is easy to miss but the line
```
getattr(self.logger, level)(stripped)
```
will retrieve a method of the `Logger` instance stored at `self.logger`. Since `level` is a log-level in lower-case, e.g., `warning` or `error`, `getattr(self.logger, level)` will return the bound _method_ `warning` or `error` in that case. We then _call_ it passing the message `...(stripped)`. So essentially it is doing something like `self.logger.warning(stripped)`.

> What's the purpose of this logger exactly?

It is configured with the associated `CalcJobNode` whose outputs are being parsed and so any log messages are stored as `Log` entries linked to the node. They will show up in the output of `verdi process report` for example. This is useful information for the user.

> this reminds me, perhaps we should end each parse method with an explicit return of a zero exit code via the exit method? Explicit is better than implicit, and then it passes through this method which emits the logs (now).

Yes, but we should then also remember to first check whether an exit status was already set by the scheduler plugin. For example if the scheduler parser detected the job was killed due to an OOW error, the associated exit status (I believe 120) is already set on the node. So before we return `ExitCode(0)`, we should check if `self.node.exit_status is not None` in which case we probably want to return `None` to keep that status, because returning `ExitCode(0)` will override it.

mbercx: Great, thanks for the clarification, @sphuber! 🙏 

> So essentially it is doing something like self.logger.warning(stripped).

Right, of course, it's quite clear after this explanation. I guess for this kind of use case, a `CalcJobNode.get_logger(level)` method would be more readable than a `logger` property. ^^

Now that I understand better, I'd be inclined to merge the `emit_logs` and `exit` methods into one method with signature:

```python
def exit(self, logs: AttributeDict = None, exit_code: ExitCode = None) -> ExitCode:
```

which:

1. Emits any logs left in the `logs` input, if provided.
2. If `exit_code` has been provided, emit its `message` via `self.logger`.
3. If not, check if there is already an `exit_status` set on the node and log this `message` (or will this have been done already?)
4. Return 
   * `exit_code` if it was provided.
   * `None` if the node already has an `exit_status`.
   * `ExitCode(0)` else

sphuber: > Right, of course, it's quite clear after this explanation. I guess for this kind of use case, a `CalcJobNode.get_logger(level)` method would be more readable than a `logger` property. ^^

Don't think that would be a correct name, because you cannot return a logger for a particular level, you log a message at a particular level through a logger. That being said, I think you _can_ easily make the current code more readable. I think the `Logger` class has a `log` method where you can pass the desired level as an argument, so that line would become `self.logger.log(level, stripped)`. Pretty sure that should work.

> Now that I understand better, I'd be inclined to merge the emit_logs and exit methods into one method with signature:

I would keep `emit_logs` as a separate function, because it can be useful to call it directly elsewhere still. Fine to move the call in all parsers to the `exit` method.

> If not, check if there is already an exit_status set on the node and log this message (or will this have been done already?)

This will already have been done by the engine before calling the `Parser`. So nothing to do here but to "keep" it if the parser itself cannot provide a more specific exit code.

mbercx: I've adapted the method as discussed, see 6529ea6d22cf532ee07496a97c66838d987b35f7.

> I would keep emit_logs as a separate function, because it can be useful to call it directly elsewhere still. Fine to move the call in all parsers to the exit method.

If we _always_ end the `parse` method with the `exit` method for clarity, and always emit _all_ the logs there, what would be the use case of calling the `emit_logs` method separate?

qiaojunfeng:
>     * @qiaojunfeng can you have a look at the `open_grid.x` parsing changes?

Had a quick look and seems good to me!
Anyhow `open_grid` is very simple so if the tests passed then I wouldn't worry.

src/aiida_quantumespresso/parsers/base.py
```diff
@@ -25,39 +88,127 @@ def emit_logs(self, logging_dictionaries, ignore=None):
...
-            logging_dictionaries = [logging_dictionaries]
+        if not isinstance(logs, (list, tuple)):
+            logs = [logs]
 
-        for logs in logging_dictionaries:
+        for logs in logs:
             for level, messages in logs.items():
                 for message in messages:
 
-                    if message is None:
```
mbercx, line 40: Note: I removed this, because to me it seems we have full control of how the `logs` are generated, and if one of them is `None` it probably indicates an issue in the parser code that we should fix.

src/aiida_quantumespresso/parsers/base.py
```diff
@@ -25,39 +88,127 @@ def emit_logs(self, logging_dictionaries, ignore=None):
...
-                        continue
-
                     stripped = message.strip()
 
-                    if not stripped or stripped in ignore:
```
mbercx, line 45: Same as https://github.com/aiidateam/aiida-quantumespresso/pull/756/files#r1192076537. If we have an empty log message something must have gone wrong that we should fix?

src/aiida_quantumespresso/parsers/base.py
```diff
@@ -25,39 +88,127 @@ def emit_logs(self, logging_dictionaries, ignore=None):
                 'error': ['Self-consistency was not achieved']
             }
 
-        :param logging_dictionaries: log dictionaries
+        :param logs: log dictionaries
         :param ignore: list of log messages to ignore
         """
         ignore = ignore or []
 
-        if not isinstance(logging_dictionaries, (list, tuple)):
```
mbercx, line 33: I was also inclined to remove this, but then noticed how the `pw.x` parser relies on this heavily and I'm trying not to touch this one too much to make your rebase of the XML work easier. It did raise a question: should we:

1. Instantiate a _single_ log container, that is passed _through_ the parsing methods, as I'm doing in this PR?
2. Instantiate _multiple_ log containers and return them as a list, as is done e.g. in the `PwParser`?

I'm marginally in favor of [1], but [2] would also work, and allow us to keep the logs separated, which might have benefits?

src/aiida_quantumespresso/parsers/base.py
```diff
@@ -25,39 +88,127 @@ def emit_logs(self, logging_dictionaries, ignore=None):
...
-                    if message is None:
-                        continue
-
                     stripped = message.strip()
 
-                    if not stripped or stripped in ignore:
+                    if stripped in ignore:
                         continue
 
-                    try:
```
mbercx, line 48: Oh, almost forgot: this one should also only really happen if we messed up the logs somehow, right?

PNOGillespie:
Hi @mbercx, I'm running some live-fire tests of XSpectra to check for issues. Tests with the `XspectraCrystalWorkChain` appear to be fine and the `output_parameters` are as expected, I'm now setting up and re-running tests of the XSpectra-specific error codes to check those too.

PNOGillespie:
@mbercx. To update the previous comment: I've tested all the XSpectra-specific exit codes, and they all work fine as well. The WorkChains themselves also rely on information in the `output_parameters` so, since the `WorkChains` run without issues, I will say that we're good for XSpectra-related stuff.

mbercx:
Great, thanks @PNOGillespie and @qiaojunfeng! I guess now she just needs a proper review, @sphuber. 😉

src/aiida_quantumespresso/parsers/base.py
```diff
@@ -3,15 +3,78 @@
...
+        """The full warning map of the parser class."""
+        warning_map = cls.base_warning_map.copy()
+        warning_map.update(cls.class_warning_map)
+        return warning_map
+
+    def parse_stdout_from_retrieved(self, logs: AttributeDict) -> Tuple[str, dict, AttributeDict]:
+        """Read and parse the ``stdout`` content of a Quantum ESPRESSO calculation.
+
+        :param logs: Logging container that will be updated during parsing.
+        :returns: size 3 tuple: (``stdout`` content, parsed data, updated logs).
```
sphuber, line 52: Why return the logs if they are passed in by reference and so are updated in place? Typically not a huge fan of tuple return values as it is often not immediately clear what is being returned.

src/aiida_quantumespresso/parsers/base.py
```diff
@@ -3,15 +3,78 @@
...
+        try:
+            parsed_data, logs = self._parse_stdout_base(stdout, logs)
+        except Exception as exception:
+            logs.error.append('ERROR_OUTPUT_STDOUT_PARSE')
+            logs.error.append(exception)
+            return stdout, {}, logs
+
+        return stdout, parsed_data, logs
+
+    def emit_logs(self, logs: AttributeDict, ignore: list = None) -> None:
```
sphuber, line 77: Incorrect type, as it still accepts a list as well
```suggestion
    def emit_logs(self, logs: list[AttributeDict] | tuple[AttributeDict] | AttributeDict, ignore: list = None) -> None:
```
Should add `from __future__ import annotations` to top of the file. I think we should start using this in favor of `typing.List` and the like as it is easier to read and cleaner.

src/aiida_quantumespresso/parsers/base.py
sphuber: I think this may have accommodated `None` versions as well, but you are probably right that this would be a problem in our code, and so we rather fix it then rely on this.

src/aiida_quantumespresso/parsers/base.py
sphuber: I may have added this because the `REPORT` level is an AiiDA only-thing. If the logging container were to be used outside of AiiDA where logging is not setup by it and so `Logger.report` wouldn't exist. But I guess it's fine to remove since this is anyway only usable by `aiida-quantumespresso` I think

src/aiida_quantumespresso/parsers/base.py
```diff
@@ -25,39 +88,127 @@ def emit_logs(self, logging_dictionaries, ignore=None):
...
+        # These exit codes have additional information that needs to be formatted in the message.
+        for exit_code in [
+            'ERROR_OUTPUT_STDOUT_READ',
+            'ERROR_OUTPUT_STDOUT_PARSE'
+        ]:
+            if exit_code in logs.error:
+                exception = logs.error[logs.index(exit_code) + 1]
+                return self.exit_codes.get(exit_code).format(exception=exception)
+
+    def exit(self, exit_code: ExitCode = None, logs: AttributeDict = None) -> ExitCode:
```
sphuber, line 138:
```suggestion
    def exit(self, exit_code: ExitCode | None = None, logs: AttributeDict | None = None) -> ExitCode:
```

src/aiida_quantumespresso/parsers/base.py
```diff
@@ -25,39 +88,127 @@ def emit_logs(self, logging_dictionaries, ignore=None):
...
+            if exit_code in logs.error:
+                return self.exit_codes.get(exit_code)
+
+        # These exit codes have additional information that needs to be formatted in the message.
+        for exit_code in [
+            'ERROR_OUTPUT_STDOUT_READ',
+            'ERROR_OUTPUT_STDOUT_PARSE'
+        ]:
+            if exit_code in logs.error:
+                exception = logs.error[logs.index(exit_code) + 1]
```
sphuber, line 135: Hmm, interesting approach. I can see how this is the most straightforward as you don't have to adapt the structure of the container and can just keep a list, but it is quite fragile I think. Ideally, the container should become a dictionary, with the key the error code, and the value can be a dictionary with additional information, for example argument to format the exit message.

src/aiida_quantumespresso/parsers/base.py
```diff
@@ -25,39 +88,127 @@ def emit_logs(self, logging_dictionaries, ignore=None):
...
 
-                    try:
-                        getattr(self.logger, level)(stripped)
-                    except AttributeError:
-                        pass
+                    getattr(self.logger, level)(stripped)
 
-    def exit(self, exit_code):
-        """Log the exit message of the give exit code with level `ERROR` and return the exit code.
+    def check_base_errors(self, logs: AttributeDict) -> Optional[ExitCode]:
```
sphuber, line 110: There might be problems with this approach in the future. For example the `PwParser`, it may be possible that the stdout could not be parsed, but the XML _can_ and so we can still parse and return useful data.

src/aiida_quantumespresso/parsers/base.py
```diff
@@ -25,39 +88,127 @@ def emit_logs(self, logging_dictionaries, ignore=None):
...
+            self.emit_logs(logs)
+
+        if exit_code is not None:
+            self.logger.error(exit_code.message)
+        elif self.node.exit_status is not None:
+            exit_code = ExitCode(self.node.exit_status, self.node.exit_message)
+        else:
+            exit_code = ExitCode(0)
+
         return exit_code
```
sphuber, line 162: This logic is not correct. Imagine the following situation: job gets submitted, but the node crashes. This is parsed by the `SlurmScheduler` and it sets `140:ERROR_NODE_FAILURE` on the `CalcJobNode`. Since the job crashed, `stdout` will either be missing or probably incomplete and so the base output parser will return the `ERROR_STDOUT_MISSING` exit code (or comparable) and in this logic, this will therefore _always_ override the scheduler status. But I would argue that the scheduler exit code is more useful in this case and we definitely want to keep it.

I think ultimately, each concrete parser implementation needs to parse the output and check _itself_ whether a scheduler exit code already exists and determine whether it can provide a more specific/informative exit code.

src/aiida_quantumespresso/parsers/dos.py
```diff
@@ -3,46 +3,36 @@
...
-        # check that the dos file is present, if it is, read it
+        base_exit_code = self.check_base_errors(logs)
+        if base_exit_code:
+            return self.exit(base_exit_code, logs)
+
+        self.out('output_parameters', Dict(parsed_stdout))
+
+        for exit_code in ['ERROR_OUTPUT_STDOUT_INCOMPLETE']:
+            if exit_code in logs.error:
+                return self.exit(self.exit_codes.get(exit_code), logs)
```
sphuber, line 28: I saw how you used this for/if construct in `check_base_errors` but here it makes little sense since there is just one exit code you check for and there is no other loop where there are multiple to keep the symmetry with. So here I see no reason to keep the loop. 
```suggestion
        if 'ERROR_OUTPUT_STDOUT_INCOMPLETE'in logs.error:
            return self.exit(self.exit_codes.ERROR_OUTPUT_STDOUT_INCOMPLETE, logs)
```
idem for other parsers

src/aiida_quantumespresso/parsers/base.py
sphuber: This currently not causing the tests to fail, because there only is such a test for `PwParser` and there you haven't made use of this method since you wanted to not touch it. But this also means that we probably won't be able to use it. Maybe we can leave this for now then as it is, but we will have to anticipate changing it in the future when other parsers will want to become aware of scheduler exit codes and not override them

sphuber (changes requested):

src/aiida_quantumespresso/parsers/base.py
mbercx: I do see your point. To me, I like making it clear that the `logs` input is updated by also passing it as an output. I guess it also related to the question I pose [here](https://github.com/aiidateam/aiida-quantumespresso/pull/756#discussion_r1192080727), and makes me inclined to simply generate separate logging containers in each parser method, but that also feels than ideal...

Maybe we should keep track of the logs on a class variable? That seems sensible. Initially I also wasn't sure why we needed this extra logging item because we can just actually use the `logger`. Then I figured this is so we can keep track of errors and then pick and choose which one to return as an exit code. But can't we just inspect the logger as well? Feels like we doing double work.

Re returning tuples: I can see your point, but what is the alternative for returning multiple outputs? A dictionary is even more opaque to me. Also: with the hover features in VSCode it's quite clear what is being returned: 🙃 

Screenshot 2023-05-18 at 20 00 20

src/aiida_quantumespresso/parsers/base.py
mbercx: Hehe, when implementing this I was already 95% sure it wouldn't pass review. Yes, I think adapting the logging container probably makes more sense. Let me resolve the other issues first and give this some thought, also in relation to the other discussions on the container. It's clear we should rethink this aspect.

src/aiida_quantumespresso/parsers/base.py
mbercx: But in the end the `Parser` implementation still has full control over what to do, no? This method simply does these basic checks and does some extra formatting. It might not be necessary anymore if we redesign the structure of the logging container, however.

src/aiida_quantumespresso/parsers/base.py
mbercx: Good points. I think the logic was indeed that in case the developer wants to return a scheduler exit code they should look for and return it in the `parse` method. This one just checks that in case the developer doesn't want to do this for whatever reason, the exit code is still added in case `self.exit()` is called at the end.

But I suppose scheduler failures will almost _always_ be more informative/useful?

src/aiida_quantumespresso/parsers/dos.py
mbercx: Hehe, fair. I guess the idea was that a future developer would add more of these later and the `for/if` constructor would make it obvious on how to add them. But those future developers will most likely be one of us and it's not that complicated anyway.

mbercx: Actually, it should probably be this:

```
        for exit_code in list(self.get_error_map().values()) + ['ERROR_OUTPUT_STDOUT_INCOMPLETE']:
            if exit_code in logs.error:
                return self.exit(self.exit_codes.get(exit_code), logs)
```

Though currently the only error in there is the `ERROR_OUT_OF_WALLTIME`. Which also isn't defined on every `NamelistsCalculation` subclass. Will go through the various parsers and see where it makes sense to add.

mbercx: See [7460b6c](https://github.com/aiidateam/aiida-quantumespresso/pull/756/commits/7460b6cf5e46dc48a0c536175305d898b5fbf22d) and [064df48](https://github.com/aiidateam/aiida-quantumespresso/pull/756/commits/064df48326e58f1a443530a4e89f2b8f5f18f78c)

src/aiida_quantumespresso/parsers/base.py
mbercx: [ef396dc](https://github.com/aiidateam/aiida-quantumespresso/pull/756/commits/ef396dcd0bfa38fb0ce5cfc9186117f0c92e7140)

src/aiida_quantumespresso/parsers/base.py
sphuber: You might not believe it, but I am actually VSCode myself now. But I think that in general making code design decisions based on a feature of the particular editor that we ourselves just happen to be using, is a bad idea, because not everyone is using the same tools. But fine, let's keep this as is.

sphuber:
@mbercx did you want to get this merged before the release? If so, is there anything to be done from your side? Otherwise I will approve. We can deal with the open points later. I also would like to have this done so I can finalize my refactor PR for `PwParser`.

mbercx:
> @mbercx did you want to get this merged before the release? If so, is there anything to be done from your side? Otherwise I will approve. We can deal with the open points later. I also would like to have this done so I can finalize my refactor PR for `PwParser`.

I've opened https://github.com/aiidateam/aiida-quantumespresso/issues/953 to keep track of the discussions that are left here. I think it makes sense to simply reconsider these decisions when integrating the `BaseParser` in the `PwParser`, we'll have a clearer idea of what to do during this process, I think.

sphuber (approved):

sphuber:
Thanks @mbercx let's get this merged

yakutovicha:
Hi 👋

Just wondering, are you guys planning on making a major release after this PR is merged?

There were API changes, as far as I can see.

sphuber:
We were getting close to releasing v4.4.0: https://github.com/aiidateam/aiida-quantumespresso/pull/951 @mbercx want to finish that one up and release it?

mbercx:
> want to finish that one up and release it?

Yessir! Sorry for the slow response.

> There were API changes, as far as I can see.

The internal parsing API was changed quite a bit, but I think the vast majority of the changes are backwards incompatible. Looking at the data regression files for the tests, the only changes in parsing output is the `code_version` and `wall_time_seconds`. Which API changes are you worried about breaking e.g. the QEapp?

yakutovicha:
> Which API changes are you worried about breaking e.g. the QEapp?

Mostly about the pp parser (that is the one I noticed). The `parse_gaussian` method, for instance. Since the name doesn't start with an underscore, it is part of the public API of the class, strictly speaking.

I didn't closely look at the PR, but there might be others.

However, it depends on how closely you want to follow the semantic versioning.

mbercx:
> However, it depends on how closely you want to follow the semantic versioning principles.

Yeah, that's fair. Perhaps we should be clearer in defining the public API. I don't expect users to use the `Parser` classes directly, since in many cases you need to parse from a node (which perhaps we should change over time, but then you definitely break the API).

The parser classes are also quite messy, and we've been working on improving that. Fixing this technical debt would be more important than strictly maintaining backwards compatibility for a part of the API that I don't think users are relying on directly.

PR #757, opened 2021-11-23
Add diagonalization errors handler to ph base workchain

unkcpz:
This is the quick workaround before PR https://github.com/aiidateam/aiida-quantumespresso/pull/756, the base stdout parsing by @mbercx get finished and merged. 

This PR fixes two issues of Ph calculation: a) when the calculation failed without 'Job Done' in the output file, the error was not caught and the calculation will be finished with exit status 0.  b) add a handler to 'problems computing cholesky' failed calculation.


sphuber (approved):
All good, thanks @unkcpz !

PR #758, opened 2021-11-24
PwBaseWorkChain: handle too high npools

unkcpz:
hi @mbercx , as I mentioned I add a handler for npool too high error. 
The implement is, I simply finish the PwBaseWorkChain and tag exit status to higher than 700 to indicate the 'warnings'.

### DIFF
--- a/aiida_quantumespresso/workflows/pw/base.py
+++ b/aiida_quantumespresso/workflows/pw/base.py
@@ -117,6 +117,8 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         spec.exit_code(710, 'WARNING_ELECTRONIC_CONVERGENCE_NOT_REACHED',
             message='The electronic minimization cycle did not reach self-consistency, but `scf_must_converge` '
                     'is `False` and/or `electron_maxstep` is 0.')
+        spec.exit_code(720, 'WARNING_NPOOLS_TOO_HIGH',
+            message='The npools set too high but the calculation still finished.')
         # yapf: enable
 
     @classmethod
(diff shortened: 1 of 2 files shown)


aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -117,6 +117,8 @@ def define(cls, spec):
         spec.exit_code(710, 'WARNING_ELECTRONIC_CONVERGENCE_NOT_REACHED',
             message='The electronic minimization cycle did not reach self-consistency, but `scf_must_converge` '
                     'is `False` and/or `electron_maxstep` is 0.')
+        spec.exit_code(720, 'WARNING_NPOOLS_TOO_HIGH',
+            message='The npools set too high but the calculation still finished.')
```
mbercx, line 121:
```suggestion
            message='The k-point parallelization `npools` is too high but the calculation still finished successfully.')
```

aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -663,3 +665,14 @@ def handle_electronic_convergence_warning(self, calculation):
...
         self.results()  # Call the results method to attach the output nodes
         return ProcessHandlerReport(True, self.exit_codes.WARNING_ELECTRONIC_CONVERGENCE_NOT_REACHED)
+
+    @process_handler(priority=320, exit_codes=[
+        PwCalculation.exit_codes.ERROR_NPOOLS_TOO_HIGH,
+    ])
+    def handle_npools_too_high(self, calculation):
+        """Handle `ERROR_NPOOLS_TOO_HIGH': consider finished."""
+        self.ctx.is_finished = True
+        action = 'npools set too high for the calculation but the calculation still finished: consider finished.'
```
mbercx, line 675:
```suggestion
        action = 'k-point parallelization `npools` set too high but the calculation still finished: consider finished.'
```

mbercx (changes requested):
All good, thanks @unkcpz! I think the approach of catching the error and finishing with a 7xx exit code is a good one, just left some minor nitpicks.

I was curious why the plugin was simply raising an error previously, and it seems that when the exit code was introduced (#444 , QE v6.4.1), QE would not continue running but simply stop the calculation:

https://github.com/greschd/aiida-quantumespresso/blob/59ded336fda6fa0cd087f88579e9d742eb231fed/tests/parsers/fixtures/pw/failed_npools_too_high/aiida.out

If you have an example for the `aiida.out` for a more recent QE version, maybe you could update the file for this test?

unkcpz:
hi @mbercx, sorry for not finishing this PR. I think it was after I start using your hyperqueue this issue never appear to me. Moreover, this problem was caused in my calculation that I only have very few kpoints but I used high npools. I changed that in my workflow and also pass by this issue. 

So in this case, I think it makes sence to raise an error for that, since set npools too high will anyway introduce the performance issue. I close this PR for now, happy to reopen it..

Issue #716, opened 2021-08-20
PDOSWorkChain - align to fermi energy


t-reents:
In the `PDOSWorkChain` there is the argument `align_to_fermi` to align the specified energy range of the DOS/PDOS around the fermi-energy (if `True`).

https://github.com/aiidateam/aiida-quantumespresso/blob/c49def7a238617439feb75afd0d0cf9fbb681613/aiida_quantumespresso/workflows/pdos.py#L220-L230

When I look at my results this shift is not performed. Moreover, I am confused, when I look at the code, how this would work in general (maybe I am overlooking something).

https://github.com/aiidateam/aiida-quantumespresso/blob/c49def7a238617439feb75afd0d0cf9fbb681613/aiida_quantumespresso/workflows/pdos.py#L450-L476

As shown above, it is checked whether `align_to_fermi` is specified in the parameters-dict. I was wondering whether the user has to add `align_to_fermi` to the parameters-dict in the inputs for the DOS/PDOS, when creating the builder? If not, why does the code not check if `self.inputs.align_to_fermi` is True or not at this point?
So basically I am wondering when the input `align_to_fermi` is actually used?

sphuber:
Hi @t-reents, thanks for the report. There indeed seems to be a problem here as the `align_to_fermi` input is completely ignored. With the current code, you can simply get it to run by adding it to the `dos.parameters` and `pdos.parameters` namespaces, e.g.:
````python
inputs = {
    'dos': {
        'parameters': Dict(dict={
            ...
            'align_to_fermi': True
        }
    }
}
```
That being said, I am not sure what the actual solution should be. It seems the top-level `align_to_fermi` input can simply be removed. I think it merely complicates the interface because since it potentially applies to both the `dos` and `pdos` calculation, and the value can also be specified for each explicit step, defining which dominates in all possible combinations becomes confusing. Also the help string of the top-level input is incomplete. It seems to be referencing the validation of the `align_to_fermi` key, but _only_ the one specified in the respective `parameters` dict, not the top-level one. @chrisjsewell what was your original reasoning here?

t-reents:
Hi @sphuber thanks for your quick reply. Sure this way it will run, was just a little bit confused whether I am overlooking something or if it's indeed ignored.

Thank you!

sphuber:
Yeah, that seems like a mistake and your correction is right.

mbercx:
[...]

PR #764, opened 2021-12-27
PDOSWorkChain - align energy range to fermi level

t-reents:
Hi,
I implemented the changes that were suggested in this comment https://github.com/aiidateam/aiida-quantumespresso/issues/716#issuecomment-909201940.

I already talked about this PR with @mbercx. I am not familiar with pytest, so I did not include additional tests. I would be happy to get some help concerning the tests.

### DIFF
--- a/src/aiida_quantumespresso/workflows/pdos.py
+++ b/src/aiida_quantumespresso/workflows/pdos.py
@@ -10,10 +10,10 @@ This requires four computations:
 
 Additional functionality:
 
-- Setting ``'align_to_fermi': True`` in the inputs will ensure that the energy range is centred around the Fermi
-  energy when `Emin` and `Emax` are provided for both the `dos` and `projwfc` inputs. This is useful when you are only
-  interested in a certain energy range around the Fermi energy. By default the energy range is extracted from the
-  NSCF calculation.
+- Setting ``'energy_range_vs_fermi'`` in the inputs allows to specify an energy range around the Fermi level that should
+    be covered by the DOS and PDOS. This is useful when you are only interested in a certain energy range around the
+    Fermi energy. By default, this is not specified and the energy range given in the `dos.x` and `projwfc.x`
+    inputs will be used.
 
 Storage memory management:
 
@@ -97,7 +97,6 @@ def validate_inputs(value, _):
 
     - Check that either the `scf` or `nscf.pw.parent_folder` inputs is provided.
     - Check that the `Emin`, `Emax` and `DeltaE` inputs are the same for the `dos` and `projwfc` namespaces.
-    - Check that `Emin` and `Emax` are provided in case `align_to_fermi` is set to `True`.
     """
     # Check that either the `scf` input or `nscf.pw.parent_folder` is provided.
     import warnings
@@ -113,10 +112,13 @@ def validate_inputs(value, _):
         if value['dos']['parameters']['DOS'].get(par, None) != value['projwfc']['parameters']['PROJWFC'].get(par, None):
             return f'The `{par}`` parameter has to be equal for the `dos` and `projwfc` inputs.'
 
-    if value.get('align_to_fermi', False):
+    if value.get('energy_range_vs_fermi', False):
         for par in ['Emin', 'Emax']:
-            if value['dos']['parameters']['DOS'].get(par, None) is None:
-                return f'The `{par}`` parameter must be set in case `align_to_fermi` is set to `True`.'
+            if value['dos']['parameters']['DOS'].get(par, None):
+                warnings.warn(
+                    f'The `{par}` parameter and `energy_range_vs_fermi` were specified.'
+                    'The value in `energy_range_vs_fermi` will be used.'
+                )
 
     if 'nbands_factor' in value and 'nbnd' in value['nscf']['pw']['parameters'].base.attributes.get('SYSTEM', {}):
         return PdosWorkChain.exit_codes.ERROR_INVALID_INPUT_NUMBER_OF_BANDS.message
@@ -158,6 +160,17 @@ def validate_projwfc(value, _):
 
     """
     jsonschema.validate(value['parameters'].get_dict()['PROJWFC'], get_parameter_schema())
+
+
+def validate_energy_range_vs_fermi(value, _):
+    """Validate specified energy_range_vs_fermi.
+
+    - List needs to consist of two float values.
+    """
+    if len(value) != 2:
+        return f'`energy_range_vs_fermi` should be a `List` of length two, but got: {value}'
+    if not all(isinstance(val, (float, int)) for val in value):
+        return f'`energy_range_vs_fermi` should be a `List` of floats, but got: {value}'
 
 
 def clean_calcjob_remote(node):
@@ -220,14 +233,15 @@ class PdosWorkChain(ProtocolMixin, WorkChain):
             help='Terminate workchain steps before submitting calculations (test purposes only).'
         )
         spec.input(
-            'align_to_fermi',
-            valid_type=orm.Bool,
+            'energy_range_vs_fermi',
+            valid_type=orm.List,
+            required=False,
             serializer=to_aiida_type,
-            default=lambda: orm.Bool(False),
+            validator=validate_energy_range_vs_fermi,
             help=(
-                'If true, Emin=>Emin-Efermi & Emax=>Emax-Efermi, where Efermi is taken from the `nscf` calculation. '
-                'Note that it only makes sense to align `Emax` and `Emin` to the fermi level in case they are actually '
-                'provided by in the `dos` and `projwfc` inputs, since otherwise the '
+                'Energy range with respect to the Fermi level that should be covered in DOS and PROJWFC calculation.'
+                'If not specified but Emin and Emax are specified in the input parameters, these values will be used.'
+                'Otherwise, the default values are extracted from the NSCF calculation.'
             )
         )
         spec.input('nbands_factor', valid_type=orm.Float, required=False,
@@ -490,10 +504,14 @@ class PdosWorkChain(ProtocolMixin, WorkChain):
         dos_inputs = AttributeDict(self.exposed_inputs(DosCalculation, 'dos'))
         dos_inputs.parent_folder = self.ctx.nscf_parent_folder
         dos_parameters = self.inputs.dos.parameters.get_dict()
+        energy_range_vs_fermi = self.inputs.get('energy_range_vs_fermi')
 
-        if dos_parameters.pop('align_to_fermi', False):
-            dos_parameters['DOS']['Emin'] = dos_parameters['Emin'] + self.ctx.nscf_fermi
-            dos_parameters['DOS']['Emax'] = dos_parameters['Emax'] + self.ctx.nscf_fermi
+        if energy_range_vs_fermi:
+            dos_parameters['DOS']['Emin'] = energy_range_vs_fermi[0] + self.ctx.nscf_fermi
+            dos_parameters['DOS']['Emax'] = energy_range_vs_fermi[1] + self.ctx.nscf_fermi
+        else:
+            dos_parameters['DOS'].setdefault('Emin', self.ctx.nscf_emin)
+            dos_parameters['DOS'].setdefault('Emax', self.ctx.nscf_emax)
 
         dos_inputs.parameters = orm.Dict(dos_parameters)
         dos_inputs['metadata']['call_link_label'] = 'dos'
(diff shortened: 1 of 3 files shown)


aiida_quantumespresso/workflows/pdos.py
```diff
@@ -219,14 +222,15 @@ def define(cls, spec):
...
         )
         spec.input(
-            'align_to_fermi',
-            valid_type=orm.Bool,
+            'fermi_energy_range',
+            valid_type=orm.List,
+            required=False,
             serializer=to_aiida_type,
-            default=lambda: orm.Bool(False),
+            default=lambda: orm.List(list=[]),
```
sphuber, line 229: Don't think this needs a default if it is going to be empty anyway.

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -219,14 +222,15 @@ def define(cls, spec):
             help='Terminate workchain steps before submitting calculations (test purposes only).'
         )
         spec.input(
-            'align_to_fermi',
-            valid_type=orm.Bool,
+            'fermi_energy_range',
+            valid_type=orm.List,
```
sphuber, line 226: This also needs a validator to check that when specified, it contains two float values

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -219,14 +222,15 @@ def define(cls, spec):
             help='Terminate workchain steps before submitting calculations (test purposes only).'
         )
         spec.input(
-            'align_to_fermi',
-            valid_type=orm.Bool,
+            'fermi_energy_range',
```
sphuber, line 225: I might be saying something stupid because I wasn't part of the discussions, but does the name `fermi_energy_range` really make sense? The values define the energy window in which the DOS is computed. Does something like just `energy_range` or `energy_window` not make more sense?

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -452,10 +456,16 @@ def _generate_dos_inputs(self):
...
-            dos_parameters['DOS']['Emin'] = dos_parameters['Emin'] + self.ctx.nscf_fermi
-            dos_parameters['DOS']['Emax'] = dos_parameters['Emax'] + self.ctx.nscf_fermi
+        if fermi_energy_range:
+            dos_parameters['DOS']['Emin'] = fermi_energy_range[0] + self.ctx.nscf_fermi
+            dos_parameters['DOS']['Emax'] = fermi_energy_range[1] + self.ctx.nscf_fermi
+        elif 'Emin' in dos_parameters['DOS'] and 'Emax' in dos_parameters['DOS']:
+            pass
+        else:
+            dos_parameters['DOS']['Emin'] = self.ctx.nscf_emin
+            dos_parameters['DOS']['Emax'] = self.ctx.nscf_emax
```
sphuber, line 468:
```suggestion
        else:
            dos_parameters['DOS'].setdefault('Emin', self.ctx.nscf_emin)
            dos_parameters['DOS'].setdefault('Emax', self.ctx.nscf_emax)
```
This solution might be a bit more elegant although it is not completely identical since now the keys are decoupled and won;t be updated only if neither are specified. But this may actually be a desirable side effect.

sphuber (commented):
Thanks @t-reents . The changes seem ok in principle, just have some suggestions.

aiida_quantumespresso/workflows/pdos.py
t-reents: I agree, I will remove this

aiida_quantumespresso/workflows/pdos.py
t-reents: Thanks @sphuber. I was not aware of this. I will add this solution.

aiida_quantumespresso/workflows/pdos.py
t-reents: I see your point. If a energy window is specified by this parameter, the DOS/PDOS is computed in this range with respect to the fermi energy. For example, the range 0 - 30 eV is specified and the fermi energy is 7 eV, then the DOS would be computed in the range 7 - 37 eV. Therefore, I think it would be useful to indicate this also by the name of the input parameter (but maybe the description in the help keyword would be already enough).

aiida_quantumespresso/workflows/pdos.py
t-reents: I added a validator. I split it into two parts. Maybe one would prefer to combine it by an `or` and return just one messsage.

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -115,10 +115,13 @@ def validate_inputs(value, _):
...
             return f'The `{par}`` parameter has to be equal for the `dos` and `projwfc` inputs.'
 
-    if value.get('align_to_fermi', False):
+    if value.get('fermi_energy_range', False):
         for par in ['Emin', 'Emax']:
-            if value['dos']['parameters']['DOS'].get(par, None) is None:
-                return f'The `{par}`` parameter must be set in case `align_to_fermi` is set to `True`.'
+            if value['dos']['parameters']['DOS'].get(par, None):
+                warnings.warn(
+                    f'The `{par}`` parameter and fermi_energy_range were specified.'
```
sphuber, line 122:
```suggestion
                    f'The `{par}` parameter and `fermi_energy_range` were specified.'
```

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -115,10 +115,13 @@ def validate_inputs(value, _):
...
+                    'The value in fermi_energy_range will be used.'
```
sphuber, line 123:
```suggestion
                    'The value in `fermi_energy_range` will be used.'
```

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -159,6 +162,17 @@ def validate_projwfc(value, _):
...
 
+def validate_fermi_energy_range(value):
+    """Validate specified fermi_energy_range.
+    
+    - List needs to consist of two float values.
+    """
+    if len(value) != 2:
+        return 'List needs to consist of two values.'
+    if not all(isinstance(val, float) for val in value):
+        return 'Fermi energy range needs to be specified as floats.'
```
sphuber, line 173:
```suggestion
    if len(value) != 2:
        return f'`fermi_energy_range` should be a `List` of length two, but got: {value}'
    if not all(isinstance(val, float) for val in value):
        return f'`fermi_energy_range` should be a `List` of floats, but got: {value}'
```

aiida_quantumespresso/workflows/pdos.py
sphuber: To me it reads a bit weird having it in the option, so I would vote to put it simply in the description that it is relative to the fermi level. I will let @mbercx have the final vote.

aiida_quantumespresso/workflows/pdos.py
sphuber: Validator looks good, just a suggestion on the error message to make it a bit clearer and more consistent

sphuber (changes requested):
Thanks @t-reents , just some final minor nitpicks that I noticed. Then I let @mbercx decide on the naming of the input. Once that is decided and applied we can merge this

t-reents:
Thanks again @sphuber. I committed your suggestions.

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -159,6 +162,17 @@ def validate_projwfc(value, _):
     jsonschema.validate(value['parameters'].get_dict()['PROJWFC'], get_parameter_schema())
 
 
+def validate_fermi_energy_range(value, _):
```
t-reents, line 165: @sphuber I added "_" in the inputs of the validator. Otherwise one would get a warning that a validator with just one input is deprecated. Please correct me if I am wrong.

sphuber: That is the correct solution indeed, we do this routinely.

aiida_quantumespresso/workflows/pdos.py
mbercx: Sorry for being later to the party! Although I agree the name isn't perfect, I still think it's better to highlight that it's an energy range around the Fermi level by adding it to the name. Maybe `energy_range_fermi` would give less of an impression that there is some range of Fermi levels considered?

t-reents: Personally, I like the last one but I leave this up to you

sphuber: I also vote for the latter :+1:  A few more characters is almost always preferable if it adds clarity. So @t-reents please feel free to go with that one, update the PR and we can merge. Thanks!

t-reents: The changes are now included

aiida_quantumespresso/workflows/pdos.py
```diff
@@ -452,10 +467,14 @@ def _generate_dos_inputs(self):
         dos_inputs = AttributeDict(self.exposed_inputs(DosCalculation, 'dos'))
         dos_inputs.parent_folder = self.ctx.nscf_parent_folder
         dos_parameters = self.inputs.dos.parameters.get_dict()
+        fermi_energy_range = self.inputs.fermi_energy_range
```
sphuber, line 470: I think you forgot to update it here

t-reents: Oh yes, I am sorry. Now there is no `fermi_energy_range` left

sphuber (approved):
Thanks a lot @t-reents Good to go.

t-reents:
Thanks a lot for your comments and instructions! @sphuber

sphuber:
The tests are failing. Apparently the validator is being called even if no value is specified. We could solve this by explicitly checking for this value in the validator and in that case just returning, but really this should not be happening. I am looking into `plumpy` which is the library responsible for this part of the code on why it is calling the validator regardless. It is really weird because the logic does seem to prevent this from happening and there are unit tests for it. But apparently, the test here in `aiida-quantumespresso` has some edge condition. Will report back once I know more.

mbercx:
Tests pass, but maybe we should add some more to properly test the functionality with and without the `energy_range_vs_fermi` input?

Also: maybe we should move the long module docstring into the documentation page of the work chain? Right now it's a bit empty:

https://aiida-quantumespresso--764.org.readthedocs.build/en/764/howto/workflows/pdos.html

t-reents:
> Tests pass, but maybe we should add some more to properly test the functionality with and without the `energy_range_vs_fermi` input?
> 
> Also: maybe we should move the long module docstring into the documentation page of the work chain? Right now it's a bit empty:
> 
> https://aiida-quantumespresso--764.org.readthedocs.build/en/764/howto/workflows/pdos.html

For sure, I will add the testst, soon.

mbercx:
@t-reents cheers! Could you also rebase the branch on `origin/main`?
=

t-reents:
Done!

src/aiida_quantumespresso/workflows/pdos.py
```diff
@@ -383,6 +397,9 @@ def setup(self):
         """Initialize context variables that are used during the logical flow of the workchain."""
         self.ctx.serial_clean = 'serial_clean' in self.inputs and self.inputs.serial_clean.value
         self.ctx.dry_run = 'dry_run' in self.inputs and self.inputs.dry_run.value
+        self.ctx.energy_range_vs_fermi = (
```
mbercx, line 400: Why do we have to add this to the context? Is it updated somehow?

mbercx (commented):
Thanks @t-reents! Just a few questions and comments, and then we can merge this after the final commit on `main` is fixed to have a better message.

tests/conftest.py
```diff
@@ -879,8 +879,8 @@ def _generate_workchain_ph(exit_code=None, inputs=None, return_inputs=False):
 def generate_workchain_pdos(generate_workchain, generate_inputs_pw, fixture_code):
     """Generate an instance of a `PdosWorkChain`."""
 
-    def _generate_workchain_pdos():
-        from aiida.orm import Bool, Dict
+    def _generate_workchain_pdos(emin=None, emax=None, energy_range_vs_fermi=None):
```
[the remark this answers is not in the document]
t-reents: Yeah, I agree. Indeed, I thought that this is the easiest solution for now. Let's open an issue and we can discuss afterwards

src/aiida_quantumespresso/workflows/pdos.py
t-reents: No, I think I just did at some point because it is used in the method to generate the inputs for the PDOS and in the one for the DOS inputs. But as it's not updated, I can also simply access the `self.inputs` directly. Will update this aspect and let you know so we can merge it

t-reents: Hi @mbercx, I just pushed the changes, should be fine now

mbercx (approved):
LGTM! Thanks a lot @t-reents

tests/conftest.py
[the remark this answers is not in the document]
mbercx: https://github.com/aiidateam/aiida-quantumespresso/issues/1068

PR #770, opened 2022-01-18
Fix XML parsing for structures with numbered kinds

### LINKED ISSUE #761: projwfc parser fails for atom labels with numbers
Hi! Everything works well with v3.4.2.

With v3.5.0, for spin polarized calculations, where I label atoms with numbers for setting initial magnetization (e.g., C1, C2, ...), I started to get the following error:

```
 |   File "/home/kristjan/opt/anaconda3/envs/aiida/lib/python3.9/site-packages/plumpy/process_states.py", line 231, in execute
 |   File "/home/kristjan/opt/anaconda3/envs/aiida/lib/python3.9/site-packages/aiida/engine/processes/calcjobs/calcjob.py", line 388, in parse
 |     exit_code_retrieved = self.parse_retrieved_output(retrieved_temporary_folder)
 |   File "/home/kristjan/opt/anaconda3/envs/aiida/lib/python3.9/site-packages/aiida/engine/processes/calcjobs/calcjob.py", line 468, in parse_retrieved_output
 |   File "/home/kristjan/opt/anaconda3/envs/aiida/lib/python3.9/site-packages/aiida_quantumespresso/parsers/projwfc.py", line 326, in parse
 |     out_info_dict['structure'] = convert_qe2aiida_structure(parsed_xml['structure'])
 |   File "/home/kristjan/opt/anaconda3/envs/aiida/lib/python3.9/site-packages/aiida_quantumespresso/parsers/parse_raw/base.py", line 149, in convert_qe2aiida_structure
 |     s.append_atom(position=tuple(atom[1]), symbols=[atom[0]])
 |   File "/home/kristjan/opt/anaconda3/envs/aiida/lib/python3.9/site-packages/aiida/orm/nodes/data/structure.py", line 1431, in append_atom
 |     kind = Kind(**kwargs)
 |   File "/home/kristjan/opt/anaconda3/envs/aiida/lib/python3.9/site-packages/aiida/orm/nodes/data/structure.py", line 2043, in __init__
 |     self.set_symbols_and_weights(kwargs.pop('symbols'), weights)
 |   File "/home/kristjan/opt/anaconda3/envs/aiida/lib/python3.9/site-packages/aiida/orm/nodes/data/structure.py", line 2287, in set_symbols_and_weights
 |     validate_symbols_tuple(symbols_tuple)
 |   File "/home/kristjan/opt/anaconda3/envs/aiida/lib/python3.9/site-packages/aiida/orm/nodes/data/structure.py", line 250, in validate_symbols_tuple
 |     raise UnsupportedSpeciesError(
 | aiida.common.exceptions.UnsupportedSpeciesError: At least one element of the symbol list ('C1',) has not been recognized.
```

The corresponding `xml` file in the output folder of the projwfc calculation is here: [aiida.xml](https://github.com/aiidateam/aiida-quantumespresso/files/7685875/aiida.xml.txt)

mbercx:
Fixes #761 

In #747 we adjusted the parser for `projwfc.x` to rely on the XML file
instead of the provenance for obtaining the structure. This worked fine
for structures that have chemical symbols as kind names, but once the
kind labels contain numbers the `convert_qe2aiida_structure` failed to
generate the `StructureData` based on the XML content since this case is
not considered.

Here we rename this method to `convert_qe_to_aiida_structure` and make
it more flexible by also adding the case of numbered kind names. This is
simply done by removing any digits from the kind names in the XML to
obtain the corresponding `symbol` of the kind name.


mbercx:
@sphuber this should fix the issue that @eimrek is experiencing in #761, but it is still not completely general, i.e. it still relies on the kind names consisting out of the element and some digits. I don't think the actual element symbols are stored in the XML though, so we'd either have to rely on the default masses defined in `aiida-core` or the pseudos.

aiida_quantumespresso/parsers/parse_raw/base.py
```diff
@@ -133,29 +140,33 @@ def convert_qe_time_to_sec(timestr):
...
+        structure = StructureData(cell=cell_dict['lattice_vectors'])
+
+        for mass, name in zip(output_dict['species']['mass'], output_dict['species']['type']):
+
+            # Remove digits from the kind name to obtain the symbol
+            symbol = re.sub('\d+', '', name)
+            structure.append_kind(Kind(symbols=symbol, mass=mass, name=name))
+
+        for kind_name, position in output_dict['atoms']:
+            structure.append_site(Site(kind_name=kind_name, position=position))
```
sphuber, line 160: I don't think you have to mess with the masses your self. You can simply define the kind name with the `name` argument in `append_atom`. It will figure out the correct `Kind`s itself
```suggestion
        for kind_name, position in output_dict['atoms']:
            symbol = re.sub('\d+', '', kind_name)
            structure.append_atom(position=position, symbols=symbol, name=kind_name))
```

sphuber (changes requested):
Thanks @mbercx . I think the implementation can be simplified. I added it as a suggestion but haven't tested this. The rest should be fine, except:



You are overencumbered!!!

Do you think it is possible to chop down those test files to the minimum required?

mbercx:
> You are overencumbered!!!

Haha, holy crap, I just now noticed the ridiculous line additions! 😨 I think the best way to do this is by running only a few k-points, let me fix that.

sphuber (approved):

PR #774, opened 2022-01-27
`PwParser`: Add support for QE v7.0

mbercx:
Add the XML schema for Quantum ESPRESSO v7.0, along with a corresponding
test for the parsing.

Because the `.xsd` filename in the XML output of QE v7.0 is incorrect,
a check is also added in the `get_schema_filename` function to see if
the user was running QE v7.0. In that case, the correct filename is
returned.

We also use `pytest.mark.parametrize` to condense all the XML schema
tests into one.


tests/parsers/test_pw.py
```diff
@@ -93,105 +93,22 @@ def test_pw_default_no_xml(
...
-
-
-def test_pw_default_xml_191206(
-    fixture_localhost, generate_calc_job_node, generate_parser, generate_inputs, data_regression
+@pytest.mark.parametrize('xml_format', [
+    '190304',
+    '191206',
+    '200420',
+    '210716',
+    '211102',
```
sphuber, line 101: Think you may have mistyped this. Shouldn't it be `211101` i.e. with a 1 at the end instead of a two. Also should probably rename the folder of the reference test files.

sphuber (changes requested):
Thanks @mbercx , all good. Thanks for the refactor of the tests, that's great. Just one minor thing to correct but then this can be merged

sphuber (approved):

PR #776, opened 2022-01-28
Devops: add GitHub Actions workflow for continuous deployment

### LINKED ISSUE #777: Add GHA workflow to automatically deploy new releases to PyPI

sphuber:
Fixes #777 

This workflow will be triggered when a new tag is pushed that matches a
semantic versioning pattern. If it is pushed to the main repository of
the `aiidateam` account, a Python script `validate_release_tag.py` is
run that will parse the `__version__` attribute of the package and
compare it to the version of the tag. If they don't match, the workflow
is aborted.

Otherwise, the pre-commit and the unit test suite is ran, and when
successfull, `flit` is used to build and publish the package to PyPI.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,0 +1,122 @@
+name: cd
+
+on:
+    push:
+        tags:
+        -   'v[0-9]+.[0-9]+.[0-9]+*'
+
+jobs:
+
+    validate-release-tag:
+
+        if: github.repository == 'aiidateam/aiida-quantumespresso'
+        runs-on: ubuntu-latest
+
+        steps:
+        -   name: Checkout source
+        -   uses: actions/checkout@v2
+
+        -   name: Set up Python 3.8
+            uses: actions/setup-python@v2
+            with:
+                python-version: '3.8'
+
+        -   name: Validate the tag version against the package version
+        -   run: python .github/workflows/validate_release_tag.py $GITHUB_REF
+
+    pre-commit:
+
+        needs: [validate-release-tag]
+        runs-on: ubuntu-latest
+
+        steps:
+        -   uses: actions/checkout@v2
+
+        -   name: Cache Python dependencies
+            uses: actions/cache@v1
+            with:
+                path: ~/.cache/pip
+                key: pip-pre-commit-${{ hashFiles('**/setup.json') }}
+                restore-keys:
+                    pip-pre-commit-
+
+        -   name: Set up Python
+            uses: actions/setup-python@v2
+            with:
+                python-version: '3.8'
+
+        -   name: Install Python dependencies
+            run: |
+                pip install --upgrade pip setuptools wheel
+                pip install -e .[pre-commit,tests]
+
+        -   name: Run pre-commit
+            run: pre-commit run --all-files || ( git status --short ; git diff ; exit 1 )
+
+    tests:
+
+        needs: [validate-release-tag]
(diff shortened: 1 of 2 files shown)


.github/workflows/cd.yml
```diff
@@ -0,0 +1,134 @@
...
+        -   name: Set up Python 3.8
+            uses: actions/setup-python@v2
+            with:
+                python-version: '3.8'
+
+        -   name: Install flit
+            run: pip install flit~=3.4
+
+        -   name: Build and publish
+            run: flit publish
```
mbercx, line 130: Don't we need to add some configuration to `pyproject.toml` for `flit`? E.g. (from the [`flit` documentation](https://flit.readthedocs.io/en/latest/#usage)):

```toml
[build-system]
requires = ["flit_core >=3.2,<4"]
build-backend = "flit_core.buildapi"

[project]
name = "foobar"
authors = [{name = "Sir Robin", email = "robin@camelot.uk"}]
dynamic = ["version", "description"]

[project.urls]
Home = "https://github.com/sirrobin/foobar"
```

mbercx (commented):
Thank @sphuber! Had to read up a bit on GitHub actions to do a proper review. Still left some small questions and comments.

.github/workflows/cd.yml
sphuber: Yep, absolutely right. I was just working on that. See PR #779 which should address it. Look at the root directory after that PR, so clean and purty :drooling_face:

sphuber: So we should merge that first, and then I will update this PR and address your comments

mbercx: beautiful

.github/workflows/cd.yml
```diff
@@ -0,0 +1,134 @@
...
+
+        strategy:
+            matrix:
+                python-version: ['3.6', '3.7', '3.8', '3.9']
+
+        services:
+            postgres:
+                image: postgres:10
+                env:
+                    POSTGRES_DB: test_${{ matrix.backend }}
```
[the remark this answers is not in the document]
sphuber: We don't set it and we are really just relying on defaults. When you don't explicitly specify `AIIDA_TEST_PROFILE`, calling `pytest` will automatically generate an entire dummy profile with full postgres database cluster on the fly. I would have to update the workflows a bit and test, because I think we can simplify them a lot. I don't think `aiida-quantumespresso` should run the tests against both backends. Anyway Django is getting removed soon.

.github/workflows/cd.yml
sphuber: Ok, I cleaned up the CI workflow in #780 Once that is merged, we should merge #779 and then we can merge this one

.github/workflows/cd.yml
```diff
@@ -0,0 +1,134 @@
...
+
+        needs: [validate-release-tag]
+        runs-on: ubuntu-latest
+
+        steps:
+        -   name: Checkout source
+        -   uses: actions/checkout@v2
+
+        -   name: Cache Python dependencies
+            id: cache-pip
```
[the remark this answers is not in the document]
sphuber: Indeed not necessary; removed.

.github/workflows/cd.yml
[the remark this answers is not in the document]
sphuber: I have updated this to be in line with the cleaned up `ci` workflow. No need to predefine the postgres service.

PR #779, opened 2022-01-28
Devops: adopt PEP 621 and move build spec to `pyproject.toml`

### LINKED ISSUE #778: Move package build procedure to `pyproject.toml` following PEP 621

sphuber:
Fixes #778 

Following PEP 621 it is now possible to fully define the build procedure
of your package in `pyproject.toml`. Since this PEP is now well
supported, and for example `pip` can use it, we migrate to it since it
provides a bunch of benefits:

 * No longer need the deprecated `setup.py` to install the package.
 * Version number is now dynamically fetched from the package. So the
   utility script to make sure the `setup.json` specifies the same
   version can be removed.
 * The `MANIFEST.in` is no longer necessary. The `flit` build tool will
   automatically include anything. The `pyproject.toml` has a tool
   section for `flit` that excludes the `tests` and `docs` folders.


mbercx (approved):
party

PR #780, opened 2022-01-28
CI: update the GHA `ci.yml` workflow

sphuber:
The workflow is cleaned with various small changes:

 * Use strings for Python version specifiers. They should be treated as
   strings, because floats will cause problems. For example the float
   `3.10` would be normalized to `3.1` which is not what is intended.
 * No need for pipe symbol if `run` is single command
 * Update `actions/checkout` action to `v2`
 * Remove unused `id` attribute of the caching step
 * Install and or update `pip`, `setuptools` and `wheel` before
   installing the package. Without `wheel` dependencies will have to be
   installed using the legacy method.
 * Remove explicit installation of postgres through system dependencies
   as this is already setup as a service.
 * Remove explicit creation of PostgreSQL database through the service
   step as this will be done on the fly when invoking `pytest`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Continuous Integration
+name: ci
 
 on: [push, pull_request]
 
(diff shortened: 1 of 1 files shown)


mbercx (approved):
Looks good, thanks @sphuber!

PR #782, opened 2022-01-31
Release `v3.5.2`

mbercx:


mbercx:
@sphuber  I wanted to make this a patch release since there are no real new features and I wanted to avoid too many rows in our compatibility matrix, but in the end I have to strictly define `v3.5.2 < v4.0.0` for the row in this case. In reality, `v3.5.0 < v3.5.2` supports QE `v6.0, <=v6.8`, and `v3.5.2 < v4.0.0` supports QE `v6.0, <= v7.0`. Maybe we can just detail only up to minor versions in the compatibility matrix, under the assumption that the user always installs the latest patch release?

sphuber (approved):
Thanks @mbercx  Think your rationale concerning patch versions in compatibility matrix makes perfect sense.

PR #783, opened 2022-01-31
Devops: fix the CD workflow

sphuber:
Certain steps accidentally had a `-` in front of the `name` as well as
the `uses` key. This made it such that they were interpreted as two
separate steps and so the first one would lack the required `uses`
attribute.

In addition, the workflow should only run on the `master` branch in case
the tagged commit also occurs on other branches, to prevent running the
workflow more than once.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -2,6 +2,8 @@ name: cd
 
 on:
     push:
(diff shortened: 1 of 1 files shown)


mbercx:
Strange... Seems this problem magically fixed itself.

PR #784, opened 2022-02-01
Devops: fix the nightly build

sphuber:
The nightly build against the `develop` branch of `aiida-core` was
failing at the installation of the Python dependencies when running
against Python 3.7. The reason is that the current `develop` branch has
dropped support for Python 3.7 and the `pip` is then failing to install
a version of `pymatgen` that matches the dependency requirements. The
solution is simply to not run against Python 3.7 anymore.

In addition, as of `aiida-core==2.0` the `develop` branch has been
renamed to `main` so the workflow is updated to install that branch. As
a result, the workflow file is also renamed from `develop` to `nightly`
as the term `develop` is no longer really applicable.

Furthermore, the workflow is updated to match the exact same setup
procedure as the `ci.yml` workflow which has also recently been updated.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,4 +1,4 @@
-name: Nightly Build
+name: nightly
 
 on:
     schedule:
@@ -9,22 +9,15 @@ jobs:
     tests:
 
         if: github.repository == 'aiidateam/aiida-quantumespresso'  # Prevent running the builds on forks as well
-
         runs-on: ubuntu-latest
 
         strategy:
             matrix:
-                python-version: [3.7, 3.8, 3.9]
+                python-version: ['3.8', '3.9', '3.10']
 
         services:
             postgres:
-                image: postgres:10
-                env:
-                    POSTGRES_DB: test_${{ matrix.backend }}
-                    POSTGRES_PASSWORD: ''
-                    POSTGRES_HOST_AUTH_METHOD: trust
-                ports:
-                -    5432:5432
+                image: postgres:12
             rabbitmq:
                 image: rabbitmq:latest
                 ports:
@@ -33,8 +26,7 @@ jobs:
         steps:
         -   uses: actions/checkout@v2
 
-        -   name: Cache python dependencies
-            id: cache-pip
+        -   name: Cache Python dependencies
             uses: actions/cache@v1
             with:
                 path: ~/.cache/pip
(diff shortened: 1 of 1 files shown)


.github/workflows/develop.yml
```diff
@@ -47,26 +39,13 @@ jobs:
...
-            continue-on-error: true
-            id: install
-            run: |
-                pip install --upgrade setuptools
+                pip install --upgrade pip setuptools wheel
                 pip install git+https://github.com/aiidateam/aiida-core@develop#egg=aiida-core[atomic_tools]
                 pip install -e .[tests]
                 reentry scan
 
-        -   name: Run pytest
```
mbercx, line 64: Aren't you removing the actual step that runs the tests here?

.github/workflows/develop.yml
```diff
@@ -47,26 +39,13 @@ jobs:
...
-            continue-on-error: true
-            id: tests
-            run:
-                pytest -sv tests
-
         -   name: Slack notification
```
mbercx, line 49: Seems the workflow had been failing for a while now, but I didn't see any notifications in the Slack channel?

.github/workflows/develop.yml
```diff
@@ -47,26 +39,13 @@ jobs:
...
-
-        -   name: Install python dependencies
-            continue-on-error: true
-            id: install
-            run: |
-                pip install --upgrade setuptools
+                pip install --upgrade pip setuptools wheel
                 pip install git+https://github.com/aiidateam/aiida-core@develop#egg=aiida-core[atomic_tools]
                 pip install -e .[tests]
                 reentry scan
```
mbercx, line 47: Do we still have to run `reentry scan` for the `develop` branch of `aiida-core`?

mbercx (commented):
Thanks @sphuber! Just have some small questions.

.github/workflows/develop.yml
sphuber: whoops, little too deletion-happy

.github/workflows/develop.yml
sphuber: Good point, not sure why because it used to work at some point, but looking at the syntax, it seems to rely on the `steps.install` and `steps.tests` identifiers. I had accidentally removed them so it would have failed, but they used to be there, so not sure why it wasn't working.

.github/workflows/develop.yml
sphuber: Absolutely, removed it. This reminds me though that merging this now might not be a good idea, as it won't work as anticipated. The install from source of `aiida-core` will work but `pip install -e .[tests]` will then install with requirements of our own develop branch and that still specified something like `aiida-core~=1.6` and so it will uninstall the `develop` branch of `aiida-core`. So we need a v2.0 release of `aiida-core` and then up our own requirements.

PR #786, opened 2022-02-07
'magnetization' not present in CP output in v7.0, CG case added stress parsing in cp 

rikigigi:
at least in cp.x, v7.0, conjugate gradient case, the magnetization key is not present in xml output.
So a default value of False is added in the parser that now does not raise an exception if the cp output is parsed.

### DIFF
--- a/src/aiida_quantumespresso/parsers/cp.py
+++ b/src/aiida_quantumespresso/parsers/cp.py
@@ -110,6 +110,7 @@ class CpParser(Parser):
                 ('cells', 'cel', CONSTANTS.bohr_to_ang, 3),
                 ('velocities', 'vel', CONSTANTS.bohr_to_ang / (CONSTANTS.timeau_to_sec * 10**12), number_of_atoms),
                 ('forces', 'for', CONSTANTS.hartree_to_ev / CONSTANTS.bohr_to_ang, number_of_atoms),
+                ('stresses', 'str', 1.0, 3) #stress in GPa
             ]
 
             for name, extension, scale, elements in trajectories:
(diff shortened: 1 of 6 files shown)


mbercx:
Thanks @rikigigi! Changes look good. Do you by any chance have a small calculation using cg and v7.0 whose output files I can use to add a test?

rikigigi:
anything new? @mbercx

aiida_quantumespresso/parsers/cp.py
```diff
@@ -121,15 +122,16 @@ def parse(self, **kwargs):
...
                         num_elements=elements, splitlines=data, prepend_name=f'{name}_traj', rescale=scale
                     )
-                    # here initialize the dictionary. If the parsing of positions fails, though, I don't have anything
-                    # out of the CP dynamics. Therefore, the calculation status is set to FAILED.
-                    if extension != 'cel':
+                    # here initialize the dictionary.
+                    if extension == 'cel':
+                        # NOTE: the trajectory output has the cell matrix transposed!!
+                        raw_trajectory['cells'] = numpy.array(traj_data['cells_traj_data']).transpose((0, 2, 1))
+                    elif extension != 'str':
```
sphuber, line 129: Think it would be more legible if you also make this a positive condition `elif extension == 'str':` and then have the generic case in the `else`.

aiida_quantumespresso/parsers/parse_xml/parse.py
```diff
@@ -138,8 +138,11 @@ def parse_xml_post_6_2(xml):
 
     lsda = inputs.get('spin', {}).get('lsda', False)
     spin_orbit_calculation = inputs.get('spin', {}).get('spinorbit', False)
-    non_colinear_calculation = outputs['magnetization']['noncolin']
-    do_magnetization = outputs['magnetization'].get('do_magnetization', False)
+    non_colinear_calculation = False
+    do_magnetization = False
+    if 'magnetization' in outputs:
+        non_colinear_calculation = outputs['magnetization']['noncolin']
+        do_magnetization = outputs['magnetization'].get('do_magnetization', False)
```
sphuber, line 145: Think the following is easier to read
```suggestion
    non_colinear_calculation = outputs.get('magnetization', {}).get('noncolin', False)
    do_magnetization = outputs.get('magnetization', {}).get('do_magnetization', False)
```

sphuber (approved):

PR #793, opened 2022-03-05
Parsers: add exception title to `ERROR_UNEXPECTED_PARSER_EXCEPTION`

### LINKED ISSUE #791: Parsers: improve the 350 exception message to include the title of the exception
When issues happened in parser, the `350` error will raise with only messaging `message='The parser raised an unexpected exception.'`. The detailed traceback can be checked by `verdi process report` however, this does not help when the issue happened in CI action. 
It would be good to include the title of the exception into the `350` exception message.

sphuber:
Fixes #791 

The exact exception and its stack trace was already added as a log to
the relevant process node, however, most often only the exit message is
shown. Currently this would only show the base message that an
unexpected exception occurred which is not very informative and one
would have to run other commands to find out more. By adding the
exception title in the exit message itself, one has more information
directly.

### DIFF
--- a/aiida_quantumespresso/calculations/neb.py
+++ b/aiida_quantumespresso/calculations/neb.py
@@ -91,7 +91,7 @@ class NebCalculation(CalcJob):
         spec.exit_code(322, 'ERROR_OUTPUT_XML_FORMAT',
             message='The XML output file has an unsupported format.')
         spec.exit_code(350, 'ERROR_UNEXPECTED_PARSER_EXCEPTION',
-            message='The parser raised an unexpected exception.')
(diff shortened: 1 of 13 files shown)


mbercx (approved):
Thanks @sphuber! I didn't field test this, but changes look pretty straightforward. 👍 

This is good to go for me, but perhaps @unkcpz would still like to take a look since he opened the issue?

sphuber:
> This is good to go for me, but perhaps @unkcpz would still like to take a look since he opened the issue?

I asked him to open it, so I think it is fine to merge.

PR #794, opened 2022-03-07
`PwParser`: Fix trajectory verification for `fixed_coords`

### LINKED ISSUE #792: `PwParser`: trajectory verification doesn't consider `fixed_coords`
The `PwParser` has an extra check to see if the calculation is properly converged, and will return exit code `500` (`ERROR_IONIC_CONVERGENCE_NOT_REACHED`) if this is not the case:

https://github.com/aiidateam/aiida-quantumespresso/blob/96356d3b767446dac3fef37b2d232b2ea4e50d89/aiida_quantumespresso/parsers/pw.py#L209-L218

This method in turn relies on the `verify_convergence_trajectory` function, which calls the `verify_convergence_forces` function:

https://github.com/aiidateam/aiida-quantumespresso/blob/96356d3b767446dac3fef37b2d232b2ea4e50d89/aiida_quantumespresso/utils/validation/trajectory.py#L40-L61

This all works fine for most calculations, however, it _does not_ consider any `fixed_coords` in the `settings` of the `PwCalculation`. This means that in case the forces on these atoms are above the threshold, the calculation will keep on failing with exit code `500`, and when run in a `PwBaseWorkChain` will restart until it reaches the max number of iterations.

Thanks to @Sokseiha  for reporting!

mbercx:
Fixes #792 

For `relax` and `vc-relax` calculations, the `PwParser` does an extra
check on the `TrajectoryData` output to see if the structure has been
properly converged. However, when comparing the final forces with the
threshold, it does not consider that some site coordinates may have been
fixed with the `fixed_coords` setting.

Here we extract the `fixed_coords` setting from the `PwCalculation`
inputs, and pass it to the `verify_convergence_trajectory` function.
This has been adapted to ignore forces for fixed coordinates by setting
the corresponding forces to zero before comparing them with the force
threshold.


mbercx:
@sphuber will add a test once we agree on the implementation!

@Sokseiha I already manually tried to parse the output files you sent me, but perhaps you can try to run another (cheap) example with the PR branch to see if it all works as expected?

aiida_quantumespresso/utils/validation/trajectory.py
```diff
@@ -54,14 +74,24 @@ def verify_convergence_forces(trajectory, index=-1, threshold=None):
...
-        forces = trajectory.get_array('forces')[index]
+        abs_forces = numpy.abs(trajectory.get_array('forces')[index])
     except (KeyError, IndexError) as exception:
         raise ValueError('the `forces` array does not exist or the given index exceeds the length.') from exception
 
-    return numpy.max(abs(forces)) < threshold
+    if fixed_coords is not None:
+        fixed_coords = numpy.array(fixed_coords)
+        # Set the forces corresponding to fixed coordinates to zero, as they should not be checked versus threshold
+        abs_forces *= numpy.invert(fixed_coords)
```
sphuber, line 84: Deducing from what you are doing here, I take it the `fixed_coords` is a list of lists where each internal member is a list of three elements with each element either a 0 or a 1? If that is the case, I think it would be good to have a short explanation of this and we should add validation of `fixed_coords`. I am not sure how liberal the format is in the actual `settings` of the `PwCalculation`. For example, maybe QE accepts any positive non-zero integer to fix a coordinate, and so if a user uses `2` instead of `1`, would `invert` still work as intended?

sphuber (commented):
I think the implementation is ok, it is quite elegant, but it needs validation and some comments explaining it.

aiida_quantumespresso/utils/validation/trajectory.py
mbercx: Actually the `fixed_coords` setting is a list of lists of booleans, e.g.:

```python
settings_dict = {
    'fixed_coords': [
        [True, False, False],
        [True, True, True],
        [False, False, False],
        [False, False, False],
        [False, False, False]
    ],
}
```

See also [this section](https://aiida-quantumespresso.readthedocs.io/en/latest/user_guide/calculation_plugins/pw.html?highlight=fixed_coords#fixing-some-atom-coordinates) in the documentation. The fact that these must be booleans is enforced during the `BasePwCpInputGenerator._generate_PWCPinputdata` step:

https://github.com/aiidateam/aiida-quantumespresso/blob/0c8a6114dfb740262fd408c7b0fc73bd0fde33ee/aiida_quantumespresso/calculations/__init__.py#L492-L497

However, this would of course result in an `Excepted` work chain, so I agree adding a validator of the `settings` input is an improvement. Will adapt (and also clean up this code a bit). 👍 

Will also extend the explanation and improve the type hinting for clarity.

sphuber (dismissed):
Pre-approving but if possible would still be good to verify this works in real example with @Sokseiha

mbercx:
> Pre-approving but if possible would still be good to verify this works in real example with @Sokseiha

I'm still testing + adding tests! And actually found some bugs, so good thing too ^^

mbercx:
There, now she's ready for another review. I've added tests for both the input file generation and the parsing (`relax` and vc-relax`).

mbercx:
@sphuber still found a flaw in the current implementation (i.e. the parsing expected `fixed_coords` lowercase) and added a test. @AriannaCantarella maybe you can give this one a test for your hydrogen restorer work chain?

mbercx:
@sphuber: As a side note: what should the "best practises" for validator methods be? i.e. do you agree with the following:

1. We add them as a `classmethod` or `staticmethod` on the process class they are validating the input ports for. Although those that do not use the `cls` could also be added as separate functions defined before the process class, I think the validators are typically directly tied to the class anyways and it is tidier to have them all in one place.
2. We do not add them as hidden methods, i.e. they are not prefaced with an underscore `_`. Although it is rare to call the validators outside of the class, one could potentially think of a use case for this, and most of the methods defined on e.g. a work chain class are rarely called outside of the class, save for e.g. testing.

sphuber:
> @sphuber: As a side note: what should the "best practises" for validator methods be? i.e. do you agree with the following:
> 
> 1. We add them as a `classmethod` or `staticmethod` on the process class they are validating the input ports for. Although those that do not use the `cls` could also be added as separate functions defined before the process class, I think the validators are typically directly tied to the class anyways and it is tidier to have them all in one place.
> 2. We do not add them as hidden methods, i.e. they are not prefaced with an underscore `_`. Although it is rare to call the validators outside of the class, one could potentially think of a use case for this, and most of the methods defined on e.g. a work chain class are rarely called outside of the class, save for e.g. testing.

I agree on both points. I also add validators as public class methods or static methods

mbercx:
@sphuber this has been tested in the field by @AriannaCantarella. If you could give one final sign off on this one I'll then rebase https://github.com/aiidateam/aiida-quantumespresso/pull/818 and we can merge that as well.

sphuber (approved):
Thanks @mbercx good to go!

PR #795, opened 2022-03-17
Implemented support for elasticity calculion in matdyn

lin-cp:


mbercx:
Thanks @lin-cp! As discussed, I also have some changes for the `matdyn.x` code in a branch, so I'll have a look at how to integrate these.

Just to be clear: these changes are for a new feature that is still in the development branch of Quantum ESPRESSO, right?

lin-cp:
@mbercx Yes, it's a new feature which is even not merged into the develop branch of QE yet. I think they will be integrated into QE in the second half of the year.

PR #801, opened 2022-04-25
New WorkChain: phonon dispersion

unkcpz:
I start working on the phonon dispersion workchain for qe-app which is better to have a corresponding workchain pre hand here that runs `scf->ph->q2r->matdyn`. Please let me know, if there any ready-to-use phonon dispersion workchain for this purpose in case I reinvent the wheel? @mbercx @bastonero

### DIFF
--- a/aiida_quantumespresso/workflows/pd.py
+++ b/aiida_quantumespresso/workflows/pd.py
@@ -1,0 +1,226 @@
+# -*- coding: utf-8 -*-
+"""Workchain to run Quantum ESPRESSO calculations that generate phonon dispersion for a structure.
+
+This requires four computations:
+
+- SCF (pw.x), to generate the initial wavefunction.
+- PH (pw.x), to generate dynamical matrices for a uniform mesh of q vectors.
+- Q2R (q2r.x), to calculate of interatomic force constants.
+- MAYDYN (matdyn.x), to generate Fourier interpolation for various q points.
+"""
+from aiida import orm, plugins
+from aiida.common import AttributeDict
+from aiida.engine import WorkChain, ToContext
+from aiida.orm.nodes.data.base import to_aiida_type
+
+from aiida_quantumespresso.utils.mapping import prepare_process_inputs
+
+from .protocols.utils import ProtocolMixin
+
+def validate_scf(value, _):
+    """Validate the scf parameters."""
+    parameters = value['pw']['parameters'].get_dict()
+    if parameters.get('CONTROL', {}).get('calculation', 'scf') != 'scf':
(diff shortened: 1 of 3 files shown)


sphuber:
Actually, it does already exist :sweat_smile: but it is in the private `aiida-quantumespresso-epfl` repo: https://github.com/aiidateam/aiida-quantumespresso-epfl/blob/master/aiida_quantumespresso_epfl/workflows/phonon_dispersion.py It also contains some nice other workchains for `ph.x` that I wrote to parallelize the work over `q-points` for example. Might want to just take that code, clean/update and test it, and then make it public here.

unkcpz:
@sphuber thanks very much! It's a good thing I just started. I'll close this one, test/update the code here. Would you mind if I copy the code and create a PR?

sphuber:
If you don't mind I will create a PR myself to copy the code and add them. Then we can go from there to improve the code and add tests in separate PRs

unkcpz:
@sphuber that will be great! thanks!

mbercx:
Hey guys, sorry for the late feedback. I've also been working with @mkotiuga on a `aiida-quantumespresso-ph` package (similar to the `aiida-quantumespresso-hp` package @sphuber was working on). Maybe we can move all `ph.x`-related work chains in there?

sphuber:
Is `aiida-quantumespresso-ph` available publicly? If so, where can I access it. What is the scope?

mbercx:
Not yet, I'm still working on it. We might also want to first make it a private repo, since it contains work chains for results that haven't been published yet. Something to discuss with @mkotiuga and Nicola.

As for the scope, it's a good question. I'd say all QE-phonon related work chains and data types. But this needs to be defined more clearly for sure.

sphuber:
I think the whole point of moving the these workchains out of `aiida-quantumespresso-epfl` is to make it public. Wouldn't make sense to move it from one private to another, would it?

mbercx:
I'd say there could be organisational benefits and it might avoid some code duplication. But I'll have a look at the contents and we can discuss next week.

unkcpz:
I guess `aiida-quantumespresso-ph` of Marnik also includes lots of useful Workchains related to PH but not well tested and may not be required for qe-app phonon-dispersion at the moment. From my point of view, what I need here is not a very fancy workchai, if @sphuber can quickly copy the code here and I can start to run a bunch of tests and start to integrate it to qe-app and it then allowed more tests easily through aiidalab. 

> But I'll have a look at the contents and we can discuss next week.

Cheers, let's meet and discuss next week.

PR #805, opened 2022-04-29
`CalcJob`: simplify `parent_folder` handling

sphuber:
As of `aiida-core==2.0` the `CalcInfo.local_copy_list` and
`CalcInfo.remote_copy_list` attributes support defining directories as
the source argument, i.e., the second element in each copy directive.
The engine will check whether the source is a file or a directory and
automatically copy all contents in the latter case.

This allows us to simplify the implementation in the `CalcJob` plugins
as now we no longer have to do the recursive discovery of directories
ourselves.

This is possible thanks to https://github.com/aiidateam/aiida-core/commit/a295616cbdc6a5aa8f0286d3c85d2408a872fa84


mbercx (approved):
Muy limpio! 🧹

PR #808, opened 2022-05-05
Remove the use of deprecated `distutils` package

### LINKED ISSUE #807: Remove use of deprecated `distutils` package
This standard library module was deprecated in Python 3.10 and will be removed in Python 3.12

sphuber:
Fixes #807 

With the acceptance of PEP 632, this module was deprecated in Python 3.10
and it will be removed in Python 3.12. The `LooseVersion` and
`StrictVersion` classes are replaced by `packaging.version.Version` as
recommended by the PEP itself, see:

    https://peps.python.org/pep-0632/#migration-advice

### DIFF
--- a/aiida_quantumespresso/parsers/cp.py
+++ b/aiida_quantumespresso/parsers/cp.py
@@ -1,8 +1,7 @@
 # -*- coding: utf-8 -*-
-from distutils.version import LooseVersion
-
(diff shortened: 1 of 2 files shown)


mbercx (approved):
I swear I'm actually looking at the code and not just approving these PRs 😝

PR #810, opened 2022-05-05
`PhBaseWorkChain`: Add a first draft for the protocols

mbercx:
Add a first draft for the `PhBaseWorkChain` protocols. These have not
been tested rigorously, but can serve as an initial setup that can be
easily adapted after using the `get_builder_from_protocol()` method.

This is also necessary to allow higher-level work chains to develop
protocols.

### DIFF
--- a/src/aiida_quantumespresso/workflows/ph/base.py
+++ b/src/aiida_quantumespresso/workflows/ph/base.py
@@ -2,14 +2,17 @@
 """Workchain to run a Quantum ESPRESSO ph.x calculation with automated error handling and restarts."""
 from aiida import orm
 from aiida.common import AttributeDict
+from aiida.common.lang import type_check
 from aiida.engine import BaseRestartWorkChain, ProcessHandlerReport, process_handler, while_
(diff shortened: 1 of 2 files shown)


aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -45,6 +48,59 @@ def define(cls, spec):
...
+
+        type_check(code, orm.Code)
+
+        inputs = cls.get_protocol_inputs(protocol, overrides)
+
+        qpoints_mesh = inputs['ph'].pop('qpoints')
+        qpoints = orm.KpointsData()
+        qpoints.set_kpoints_mesh(qpoints_mesh)
+
+        # pylint: disable=no-member
```
sphuber, line 89: Where are these warnings coming from?

sphuber (approved):
Approving this. Just a question about the pylint disable. Wonder if it is necessary

sphuber (approved):
Gracias señor Bercx!

PR #812, opened 2022-05-05
DevOps: Move the source directory into `src/`

sphuber:
This package directory structure is fast becoming the standard in the
Python packaging community. Even though the PyPA is not yet officially
recommending it as the standard, there are various articles out there
outlining the advantages of this structure over having the source folder
at the root of the repository.


mbercx (approved):
So clean, so purty. 🤩

PR #813, opened 2022-05-05
`PhCalculation`: Fix bug when `only_initialization=True`

### LINKED ISSUE #811: `PhCalculatio` is bugged when starting from `parent_folder` of `PwCalculation` and `only_initialization = True`
In this case the plugin calls:
```python
remote_copy_list.append((
    parent_folder.computer.uuid,
    os.path.join(parent_folder.get_remote_path(), self._FOLDER_DYNAMICAL_MATRIX), '.'
))
```
But this file won't exist in the folder of a `PwCalculation` and so will fail.

sphuber:
Fixes #811 

When the `only_initialization` setting is set to `True` the plugin would
add an instruction to the `remote_copy_list` to copy `DYN_MAT` folder
from the `parent_folder`. However, this would fail if the `parent_folder`
was that of a `PwCalculation`, which is one of the main use cases of the
`PhCalculation`.

It is not entirely clear why this was added in the commit that added it
https://github.com/aiidateam/aiida-quantumespresso/commit/dfdca7e1317484afedcf522b5adea17dc3f8533a which was just adding a plugin
for EPW. We assume it was added by mistake and so remove it.

### DIFF
--- a/src/aiida_quantumespresso/calculations/ph.py
+++ b/src/aiida_quantumespresso/calculations/ph.py
@@ -298,11 +298,6 @@ class PhCalculation(CalcJob):
             with folder.open(f'{self._PREFIX}.EXIT', 'w') as handle:
                 handle.write('\n')
 
-                remote_copy_list.append((
-                    parent_folder.computer.uuid,
-                    os.path.join(parent_folder.get_remote_path(), self._FOLDER_DYNAMICAL_MATRIX), '.'
-                ))
-
         codeinfo = datastructures.CodeInfo()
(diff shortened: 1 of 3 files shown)


mbercx (dismissed):
Cheers @sphuber! I have do double check, but I think this may be also required for my `ElectronPhononWorkChain`. However, ok to remove for now, I'll re-add it in a correct manner in case I need it.

sphuber:
> However, ok to remove for now, I'll re-add it in a correct manner in case I need it.

In that case I maybe better add a test to ensure the current functionality is preserved.

mbercx (approved):
Let's get this baby merged then 🚀

PR #818, opened 2022-06-02
`PhParser`: parse symmetry labels from the `stdout` of `ph.x`

mkotiuga:
Add parsing of the symmetry labels to the `PhParser`, which are _only_ printed in the `stdout` after the mode frequencies for each computed q-point.

For each q-point, there is already a `dynamical_matrix_<#qpoint>` key in the `output_parameters` output of the `PhCalculation`. Here we add the `mode_symmetry` and `point_group` outputs, parsed from the `stdout`.

As the symmetry labels are _only_ printed for each computed q-point of a `ph.x` run, this adds an extra challenge when not all q-points are calculated in a single run, for example in case the calculation ran into the walltime and hence is restarted by the `PhBaseWorkChain`. This is dealt with by adding a calculation function called `merge_ph_outputs`, which:

* Merges the `dynamic_matrix_X` keys.
* Combines all the `number_of_irr_representations_for_each_q` lists into one.
* Adds all the `number_of_qpoints`.
* Adds all the `wall_time_seconds`.

This calculation function is called in an additional step in the outline of the `PhBaseWorkChain` called `create_merged_output`. This step also:

* Is skipped in case the `ph.x` is only used for the initialization (`only_initialization` == `True`).
* Is skipped in case the `PhBaseWorkChain` only ran one `PhCalculation`.
* Checks that the number of parsed q-points matches the total number, and returns the `ERROR_MERGING_QPOINTS` exit code if this is not the case.

Some additional logic is added for properly parsing `ph.x` calculations that only run a single q-point, for which no list of q-points is provided at the start of the `stdout`. The parsing and output merging also works as expected in case such a run is restarted in a `PhBaseWorkChain`.

The `wall_time` string output is also removed from the `output_parameters`. This serves little purpose and is convoluted to merge.


mbercx:
@sphuber I'm first going to clean this up and make sure it's properly rebased on `develop`. :)

mbercx:
Changes to regression test files look good in my opinion, i.e. the expected parsed symmetry labels are added. We should still add a test to see if multiple separate `PhCalculation`s in one `PhBaseWorkChain` are added properly.

mbercx:
The code is already in good shape I think (tests still need to be fixed though), but it seems I can't just override the `output_parameters` output with the `self.out()` method. @sphuber is there a clean way of doing this?

sphuber:
> but it seems I can't just override the `output_parameters` output with the `self.out()` method. @sphuber is there a clean way of doing this?

Don't understand what you mean here. What line are you referring to in the changes and what is the problem?

mbercx:
> Ideally, the BaseRestartWorkChain provides a hook, like get_outputs, that is called in results and returns the final output mapping. You could then just override this and the rest of results would remain the same. As for now, I think we do it here by overriding results entirely and add a feature request in aiida-core to add this and then in the next release we can improve the implementation here. Thoughts?

Sure, the hook approach would work. 👍 For this PR, maybe we can just call the `create_merged_output` step first, so the `output_parameters` are already set before the `results`?

sphuber:
> Sure, the hook approach would work. 👍 For this PR, maybe we can just call the `create_merged_output` step first, so the `output_parameters` are already set before the `results`?

Does that work? As in the second time it is being added it simply is ignored without error (would surprise me tbh)? If so, that works as well and is less messy, but it would require a big fat comment that it is relying on this.

mbercx:
>  As in the second time it is being added it simply is ignored without error (would surprise me tbh)?

I think I tried overriding it in the merge step, after the results one, and I don't think it raised an error. But I'll give it a go. ^^

mbercx:
Seems first setting the `output_parameters` in the `create_merged_output` step and then adding the `results` step to the outline actually works. The following work chain:

```console
$ verdi process status 9696
PhBaseWorkChain<9696> Finished [0] [4:results]
    ├── PhCalculation<9699> Finished [400]
    ├── PhCalculation<9705> Finished [0]
    └── merge_ph_outputs<9709> Finished [0]
```

Will nicely have the merged output parameters:

```python
In [1]: ph_base = load_node(9696)

In [2]: ph_base.outputs.output_parameters.get_dict()
Out[2]: 
{'wall_time': '     55.47s ',
 'num_q_found': 2,
 'number_of_atoms': 1,
 'number_of_qpoints': 4,
 'wall_time_seconds': 113.61,
 'dynamical_matrix_1': {'q_point': [0.0, 0.0, 0.0],
  'frequencies': [-3.205876, -3.194079, -3.194079],
  'point_group': 'D_3d (-3m)',
  'mode_symmetry': ["L_2'", "L_3'", "L_3'"],
  'q_point_units': '2pi/lattice_parameter',
  'frequencies_units': 'cm-1'},
 'dynamical_matrix_2': {'q_point': [0.0, 0.0, -0.612372436],
  'frequencies': [65.862765, 65.862765, 112.146012],
  'point_group': 'D_3d (-3m)',
  'mode_symmetry': ["L_3'", "L_3'", "L_2'"],
  'q_point_units': '2pi/lattice_parameter',
  'frequencies_units': 'cm-1'},
 'dynamical_matrix_3': {'q_point': [0.0, -0.577350269, 0.204124145],
  'frequencies': [65.862769, 65.863417, 112.148301],
  'point_group': 'C_2h (2/m)',
  'mode_symmetry': ['A_u', 'B_u', 'B_u'],
  'q_point_units': '2pi/lattice_parameter',
  'frequencies_units': 'cm-1'},
 'dynamical_matrix_4': {'q_point': [0.0, -0.577350269, -0.40824829],
  'frequencies': [83.202774, 83.20789, 105.820001],
  'point_group': 'C_2h (2/m)',
  'mode_symmetry': ['A_u', 'B_u', 'B_u'],
  'q_point_units': '2pi/lattice_parameter',
  'frequencies_units': 'cm-1'},
 'number_of_irr_representations_for_each_q': [2, 2, 3, 3]}
```

@sphuber so I can use this workaround for now, and open a new issue/PR that would rely on https://github.com/aiidateam/aiida-core/pull/5618 once this feature has been released.

@mkotiuga I've given the code a try, and it seems to work fine. Could you also give it a go, and closely check how the outputs are formatted to see if you still have any comments? E.g. should `mode_symmetry` perhaps be called `mode_symmetries`? ^^

mbercx:
@sphuber quick question: what's the use of the `generate_parser` fixture?

```python
@pytest.fixture(scope='session')
def generate_parser():
    """Fixture to load a parser class for testing parsers."""

    def _generate_parser(entry_point_name):
        """Fixture to load a parser class for testing parsers.

        :param entry_point_name: entry point name of the parser class
        :return: the `Parser` sub class
        """
        return ParserFactory(entry_point_name)

    return _generate_parser
```

This seems to just import the parser class, no? Can't we just remove this fixture and directly load the parser class?

sphuber:
> This seems to just import the parser class, no? Can't we just remove this fixture and directly load the parser class?

One would think yeah :sweat_smile: Saves an explicit import in most test files, but that is not really good enough, fine to remove it.

sphuber:
> @sphuber any chance we can push for a v2.1 soonish?

I am trying my hardest. Almost everything is there. Just waiting for a final release of `plumpy` and `kiwipy`, but for the latter we need an important fix, which I am not sure how fast we can deliver.

sphuber:
@mbercx `aiida-core==2.1` is out

mbercx:
@mkotiuga could you double check that the point group and symmetry labels for the non-symmetry case are to your liking, and maybe how we could expand on the comment here a bit?

https://github.com/mkotiuga/aiida-quantumespresso/blob/94d6839756ae1810fd581672cbed38cf44093977/src/aiida_quantumespresso/parsers/parse_raw/ph.py#L209-L212

@sphuber still had to make some fixes to cover all the different variants of stdout for `ph.x`. With `aiida-core` v2.1 released, I think this one is almost good to go.

mbercx:
@sphuber any idea why the integration CI test is now failing to install the SSSP? 🤔

sphuber:
> @sphuber any idea why the integration CI test is now failing to install the SSSP? thinking

It's because the CLI is broken with `aiida-core==2.1`, see this issue: https://github.com/aiidateam/aiida-pseudo/issues/130
I will fix it now and make a release that is compatible with v2.1. The same problem applies to `aiida-quantumespresso`. I already made a branch with the fix long ago (https://github.com/aiidateam/aiida-quantumespresso/tree/fix/834/cli-verdi-context) was just waiting for the release to happen.

sphuber:
Ok, `aiida-pseudo==0.8.0` is released and I updated `main` branch to add the dependency requirement. You can rebase now and tests should run without problems.

pyproject.toml
```diff
@@ -22,8 +22,8 @@ classifiers = [
 keywords = ['aiida', 'workflows']
 requires-python = '>=3.8'
 dependencies = [
-    'aiida_core[atomic_tools]~=2.1',
-    'aiida-pseudo~=0.8.0',
+    'aiida_core[atomic_tools]~=2.0,>=2.1.0',
+    'aiida-pseudo~=0.7.0',
```
sphuber, line 26:
```suggestion
    'aiida_core[atomic_tools]~=2.1',
    'aiida-pseudo~=0.7.0',
```
Should `aiida-pseudo` be downgraded?

src/aiida_quantumespresso/parsers/parse_raw/ph.py
```diff
@@ -172,6 +177,42 @@ def detect_important_message(logs, line):
...
+                    symm_label = line.split()[-2]
+                else:
+                    symm_label = line.split()[-1]
+
+                q_data['mode_symmetry'].extend([symm_label] * (freq_end - freq_start + 1))
+
+        # If the mode symmetries are not printed, set those for the non-symmetry case
+        if 'point_group' not in q_data:
+            q_data['point_group'] = 'C_1'
+            q_data['mode_symmetry'] = ['A_1'] * (3 * num_atoms)
```
mkotiuga, line 212:
```suggestion
            q_data['mode_symmetry'] = ['A'] * (3 * num_atoms)
```

mkotiuga: to be consistent with point group tables

src/aiida_quantumespresso/calculations/functions/merge_ph_outputs.py
```diff
@@ -0,0 +1,41 @@
...
+        number_of_qpoints += output.pop('num_q_found', 0)
+
+        for key, value in output.items():
+
+            if key.startswith('dynamical_matrix_') and 'mode_symmetry' in value.keys():
+                if '0' in key:
+                    merged.pop('dynamical_matrix_0', None)
+                    merged[f'dynamical_matrix_{index + 1}'] = value
+                else:
+                    merged[key] = value
```
mkotiuga, line 34: so keys that are not dynamical_matrix are merged
```suggestion
            if key.startswith('dynamical_matrix_') and 'mode_symmetry' in value.keys():
                if '0' in key:
                    merged.pop('dynamical_matrix_0', None)
                    merged[f'dynamical_matrix_{index + 1}'] = value
            else:
                merged[key] = value
```

mkotiuga:
When a PhBaseCalculation for 1 q-point has a PhCalculation that hits the wall time and then properly restarts, the final mode symmetries are not parsed. This is probably due to some logic in parses/parse_raw/ph.py - will try look into it shortly

mkotiuga:
When a PhBaseCalculation has multiple PhCalcuations and the wall-time is hit mid q-point calculation there is an extra entry in `number_of_irr_representations_for_each_q` as there are two 'headers' for the same q-point in two different `PhCalculations`   - This is probably also due to an error in the parser

src/aiida_quantumespresso/parsers/parse_raw/ph.py
```diff
@@ -236,6 +276,32 @@ def detect_important_message(logs, line):
...
+            parsed_data['num_q_found'] += 1
+
+    # Remove the q-points from the parsed data; these are only used to assign the symmetry labels to the right index
+    parsed_data.pop('q_points', None)
+
+    # Trim the number of irreps to the number of q-points finished in this run
+    if parsed_data['num_q_found'] > 0:
+        parsed_data['number_of_irr_representations_for_each_q'] = \
+            parsed_data['number_of_irr_representations_for_each_q'][:parsed_data['num_q_found']]
+
```
mkotiuga, line 304: In case no q-points are found to parse (should fix the duplication in 'number_or_irr_representations_for_each_q' in the merged output mentioned in previous comment)

```suggestion
    if parsed_data['num_q_found'] >= 0:
        parsed_data['number_of_irr_representations_for_each_q'] = \
            parsed_data['number_of_irr_representations_for_each_q'][:parsed_data['num_q_found']]
 
```

mkotiuga: should fix:
#
> When a PhBaseCalculation has multiple PhCalcuations and the wall-time is hit mid q-point calculation there is an extra entry in `number_of_irr_representations_for_each_q` as there are two 'headers' for the same q-point in two different `PhCalculations` - This is probably also due to an error in the parser

src/aiida_quantumespresso/parsers/parse_raw/ph.py
```diff
@@ -190,35 +231,34 @@ def detect_important_message(logs, line):
...
+                parsed_data['q_points'] = parse_qpoints(lines[count + 2:count + parsed_data['number_of_qpoints'] + 2])
+
+            except Exception as exc:
+                logs.warning.append(f'Error while parsing number of q points: {exc}')
+
+        # In case a single q-point is provided at the end of the input file, there is no summary of the q-points at the
+        # start of the stdout. Then we have to parse the q-point further down the output file.
+        elif 'number_of_qpoints' not in parsed_data and 'Calculation of q =' in line:
+            parsed_data['number_of_qpoints'] = 1
+            parsed_data['q_points'] = {0: [float(coord) for coord in line.split('=')[-1].split()]}
```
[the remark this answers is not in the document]
mkotiuga: Need to test with an actual run - but I think the above should fix the following comment

> When a PhBaseCalculation for 1 q-point has a PhCalculation that hits the wall time and then properly restarts, the final mode symmetries are not parsed. This is probably due to some logic in parses/parse_raw/ph.py - will try look into it shortly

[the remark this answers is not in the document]
mkotiuga: seems good to go

src/aiida_quantumespresso/calculations/functions/merge_ph_outputs.py
```diff
@@ -0,0 +1,41 @@
...
+                if '0' in key:
+                    merged.pop('dynamical_matrix_0', None)
+                    merged[f'dynamical_matrix_{index + 1}'] = value
+            else:
+                merged[key] = value
+
+    merged['wall_time_seconds'] = total_walltime
+    merged['number_of_irr_representations_for_each_q'] = number_irreps
+    merged['number_of_qpoints'] = number_of_qpoints
+
```
mkotiuga, line 39: remove wall_time if we are not combining it
```suggestion
    merged.pop('wall_time',0)
```

src/aiida_quantumespresso/calculations/functions/merge_ph_outputs.py
```diff
@@ -0,0 +1,41 @@
...
+        for key, value in output.items():
+
+            if key.startswith('dynamical_matrix_') and 'mode_symmetry' in value.keys():
+                if '0' in key:
+                    merged.pop('dynamical_matrix_0', None)
+                    merged[f'dynamical_matrix_{index + 1}'] = value
+            else:
+                merged[key] = value
+
+    merged['wall_time_seconds'] = total_walltime
```
mkotiuga, line 36: as written, this will miss the wall_time of the initialization PhCalc and any PhCalculations that hit the wall time

src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -146,6 +153,40 @@ def prepare_process(self):
...
+        output_dict = {
+            'output_' + str(index + 1): child.outputs.output_parameters
+            for index, child in enumerate(self.ctx.children)
+        }
+
+        num_qpoints = self.ctx.children[0].outputs.output_parameters['number_of_qpoints']
+        num_qpoints_final = self.ctx.children[-1].outputs.output_parameters['num_q_found']
+
+        if num_qpoints == num_qpoints_final:
+            return
```
mkotiuga, line 170: maybe remove this? The logic is if all of the q-points found, then one does not need to merge - however, this is not true for the wall time. 

As we are merging the wall time in other places (or in the case where the q-points are split between files) seems best to do it here as well

src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -146,6 +153,40 @@ def prepare_process(self):
...
+
+        num_qpoints_found = sum(output['num_q_found'] for output in output_dict.values())
+
+        if num_qpoints_found == num_qpoints:
+            self.report(f'Merging {num_qpoints} q-points data from `PhCalculation`s.')
+            self.ctx.merged_output_parameters = merge_ph_outputs(**output_dict)
+        else:
+            self.report(f'Only {num_qpoints_found} of {num_qpoints} q-points were parsed.')
+            return self.exit_codes.ERROR_MERGING_QPOINTS
```
mkotiuga, line 179: The user could have chosen just to run a few q-points with start_q/last_q and in this case, the number of qpoints parsed from the beginning of the file will be determined by the qmesh and it will appear as if there are q-points missing. Maybe we remove this error, or check if the user specified this input

pyproject.toml
mbercx: Eh, not sure how that happened. Fixing now

src/aiida_quantumespresso/parsers/parse_raw/ph.py
mbercx: This unfortunately breaks one of my tests 😅. Could you provide some example output files for this case? Then I can also add a test for it and see how I can make sure both run successfully.

src/aiida_quantumespresso/workflows/ph/base.py
mbercx: This case is now also considered by checking if either of these inputs have been set and recalculating the `num_qpoints` a few lines above.

src/aiida_quantumespresso/workflows/ph/base.py
mbercx: I've removed this logic. Now the step checks if there is more than one child calculation for the restart work chain.

src/aiida_quantumespresso/parsers/parse_raw/ph.py
mbercx: Never mind! I generated a simple test case myself. But it's clear the parsing isn't giving us the desired outputs:

```yaml
dynamical_matrix_0:
  frequencies:
  - -3.151844
  - -3.151844
  - -3.139229
  frequencies_units: cm-1
  q_point:
  - 0.0
  - 0.0
  - 0.0
  q_point_units: 2pi/lattice_parameter
number_of_atoms: 1
number_of_irr_representations_for_each_q:
- 2
- 2
number_of_qpoints: 1
wall_time_seconds: 27.53

```

src/aiida_quantumespresso/parsers/parse_raw/ph.py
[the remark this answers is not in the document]
mbercx: Just amazing. So QE _doesn't_ print the `1 q-points for this run` for the first run of a single q-point, but it _does_ for a restart run. And why not just call the file `dynamical-matrix-1`? Too consistent?

I'm going to ask @giovannipizzi if it's possible to organise a Trieste visit and just spend as much time as is needed to move all the outputs to XML. At this point I think it's more efficient 😂.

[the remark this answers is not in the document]
mbercx: I propose an alternative fix in [47f5cc7](https://github.com/aiidateam/aiida-quantumespresso/pull/818/commits/47f5cc7c92f491ea5bf5afa19b4599927de669d9). It's not super clean, but I'm loathe to start fidgeting with the dynamical matrix parsing too much since we don't have many tests for other use cases of `ph.x`. This change has the added benefit that the output is more consistent, i.e. now for both single and multiple kpoint(s) the first output key is `dynamical_matrix_1`.

src/aiida_quantumespresso/parsers/parse_raw/ph.py
mbercx: The basic idea was solid, but I propose an alternative solution in [0c5cc84](https://github.com/aiidateam/aiida-quantumespresso/pull/818/commits/0c5cc8449c4c4d412bc291a2cdc7c88d11473cff), which basically boils down to the same thing but doesn't break for cases where the `ph.x` calculation didn't print the irr repr line at all. Now the outputs are starting to look pretty good:

```yaml
dynamical_matrix_1:
  frequencies:
  - -3.151844
  - -3.151844
  - -3.139229
  frequencies_units: cm-1
  mode_symmetry:
  - L_2'
  - L_3'
  - L_3'
  point_group: D_3d (-3m)
  q_point:
  - 0.0
  - 0.0
  - 0.0
  q_point_units: 2pi/lattice_parameter
number_of_atoms: 1
number_of_irr_representations_for_each_q:
- 2
number_of_qpoints: 1
wall_time_seconds: 27.53

```

mbercx:
It took a lot of tweaking, but I think the PR is now _very_ close to being finalised. I think the most work now is to write a proper commit message. ^^

The parsing can now deal with:

- Calculations that finish in one go for multiple q-points or a single one.
- Calculations that are restarted between q-points.
- Calculations that are restarted while running a single q-point.

In all cases, the outputs are identical, which should be the goal.

mbercx:
@mkotiuga I've updated the title and OP to what I think the commit header and message should be. Have a look and let me know if you have any comments!

mbercx:
@sphuber let me know if you still want to have a look at this PR, but hopefully my beautiful and extensive testing convinces you that this is quality code. 😁

mkotiuga:
@mbercx looks good

sphuber (dismissed):
Thanks a lot @mbercx and @mkotiuga . Fine for me to squash merge this

mbercx (approved):
Thanks a lot again for all the work @mkotiuga! It took a while to get this one right :)

PR #822, opened 2022-07-11
`PwParser`: Add support for QE v7.1

mbercx:
Add the XML schema for Quantum ESPRESSO v7.1 `pw.x`, along with a corresponding
test for the parser.

Note that the date of the file found in the `qeschemas` repository:

https://github.com/QEF/qeschemas/tree/master/PW_CPV/previous_schemas

doesn't correspond to the date in the header of the actual XML output.
Since this header is used for determining the correct file name, we
adapted the `.xsd` file name accordingly.


mbercx:
Note: This currently does not yet provide support for the new `HUBBARD` card input, see:

https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm1699

sphuber (approved):
Noice, thanks @mbercx

PR #824, opened 2022-07-19
`Q2rCalculation`: Fix copying from `parent_folder`

### LINKED ISSUE #823: `Q2rCalculation`: parent folder copying broken
In https://github.com/aiidateam/aiida-quantumespresso/pull/805 we simplified the handling of copying files from `parent_folder` directories, now that `aiida-core=2.0.0` supports recursively copying contents from directories.

However, in case the parent folder is a `RemoteData`, the outputs are now (correctly) copied from the `_default_parent_output_folder`:

https://github.com/aiidateam/aiida-quantumespresso/blob/15967da5431097c1dee043f104723656e8f53eca/src/aiida_quantumespresso/calculations/namelists.py#L174-L178

and the `Q2rCalculation` and `MatdynCalculation` plugins do not specify this class variable, so it reverts to the default set by `NamelistsCalculation` (`'./out/'`). So currently the `out` directory contents are copied to the `DYN_MAT` directory, which breaks the use of both calculation plugins.

mbercx:
Fixes #823 

When providing a `RemoteData` node as a `parent_folder` to the
`Q2rCalculation`, the contents of the `out` folder are now copied to the
`DYN_MAT` folder for the `q2r.x` calculation. This is because for
`RemoteData` input the `_default_parent_output_folder` is used, which is
defined as `out` for the parent `NamelistsCalculation` class, but not
overriden by `Q2rCalculation`.

Here we override the `_default_parent_output_folder` class variable so
the contents of the `DYN_MAT` folder are also copied when a `RemoteData`
input is provided.

### DIFF
--- a/src/aiida_quantumespresso/calculations/q2r.py
+++ b/src/aiida_quantumespresso/calculations/q2r.py
@@ -15,6 +15,7 @@ class Q2rCalculation(NamelistsCalculation):
     _FORCE_CONSTANTS_NAME = 'real_space_force_constants.dat'
     _OUTPUT_SUBFOLDER = PhCalculation._FOLDER_DYNAMICAL_MATRIX  # pylint: disable=protected-access
     _INPUT_SUBFOLDER = os.path.join('.', PhCalculation._FOLDER_DYNAMICAL_MATRIX)  # pylint: disable=protected-access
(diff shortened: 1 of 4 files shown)


mbercx:
Heh, yeah, some redesign might be useful at some point, but for now we'll have to bear the technical debt.

PR #827, opened 2022-07-20
`Q2rParser`: Set `filename` for `force_constants`

### LINKED ISSUE #826: `Q2rParser`: incorrect `filename` for `force_constants` output
Currently, the `filename` of the `force_constants` output of a `Q2rCalculation` seems to be a very long random string, i.e.:

```python
In [1]: q2r_calc = load_node(832465)
   ...: force_constants = q2r_calc.outputs.force_constants
   ...: force_constants.filename
Out[1]: 'c986fb8b4cecc82a0c2389f4ccdd4d67bcbdc93da7d358d9c500a9e54550bd'
```

This `filename` is also used in the `prepare_for_submission` method of the `MatdynCalculation`, resulting in a calculation input that _works_, but is somewhat confusing for the user.

mbercx:
Fixes #826 

Currently, the `force_constants` output has a filename that corresponds
to a long obscure string. Here we set the filename explicitly to the
`_FORCE_CONSTANTS_NAME` class variable defined by the `Q2rCalculation`
class.


mbercx:
@sphuber if we merge #824 first, then I'll add a test for this bug as well.

sphuber (approved):
Thanks @mbercx . Don't think a test is necessary for this

PR #828, opened 2022-07-20
Add support for electron-phonon calculations

### LINKED ISSUE #825: Provide support for electron-phonon calculations
Currently, the `Q2rCalculation` and `MatDynCalculation` plugins do not support electron-phonon calculations, since data related to the electron-phonon coefficients is stored in the `elph_dir` and this is not symlinked/copied from the remote folder of the `PhCalculation`.

mbercx:
Fixes #825 

Make several changes in the plugins for `ph.x`, `q2r.x` and `matdyn.x` to provide
support for electron-phonon calculations:

* Adapt the `prepare_for_submission` scripts to allow remote copying of
the `elph_dir` for both plugins in case `la2F` is set to true.
* Add support to the `MatDynCalculation` plugin for setting `dos` to
true, and in this case converting the `kpoints` input to the `nkX`
tags instead of providing them as a list.
* For `matdyn.x` calculations where `dos` is true, parse the phonon DOS
instead of the phonon bands, and provide this as an output.
* For the `MatdynCalculation` plugin, providing the `parent_folder` input
currently only makes sense for electron-phonon calculations, so a
validator is added to check this. The remote copy/symlink list is also
overriden by the `elph_dir` to avoid adding it to the default `out`
directory of the `NamelistsCalculation`, which is typically no longer
required for the `matdyn.x` step.


src/aiida_quantumespresso/calculations/matdyn.py
```diff
@@ -55,11 +62,20 @@ def _validate_inputs(value, _):
         if 'parameters' in value:
             parameters = value['parameters'].get_dict()
         else:
-            parameters = {}
+            parameters = {'INPUT': {}}
```
sphuber, line 65: Note that if the `parameters` contains an empty dict, this will now fail. Rather initialize empty dict and ensure the `INPUT` sub dict is there in both bases
```suggestion
            parameters = {}

        parameters.setdefault('INPUT', {})
```

mbercx: That said, if the user provides an _empty_ `Dict` node for `parameters`, they might as well have not provided the input (cleaner provenance). So perhaps we should validate that the `INPUT` key is in the `parameters` input instead?

mbercx: See f659d2c53922d0340fb3c9f7905ecb8711a52974

mbercx:
This has been rigorously field-tested with ~5500 materials. The corresponding workflow will be published in 

https://github.com/aiidateam/aiida-quantumespresso-ph

PR #829, opened 2022-07-20
Update the "Get Started" example of the `ph.x` code

mbercx:


mbercx:
@mkotiuga you want to give this a quick read to see if it is clear? 🙃 

https://aiida-quantumespresso--829.org.readthedocs.build/en/829/user_guide/get_started/examples/ph_tutorial.html#parameters

sphuber:
@mbercx @mkotiuga would be good to merge this soon so we can release it with 4.0.1

mkotiuga:
@mbercx @sphuber it looks OK to me as a very basic example. 

If I'm being picky, there is some wording I would change/details I would clarify. If I were to add one thing, it would be a little more about the q-points - specifically that a 1x1x1 grid gives the gamma point (just to be very clear) and how you can specify a list of explicit q-points.  

At some point it would be good to add the q2r and matdyn steps for calculating a dispersion...

sphuber (approved):

PR #833, opened 2022-08-10
Release/4.0.0

mbercx:


mbercx:
@sphuber tests are failing because `qe-tools` v2.0 hasn't been released yet.

sphuber:
> @sphuber tests are failing because `qe-tools` v2.0 hasn't been released yet.

Now it has been, restart?

mbercx:
Hmm, seems `aiida-quantumespresso` isn't one of my RTD projects, so I can't restart the build there.

However, I was still wondering about the Python version badges. Since we now support v3.8-v3.10, we have to adapt the compatibility matrix for the older versions, I suppose? Is there a way to automatically check the Python versions supported for a specific version of a package?

mbercx:
Couldn't find an automated sexy solution, so I caved and fixed the Python version shields manually. 😝

mbercx:
Just realised that QE v6.5 was released over two years ago:

https://www.quantum-espresso.org/quantum-espresso-v-6-5/

So we can restrict support to >v6.6. This has as a main benefit that we can assume that the `paro` `diagonalization` has been implemented.

sphuber:
> Hmm, seems `aiida-quantumespresso` isn't one of my RTD projects, so I can't restart the build there.
> 
> However, I was still wondering about the Python version badges. Since we now support v3.8-v3.10, we have to adapt the compatibility matrix for the older versions, I suppose? Is there a way to automatically check the Python versions supported for a specific version of a package?

Don't understand why you would have to do manually update the Python version badges? They are created automatically based on the link which takes the versions from PyPI.

mbercx:
> Don't understand why you would have to do manually update the Python version badges? They are created automatically based on the link which takes the versions from PyPI.

Sure, but won't it always that the _latest_ version?

sphuber:
> > Don't understand why you would have to do manually update the Python version badges? They are created automatically based on the link which takes the versions from PyPI.
> 
> Sure, but won't it always take the _latest_ version?

Yep, just realized this, you are right and now understand what you changed. All good.

sphuber (approved):
Let her rip!

PR #836, opened 2022-08-29
Hotfix missing aiida 2.0 compatibility update

### LINKED ISSUE #835: Missing 2.0 compatibility update
I am getting the following warning when submitting a `quantumespresso.pw.bands` workchain with `aiida-quantumespresso==4.0.0`:

``` python
In [26]: from aiida.engine import submit
    ...: workchain_node = submit(builder)
/home/framirez/miniconda3/envs/tutorial2022/lib/python3.8/site-packages/aiida_quantumespresso/workflows/pw/bands.py:21: AiidaDeprecationWarning: `Dict.get_attribute` is deprecated, use `Dict.base.attributes.get` instead. (this will be removed in v3)
  if 'nbands_factor' in inputs and 'nbnd' in inputs['bands']['pw']['parameters'].get_attribute('SYSTEM', {}):
```

ramirezfranciscof:
Close #835

I can contribute this but I can't do a bugfix release and I think we might want this for the upcoming tutorial (although perhaps it is better to wait until all the sections are reviewed and so we know if there is any other fix needed here).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,4 +66,6 @@ jobs:
             run: pip install -e .[tests]
 
         -   name: Run pytest
+            env:
+                AIIDA_WARN_v3: 1
             run: pytest -sv tests
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -54,6 +54,8 @@ jobs:
 
         -   name: Run pytest
             id: tests
+            env:
+                AIIDA_WARN_v3: 1
             run: pytest -sv tests
 
         -   name: Slack notification
--- a/src/aiida_quantumespresso/calculations/__init__.py
+++ b/src/aiida_quantumespresso/calculations/__init__.py
@@ -576,7 +576,7 @@ class BasePwCpInputGenerator(CalcJob):
             try:
                 structure_parameters = get_parameters_from_cell(
                     ibrav=ibrav,
-                    cell=structure.get_attribute('cell'),
+                    cell=structure.base.attributes.get('cell'),
                     tolerance=settings.pop('IBRAV_CELL_TOLERANCE', 1e-6)
                 )
             except ValueError as exc:
--- a/src/aiida_quantumespresso/calculations/ph.py
+++ b/src/aiida_quantumespresso/calculations/ph.py
@@ -103,7 +103,7 @@ class PhCalculation(CalcJob):
             settings = {}
 
         parent_folder = self.inputs.parent_folder
-        parent_calcs = parent_folder.get_incoming(node_class=orm.CalcJobNode).all()
+        parent_calcs = parent_folder.base.links.get_incoming(node_class=orm.CalcJobNode).all()
 
         if not parent_calcs:
             raise exceptions.NotExistent(f'parent_folder<{parent_folder.pk}> has no parent calculation')
--- a/src/aiida_quantumespresso/calculations/pwimmigrant.py
+++ b/src/aiida_quantumespresso/calculations/pwimmigrant.py
@@ -301,7 +301,7 @@ class PwimmigrantCalculation(PwCalculation):
         if settings_dict:
             self.use_settings(Dict(settings_dict))
 
-        self.set_attribute('input_nodes_created', True)
+        self.base.attributes.set('input_nodes_created', True)
 
     def _prepare_for_retrieval(self, open_transport):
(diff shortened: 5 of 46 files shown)


ramirezfranciscof:
Ok, this is ready I think

sphuber:
Thanks @ramirezfranciscof . Guess we forgot to enable the deprecation warnings when we released 4.0, which was done in a bit of a rush. We need to address all of them instead of just these three.

ramirezfranciscof:
All warnings seem to have been addressed @sphuber (also tagging @mbercx now, didn't do it at first because it was a smaller corrections and thought we would just merge it yesterday).

There are 4 remaining but that seem to come from AiiDA core, so I think not much to do here.

```
=============================== warnings summary ===============================
tests/calculations/test_autoinvalidate_cache.py::test_exit_code_invalidates_cache[quantumespresso.cp]
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/aiida/manage/tests/main.py:125: AiidaDeprecationWarning: reset_db() is deprecated, use clear_profile() instead (this will be removed in v3)
    warn_deprecation('reset_db() is deprecated, use clear_profile() instead', version=3)

tests/calculations/test_autoinvalidate_cache.py::test_exit_code_invalidates_cache[quantumespresso.xspectra]
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/aiida/plugins/entry_point.py:307: AiidaDeprecationWarning: The entry point `array.kpoints` is deprecated. Please replace it with `core.array.kpoints`. (this will be removed in v3)
    warn_deprecation(

(...)
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/aiida/orm/entities.py:174: AiidaDeprecationWarning: This property is deprecated, use `.collection` instead. (this will be removed in v3)
    warn_deprecation('This property is deprecated, use `.collection` instead.', version=3, stacklevel=2)

tests/parsers/test_neb.py::test_neb_default
tests/parsers/test_neb.py::test_neb_all_iterations
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/aiida/orm/utils/managers.py:101: AiidaDeprecationWarning: dereferencing nodes with links containing double underscores is deprecated, simply replace the double underscores with a single dot instead. For example: 
  `self.inputs.some__label` can be written as `self.inputs.some.label` instead.
   (this will be removed in v3)
    warn_deprecation(

-- Docs: https://docs.pytest.org/en/stable/warnings.html
=========== 253 passed, 1 skipped, 187 warnings in 113.78s (0:01:53) ===========
```

sphuber:
I think all of these warnings should actually still be changed.

* The first one, `reset_db`, is probably being used in our `conftest.py` somewhere. Change it to `clear_profile`
* There is an instance of `array.kpoints` -> `core.array.kpoints`
* There is an instance of `Entity.collection` which should be `Entity.objects`
* Search for a label with double underscore `__`, as said in the warning that should be replaced with a `.`. Or did you look for those and not find anything?

ramirezfranciscof:
Ah, wait, now reading closely I see these don't seem to be from `aiida-core` using its own deprecated functions in calls to non-deprecated ones, but still from `aiida-quantumespresso` using deprecated functions. Then why is this not showing me the lines of `aiida-quantumespresso` that call the deprecated functions like the other ones did?

I'll try to check this on Friday.

sphuber:
@ramirezfranciscof can you still have a look at the final warnings so we can merge this?

ramirezfranciscof:
@sphuber yes, sorry, all warning messages have been taken care of now.

I will say that I think the ones from the last batch would merit the consideration of why there were not being shown with a full stack trace and if there is some way to fix them. Moreover, some of the warning messages could be improved: 'This property is deprecated, use `.collection` instead' doesn't explicitly mention that the property is '.objects', I had to go into 'aiida/orm/entities.py' to figure it out.

sphuber:
> I will say that I think the ones from the last batch would merit the consideration of why there were not being shown with a full stack trace and if there is some way to fix them. Moreover, some of the warning messages could be improved: `This property is deprecated, use .collection instead` doesn't explicitly mention that the property is `.objects`, I had to go into `aiida/orm/entities.py` to figure it out.

Fair enough, but that is a problem for `aiida-core`. Nothing we can do about it here

sphuber (approved):
Thanks a lot @ramirezfranciscof

PR #839, opened 2022-09-05
`PwCalculation/PhCalculation`: add validation for `pseudos` port

### LINKED ISSUE #838: Add validation of `pseudos` input to `PwCalculation`
Currently it is possible to submit a `PwCalculation` without pseudos and it will fail in the upload stage, excepting the process.

sphuber:
Fixes #838 

The `pseudos` port was not marked as required, making it possible to
submit a calculation without any pseudos, but this would always fail in
the `prepare_for_submission`.

The port is made required and a validator is added to check that the
pseudos provided match the set of kinds of the `structure` input.

### DIFF
--- a/src/aiida_quantumespresso/calculations/__init__.py
+++ b/src/aiida_quantumespresso/calculations/__init__.py
@@ -123,7 +123,7 @@ class BasePwCpInputGenerator(CalcJob):
             help='An optional working directory of a previously completed calculation to restart from.')
         spec.input('vdw_table', valid_type=orm.SinglefileData, required=False,
             help='Optional van der Waals table contained in a `SinglefileData`.')
-        spec.input_namespace('pseudos', valid_type=(LegacyUpfData, UpfData), dynamic=True,
+        spec.input_namespace('pseudos', valid_type=(LegacyUpfData, UpfData), dynamic=True, required=True,
             help='A mapping of `UpfData` nodes onto the kind name to which they should apply.')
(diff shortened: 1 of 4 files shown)


src/aiida_quantumespresso/calculations/__init__.py
```diff
@@ -136,11 +136,29 @@ def define(cls, spec):
...
+            return 'required value was not provided for the `pseudos` namespace.'
+
+        # Some wrapping processes expose the inputs excluding the ``structure``, in which case we can't validate.
+        if 'structure' not in value:
+            return
+
+        structure_kinds = value['structure'].get_kind_names()
+        pseudo_kinds = list(value['pseudos'].keys())
+
+        if structure_kinds != pseudo_kinds:
```
mbercx, line 156: Should we sort the kinds list first, or use sets? They don't necessarily have the same order.

mbercx (changes requested):
Thanks @sphuber! Looking good, just a small comment regarding comparing the kind lists.

src/aiida_quantumespresso/calculations/__init__.py
sphuber: Astute observation and you are absolutely correct. I used sets initially, but had some kind of error, so removed it. But we should use it yeah, will adapt.

mbercx (approved):
All good to go then, thanks @sphuber! 🚀

Issue #745, opened 2021-09-24
Protocol: Clarify/Improve API for custom protocols and overrides


mbercx:
The current API for the `get_builder_from_protocol()` method doesn't really allow the user to specify their own protocol, only change inputs using the `overrides` argument or once the builder has been obtained. Moreover, there is some discussion on how the overrides should work, and if there should be a clear separation between actual inputs of the work chain and so-called meta parameters. Below I try to separate these issues and give a clear description of each.

### "True" protocols vs "override" protocols

Defining your own protocol files is different from the overrides, since overrides are always defined in relation to the protocol. This means that e.g. not specifying a tag will lead to the default value being used, not the value being unset (although we could look into allowing this by setting the value to `null`, see https://github.com/aiidateam/aiida-quantumespresso/issues/653).

The challenge here is that there are multiple protocol files, i.e. one for each work chain. And of these, only the `PwBaseWorkChain` is a "true" protocol, i.e. the others are defined in relation to the `PwBaseWorkChain` one, as they override its inputs. This is for a good reason: The protocol files are much easier to maintain this way, and can be used across packages.

The question becomes if users should be able to define a "pure" protocol, and if so: how they can define this. Do we have one large YAML file where the user can specify the "pure" protocol for the `PwBaseWorkChain`, as well as the protocol overrides for the higher level ones? What if a certain work chain is missing from this file? One option for the latter would be to simply revert to the default in case the requested protocol is missing for a certain work chain.

### Meta parameters vs actual inputs

The protocol files specify two different kinds of inputs:

1. Meta parameters: these are not actual input ports of the work chain or any process it wraps. Examples include:
  * `pseudo_family`
  * `conv_thr_per_atom`
  * `etot_conv_thr_per_atom`
2. Actual inputs of the work chain or its subprocesses, i.e. those defined by their spec.

There has been some confusion on the fact that the `overrides` are used to specify both meta parameters and actual inputs, see #712. The example here is especially confusing since there is both a `pseudo_family` meta parameters and (deprecated) `pseudo_family` input. Still, it would be better to separate how these are overridden. A protocol should still define both, however.

### How to override inputs

In order to override the inputs of the protocol, there are two approaches:

1. Use the `overrides` input argument, which is either a `dict` or `pathlib.Path` to a YAML file.
2. After obtaining the builder, override the desired inputs in the builder.

[...]

PR #846, opened 2022-10-07
Add the `options` argument to all `get_builder_from_protocol`

### LINKED ISSUE #844: Add `options` argument to all `get_builder_from_protocol`
Often a user will need to use the same options for all the jobs that will be run in a nested workflow, such as specifying the `queue_name` or the `account`, etc. Currently, the user will have to manually specify those on the builder, or in the `overrides`, but this means that the user knows exactly where in the nesting the jobs are. There should be a way to define these options once, and they get automatically propagated to all nested `CalcJob` inputs.

sphuber:
Fixes #844 

Often a user will need to use the same options for all the jobs that will be run in a nested workflow, such as specifying the `queue_name` or the `account`, etc. Currently, the user will have to manually specify those on the builder, or in the `overrides`, but this means that the user knows exactly where in the nesting the jobs are.

The `options` argument is added to all `get_builder_from_protocol` implementations, which takes a dictionary just as one would pass to the `metadata.options` input for a `CalcJob`. The implementation ensures that these inputs are recursively set on all the inputs of all the `CalcJob`s that are part of the nested namespace.


sphuber:
@ltalirz quick implementation of the first feature request. If you could verify this satisfies your requirements and works in the wild, that would be great.

mbercx:
I can definitely see the use case, have thought about implementing something similar in the past. I had some reservations at the time, since there are many inputs one might want to set for all calculation jobs (e.g. `parallellization`, `settings`, `parameters`, ...) and I didn't want the input signature to be extended too much. We also hadn't fully agreed on the `get_builder_from_protocol()` API yet (https://github.com/aiidateam/aiida-quantumespresso/issues/745) and the options aren't necessarily the same for all calculations in the work chain (e.g. resources, you might want to run the relaxation on 4 nodes, but the NSCF on a single node). So any serious user of the work chain would still have to figure out where the calcjob options are in the work chain hierarchy, which is easier now that the builder representation shows the hierarchy and where the options are.

That said, I don't want to block this feature, it can definitely be useful for people that just begin to use the work chain, and the options are one of the inputs you pretty much always have to override.

sphuber:
The points you raise are indeed why we at the time didn't implement this feature. There simply doesn't seem to be a single solution that fits all use-cases, which is what I mentioned to @ltalirz as to why this didn't exist yet. That being said though, currently it is quite difficult to use the `get_builders_from_protocol` for novice users precisely because they most likely will have to set some options for all calcjobs and they will have to find them in the namespace. It think @ltalirz mentioned, rightfully so, that we should try and make it as easy to use for the majority of use cases (novice users) while still allowing expert users to tune the behavior. This was also the main design goal of the common workflows. So even though for some workflows you might want to use different resources, expert users that know they want to do this, will therefore know that there *even are* different calcjobs and so they will be more likely to find them in the namespace. Overall then, I think this change will make life of the average user easier.

Regarding the parallel with the `parallelization`, `settings` and `parameters`. The `parameters` are essentially entirely taken care of by the protocols, so these don't really count I think. The `parallelization` and `settings` are really more for advanced users, and so again I think it might be reasonable to expect them to understand the workchain in more detail and be able to address the calcjobs in the various namespaces.

sphuber:
Verdict is that it is not a 100% sure that this is the best solution, but let's merge it for now as it brings considerable advantages at the risk we may have to change things forcefully in the future.

PR #852, opened 2022-10-21
Implement a Base WorkChain for XSpectra

PNOGillespie:
# Overview
In this PR, we propose to add a core WorkChain for XSpectra. This WorkChain will take a given input structure where one atomic site is designated as the site for the absorbing atom (see inputs for more info) and run all XSpectra calculations requested. We designate this as a "Base" WorkChain since this is intended to be paired with a future "Relax" WorkChain which will start from a simple input structure, relax the structure, prepare a marked structure for all potential absorbing sites (depending on symmetry), and use the BaseWorkChain to compute all the required spectra for each structure. Tools to generate structures for the BaseWorkChain will be included in such a future addition. The BaseWorkChain however can, in principle, be run on its own with a suitable input structure (see inputs section).

# Process Flow
1. Perform an SCF calculation using the `PwBaseWorkChain` for the structure with a core-hole present.
2. (Optionally) run the `upf2plotcore.sh` utility script on the ground-state pseuopotential assigned to the absorbing elemental species (i.e. the non-absorbing-atoms of the same element type as the absorbing atom). This is only called if `core_wfc_data` is not provided as an input. Requires `AiiDA-Shell` and a code node for the script using the plugin in order to run.
3. Perform an XSpectra calculation for each requested polarisation vector (see `eps_vectors` in the inputs). This will run with the broadening parameter `xgamma` set relatively high (0.8 eV) - see the Notes section for more info.
4. Perform a second XSpectra calculation to re-plot the spectra from the XSpectra save files produced in step 3 - see the Notes section for more info.
5. Collate all output data: return the `output_parameters` of the SCF calculation and all XSpectra calculations from step 3, return all final spectra from step 4 in a single `XyData` node for easier plotting, and produce a representation of the XANES of a powder sample of the structure (if requested).

# Inputs
Important inputs to take notice of are:
* `structure` - the input `StructureData` node for the WorkChain. The structure must contain at least one unique `Kind` which represents the absorbing atom and appears at one `Site` in the structure. The name of the absorbing atom `Kind` must match the `abs_atom_marker` in the inputs.
* `abs_atom_marker` - a `Str` node designating the name of the `Kind` to be used as the absorbing atom. Defaults to "X" but can be overwritten
* `core_wfc_data` - a `SinglefileData` node containing the core wavefunction of the ground-state of the absorbing element, obtained by parsing the ground-state pseudopotential for the absorbing element with `upf2plotcore.sh`. Since this is normally obtained by `AiiDA-Shell` and thus (as far as I remember) can only be the `stdout`, the filename of this must be "stdout".
* `eps_vectors` - a `List` node containing all the 3-vectors to be computed for `xepsilon(1-3)`, thus also defining the number of XSpectra sub-processes to launch. Defaults to `[[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]]`
* `get_powder_spectrum` - a `Bool` node which requests if the powder spectrum of the structure should be computed. Defaults to `False`. The `CalcFunction` associated to this will operate differently depending on the number of given spectra (from 1 to 3) using symmetry rules and, since no symmetry information can be obtained by the WorkChain in advance, the WorkChain will filter out the correct spectra to use and potentially even skip the process with a warning in the logs if not enough of the required spectra are available.

# Protocol Inputs
When computing XANES spectra, some type of core-hole screening is typically used (e.g. adding a counter charge to the calculation inputs). Since there are a few different recipes out there to do this (e.g. "full-core-hole", "half-core-hole", etc) we propose to provide pre-written inputs to take care of this. In addition to setting the `XspectraCalculation` inputs automatically, the `XspectraBaseWorkChain` protocols also define inputs for the `PwBaseWorkChain` to represent a few potential core-hole treatments (including no treatment). Crucially however, the `get_builder_from_protocol()` step *does not* define all the inputs for the `PwBaseWorkChain` but rather uses the existing `get_protocol_inputs` for the `PwBaseWorkChain` and simply merges the inputs for the core-hole treatment over the top of these (via `recursive_merge()`). This therefore prevents the need to duplicate the `PwBaseWorkChain` protocol inputs and only changes what is needed for the core-hole treatment specified by this WorkChain. `XspectraBaseWorkChain.get_builder_from_protocol()` therefore requires `xs_protocol` and `pw_protocol` as separate inputs.

Finally, we include optional inputs to `get_builder_from_protocol()` for `core_wfc_data`, `core_hole_pseudos`, and `options` (in line with recent commit 0496b5f8b8d10f60aa46ee41b100e60197f6bf75) which provide shortcuts for more frequently-used overrides.

# Tests
This PR also includes tests for the WorkChain, both for basic process flow and for `get_builder_from_protocol()` (see commit 3cab1f85b22bdfce9b73b219fd93da722556af9c). This also includes adding new fixtures required to run the WorkChain test itself. Note that the WorkChain process flow test does not include the `run_upf2plotcore` step, since this is mainly handled by `AiiDA-Shell` and it made sense to keep the tests within the `AiiDA-QE` environment.

# Notes
## Why Re-Plot?
The rationale for running two `XspectraCalculation`s for one final spectrum is because xspectra.x uses the parameter `xgamma` in the input file as a broadening parameter in *both* the Lanczos recursion method and the convolution of the final XANES spectrum. Increasing the value of `xgamma` to 0.8 eV (as we have done in the `xs_prod` step) significantly increases the convergence rate of the Lanczos procedure but yields an over-broadened final spectrum, while a low value of `xgamma` yields a more reasonable spectrum but takes significantly longer (particularly for calculations using a large number of k-points). We therefore propose to compute the Lanczos in one step with a high broadening parameter and then re-plot at a lower value in order to reduce calculation time. Note that XSpectra always applies a Lorenzian convolution to the final spectrum, thus some level of broadening is required regardless.

## Changes to the `XspectraParser`
See commits bde5b328b6a06df1cacfc9142d642996ac4cbef7 and 66cfa3957d0f0325f4ebf2d593acc0e468017931

In the course of writing the WorkChain, it was noticed that when XSpectra performs a calculation re-start and reports information from the XSpectra save file, the returned information on the polarisation vector used is reported on a cartesian basis - even if the parent calculation uses a "crystal system" basis. This is because the polarisation vectors are always stored in the save file using a cartesian basis regardless. In light of this, it was decided to change the `XspectraParser` to parse certain information on the calculation from the inputs rather than the output file. This PR thus also changes the `XspectraParser` and associated tests to improve the reliability of the Parser.


sphuber:
Thanks a lot for the contribution @PNOGillespie . Especially the detailed PR message is very useful. Before I go into a detailed review, there was one point that I wanted to raise. So far, we have been using the `BaseWorkChain` terminology for a workchain that wraps around a `CalcJob` and just deals with error handling. It is not supposed to be a workchain that comprises multiple steps of different codes. This is such a common paradigm in AiiDA (having to recover from errors in a `CalcJob`) that I wrote a base implementation of `WorkChain` in `aiida-core` called the `BaseRestartWorkChain`. With this, you can easily add error handling for any `CalcJob`. For `XspectraCalculation`, it would look something like this:

```python
# -*- coding: utf-8 -*-
"""Workchain to run a Quantum ESPRESSO xspectra.x calculation with automated error handling and restarts."""
from aiida.common import AttributeDict
from aiida.engine import BaseRestartWorkChain, ProcessHandlerReport, process_handler, while_
from aiida.plugins import CalculationFactory

XspectraCalculation = CalculationFactory('quantumespresso.xspectra')


class MatdynBaseWorkChain(BaseRestartWorkChain):
    """Workchain to run a Quantum ESPRESSO xspectra.x calculation with automated error handling and restarts."""

    _process_class = XspectraCalculation

    @classmethod
    def define(cls, spec):
        """Define the process specification."""
        super().define(spec)
        spec.expose_inputs(XspectraCalculation, namespace='xspectra')
        spec.expose_outputs(XspectraCalculation)
        spec.outline(
            cls.setup,
            while_(cls.should_run_process)(
                cls.run_process,
                cls.inspect_process,
            ),
            cls.results,
        )
        spec.exit_code(
            300, 'ERROR_UNRECOVERABLE_FAILURE',
            message='The calculation failed with an unrecoverable error.'
        )

    def setup(self):
        """Call the `setup` of the `BaseRestartWorkChain` and then create the inputs dictionary in `self.ctx.inputs`.

        This `self.ctx.inputs` dictionary will be used by the `BaseRestartWorkChain` to submit the calculations in the
        internal loop.
        """
        super().setup()
        self.ctx.restart_calc = None
        self.ctx.inputs = AttributeDict(self.exposed_inputs(XspectraCalculation, 'xspectra'))

    def report_error_handled(self, calculation, action):
        """Report an action taken for a calculation that has failed.

        This should be called in a registered error handler if its condition is met and an action was taken.

        :param calculation: the failed calculation node
        :param action: a string message with the action taken
        """
        arguments = [calculation.process_label, calculation.pk, calculation.exit_status, calculation.exit_message]
        self.report('{}<{}> failed with exit status {}: {}'.format(*arguments))
        self.report(f'Action taken: {action}')

    @process_handler(priority=600)
    def handle_unrecoverable_failure(self, node):
        """Handle calculations with an exit status below 400 which are unrecoverable, so abort the work chain."""
        if node.is_failed and node.exit_status < 400:
            self.report_error_handled(node, 'unrecoverable error, aborting...')
            return ProcessHandlerReport(True, self.exit_codes.ERROR_UNRECOVERABLE_FAILURE)
```

For more information, you can read [this section of the documentation](https://aiida.readthedocs.io/projects/aiida-core/en/latest/howto/workchains_restart.html) which explains the concept in more detail.

What I would suggest then, to make this PR more manageable to review and merge:

* Ideally move the parser changes to a separate PR.
* Add the `XspectraBaseWorkChain` using the code I sketched above
* Rename the workchain you added to `XspectraWorkChain` (another name is also fine, just don't think we should use the `BaseWorkChain` suffix since that already has special meaning. With `XspectraWorkChain` you could still add `XspectraRelaxWorkChain` later to include relax steps if you'd like)

If you need help with `git` to squash the commits and separate out the parser changes to create a separate PR, let me know. Also happy to do it for you if you prefer.

PNOGillespie:
Thanks for the review @sphuber. I had thought originally about structuring the proposed WorkChain in a similar way to how you suggest - a `BaseRestartWorkChain` to handle XSpectra sub-processes and a main WorkChain to essentially handle the process flow for a full analysis run. My initial thinking however was that it would be controversial to have a WorkChain which operates only on a single post-processing rather than the entire workflow from `PwCalculation` to final post-processing (since this would not really be a WorkChain starting from "first-principles") - much like `PwBandsWorkChain` and `PdosWorkChain`, which I took more inspiration from in this case. However, looking through your example and the other uses of BaseRestartWorkChain for the phonon codes I can see the mistake in my thinking there (in my defense, my work is focused on X-ray spectra so I haven't made use of the phonon WorkChains yet :sweat_smile:). 

I agree to the change in name, where I could change my originally-proposed `XspectraBaseWorkChain` to `XspectraCoreWorkChain` - principally to distinguish it more clearly from the upcoming "Relax" WorkChain (the name for which is provisional, and remains a minor point of discussion in our group) - and add a new `XspectraBaseWorkChain` to wrap the `XspectraCalculation` sub-processes. Hopefully I can get all the changes sorted and tested by tomorrow.

Regarding the changes to `XspectraParser`: since I pushed those commits out on a particular branch (`xspectra_base_workchain`) rather than `develop`, is there a way to merge only those commits to `develop` (or another branch) and create a separate PR that way, or would I need to make new commits to a different branch to bring the changes in-line with the branch used for this PR? I'm still learning my way around `git`, so apologies if I'm missing an obvious trick here.

sphuber:
> My initial thinking however was that it would be controversial to have a WorkChain which operates only on a single post-processing rather than the entire workflow from `PwCalculation` to final post-processing (since this would not really be a WorkChain starting from "first-principles")

It's a fair point, and maybe having a base workchain for `XspectraCalculation` may be overkill. Does `xspectra.x` ever fail? Can it exit with any error? If so, then it might make sense, especially if there is an automatic way of recovering from the error by relaunching the calculation (with some updated inputs perhaps).

> I agree to the change in name, where I could change my originally-proposed `XspectraBaseWorkChain` to `XspectraCoreWorkChain` - principally to distinguish it more clearly from the upcoming "Relax" WorkChain (the name for which is provisional, and remains a minor point of discussion in our group) - and add a new `XspectraBaseWorkChain` to wrap the `XspectraCalculation` sub-processes. Hopefully I can get all the changes sorted and tested by tomorrow.

`XspectraCoreWorkChain` works just fine :+1:


> Regarding the changes to `XspectraParser`: since I pushed those commits out on a particular branch (`xspectra_base_workchain`) rather than `develop`, is there a way to merge only those commits to `develop` (or another branch) and create a separate PR that way, or would I need to make new commits to a different branch to bring the changes in-line with the branch used for this PR? I'm still learning my way around `git`, so apologies if I'm missing an obvious trick here.

It certainly is possible. I think you put the changes in separate commits, which will help you here. What you can do is the following:
```
# Make sure remotes are up-to-date
git fetch --all
# Check out new branch from latest `develop`
git checkout -b fix/xspectra-parsers origin/develop
# cherry-pick the commits from your current branch
git cherry-pick 3c83c4d9bc62d3bc7f752d9add509fe001e0a8be
```
For the last command, it will take the commit hash of the commit with your parser changes and try to apply them to the new branch `fix/xspectra-parsers`. It might be possible that there are merge conflicts. Simply address them in the relevant files and do `git add` to stage them. Once done, you do `git cherry-pick --continue`. It will then recreate the commit on the new branch. Repeat the `git cherry-pick` command for each commit you want to copy to the new branch

PNOGillespie:
> It's a fair point, and maybe having a base workchain for `XspectraCalculation` may be overkill. Does `xspectra.x` ever fail? Can it exit with any error? If so, then it might make sense, especially if there is an automatic way of recovering from the error by relaunching the calculation (with some updated inputs perhaps).

The only recoverable failure seems to be a clean exit after hitting the assigned time limit (`INPUT_XSPECTRA/time_limit` in this case) though it does occur to me now that I haven't yet tested what happens in the case where the Lanczos procedure fails to converge (I'm doing so now, so probably I will find a new exit code to add for that). That said, the procedure always converges eventually if even a small (e.g. 0.2 eV) smearing is applied, which happens quite quickly if one uses the default value in the proposed `CoreWorkChain` (0.8 eV). I think wrapping it in a `BaseWorkChain` to handle restarts is still reasonable especially because, as  `xspectra.x` is unfortunately not parallelized on k-points, calculations on dense k-point meshes can take a while to finish. All that is required is to change `INPUT_XSPECTRA/restart_mode` to `restart` and the code then reads the previous Lanczos coefficients from a save file produced by the previous calculation, fast-forwards through the already-computed k-points, and resumes where it stopped last time. The `XspectraCalculation` class looks for and copies the save file from the parent folder if the parent was an `XspectraCalculation` (see previous PR: aiidateam#837 for more info).

> For the last command, it will take the commit hash of the commit with your parser changes and try to apply them to the new branch fix/xspectra-parsers. It might be possible that there are merge conflicts. Simply address them in the relevant files and do git add to stage them. Once done, you do git cherry-pick --continue. It will then recreate the commit on the new branch. Repeat the git cherry-pick command for each commit you want to copy to the new branch

Thanks for the help with that. I will get on with this now and setup a separate PR shortly afterwards.

PNOGillespie:
Hi @sphuber, just a little update: I have now a working version of a `BaseRestartWorkChain` for XSpectra which I am currently writing some testing fixtures for. None of the work has been committed yet, as I wanted to check a point of protocol first: Since this PR now really refers to implementing a `CoreWorkChain`, and I'm now planning to add a BaseRestartWorkChain as well, would you prefer me to open a separate PR for the Restart WorkChain itself or just collect everything here for review?

sphuber:
> Hi @sphuber, just a little update: I have now a working version of a `BaseRestartWorkChain` for XSpectra which I am currently writing some testing fixtures for. None of the work has been committed yet, as I wanted to check a point of protocol first: Since this PR now really refers to implementing a `CoreWorkChain`, and I'm now planning to add a BaseRestartWorkChain as well, would you prefer me to open a separate PR for the Restart WorkChain itself or just collect everything here for review?

It'd be great if you could open a separate PR with just the `BaseRestartWorkChain` implementation. When that is merged, making the PR for the core workchain will then be a lot easier to review

PNOGillespie:
Ok @sphuber, I will close this PR - since especially now the name, branch name, and description has diverged from the original intent - and open a new one for just the `BaseRestartWorkChain`. After finalising that one, I'll start a new one for the `CoreWorkChain`.

sphuber:
Thanks a lot for the great work @PNOGillespie really appreciated :+1:

PR #853, opened 2022-10-27
`XspectraParser`: Parse parameters from CalcJob inputs

PNOGillespie:
Addresses a request made in PR: aiidateam#852

Updates the method for the `XspectraParser` class to parse the following parameters from the input of the CalcJob class, rather than the stdout:

* The polarisation (epsilon) vectors used in the calculation
* The coordinate system used for the calculation (either crystallographic or cartesian)
* Whether the calculation was a re-plot or a full calculation (`xonly_plot`)

Keyword `plot_only` in `output_parameters` is also changed to `only_plot_spectrum` in order to clarify its meaning.

The changes were made necessary as, during further testing involving calculation restarts, it was found that XSpectra stores the polarisation and k-vector used for the calculation in its save file using a cartesian basis regardless of the setting used in the calculation itself. When performing a re-start or re-plot calculation the code reports the polarisation/k-vector stored in the save file, on a cartesian basis, which causes confusion for the Parser - particularly in the case of a restarted calculation as the polarisation vector is then reported twice: once at the initial read of the input file and then again when the save file is read (and in potentially different coordinate systems, depending on calculation settings). It was therefore decided to move the parsing of these parameters from the raw output to just taking them directly from the CalcJob inputs.

### DIFF
--- a/src/aiida_quantumespresso/parsers/xspectra.py
+++ b/src/aiida_quantumespresso/parsers/xspectra.py
@@ -87,6 +87,19 @@ class XspectraParser(Parser):
         xy_data.set_y(y_arrays, y_names, y_units)
 
         parsed_data, logs = parse_stdout_xspectra(filecontent=out_file, codename='XSpectra')
+
+        # Parse some additional info which the stdout does not reliably report
+        parameters = self.node.inputs.parameters.base.attributes.get('INPUT_XSPECTRA', {})
+
+        xepsilon_defaults = {
+            '1': 0,
+            '2': 0,
+            '3': 1,
+        }
+        raw_xepsilon = [parameters.get(f'xepsilon({n})', xepsilon_defaults[n]) for n in ['1', '2', '3']]
+        parsed_data['xepsilon'] = [float(n) for n in raw_xepsilon]
+        parsed_data['xcoordcrys'] = parameters.get('xcoordcrys', True)
+        parsed_data['xonly_plot'] = parameters.get('xonly_plot', False)
         self.emit_logs(logs)
 
         self.out('spectra', xy_data)
@@ -165,16 +178,6 @@ def parse_stdout_xspectra(filecontent, codename=None, message_map=None):
     # Parse the necessary information for data plotting: core level energy of the
     # absorbing atom and the energy zero of the spectrum (typically the Fermi level)
     for line in lines:
-        if 'xepsilon' in line:
-            eps_vector_string = line.split(':')[-1].strip()
-            eps_vector_list = [float(i) for i in eps_vector_string.split('   ')]
-            parsed_data['epsilon_vector'] = eps_vector_list
-        if 'xonly_plot:' in line:
-            is_only_plot_string = line.split(':')[-1].strip()
-            if is_only_plot_string == 'FALSE':
-                parsed_data['plot_only'] = False
-            elif is_only_plot_string == 'TRUE':
-                parsed_data['plot_only'] = True
         if 'From SCF save directory' in line:
             if '(spin polarized work)' in line:
                 spin = True
@@ -206,7 +209,7 @@ def parse_stdout_xspectra(filecontent, codename=None, message_map=None):
             parsed_data['fermi_energy'] = ef_energy
             parsed_data['fermi_energy_units'] = ef_energy_units
 
-        # parse dynamical RAM estimates
+        # parse per-process dynamical RAM estimates
         if 'Estimated max dynamical RAM per process' in line:
             value = line.split('>')[-1]
             match = re.match(r'\s+([+-]?\d+(\.\d*)?|\.\d+([eE][+-]?\d+)?)\s*(Mb|MB|GB)', value)
@@ -217,7 +220,7 @@ def parse_stdout_xspectra(filecontent, codename=None, message_map=None):
                 except (IndexError, ValueError):
                     pass
 
-        # parse dynamical RAM estimates
+        # parse total dynamical RAM estimates
         if 'Estimated total dynamical RAM' in line:
             value = line.split('>')[-1]
             match = re.match(r'\s+([+-]?\d+(\.\d*)?|\.\d+([eE][+-]?\d+)?)\s*(Mb|MB|GB)', value)
@@ -247,9 +250,8 @@ def parse_stdout_xspectra(filecontent, codename=None, message_map=None):
                 parsed_data['code_version'] = line.split(codestring)[1].split('starts on')[0].strip()
 
             # Parse the walltime
-            # XSpectra does not appear next to the timing data, so we must find either 'xanes' or 'nrixs' instead.
-            if 'xanes' in line or 'nrixs' in line:
-                if 'WALL' in line:
+            # XSpectra does not appear next to the timing data, so we must find 'xanes' instead.
+            if 'xanes' in line and 'WALL' in line:
                     try:
                         time = line.split('CPU')[1].split('WALL')[0].strip()
                         parsed_data['wall_time'] = time
--- a/tests/parsers/test_xspectra.py
+++ b/tests/parsers/test_xspectra.py
@@ -7,7 +7,11 @@ from aiida.common import AttributeDict
 
 def generate_inputs():
     """Return only those inputs that the parser will expect to be there."""
-    return AttributeDict()
+
+    parameters = {'INPUT_XSPECTRA': {}, 'PLOT': {}, 'PSEUDOS': {}, 'CUT_OCC': {}}
+
+    inputs = {'parameters': orm.Dict(parameters)}
+    return AttributeDict(inputs)
 
 
 def test_xspectra_default(fixture_localhost, generate_calc_job_node, generate_parser, data_regression, num_regression):
--- a/tests/parsers/test_xspectra/test_xspectra_default.yml
+++ b/tests/parsers/test_xspectra/test_xspectra_default.yml
@@ -4,10 +4,6 @@ parameters:
   core_level_energy_units: eV
   energy_zero: '4.1718'
   energy_zero_units: eV
-  epsilon_vector:
-  - 1.0
-  - 0.0
-  - 0.0
   estimated_ram_per_process: 4.54
   estimated_ram_per_process_units: MB
   estimated_ram_total: 72.59
@@ -18,9 +14,14 @@ parameters:
   highest_occupied_level_units: eV
   lsda: false
   lumo_found: false
-  plot_only: false
   wall_time: 5.77s
   wall_time_seconds: 5.77
+  xcoordcrys: true
+  xepsilon:
+  - 0.0
+  - 0.0
+  - 1.0
+  xonly_plot: false
 xspectra:
   labels:
   - energy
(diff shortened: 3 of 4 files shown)


src/aiida_quantumespresso/parsers/xspectra.py
```diff
@@ -87,6 +87,28 @@ def parse(self, **kwargs):
         xy_data.set_y(y_arrays, y_names, y_units)
 
         parsed_data, logs = parse_stdout_xspectra(filecontent=out_file, codename='XSpectra')
+
+        # Parse some additional info which the stdout does not reliably report
+        input_params = self.node.get_incoming().get_node_by_label('parameters').get_dict()
```
sphuber, line 92: To get an input node, `node.inputs` is easier to read and more elegant.
```suggestion
        parameters = self.node.inputs.parameters.get_attribute('INPUT_XSPECTRA')
```
Since you also only ever use the `INPUT_XSPECTRA`, might as well get that subdictionary straight away.

src/aiida_quantumespresso/parsers/xspectra.py
```diff
@@ -87,6 +87,28 @@ def parse(self, **kwargs):
...
+
+        epsilon_vector = [
+            float(input_params['INPUT_XSPECTRA'][f'xepsilon({n})']) for n in [1, 2, 3]
```
sphuber, line 95:
```suggestion
            float(parameters[f'xepsilon({n})']) for n in [1, 2, 3]
```

src/aiida_quantumespresso/parsers/xspectra.py
```diff
@@ -87,6 +87,28 @@ def parse(self, **kwargs):
...
+                parsed_data['vector_coord_system'] = 'crystal'
+            else:
+                parsed_data['vector_coord_system'] = 'cartesian'
+        else:
+            parsed_data['vector_coord_system'] = 'crystal'
+
+        if 'xonly_plot' in input_params['INPUT_XSPECTRA']:
+            parsed_data['only_plot_spectrum'] = input_params['INPUT_XSPECTRA']['xonly_plot']
+        else:
+            parsed_data['only_plot_spectrum'] = False
```
sphuber, line 110: This common pattern can be simplified with the `get(key, default)` paradigm:
```suggestion
        parsed_data['only_plot_spectrum'] = parameters.get('xonly_plot', False)
```

src/aiida_quantumespresso/parsers/xspectra.py
```diff
@@ -87,6 +87,28 @@ def parse(self, **kwargs):
...
+        ]
+        parsed_data['epsilon_vector'] = epsilon_vector
+
+        if 'xcoordcrys' in input_params['INPUT_XSPECTRA']:
+            if input_params['INPUT_XSPECTRA']['xcoordcrys']:
+                parsed_data['vector_coord_system'] = 'crystal'
+            else:
+                parsed_data['vector_coord_system'] = 'cartesian'
+        else:
+            parsed_data['vector_coord_system'] = 'crystal'
```
sphuber, line 105: This logic seems a bit weird to me, is it correct? Basically `vector_coord_system` is getting set to `crystal` if `xcoordcrys` is not set *or* `xcoordcrys` is "truthy" (i.e. not `None`, empty string etc.). Otherwise it is `cartesian`.

src/aiida_quantumespresso/parsers/xspectra.py
```diff
@@ -159,22 +181,16 @@ def parse_stdout_xspectra(filecontent, codename=None, message_map=None):
 
     logs = get_logging_container()
     parsed_data = {}
+    # We need to assign this before the loop over each line due to a bug spotted in the code
+    # when `only_plot=True`. Since the epsilon vector is reported before the value of
+    # `xonly_plot`, we need a starting value before the loop begins.
+    # parsed_data['plot_only'] = False
```
sphuber, line 187: The code seems commented out? Is the comment actually correct? Or can this be removed?

tests/parsers/test_xspectra.py
```diff
@@ -7,7 +7,17 @@
 
 def generate_inputs():
     """Return only those inputs that the parser will expect to be there."""
-    return AttributeDict()
+
+    parameters = {
+        'INPUT_XSPECTRA': {
+            'xepsilon(1)': 1,
+            'xepsilon(2)': 0,
+            'xepsilon(3)': 0,
```
sphuber, line 15: Are these keys now guaranteed to be in the parameters always? Because if not, the newly added line 95 in the parser will throw a `KeyError` if one of these is not defined.

tests/parsers/test_xspectra.py
```diff
@@ -7,7 +7,17 @@
...
+        }
+    }
+
+    inputs = {'parameters': orm.Dict(dict=parameters)}
```
sphuber, line 19:
```suggestion
    inputs = {'parameters': orm.Dict(parameters)}
```

src/aiida_quantumespresso/parsers/xspectra.py
sphuber: Should maybe consider `parameters.get(f'xepsilon({n})', default)` here to prevent `KeyError` if the key is not defined. But don't know what a sensible default could be here.

src/aiida_quantumespresso/parsers/xspectra.py
sphuber: I see in the input doc that this is a boolean. Why convert to a string? Why not simply do
```python
parsed_data['xcoordcrys'] = parameters.get('xcoordcrys', True)
```

Actually, now that I read all this. These are all input parameters defined by the user. Why do we "parse" them and add them to the output parameters? If they are already in the input, why not get them from there if needed?

PNOGillespie:
Hi @sphuber, thanks for the review. I'll add an explanation here since there are a number of comments directed at the idea of parsing of the vector coordinate system and the polarisation vectors from the inputs - particularly:

> Actually, now that I read all this. These are all input parameters defined by the user. Why do we "parse" them and add them to the output parameters? If they are already in the input, why not get them from there if needed?

The main reason for the changes proposed are for user convenience, as knowing the polarisation vector and its coordinate system used for the calculation is crucial information that (I felt, at least) should be added to the output parameters in order to make them easier to access. This is particularly useful for the polarisation vector since the Parser automatically compiles each value of `xepsilon(n)` individually into a list object. Similarly, stating in the output parameters whether the coordinate system is "crystal" or "cartesian" makes it more obvious which system was used, and stating if the calculation was a re-plot or a full calculation also helps.

Ultimately, the two options to address the issues with parsing were: (1) look in the inputs to get the parameters, to avoid the problems stated in parsing from the stdout, or (2) remove the parsing of these parameters altogether and leave it to the user to search for them manually.

Unless I've missed a trick somewhere, is there a way to access these inputs as easy as one could access the output parameter dictionary?

sphuber:
Thanks for the reply @PNOGillespie that helps a lot.

> Unless I've missed a trick somewhere, is there a way to access these inputs as easy as one could access the output parameter dictionary?

Well, if the user defines them in the input parameters, then it would be as easy as `node.inputs.parameters.get_attribute('INPUT_XSPECTRA')['xcoordcrys']`.

But as I understand now, these parameters are not required to be defined in the input parameters because a lot of them have defaults that are automatically set by `xspectra.x`. So if you think, some of these input parameters are important to have for the user, I can see why you might want to have the parser add them in the output parameters.

However, that means you have to a be a lot more careful and I would recommend the following rules:

* If an input is optional for `xspectra.x` don't assume in the parser that it will be in the input parameters (such as you have done for `xepsilon({n})`) because that will raise `KeyError`s
* I wouldn't change the name of the input. So if in `xspectra.x` it is called `xcoordcrys` I wouldn't use the `vector_coord_system` key. If you do this, you will have to start documenting all the conversions, which is a lot of work.

So to summarize I would change the parsing code to:
```python
parameters = self.node.inputs.parameters.get_attribute('INPUT_XSPECTRA')
xepsilon_defaults = {
    '1': 0,
    '2': 0,
    '3': 1,
}
parsed_data['xepsilon'] = [float(parameters.get(f'xepsilon({n})', xepsilon_defaults[n]) for n in (1, 2, 3)]
parsed_data['xcoordcrys'] = parameters.get('xcoordcrys', True)
parsed_data['xonly_plot'] = parameters.get('xonly_plot', False
```

PNOGillespie:
Thanks, I think these are reasonable suggestions for the Parser. I'll bundle these into a commit with your other suggestions and push it shortly.

> But as I understand now, these parameters are not required to be defined in the input parameters because a lot of them have defaults that are automatically set by xspectra.x. So if you think, some of these input parameters are important to have for the user, I can see why you might want to have the parser add them in the output parameters.

Each of these do have defaults, though one thing to bear in mind is that most analysis using XSpectra involves computing the spectra for different polarisation vectors - particularly in the case of crystal systems with low symmetry (e.g. Monoclinic) or ones which contain defects since the spectrum will not be invariant of polarisation vector (see past works on [SiO<sub>2</sub> & La<sub>2</sub>CuO<sub>4</sub>](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.80.075102) as well as [NiO](link.aps.org/doi/10.1103/PhysRevB.79.045118)) - so it would be reasonable to expect the user to always define this themselves. In particular, calculations for different polarisation vectors and analysing polarisation dependence is part of the `XspectraCoreWorkChain` logic and is a key feature that XSpectra has in its favour.

In any case, I'll make sure the defaults are set properly

src/aiida_quantumespresso/parsers/xspectra.py
```diff
@@ -87,6 +87,19 @@ def parse(self, **kwargs):
         xy_data.set_y(y_arrays, y_names, y_units)
 
         parsed_data, logs = parse_stdout_xspectra(filecontent=out_file, codename='XSpectra')
+
+        # Parse some additional info which the stdout does not reliably report
+        parameters = self.node.inputs.parameters.get_attribute('INPUT_XSPECTRA', None)
```
sphuber, line 92:
```suggestion
        parameters = self.node.inputs.parameters.get_attribute('INPUT_XSPECTRA', {})
```

sphuber (changes requested):
Thanks a lot @PNOGillespie , just one final problem that would actually cause exceptions. Once you fix this, I will merge.

I just want to still note that with this new method of parsing the parameters from the inputs and resorting to default if not defined, it means that the parser can set the wrong defaults if the defaults in `xspectra.x` change. For example if QE 7.0 changes default of `xcoordcrys` to False, then the Parser would still set `True` if not explicitly defined by the user. Again, not sure how likely the defaults are subject to change, but something to be aware of.

PNOGillespie:
Thanks again @sphuber.

> I just want to still note that with this new method of parsing the parameters from the inputs and resorting to default if not defined, it means that the parser can set the wrong defaults if the defaults in xspectra.x change. For example if QE 7.0 changes default of xcoordcrys to False, then the Parser would still set True if not explicitly defined by the user. Again, not sure how likely the defaults are subject to change, but something to be aware of.

I understand, it's something I bear in mind while I'm working on the implementation. If it's any comfort, development work on `xspectra.x` itself seems to have stopped quite a while ago (which maybe explains why it doesn't use an xml file as well) so it's likely that the FORTRAN code is frozen in its current state and won't change for some time yet.

sphuber (dismissed):

sphuber (approved):
Thanks a lot @PNOGillespie

sphuber (approved):

PR #854, opened 2022-11-04
Add feature: `XspectraBaseWorkChain`

PNOGillespie:
In this PR we propose a `BaseRestartWorkChain` implementation for `xspectra.x` calculations to handle automatic restarts and automate some processes for setting `XspectraCalculation` inputs:
* Uses an input `StructureData` node in both the main WorkChain and its `get_builder_from_protocol()` method to generate the kpoints mesh in the former and set the parameter `xiabs` in the `XspectraCalculation` input parameters in the latter.
* Passes the `core_wavefunction_data` required for the `xspectra.x` calculation via the given shortcut.

Includes three pre-written protocols:
* `one_shot` (default): inputs for an `XspectraCalculation` using a moderate broadening parameter, intended to be used in most cases.
* `production`: inputs for an `XspectraCalculation` with a high broadening parameter and minimal spectra points, intended for faster computation of the Lanczos with the aim to re-plot the spectrum in a follow-up process.
* `re_plot`: inputs for an `XspectraCalculation` with a low broadening parameter and instructions to only re-plot from computed Lanczos coefficients, intended for re-plotting already-finished `xspectra.x` calculations


src/aiida_quantumespresso/workflows/xspectra/base.py
```diff
@@ -0,0 +1,277 @@
...
+        kinds_present = sorted([Kind.name for Kind in structure.kinds])
+        if abs_atom_marker not in kinds_present:
+            raise AttributeError(f'given abs_atom_marker ({abs_atom_marker}) does not match a Kind in the structure.')
+        else:
+            parameters['INPUT_XSPECTRA']['xiabs'] = kinds_present.index(abs_atom_marker) + 1
+
+        # pylint: disable=no-member
+        builder = cls.get_builder()
+        builder.xspectra['code'] = code
+        if parent_folder is not None:
```
sphuber, line 146: Should the `parent_folder` be optional? It should always require one, no?

sphuber (changes requested):
Thanks @PNOGillespie . Looking very good, just have some minor questions and comments

src/aiida_quantumespresso/workflows/xspectra/base.py
PNOGillespie: This was my initial thought, however the `get_builder_from_protocol()` tests then run into a problem because there's no method to serialize the `parent_folder` input when serializing the builder using the fixture in `conftest.py`:

      @pytest.fixture
      def serialize_builder():
          """Serialize the given process builder into a dictionary with nodes turned into their value representation.
          :param builder: the process builder to serialize
          :return: dictionary
          """
          def serialize_data(data):
          # pylint: disable=too-many-return-statements
            from aiida.orm import BaseType, Code, Dict
            from aiida.plugins import DataFactory

            StructureData = DataFactory('core.structure')
            UpfData = DataFactory('pseudo.upf')
    
            if isinstance(data, dict):
                return {key: serialize_data(value) for key, value in data.items()}
    
            if isinstance(data, BaseType):
                return data.value
    
            if isinstance(data, Code):
                return data.full_label
    
            if isinstance(data, Dict):
                return data.get_dict()
    
            if isinstance(data, StructureData):
                return data.get_formula()
    
            if isinstance(data, UpfData):
                return f'{data.element}<md5={data.md5}>'
    
            return data
    
        def _serialize_builder(builder):
            return serialize_data(builder._inputs(prune=True))  # pylint: disable=protected-access

The only options I see here are to either leave the `parent_folder` input as it is, or figure out some standard method to serialize the `RemoteData` input such that the tests can parse that information correctly. I guess this depends on what would be preferable in your opinion.

tests/conftest.py
```diff
@@ -308,6 +312,48 @@ def _generate_upf_data(element):
     return _generate_upf_data
 
 
+@pytest.fixture(scope='session')
+def generate_singlefile_data():
```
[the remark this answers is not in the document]
sphuber: Well but it is not generic is it? Because it will always return a `SinglefileData` with content that is specifically that of a core_wfc file? Why would any other test use it that would require other content? But even more so, creating a `SinglefileData` is as easy as `SinglefileData(io.StringIO('some content'))` so do we really even need a fixture for this?

src/aiida_quantumespresso/workflows/protocols/xspectra/base.yaml
```diff
@@ -0,0 +1,46 @@
...
+                cut_occ_states: True
+            CUT_OCC:
+                cut_desmooth: 0.1
+        metadata:
+            options:
+                resources:
+                    num_machines: 1
+                max_wallclock_seconds: 43200  # Twelve hours
+                withmpi: True
+default_protocol: one_shot
```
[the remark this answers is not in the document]
PNOGillespie: If `moderate` refers to, say, a "general-purpose" method suitable for any given system, then I think that would be reasonable to change `one_shot` to `moderate`. `production` in this case would be better defined as `fast`, since the aim is to speed up the Lanczos recursion and reduce the amount of time spent plotting the spectrum at the end - as evidenced by the fact that `production` only plots 5 points in total - which works well if it's being used for testing purposes or to re-plot the spectrum in a follow-up calculation. If we're sticking to the convention of `fast`, `moderate`, and `precise` then I could maybe add a `precise` method with tighter convergence criteria and shorter `kpoints_distance` parameter, if that would be welcomed? Keeping the `re_plot` protocol is a useful shortcut in any case, so I think it would be best to keep this as an additional protocol.

src/aiida_quantumespresso/workflows/xspectra/base.py
sphuber: The `serialize_builder` is just used in our own tests to have an easy way to compare builders. The solution is just to add a fallback case for any `Data` nodes that are not handled explicitly. Just add the following below the conditional for `UpfData`:
```python
if isinstance(data, Data):
    return f'{data.__class__.__name__}<{data.uuid}>'
```
That should solve the issue

PNOGillespie: Ok, I'll add it to the list of revisions in that case.

src/aiida_quantumespresso/workflows/protocols/xspectra/base.yaml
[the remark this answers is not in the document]
sphuber: Yeah, the idea is that `moderate` is the default and should give a calculation with reasonable precision while not being too expensive, a good middle ground. The `fast` protocol should be faster, and should typically be used for testing or if precision is not that important. The `precise` option is reserved for when you need really precise calculations for production, such as actual research. I agree that keeping `re_plot` as an additional protocol is useful.

src/aiida_quantumespresso/workflows/xspectra/base.py
sphuber: Very good point, I missed that. However, I am not sure that your proposed solution would work either, because I think the `get_remote_path` of the `RemoteData` may also be variable between tests. That being said, I think there still is a solution. What we are after is a value that "represents" the contents of the node, so that we can easily compare it. Luckily we already have this in `aiida-core` as there is functionality for the caching feature that computes the hash of a node, which is representative for its contents. You can call `data.get_hash` to get it and use that. Note that this will raise if the node is not stored, so maybe it is better to use `data._get_hash()` which bypassed the protection. You can add the `# pylint: disable=protected-access` to disable the pylint warning in this case.

PNOGillespie: Ok, I tried this approach and it works fine for everything apart from the `RemoteData` input, as the hash of the node changes with each test. I have tried to set the `remote_path` of the node explicitly, but the tests return a different hash each time. To make clear, this is the code that I am trying to run:

      @pytest.fixture
      def generate_xspectra_calc_job_node(generate_calc_job_node, generate_structure, fixture_localhost, fixture_sandbox, generate_inputs_xspectra):
          """Generate a ``CalcJobNode`` that would have been created by an ``XspectraCalculation``."""
          from aiida.orm import RemoteData, FolderData, Dict
      
          def _generate_xspectra_calc_job_node():
              node = generate_calc_job_node(
                  entry_point_name='quantumespresso.xspectra',
                  inputs=generate_inputs_xspectra()
              )
      
              remote_folder = RemoteData(computer=fixture_localhost, remote_path='/tmp')
              remote_folder.base.links.add_incoming(node, link_type=LinkType.CREATE, link_label='remote_folder')
              remote_folder.store()
       
              return node
      
          return _generate_xspectra_calc_job_node
      
      @pytest.fixture
      def get_xspectra_generator_inputs(fixture_code, fixture_localhost, generate_xspectra_calc_job_node, generate_inputs_xspectra):
          """Generate a set of default inputs for the ``XspectraBaseWorkChain.get_builder_from_protocol()`` method."""
          from aiida.orm import RemoteData
          SinglefileData = DataFactory('core.singlefile')
          
          parent_calc = generate_xspectra_calc_job_node()
      
          return {
              'code': fixture_code('quantumespresso.xspectra'),
              'parent_folder' : parent_calc.outputs.remote_folder,
              'core_wfc_data': SinglefileData(io.StringIO(
                  '# number of core states 3 =  1 0;  2 0;'
                  '\n6.51344e-05 6.615743462459999e-3'
                  '\n6.59537e-05 6.698882211449999e-3'
              )),
              'abs_atom_marker': 'Si',
          }
      
I have tried this by also using the `SandboxFolder` fixture, but the same problem arises. Is there some other property of `RemoteData` or generic `Data` nodes that could be exploited instead? or would it be better to try and avoid serializing the `parent_folder` instead?

PNOGillespie: Very much appreciated @sphuber, I'll have this done soon.

PNOGillespie:
Hi @sphuber. Given the short-term fix for the tests made recently in #861, would it be acceptable to do something similar here? The only thing that needs to be done is just to change the serialization of the `core_wfc_data` in the test, since the `kpoints` input has been handled already in efafe54a1654434f04ba189715366d3a2424e363, for which I could just serialize the contents of the node with `node.get_content()`

sphuber:
> Hi @sphuber. Given the short-term fix for the tests made recently in #861, would it be acceptable to do something similar here? The only thing that needs to be done is just to change the serialization of the `core_wfc_data` in the test, since the `kpoints` input has been handled already in [efafe54](https://github.com/aiidateam/aiida-quantumespresso/commit/efafe54a1654434f04ba189715366d3a2424e363), for which I could just serialize the contents of the node with `node.get_content()`

Yes, if you need to add another case for another Data plugin, go ahead and add it to the `serialize_builder` fixture.

src/aiida_quantumespresso/workflows/xspectra/base.py
```diff
@@ -0,0 +1,263 @@
...
+        if options:
+            metadata['options'] = recursive_merge(inputs['xspectra']['metadata']['options'], options)
+
+        parent_calc = parent_folder.creator
+        if parent_calc.process_type == 'aiida.calculations:quantumespresso.xspectra':
+            inputs['xspectra']['kpoints'] = parent_calc.inputs.kpoints
+            xiabs = parent_calc.inputs.parameters.get_dict()['INPUT_XSPECTRA'].get('xiabs', 1)
+            parameters['INPUT_XSPECTRA']['xiabs'] = xiabs
+        elif parent_calc.process_type == 'aiida.calculations:quantumespresso.pw':
+            structure = parent_folder.creator.inputs.structure
```
sphuber, line 118:
```suggestion
            structure = parent_calc.inputs.structure
```

src/aiida_quantumespresso/workflows/xspectra/base.py
```diff
@@ -0,0 +1,263 @@
...
+            kinds_present = sorted([Kind.name for Kind in structure.kinds])
+            if abs_atom_marker not in kinds_present:
+                raise AttributeError(
+                    f'given abs_atom_marker ({abs_atom_marker}) does not match a Kind in the structure.'
+                )
+            else:
+                parameters['INPUT_XSPECTRA']['xiabs'] = kinds_present.index(abs_atom_marker) + 1
+        else:
+            raise AttributeError(f'Parent calculation type ({parent_calc.process_type}) is not  suitable.')
```
sphuber, line 127:
```suggestion
            raise ValueError(f'process type of `parent_folder` creator `{parent_calc.process_type}` is not supported.')
```

src/aiida_quantumespresso/workflows/xspectra/base.py
```diff
@@ -0,0 +1,263 @@
...
+
+        # pylint: disable=no-member
+        builder = cls.get_builder()
+        builder.xspectra['code'] = code
+        builder.xspectra['parent_folder'] = parent_folder
+        builder.xspectra['core_wfc_data'] = core_wfc_data
+        builder.xspectra['parameters'] = orm.Dict(inputs['xspectra']['parameters'])
```
sphuber, line 134:
```suggestion
        builder.xspectra['parameters'] = orm.Dict(parameters)
```

src/aiida_quantumespresso/workflows/xspectra/base.py
```diff
@@ -0,0 +1,263 @@
...
+        builder.xspectra['metadata'] = metadata
+        if 'settings' in inputs['xspectra']:
+            builder.xspectra['settings'] = orm.Dict(inputs['xspectra']['settings'])
+        builder.clean_workdir = orm.Bool(inputs['clean_workdir'])
+        if 'kpoints' in inputs['xspectra']:
+            builder.kpoints = parent_calc.inputs.kpoints
```
sphuber, line 140: Is this logic correct? Should it not be
```python
builder.kpoints = inputs['xspectra']['kpoints']
```
otherwise you are ignoring `inputs['xspectra']['kpoints']`

sphuber (changes requested):
Thanks @PNOGillespie , just some final suggestions and questions about the `get_builder_from_protocol`. After that it is good to go

src/aiida_quantumespresso/workflows/xspectra/base.py
PNOGillespie: Yes, this actually has been superseded by the changes made previously, but I forgot to remove it. As the check of the `parent_calc` now sets the logic for the `kpoints` input (use parent mesh if `XspectraCalculation` or use structure to generate a new mesh if `PwCalculation`), I think it would make sense to simply merge these two parts together:

        if parent_calc.process_type == 'aiida.calculations:quantumespresso.xspectra':
            inputs['xspectra']['kpoints'] = parent_calc.inputs.kpoints
            xiabs = parent_calc.inputs.parameters.get_dict()['INPUT_XSPECTRA'].get('xiabs', 1)
            parameters['INPUT_XSPECTRA']['xiabs'] = xiabs
        elif parent_calc.process_type == 'aiida.calculations:quantumespresso.pw':
            structure = parent_folder.creator.inputs.structure
            kinds_present = sorted([Kind.name for Kind in structure.kinds])
            if 'kpoints' in inputs['xspectra']:
                builder.kpoints = parent_calc.inputs.kpoints
            else:
                builder.kpoints_distance = orm.Float(inputs['kpoints_distance'])
                builder.kpoints_force_parity = orm.Bool(inputs['kpoints_force_parity'])

sphuber: If it makes sense to have a different mesh for the `xspectra.x` code then is used for `pw.x` that makes sense to me.

PNOGillespie: It is the case that some users might still want to (for instance) use a denser mesh for the `xspectra.x` calculation than the `pw.x` calculation - particularly some of the published works and tutorial examples that I've seen do this as well - so allowing this does indeed make sense.

src/aiida_quantumespresso/workflows/xspectra/base.py
```diff
@@ -0,0 +1,263 @@
...
+        parameters = inputs['xspectra']['parameters']
+
+        if options:
+            metadata['options'] = recursive_merge(inputs['xspectra']['metadata']['options'], options)
+
+        parent_calc = parent_folder.creator
+        if parent_calc.process_type == 'aiida.calculations:quantumespresso.xspectra':
+            inputs['xspectra']['kpoints'] = parent_calc.inputs.kpoints
+            xiabs = parent_calc.inputs.parameters.get_dict()['INPUT_XSPECTRA'].get('xiabs', 1)
+            parameters['INPUT_XSPECTRA']['xiabs'] = xiabs
```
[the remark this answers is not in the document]
PNOGillespie: Originally, I just wanted to ensure that the parameter was passed automatically from parent calculation to new calculation, and the form used there was simpler than checking if the parameter was set in the first place. I can remove it without causing any problems, if that would be preferable.

PNOGillespie:
Hi @sphuber, apologies for the slow response. I have added the changes to `get_builder_from_protocol()` that you suggested. Regarding the previous comment about this line:

            xiabs = parent_calc.inputs.parameters.get_dict()['INPUT_XSPECTRA'].get('xiabs', 1)
            parameters['INPUT_XSPECTRA']['xiabs'] = xiabs

> You are essentially just setting the value to 1 if not already defined. But 1 is the default set by xspectra.x itself. Is there any particular reason you are doing this?

I decided to just remove the lines from that section since, on closer inspection, they don't do anything useful.

Let me know if there are any other changes needed here.

sphuber (approved):
Thanks a lot @PNOGillespie . Fantastic work as always :+1:

Issue #859, opened 2022-11-11
The `serialize_builder` fixture is susceptible to version changes in `aiida-core` causing the `nightly` build to fail


sphuber:
In #858 the `serialize_builder` fixture was updated to add support for arbitrary `Data` plugins by using their hash as a fallback. However, this hash includes the version of the plugin package and so when that changes, for example when `aiida-core` releases a new version, the hashes of its data types become invalid.

PR #860, opened 2022-11-14
Fix bug in `wrap_bare_dict_inputs`

mbercx:
From `aiida-core==2.1.0`, the `valid_type` of an input is extended on the fly with `None` in case the port is not required. This causes an issue in the `wrap_bare_dict_inputs` function, which checks if the `valid_type` is _equal_ to `Dict`. Here we adapt the function to check if `Dict` is one of the valid types after converting the valid types into a tuple if necessary.

### DIFF
--- a/src/aiida_quantumespresso/utils/mapping.py
+++ b/src/aiida_quantumespresso/utils/mapping.py
@@ -88,10 +88,11 @@ def wrap_bare_dict_inputs(port_namespace, inputs):
             continue
 
         port = port_namespace[key]
+        valid_types = port.valid_type if isinstance(port.valid_type, (list, tuple)) else (port.valid_type,)
 
         if isinstance(port, PortNamespace):
             wrapped[key] = wrap_bare_dict_inputs(port, value)
-        elif port.valid_type == Dict and isinstance(value, dict):
+        elif Dict in valid_types and isinstance(value, dict):
             wrapped[key] = Dict(value)
         else:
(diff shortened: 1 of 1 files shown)


mbercx:
@sphuber I didn't add a test, since we plan to remove these methods in favour of automated serialisation in the future.

mbercx:
Tests are failing because the hash of the `KpointsData` is dependent on the version of `aiida-core`:

```python
In [3]: kpoints = orm.KpointsData()
   ...: kpoints.set_kpoints_mesh([2, 2, 2])
   ...: 
   ...: kpoints.base.caching._get_objects_to_hash()
Out[3]: 
['2.1.1',
 {'mesh': [2, 2, 2], 'offset': [0.0, 0.0, 0.0]},
 '076c4a75b0e7f3de89d0f21d3c4fad9c09fb83054f8e6c467b875095d93fd154',
 None]
```

Opened https://github.com/aiidateam/aiida-quantumespresso/pull/861 to add serialization for `KpointsData` to the `serialize_builder` fixture.

sphuber:
> Tests are failing because the hash of the `KpointsData` is dependent on the version of `aiida-core`:

Indeed, had opened #859 for this. We should still try to find a way to have a fallback serialization for all nodes that is independent of the version. But for now as a quick fix this will work.

sphuber (approved):
Thanks for the fix @mbercx

PR #861, opened 2022-11-14
Add `KpointsData` serialization to `serialize_builder` fixture

mbercx:


sphuber (approved):
Thanks for the temporary fix until we find a long-term solution for #859

PR #866, opened 2022-11-18
`BasePwCpInputGenerator`: do not require pseudos in `validate_inputs`

### LINKED ISSUE #865: The validator of `BasePwCpInputGenerator` should not fail if no pseudos are provided
Just like for the `structure`, a superclass may exclude the port from being exposed.

sphuber:
Fixes #865 

The `validate_inputs` of the `BasePwCpInputGenerator` base `CalcJob` validates that the `structure` and `pseudos` inputs are commnesurate. However, wrapping processes may decide to exclude these inputs when exposing the base class in which case they will not be present when the validator is called. The validator should not return an error message in this case as the burden of validation rests on the shoulders of the wrapping process.

This check was already applied for the `structure` input, but is now extended to the `pseudos` as well.

### DIFF
--- a/src/aiida_quantumespresso/calculations/__init__.py
+++ b/src/aiida_quantumespresso/calculations/__init__.py
@@ -141,14 +141,19 @@ class BasePwCpInputGenerator(CalcJob):
         spec.inputs.validator = cls.validate_inputs
 
     @classmethod
-    def validate_inputs(cls, value, _):
+    def validate_inputs(cls, value, port_namespace):
         """Validate the entire inputs namespace."""
-        if 'pseudos' not in value:
-            return 'required value was not provided for the `pseudos` namespace.'
 
-        # Some wrapping processes expose the inputs excluding the ``structure``, in which case we can't validate.
-        if 'structure' not in value:
+        # Wrapping processes may choose to exclude certain input ports in which case we can't validate. If the ports
+        # have been excluded, and so are no longer part of the ``port_namespace``, skip the validation.
+        if any(key not in port_namespace for key in ('pseudos', 'structure')):
             return
+
+        # At this point, both ports are part of the namespace, and both are required so return an error message if any
+        # of the two is missing.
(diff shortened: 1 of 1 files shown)


unkcpz (approved):
@sphuber All good and thanks a lot for debugging together.
Just second thought, instead of returning non-empty strings to fail validation, can we just raise exceptions for all the port validation? At least for this case, I think the trace will point to where the error happened which is more straightforward to find debug.

sphuber:
>  Just second thought, instead of returning non-empty strings to fail validation, can we just raise exceptions for all the port validation? At least for this case, I think the trace will point to where the error happened which is more straightforward to debug.

Not sure this is possible as this would break everything. This has been the design in `plumpy` for a very long time. I can see the point though. Should maybe ask @muhrin why he decided to go for returning strings instead of raising exceptions.

sphuber:
>  I think the failed test is related to the change, says the "pseudos" not in the AttributesDict.

And for a good reason. I put in that check to make sure the calculation cannot be called without `pseudos`, which is sensible.

My original changes in this PR were actually wrong. In fact, we shouldn't skip the validation if the `pseudos` or `structure` are not in the *value* (which are the inputs), but if they are no longer in the *port namespace*. Now, one would think that we could call `cls.spec().inputs` to get the spec and check, but that would always get you the spec of the class itself and not the exposed spec, so this wouldn't work. But luckily, the actual port namespace is passed into the validator as the third argument. Actually it is not luck, I added this way back when as the solution to this very problem :sweat_smile: I had just forgotten. The correct solution then is to check this argument to see if the ports are still present. If not, the validator is being called in a parent workchain that has apparently excluded the ports, so we should skip validation. And all is well in the world :smile: :+1:

unkcpz:
@sphuber thanks! I test with my sssp plugin and all works fine.

unkcpz (approved):

PR #867, opened 2022-11-18
`PwBaseWorkChain`: Set `restart_mode` in parameters if `parent_folder`

### LINKED ISSUE #864: `PwBaseWorkChain`: does not automatically set restart mode in `parameters` when `parent_folder` is specified
I have the feeling that at some point this functionality was there, that before launching the first `PwCalculation` if a `parent_folder` was defined and there was no `parameters.CONTROL.restart_mode` then the workchain would set this automatically. Maybe this got lost in the recent refactoring with the `set_restart_type` method? @mbercx ?

sphuber:
Fixes #864 

Typically, if a user launches a `PwBaseWorkChain` with a `parent_folder` they want to restart from a previously completed but unconverged run. In this case the `CONTROL.restart_mode` parameter should be set to `from_scratch`. This used to be done automatically but was removed in commit cb32be5361afa05bad617f00f8b187c96eb365ec.

Here we reinstate the behavior by calling `set_restart_type` to `RestartType.FULL` if the `parent_folder` is defined and the `CONTROL.restart_mode` hasn't already been set explicitly in the parameters to `from_scratch`. In this case, it was likely set by the caller and we don't want to silently override this.


mbercx (approved):
All good, thanks @sphuber!

PR #870, opened 2022-12-14
Fix nightly build tests

mbercx:
Due to the changes in https://github.com/aiidateam/aiida-core/commit/fda91e2e79bc0c84f43b83dc22894ad4cc77cd36, the `aiida_localhost` fixture now has the label `localhost` instead of `localhost-test`. Here we simply update the pytest data regression files to reflect this change.


sphuber (approved):
Thanks @mbercx for the quick fix

PR #872, opened 2022-12-16
Add feature: `XspectraCoreWorkChain`

PNOGillespie:
# Overview
In this PR, we propose to implement a `CoreWorkChain` for XSpectra. This WorkChain will take a given input structure in which one atomic site is designated as the site for the absorbing atom (see inputs for more info) and run all XSpectra calculations requested. The aim is to provide a generic workflow process to cover the necessary parts for computing the XANES spectrum for a particular absorbing atom in a structure, which can then be used either as a standard sub-process in other WorkChains or self-standing as a means to simply automate analysis with XSpectra for a given structure. This WorkChain makes use of as many features already implemented in `aiida-quantumespresso` as possible, however it does require the [AiiDA-Shell plugin](https://github.com/sphuber/aiida-shell/) for `run_upf2plotcore` and thus will add it as a dependency.

# Process Flow

1.  Perform an SCF calculation using the PwBaseWorkChain for the structure with a core-hole present.
2.  (Optionally) run the upf2plotcore.sh utility script on the ground-state pseuopotential assigned to the absorbing elemental species (i.e. the non-absorbing-atoms of the same element type as the absorbing atom). This is only called if core_wfc_data is not provided as an input. Requires AiiDA-Shell and a code node for the script using the plugin in order to run.
3.  Perform an XSpectra calculation using the XspectraBaseWorkChain for each requested polarisation vector (see eps_vectors in the inputs).
4.  (Optionally) Perform a second XSpectra calculation to re-plot the spectra from the XSpectra save files produced in step 3 - useful if the user intends to re-plot the final spectrum using a different degree of broadening.
5.  Collate all output data: return the output_parameters of the SCF calculation and all XSpectra calculations from step 3, return all final spectra from step 3 (or step 4, if `inputs.run_replot = True`) in a single XyData node for easier plotting, and produce a representation of the XANES of a powder sample of the structure (if requested).

# Inputs
## Important Inputs
Important inputs to take notice of are:

1.  `structure` - the input StructureData node for the WorkChain. The structure must contain at least one unique Kind which represents the absorbing atom and appears at one Site in the structure. The name of the absorbing atom Kind must match the abs_atom_marker in the inputs.
2.  `abs_atom_marker` - a Str node designating the name of the Kind to be used as the absorbing atom. Defaults to "X".
3.  `core_wfc_data` - a SinglefileData node containing the core wavefunction of the ground-state of the absorbing element, obtained by parsing the ground-state pseudopotential for the absorbing element with upf2plotcore.sh. Since this is normally obtained by AiiDA-Shell and thus (as far as I remember) can only be the stdout, the filename of this must be "stdout".
4.  `eps_vectors` - a List node containing all the 3-vectors to be computed for `xepsilon(1-3)` in the `xspectra.x` calculation, thus also defining the number of XSpectra sub-processes to launch. Defaults to [[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]].
5.  `get_powder_spectrum` - a Bool node which requests if the powder spectrum of the structure should be computed. Defaults to False. The CalcFunction associated to this will operate differently depending on the number of given spectra (from 1 to 3) using symmetry rules which, since no symmetry information can be obtained by the WorkChain in advance, the WorkChain will filter out the correct spectra to use. The process will be skipped with a warning in the logs if the spectra required to produce the powder spectrum are not present.
6. `run_replot` - a Bool node which, if `True`, requests a second `xs_plot` step to re-plot the final spectrum produced by `xs_prod`. Set to `False` by default. Uses the `replot` protocol within the `XspectraBaseWorkChain` to generate inputs when using `get_builder_from_protocol()`.

## Core-Hole Treatments
The WorkChain contains a set of pre-defined input settings for the SCF step which correspond to different core-hole approximations (referred to as `treatment`s in the code) needed to handle the presence of the excited electron in the calculation. The method employed here in the code is a copy of the YAML-format system used to define and retrieve protocol inputs and uses the following presets:

* `full` (default) - "full-core-hole" (FCH) method, in which the excited electron is balanced out by a simple +1 counter charge. This approximation is acceptable for most elements and avoids the need for a spin-polarised simulation of the excited state.
* `half` - "half-core-hole" (HCH) method, which works the same as `full` but uses a +0.5 charge - requires a core-hole pseudopotential with a core-hole of half an electron (e.g. 1s1.5). Useful for some core-holes (e.g. carbon) where the FCH method works poorly
* `xch_fixed` & `xch_smear` - "excited-core-hole" (XCH) method (for systems of fixed and smeared populations, respectively), in which the excited electron is included explicitly in the system - typically more costly as this makes the calculation spin-polarised but accurate for any element.
* `none` - no core-hole, which simply counteracts the default value and provides no core-hole approximation - only usable if there is no core-hole in the simulation.

The settings are applied to `scf.pw.parameters` during the `get_builder_from_protocol` method and acts after the `protocol` defaults are obtained but before the `overrides` are applied - thus ensuring that the overrides are always applied as a final step regardless.


src/aiida_quantumespresso/workflows/xspectra/core.py
```diff
@@ -0,0 +1,910 @@
...
+from aiida_quantumespresso.workflows.protocols.utils import ProtocolMixin, recursive_merge
+
+PwCalculation = CalculationFactory('quantumespresso.pw')
+PwBaseWorkChain = WorkflowFactory('quantumespresso.pw.base')
+XspectraBaseWorkChain = WorkflowFactory('quantumespresso.xspectra.base')
+XyData = DataFactory('array.xy')
+
+
+@calcfunction
+def get_all_spectra(**kwargs):
```
sphuber, line 26: Could you please move each of these calcfunctions to a separate file in the directory `aiida_quantumespresso/calculations/functions/xspectra`. You need to create the directory first.

sphuber: Also would a slightly more specific name be useful? Maybe something like `merge_spectra`?

src/aiida_quantumespresso/workflows/xspectra/core.py
```diff
@@ -0,0 +1,910 @@
...
+    """Compile all calculated spectra into a single XyData node for easier plotting.
```
sphuber, line 27: Can you please describe in the docstring what nodes will be accepted/expected in `**kwargs`. For example, should be `XyData` nodes that are the `output_spectra` of an `XspectraCalculation`. Not sure if these labels are correct, but just an example.

src/aiida_quantumespresso/workflows/xspectra/core.py
```diff
@@ -0,0 +1,910 @@
...
+    This will take all the data for each spectrum produced during the re-plot step and output
+    them into a single XyData node.
+    """
+
+    output_spectra = XyData()
+    y_arrays_list = []
+    y_units_list = []
+    y_labels_list = []
+
+    spectra = [node for label, node in kwargs.items() if isinstance(node, orm.Data)]
```
sphuber, line 38: Should this be a more specific test? I think you only want `XyData` nodes right?

src/aiida_quantumespresso/workflows/xspectra/core.py
```diff
@@ -0,0 +1,910 @@
...
+    a second xspectra.x calculation which replots the spectra produced from the ``xs_prod``
+    step. This option can be very useful for obtaining a final spectrum at low levels of
+    broadening (relative to the default of 0.5 eV), particularly as higher levels of
+    broadening significantly speed up the convergence of the Lanczos procedure. Inputs for the
+    replot calculation are found in the ``xs_plot`` namespace.
+
+    The the core-wavefunction plot derived from the ground-state of the absorbing element can
+    be provided as a top-level input or produced by the WorkChain. If left to the WorkChain,
+    the ground-state pseudopotential assigned to the absorbing element will be used to
+    generate this data using the upf2plotcore.sh utility script (via the ``AiiDA-Shell``
```
sphuber, line 211:
```suggestion
    generate this data using the upf2plotcore.sh utility script (via the ``aiida-shell``
```

src/aiida_quantumespresso/workflows/xspectra/core.py
```diff
@@ -0,0 +1,910 @@
...
+            self.report(
+                f'launching XspectraWorkChain<{future_xspectra.pk}> for epsilon vector {vector}'
+                ' (Lanczos production)'
+            )
+            xspectra_prod_calcs[f'{label}_prod'] = future_xspectra
+            calc_number += 1
+
+        if 'xspectra_calc_labels' not in self.ctx.keys():
+            self.ctx.xspectra_calc_labels = xspectra_calc_labels
+        return ToContext(**xspectra_prod_calcs)
```
sphuber, line 706: Instead of manually creating labels and adding them to the context as this, if you just need a list, you can use the following in the for-loop as well
```python
from aiida.engine import append_

for index, vector in enumerate(eps_vectors):
    ...
    future_xspectra = self.submit(XspectraBaseWorkChain, **xspectra_inputs)
    self.to_context(xspectra_calculations=append_(future_xspectra)
```
This will automatically create a list in the `ctx` under the key `xspectra_calculations` to which the xspectra nodes are appended. You can then use this in `inspect_all_xspectra_prod` as follows:
```python
for xspectra_calculation in self.ctx.xspectra_calculations`
```
Think that makes for a lot more readable code

src/aiida_quantumespresso/workflows/xspectra/core.py
```diff
@@ -0,0 +1,910 @@
...
+
+        calculations = []
+        labels = self.ctx.xspectra_calc_labels
+        for label in labels:
+            calculation = self.ctx[f'{label}_plot']
+            calculations.append(calculation)
+        unrecoverable_failures = False
+
+        finished_replots = []
+        for calculation in calculations:
```
sphuber, line 789: Same here. Instead of manually having to retrieve a list of calculations through the labels, simply use the `append_` method to directly create a list of nodes in the context.

src/aiida_quantumespresso/workflows/xspectra/core.py
```diff
@@ -0,0 +1,910 @@
...
+            valid_type=orm.Dict,
+            help='The output dictionaries of each `XspectraBaseWorkChain` performed',
+            dynamic=True
+        )
+        spec.output(
+            'output_spectra',
+            valid_type=orm.XyData,
+            help='An XyData node containing all the final spectra produced by the WorkChain.'
+        )
+        spec.output('powder_spectrum', valid_type=orm.XyData, required=False, help='The simulated powder spectrum')
```
sphuber, line 368: Let's just get rid of the `output_` prefix for all the outputs. We still have this in some other workchains for historical reasons that we cannot change because of backwards compatibility, but it is redundant. If you do `node.outputs.xspectra` it is clear that is an output and you don't need `node.outputs.output_xspectra` to tell you.

sphuber (changes requested):
Thanks @PNOGillespie , great contribution. I have gone through most of the code but not all the details, but I trust this to be ok or we can fix it later. Just have some comments for now for slight improvements.

src/aiida_quantumespresso/workflows/protocols/core_hole_treatments.yaml
```diff
@@ -0,0 +1,27 @@
+default_inputs:
```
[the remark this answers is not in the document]
PNOGillespie: I think the plan here is to provide this for other classes/functions to access, notably the in-progress WorkChain for XPS (#879), though admittedly this is only meaningful for core-level spectroscopy calculations so it would make sense for it to go under `xspectra`

@superstar54, do you agree with all that?

src/aiida_quantumespresso/workflows/xspectra/core.py
PNOGillespie: Something like this, potentially?

    """Compile all calculated spectra into a single XyData node for easier plotting.

    The keyword arguments must be an arbitrary number of XyData nodes from 
    the `output_spectra` of `XspectraCalculation`s, all other `kwargs` will be discarded at 
    runtime.
    
    Returns a single XyData node where each set of y values is labelled
    according to the polarisation vector used for the `XspectraCalculation`.
    """

src/aiida_quantumespresso/workflows/protocols/core_hole_treatments.yaml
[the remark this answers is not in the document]
superstar54: Since it will also be accessed by XPS (maybe other core-level spectroscopy methods in the future), I think we can keep it here instead of moving to xspectra.

PNOGillespie:
Hi @sphuber. I addressed the points that you mentioned. Regarding [this converstation](https://github.com/aiidateam/aiida-quantumespresso/pull/872#discussion_r1081283999), I decided to leave `core_hole_treatments.yaml` where it is for now, given @superstar54's comment regarding the intended XPS WorkChain, but everything else has been addressed.

Let me know if there is anything else that needs changing.

tests/workflows/xspectra/test_core.py
```diff
@@ -0,0 +1,161 @@
...
+import io
+
+from aiida import engine, orm
+from aiida.common import LinkType
+from aiida.engine.utils import instantiate_process
+from aiida.manage.manager import get_manager
+from plumpy import ProcessState
+import pytest
+
+from aiida_quantumespresso.calculations.helpers import pw_input_helper
```
sphuber, line 13:
```suggestion
```

sphuber (changes requested):
Thanks a lot @PNOGillespie . Just a couple final minor touches and this can be merged.

tests/workflows/xspectra/test_core.py
```diff
@@ -0,0 +1,151 @@
...
+from aiida.manage.manager import get_manager
+from plumpy import ProcessState
+import pytest
+
+
+def instantiate_process(cls_or_builder, inputs=None):
+    """Instantiate a process, from a ``Process`` class or ``ProcessBuilder`` instance."""
+    manager = get_manager()
+    runner = manager.get_runner()
+    return instantiate_process(runner, cls_or_builder **{inputs or {})
```
sphuber, line 18:
```suggestion
from aiida.manage.manager import get_manager
from plumpy import ProcessState
import pytest


def instantiate_process(cls_or_builder, inputs=None):
    """Instantiate a process, from a ``Process`` class or ``ProcessBuilder`` instance."""
    from aiida.engine.utils import instantiate_process as _instantiate_process
    manager = get_manager()
    runner = manager.get_runner()
    return _instantiate_process(runner, cls_or_builder **{inputs or {})
```

src/aiida_quantumespresso/workflows/protocols/core_hole_treatments.yaml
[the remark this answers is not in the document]
PNOGillespie: Ah, perhaps there is a slight misunderstanding. The XPS workflow itself actually just uses the PwBaseWorkChain and does not use `xspectra.x` to compute XPS - essentially all that happens there is we calculate the difference in total energy between `pw.x` calculations where the localised core-hole is on different symmetrically non-equivalent atoms of the same species.

The future plan involves expanding the theory further to calculate absolute values for the XPS binding energy, but even that will only be using `pw.x` and not `xspectra.x`.

Hope that makes things clearer.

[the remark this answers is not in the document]
sphuber: Ah, that does clear things up! Thanks a lot for the explanation. All good then!

sphuber (approved):

PR #875, opened 2022-12-21
`PwBandsWorkChain`: Fix `nbnd` for spin-polarised calculations

### LINKED ISSUE #874: `PwBandsWorkChain`: Fix setting number of bands for spin-polarised calculations
Currently, the `PwBandsWorkChain` will double the number of bands for the `bands` step when running spin-polarised calculations:

https://github.com/aiidateam/aiida-quantumespresso/blob/b904cc61a4dd384d69617c3682638617d9755192/src/aiida_quantumespresso/workflows/pw/bands.py#L278-L291

However, as noted in the [documentation](), in QE the number of k-points is doubled for spin-polarised calculations, not the number of bands `nbnd`:

> Number of electronic states (bands) to be calculated.
Note that in spin-polarized calculations the number of
k-point, not the number of bands per k-point, is doubled

So the `nspin_factor` should be removed, at least for collinear calculations. I'm not sure how QE treats the case for non-collinear calculations.

mbercx:
Fixes #874

When running spin-polarised calculations (`nspin != 2`) with the `PwBandsWorkChain` and providing an `nbands_factor`, the number of bands should _not_ be doubled, since QE doubles the number of treated k-points instead.

Here we fix the `PwBandsWorkChain` by removing the `nspin_factor` from the equations to determine the number of bands for the `bands` calculation in the work chain.

### DIFF
--- a/src/aiida_quantumespresso/workflows/pw/bands.py
+++ b/src/aiida_quantumespresso/workflows/pw/bands.py
@@ -38,8 +38,8 @@ class PwBandsWorkChain(ProtocolMixin, WorkChain):
         used. If the `nbands_factor` is specified the maximum value of the following values will be used:
 
         * `nbnd` of the preceding SCF calculation
-        * 0.5 * nspin * nelectrons * nbands_factor
-        * 0.5 * nspin * nelectrons + 4 * nspin
+        * 0.5 * nelectrons * nbands_factor
+        * 0.5 * nelectrons + 4
 
     Kpoints:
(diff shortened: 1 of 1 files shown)


sphuber (approved):
Thanks @mbercx Changes look good, just a small typo in the commit message perhaps?
```
When running spin-polarised calculations (`nspin != 2`)
```
should that be
```
When running spin-polarised calculations (`nspin >= 2`)
```
?

mbercx:
Thanks @sphuber! Fixed in the final commit message.

PR #878, opened 2023-01-12
Feature/get xspectra structures

PNOGillespie:
# Overview
In this PR, we propose a new CalcFunction to automatically prepare input structures suitable for the calculation of core-level spectroscopic data such as X-Ray Absorbtion (XAS) and X-Ray Photoelectron (XPS) spectra. The CF takes a StructureData object as input and performs the following steps:

* Converts the crystal structure to its crystallographic standard using the `spglib` library.
* Analyses the symmetry of the standardized crystal using `spglib` to obtain all symmetrically non-equivalent atomic sites.
* Creates a StructureData object of the supercell of the standardized structure such that the final cell dimensions are equal to or greater than a given minimum length in angstrom required to minimize spurious core-hole image interactions (8.0 by default)
* Creates a new StructureData object for each symmetrically non-equivalent `Site` where the `Kind` at the `Site` is replaced with a new `Kind` for the absorbing atom (using 'X' as a default `Kind` name)
* Returns all StructureData generated as well as an `output_parameters` dictionary containing information on the structures generated and the symmetric properties of each non-equivalent site.

For molecular systems, the input structure is instead converted to a `centered_structure`, where the molecule is placed at the centre of the new periodic box generated for the marked structures. Instead of `spglib`, the `Pymatgen` library is used to analyse the symmetry of molecular systems.

# Usage Notes
The CF takes `structure` as a positional argument and can take an arbitrary dictionary of keyword arguments which control the behaviour of the function:
* `abs_atom_marker` (`Str`) - sets the `Kind` name of the absorbing atom for all output StructureData (defaults to 'X'), will except with an error message if the marker corresponds to the name of any existing `Kind` in the structure.
* `supecell_min_parameter` (`Float`) - sets the minimum cell length required for the supecell (defaults to 8.0 ang), or `centered_structure` in the case of molecules.
* `standardize_structure` (`Bool`) - sets whether or not to standardize the input structure before creating the new supercell (defaults to `True`), though is not meaningful in the case for molecules.
* `absorbing_elements_list` (`List`) - sets the list of elements to consider when creating marked structures (defaults to all Kinds present in the structure). Will return a `ValidationError` if the list contains `Kind`s not present in the structure.
* `molecule_input` (`Bool`) - sets whether or not the input structure is a molecule or otherwise non-periodic (defaults to `False`).
* `spglib_settings` (`Dict`) - sets symmetry tolerance parameters used by `spglib` (none by default). Will return a `ValidationError` if the dictionary contains invalid keywords, though not all valid keywords need to be defined. Only meaningful in the case of periodic systems
* `pymatgen_settings` (`Dict'`) - sets symmetry tolerance parameters used by `pymatgen` (none by default). Will return a `ValidationError` if the dictionary contains invalid keywords, though not all valid keywords need to be defined. Only meaningful in the case of non-periodic systems.

In addition to these user-defined options, the CF will also do the following as pre-set methods:
* If the space group of a system not declared as a molecule is found to be 1 or 2 (triclinic), the structure will not be standardized in any case, regardless of user input. This is intended in order to preserve the structures of improper crystals and defected systems when they are used as input.
* If `supercell_min_parameter` is set to exactly 0, the supercell creation step will be skipped cleanly in the case of periodic systems.
* The box size used in the case of molecular systems will be the greater of `supercell_min_parameter` and 2x the extent of the molecular structure in x, y, and z. This is intended to meet the requirement for SCF calculations of molecular systems in `pw.x` using the Martyna-Tuckerman method for `assume_isolated`.


src/aiida_quantumespresso/workflows/functions/get_xspectra_structures.py
```diff
@@ -0,0 +1,336 @@
...
+        spglib_settings_dict = unwrapped_kwargs['spglib_settings'].get_dict()
+        valid_keys = ['symprec', 'angle_tolerance']
+        spglib_kwargs = {key: value for key, value in spglib_settings_dict.items() if key in valid_keys}
+    else:
+        spglib_kwargs = {}
+
+    output_params = {}
+    result = {}
+    if not elements_defined:
+        abs_elements_list = [Kind.symbol for Kind in structure.kinds]
```
superstar54, line 118: move this to line 97.

superstar54 (changes requested):

PNOGillespie:
Hi @superstar54. I addressed the changes you requested. I deviated a little by changing `molecule_input` to `is_molecule_input` and `input_standardized` to `structure_is_standardized` since I felt that it was better to be more explicit. The rest of the changes are as you asked for.

Let me know if there is anything else that needs changing.

src/aiida_quantumespresso/workflows/functions/get_xspectra_structures.py
```diff
@@ -0,0 +1,332 @@
...
+        result['centred_structure'] = centred_structure
+
+        for value in equivalency_dict.values():
+            target_site = value['site_index']
+            marked_structure = StructureData()
+            centred_kinds = centred_structure.kinds
+
+            marked_structure.set_cell(centred_structure.cell)
+            for kind in centred_kinds:
+                marked_structure.append_kind(kind)
```
superstar54, line 207: Here you copy all kinds to the new structure, thus in the end, you need to validate that all kinds have at least one site. 
For example, for the `CH4` molecule, since there is only one `C` atom. When this site was assigned as `X`,  there is no more site with the kind `C`. Then it shows this error.
```
aiida.common.exceptions.ValidationError: The following kinds are defined, but there are no sites with that kind: ['C']
```
This also applies to larger crystal structures which have only one site for one of the kinds.

src/aiida_quantumespresso/workflows/functions/get_xspectra_structures.py
```diff
@@ -0,0 +1,332 @@
...
+                    absorbing_kind = Kind(name=abs_atom_marker, symbols=site.kind_name)
+                    absorbing_site = Site(kind_name=absorbing_kind.name, position=site.position)
+                    marked_structure.append_kind(absorbing_kind)
+                    marked_structure.append_site(absorbing_site)
+                else:
+                    new_site = Site(kind_name=site.kind_name, position=site.position)
+                    marked_structure.append_site(new_site)
+                index_counter += 1
+
+            result[f'site_{target_site}'] = marked_structure
```
superstar54, line 221: I suggest adding the symbol (kind) to the name too, so that one can easily know which element is calculated.

e.g. `site_2` to `site_2_Ti`

src/aiida_quantumespresso/workflows/functions/get_xspectra_structures.py
PNOGillespie: Yes, that is a fair point. I think it would make sense then to dynamically add `kind`s to the new structure as required, rather than adding an extra check at the end to ensure that no `kind`s are present that do not have a `site`. That seems the simpler option to me at least.

src/aiida_quantumespresso/workflows/functions/get_xspectra_structures.py
PNOGillespie: I think it would be useful to add this information to the link label of the CalcFunction's outputs, for instance:

    Outputs                    PK  Type
    ----------------------  -----  -------------
    output_parameters       11298  Dict
    site_0_Li               11295  StructureData
    site_1_P                11296  StructureData
    site_2_F                11297  StructureData
    standardized_structure  11293  StructureData
    supercell               11294  StructureData

This would be useful for quick reference of the outputs of the CalcFunction at no extra effort. For the site labels in the `output_parameters` Dict node, the element information is kept in the dictionary itself, so I don't think adding it would be helpful - especially as the dictionary keys can return this information more reliably.

superstar54: Yes, I agree. Please add this.

PNOGillespie:
Ok @superstar54, I've added the suggested changes. I also tidied up some of the code here and there (such as replacing `index_counter` variables with `enumerate(list)`)

Let me know if there are any other changes needed.

src/aiida_quantumespresso/workflows/functions/get_xspectra_structures.py
```diff
@@ -0,0 +1,325 @@
...
+    output_params['input_structure_num_sites'] = incoming_structure_size
+    output_params['input_structure_cell_matrix'] = incoming_structure_cell
+    output_params['input_structure_cell_lengths'] = incoming_structure_params
+
+    # Process a non-periodic system
+    if is_molecule_input:
+        from pymatgen.symmetry.analyzer import PointGroupAnalyzer
+
+        pymatgen_molecule = structure.get_pymatgen_molecule()
+        centred_molecule = pymatgen_molecule.get_centered_molecule()
```
superstar54, line 137:
```suggestion
        centered_molecule = pymatgen_molecule.get_centered_molecule()
```
Please change all of them.

src/aiida_quantumespresso/workflows/functions/get_xspectra_structures.py
```diff
@@ -0,0 +1,325 @@
...
+            marked_structure = StructureData(pbc=False)
+            centred_kinds = {kind.name: kind for kind in centred_structure.kinds}
+
+            for index, site in enumerate(centred_structure.sites):
+                if index == target_site:
+                    absorbing_kind = Kind(name=abs_atom_marker, symbols=site.kind_name)
+                    absorbing_site = Site(kind_name=absorbing_kind.name, position=site.position)
+                    marked_structure.append_kind(absorbing_kind)
+                    marked_structure.append_site(absorbing_site)
+                else:
```
superstar54, line 211: You use `centered_structure` for the molecule. I suggest you use `supercell`, the same name as for the crystal. Here, the main goal is to create a larger unit cell for molecules, and shifting molecules to the center is only for better visualization, and will not affect the calculation (correct me if I am wrong).
If you use `supercell` for both of them, after creating the supercell, the step to generate the marked structures should be the same for molecule and crystal, thus lines 196-215 and lines 297-320 are the same and can be written only once.

src/aiida_quantumespresso/workflows/functions/get_xspectra_structures.py
```diff
@@ -0,0 +1,325 @@
...
+                    if site.kind_name not in [kind.name for kind in marked_structure.kinds]:
+                        marked_structure.append_kind(centred_kinds[site.kind_name])
+                    new_site = Site(kind_name=site.kind_name, position=site.position)
+                    marked_structure.append_site(new_site)
+
+            result[f'site_{target_site}'] = marked_structure
```
superstar54, line 217:
```suggestion
            result[f'site_{target_site}_{value["symbol"]}'] = marked_structure
```

superstar54 (changes requested):
@PNOGillespie Thanks! I added a few comments. I think we can merge after you address them.

src/aiida_quantumespresso/workflows/functions/get_xspectra_structures.py
PNOGillespie: Ah, yes, apologies about those. That was me using the British English spelling, but I will change the spelling over for the sake of consistency.

PNOGillespie:
Ok @superstar54, I have added the requested changes. Let me know if anything else needs changing.

superstar54 (approved):
LGTM!

src/aiida_quantumespresso/workflows/functions/get_xspectra_structures.py
```diff
@@ -0,0 +1,306 @@
...
+    Takes an incoming StructureData node and prepares structures suitable for calculation with
+    xspectra.x. The basic criteria for a suitable structure is for the cell parameters to be
+    sufficiently large to reduce spurious interactions between localised core-holes in
+    neighbouring periodic images and for one atomic site to be marked as the absorbing atom
+    site.
+
+    The default setting for the cell size is 8.0 angstrom, which can be changed by including
+    'supercell_min_parameter' as a Float node in the keyword arguments. Accepted keyword
+    arguments are:
+        - abs_atom_marker: a Str node defining the name of the absorbing atom Kind. The
```
superstar54, line 29: Here is the line with the error reported from the docs check.
```suggestion
    arguments are:

        - abs_atom_marker: a Str node defining the name of the absorbing atom Kind. The
```

src/aiida_quantumespresso/workflows/functions/get_xspectra_structures.py
```diff
@@ -0,0 +1,306 @@
...
+        - is_molecule_input: a Bool object to define for the function that the input structure is
+                    a molecule and not a periodic solid system. Required in order to instruct
+                    the CF to use Pymatgen rather than spglib to determine the symmetry. The
+                    CF will assume the structure to be a periodic solid if no input is given.
+        - spglib_settings: an optional Dict object containing overrides for the symmetry
+                            tolerance parameters used by spglib (symmprec, angle_tolerance).
+        - pymatgen_settings: an optional Dict object containing overrides for the symmetry
+                             tolerance parameters used by Pymatgen when processing molecular
+                             systems (tolerance, eigen_tolerance, matrix_tolerance).
+    :param structure: the StructureData object to be analysed
```
superstar54, line 52:
```suggestion
                             systems (tolerance, eigen_tolerance, matrix_tolerance).

    :param structure: the StructureData object to be analysed
```

superstar54 (changes requested):

superstar54 (approved):

PR #879, opened 2023-01-13
Add Feature: `XpsWorkChain`

mhdzbert:
Adds a WorkChain to compute XPS spectra using core-hole pseudopotentials within pw.x. Perfoms the following steps:
* Relaxes the input structure (optional)
* Calls `get_xspectra_structures()` to get all symmetrically-nonequivalent sites for each element and prepare structures suitable for core-hole SCF calculations
* Runs all core-hole SCF calculations
* Calls `get_spectra_by_element()` to return an XPS spectrum on a relative energy axis for each structure produced in step 2

Returns both a broadened spectrum using a Voight profile (combined Lorenzian and Gaussian) as well as the chemical shift values (differences in total energy relative to the lowest value) for each element.


superstar54 (dismissed):
Hi @mhdzbert Thanks for the XPS PR. This is nice work. I made some changes. 
- support calculating absolute binding energy
- update the validate_input
- clean some old code from the XAS part.

I think it is ready to merge.

sphuber:
Thanks to both. Just two quick comments:

* It might be useful to have the calcfunction in a separate file in `aiida_quantumespresso.calculations.functions.xspectra`. There are already other functions related to the xspectra workchains
* There are no tests for the workchain. I understand that this is not the easiest to add, but maybe at least for the protocol methods that do not require running the workchain, would be very useful to catch simple mistakes and bugs. Have a look at how @PNOGillespie did this for the `Xspectra*WorkChains`.

superstar54:
Hi @sphuber Thanks for your comments. It's fine for me to move the calcfunction. I will add a test for the workchain.

superstar54 (approved):
Hi @sphuber I moved the calcfunction to the `xspectra` folder, and added a test for the workchain.

I will merge the PR if no more comments.

PR #881, opened 2023-01-25
Deal with hard interruptions of ionic calculations with `pw.x`

sphuber:


sphuber:
The tests are failing because they require https://github.com/aiidateam/aiida-core/pull/5866 to be merged and released.

src/aiida_quantumespresso/parsers/pw.py
```diff
@@ -106,16 +108,35 @@ def parse(self, **kwargs):
...
+            # First check whether the scheduler already reported an exit code.
+            if self.node.exit_status is not None:
 
-        if self.exit_code_xml:
-            return self.exit(self.exit_code_xml)
+                # The following scheduler errors should correspond to cases where we can simply restart the calculation
+                # and have a chance that the calculation will succeed as the error can be transient.
+                recoverable_scheduler_error = self.node.exit_status in [
+                    PwCalculation.exit_codes.ERROR_SCHEDULER_OUT_OF_WALLTIME.status,
+                    PwCalculation.exit_codes.ERROR_SCHEDULER_NODE_FAILURE.status,
```
ltalirz, line 121: Where do these exit codes come from, does the scheduler plugin define them or are they actually defined by the calculation plugin?

If the scheduler plugin defines them, can we assume that all scheduler plugins define these exit codes?

tests/parsers/test_pw.py
```diff
@@ -364,6 +366,64 @@ def test_pw_failed_interrupted_xml(
...
+        ('default', None),
+        ('failed_interrupted', PwCalculation.exit_codes.ERROR_SCHEDULER_OUT_OF_WALLTIME),
+    )
+)
+def test_pw_failed_interrupted_scheduler(
+    fixture_localhost, generate_calc_job_node, generate_parser, generate_inputs, test_case, expected_exit_code
+):
+    """Test that an exit code set by the scheduler is not overridden unless a more specific error is parsed.
+
+    The test is run twice, one for the ``default`` test case and once for ``failed_interrupted``, which correspond to a
```
ltalirz, line 380:
```suggestion
    The test is run twice, once for the ``default`` test case and once for ``failed_interrupted``, which correspond to a
```

tests/parsers/test_pw.py
```diff
@@ -364,6 +366,64 @@ def test_pw_failed_interrupted_xml(
...
+
+    if expected_exit_code is None:
+        assert calcfunction.is_finished_ok, calcfunction.exception
+    else:
+        assert not calcfunction.is_finished_ok, calcfunction.exception
+        assert calcfunction.exit_status == expected_exit_code.status
+
+
+def test_pw_failed_interrupted_relax(fixture_localhost, generate_calc_job_node, generate_parser, generate_inputs):
+    """Test a `relax` that was interrupted because the scheduler killed the job since it was out of walltime.
```
ltalirz, line 404:
```suggestion
    """Test that a `relax` calculation that was interrupted by the scheduler is nevertheless parsed (partially).
```

tests/parsers/test_pw.py
```diff
@@ -364,6 +366,64 @@ def test_pw_failed_interrupted_xml(
...
+    node.set_exit_status(PwCalculation.exit_codes.ERROR_SCHEDULER_OUT_OF_WALLTIME.status)
+    parser = generate_parser(entry_point_parser)
+    results, calcfunction = parser.parse_from_node(node, store_provenance=False)
+
+    assert calcfunction.is_failed
+    assert calcfunction.exit_status == PwCalculation.exit_codes.ERROR_IONIC_INTERRUPTED_PARTIAL_TRAJECTORY.status
+    assert orm.Log.collection.get_logs_for(node)
+    assert 'output_parameters' in results
+    assert 'output_structure' in results
+    assert 'output_trajectory' in results
```
ltalirz, line 424: I guess it'll be fine, but maybe test that it has the expected length?

ltalirz (commented):
thanks a lot @sphuber , this looks great!

just one question

ltalirz (approved):
I'll approve, as I don't have any objections to the direction; feel free to complete without further review.

src/aiida_quantumespresso/parsers/pw.py
sphuber: They are defined by the `CalcJob` base class. However, it is then up to the scheduler plugin to implement the functionality to parse the output and return them if applicable. This is currently done for the `SlurmScheduler` though.

tests/parsers/fixtures/pw/failed_interrupted_relax/aiida.out
```diff
@@ -0,0 +1,926 @@
...
+             0 MiB allocation reported by mallinfo(arena+hblkhd)
+        218087 MiB available memory on the node where the printing process lives
+------------------
+     ethr =  1.79E-11,  avg # of iterations =  1.0
+
+     total cpu time spent up to now is        6.2 secs
+
+     End of self-consistent calculation
+
+          k = 0.0000 0.0000 0.0000 (   749 PWs)   bands (ev):
```
mbercx, line 347: Just a reminder to still remove occupations etc. We should really write a script for this `stdout` trimming that we can update as move more and more to the XML ^^

sphuber:
@mbercx this is now ready for review

tests/parsers/test_pw.py
```diff
@@ -496,6 +498,64 @@ def test_pw_failed_interrupted_xml(
...
+def test_pw_failed_interrupted_scheduler(
+    fixture_localhost, generate_calc_job_node, generate_parser, generate_inputs, test_case, expected_exit_code
+):
+    """Test that an exit code set by the scheduler is not overridden unless a more specific error is parsed.
+
+    The test is run twice, once for the ``default`` test case and once for ``failed_interrupted``, which correspond to a
+    successful run and a run that got interrupted (usually due to scheduler killing the job). Before calling the parser
+    the ``ERROR_SCHEDULER_OUT_OF_WALLTIME`` is set on the node. In the case of the ``default`` test case, this should be
+    ignored and the parser should return ``ExitCode(0)``. For the interrupted case, the exit code of the scheduler
+    should not be overridden so the parser should return ``None``.
```
mbercx, line 516: Doesn't the parser return `PwCalculation.exit_codes.ERROR_SCHEDULER_OUT_OF_WALLTIME` in this case?

mbercx (approved):
Thanks @sphuber! Implementation looks solid, just have one little question regarding the tests and then it's good to go. 🚀

PR #882, opened 2023-01-25
Dependencies: Update requirement `aiida-pseudo~=1.0`

sphuber:
The package is now considered stable. By upgrading to the first major version, future minor version updates which often contain new established family versions, will be automatically supported without explicitly having to update the version requirement in this package.


mbercx (approved):
Thanks @sphuber!

PR #884, opened 2023-01-31
Protocol: Use v1.2 of SSSP and PBEsol by default

mbercx:
Now that `aiida-pseudo` v1.0 has been released with support for v1.2 of the SSSP, we can switch to using this as the default for our protocols.

A recent discussion with @giovannipizzi and Nicola also led to the decision to use PBEsol by default for the protocols, since this functional typically leads to improved results for solid state systems.


sphuber (approved):
Thanks @mbercx . I take it the motivation for this has been discussed with the higher-ups ^^

PR #888, opened 2023-02-15
Add Feature: `XspectraCrystalWorkChain`

PNOGillespie:
# Overview

In this PR we propose to implement a new WorkChain intended to fully automate the calculation of K-edge XANES spectra for each element in a given input structure. This new WorkChain essentially automates the generation of input structures for `XspectraCoreWorkChain` sub-processes, as well as handling generation and setting of inputs for each sub-process, in order to compute a complete K-edge XANES spectrum for each element requested. The workflow process defined herein applies specifically to periodic systems as it uses the `get_xspectra_structures` CalcFunction to prepare input structures under the assumption that the input structure is not molecular. Due to the lack of pseudopotentials currently available which either contain GIPAW information or define a core-hole state, such pseudopotentials are presently required to be given as input. A code node to calculate the core wavefunction using `upf2plotcore.sh` in [AiiDA-Shell](https://github.com/sphuber/aiida-shell/) can be provided to produce the core wavefunction data for each pseudopotential on-the-fly if desired, however pre-computed core wavefunction files can also be provided as input, either directly or via `get_builder_from_protocol`.

# Process Flow

1. Perform a vc-relax of the input structure using the `PwRelaxWorkChain` (optional)
2. Analyse the symmetry of the input/relaxed structure using `get_xspectra_structures` and return a marked structure for each symmetrically inequivalent absorbing atom site. Use the spacegroup number obtained to determine the list of polarisation vectors to calculate for each marked structure.
3. Collect the core wavefunction data from each GIPAW pseudopotential provided as input (optional).
4. Launch an `XspectraCoreWorkChain` sub-process for each marked structure
5. Collect the powder spectra generated by each `XspectraCoreWorkChain` sub-process and compile a final spectrum for each element using a weighted average, based on the multiplicity of each absorbing atom site of the same element.

# Inputs
In addition to inputs for the `XspectraCoreWorkChain` and `PwRelaxWorkChain`, other inputs to take notice of are:

- `gipaw_pseudos` - an input namespace for `UpfData` nodes, taking the form `{{element} : {upf}}`, which defines the ground-state pseudopotential for the associated element.
- `core_hole_pseudos` - an input namespace analagous to `gipaw_pseudos`, used to define each pseudopotential containing the core-hole. The namespace keys must be those of their associated element, **not** the marker to be used for the absorbing atom.
- `core_wfc_data` - an optional input namespace for `SinglefileData` nodes, taking the form `{{element} : {singlefile}}`. The WorkChain will automatically skip the `upf2plotcore.sh` step if this input is provided. Note that this input **must** be complete for all elements to be calculated.
- `core_hole_treatments` - an optional `Dict` input which defines the core-hole treatment (approximation) to be used in the `XspectraCoreWorkChain` for each element. Note that this does not need to be complete as any elements not defined in this input will use the default full-core-hole (FCH) approximation instead.
- `spglib_settings` - an optional `Dict` input which defines symmetry tolerances to be used by [`spglib`](https://github.com/spglib/spglib). Must contain valid options used by `spglib.get_symmetry_dataset()` and `spglib.standardize_structure()`, though not all options must be defined.
- `structure_preparation_settings` - an optional `Dict` input which defines parameters to be used by `get_xspectra_structures()`. Must contain valid options used by `get_xspectra_structures`, though not all options must be defined.
- `return_all_powder_spectra` - an optional `Bool` input which requests that all powder spectra generated during the runtime of the entire WorkChain be returned as output to the main WorkChain. Optional and set to `False` by default in order to keep the terminal display of the WorkChain's outputs in a condensed format, though potentially useful if spectra for each absorbing atom site are of interest.


src/aiida_quantumespresso/workflows/xspectra/crystal.py
```diff
@@ -0,0 +1,676 @@
...
+            optional_cell_prep = self.inputs.structure_preparation_settings
+            for key, value in optional_cell_prep.items():
+                if isinstance(value, float):
+                    node_value = orm.Float(value)
+                if isinstance(value, int):
+                    node_value = orm.Int(value)
+                if isinstance(value, bool):
+                    node_value = orm.Bool(value)
+                if isinstance(value, str):
+                    node_value = orm.Str(value)
```
superstar54, line 490: Creating new node data here causes a loss of data provenance.

src/aiida_quantumespresso/workflows/xspectra/crystal.py
```diff
@@ -0,0 +1,676 @@
...
+                inputs[key] = node_value
+
+        if 'spglib_settings' in self.inputs:
+            spglib_settings = self.inputs.spglib_settings
+            inputs['spglib_settings'] = spglib_settings
+        else:
+            spglib_settings = None
```
superstar54, line 497:
```suggestion
        if 'spglib_settings' in self.inputs:
            inputs['spglib_settings'] = self.inputs.spglib_settings

```

src/aiida_quantumespresso/workflows/xspectra/crystal.py
```diff
@@ -0,0 +1,676 @@
...
+        """
+
+        labels = self.ctx.elements_list
+        for label in labels:
+            shelljob_node = self.ctx[f'upf2plotcore_{label}']
+            core_wfc_data = shelljob_node.outputs.stdout
+            header_line = core_wfc_data.get_content()[:40]
+            num_core_states = int(header_line.split(' ')[5])
+            if num_core_states == 0:
+                return self.exit_codes.ERROR_NO_GIPAW_INFO_FOUND
```
superstar54, line 566: The code here is repeated in the `XspectraCoreWorkChain`. Is it possible to avoid this?

src/aiida_quantumespresso/workflows/xspectra/crystal.py
```diff
@@ -0,0 +1,676 @@
...
+            else:
+                inputs.core_wfc_data = self.ctx[f'upf2plotcore_{abs_element}'].outputs.stdout
+
+            # Get the given settings for the SCF inputs and then overwrite them with the
+            # chosen core-hole approximation, then apply the correct pseudopotential pair
+            scf_params = inputs.scf.pw.parameters.get_dict()
+            ch_inputs = XspectraCoreWorkChain.get_treatment_inputs(treatment=ch_treatment)
+            new_scf_params = recursive_merge(left=scf_params, right=ch_inputs)
+
+            # Set the parameter `xiabs` correctly, now that we know which structure we have
```
superstar54, line 600: Please add a comment to describe what `xiabls` is and how you set the value.

src/aiida_quantumespresso/workflows/xspectra/crystal.py
```diff
@@ -0,0 +1,676 @@
...
+
+            # Set `starting_magnetization` if we are using an XCH approximation, using the
+            # absorbing species as a reasonable place for the unpaired electron.
+            if ch_treatment == 'xch_smear':
+                new_scf_params['SYSTEM'][f'starting_magnetization({abs_species_index})'] = 1
+
+            core_hole_pseudo = self.inputs.core_hole_pseudos[abs_element]
+            gipaw_pseudo = self.inputs.gipaw_pseudos[abs_element]
+            inputs.scf.pw.pseudos[abs_atom_marker] = core_hole_pseudo
+            inputs.scf.pw.pseudos[abs_element] = gipaw_pseudo
```
superstar54, line 614: If a user has a larger cell, so no supercell is created. For a special element, there is only one atom. In the default `inputs.scf.pw`, there is a pseudo for this element. Now this atom is replaced by the marker. So there will be pseudo corresponding to no kind anymore. This will raise an error.

src/aiida_quantumespresso/workflows/xspectra/crystal.py
```diff
@@ -0,0 +1,676 @@
...
+
+            inputs.scf.pw.parameters = orm.Dict(dict=new_scf_params)
+            inputs.xs_prod.xspectra.parameters = orm.Dict(dict=new_xs_params)
```
superstar54, line 617: Lose data provenance.

src/aiida_quantumespresso/workflows/xspectra/crystal.py
```diff
@@ -0,0 +1,676 @@
...
+
+            inputs = prepare_process_inputs(XspectraCoreWorkChain, inputs)
+
+            future = self.submit(XspectraCoreWorkChain, **inputs)
+            self.to_context(xspectra_core_workchains=append_(future))
```
superstar54, line 622: Is there any special reason to use `append_` instead of using a `dict`? In line 628, you load the `keys` of `structures_to_process` again, thus you assume the keys are always in the same order. This has been the case since Python 3.6, but better to not rely on this.

src/aiida_quantumespresso/workflows/xspectra/crystal.py
```diff
@@ -0,0 +1,676 @@
...
+    def results(self):
+        """Compile all output spectra, organise and post-process all computed spectra, and send to outputs."""
+
+        site_labels = self.ctx.structures_to_process.keys()
+        work_chain_nodes = self.ctx.xspectra_core_workchains
+        spectra_nodes = {label : node.outputs.powder_spectrum for label, node in zip(site_labels, work_chain_nodes)}
+        spectra_nodes['metadata'] = {'call_link_label' : 'compile_final_spectra'}
+
+        equivalent_sites_data = self.ctx.equivalent_sites_data
+        elements_list = orm.List(list=self.ctx.elements_list)
```
superstar54, line 647: Lose data provenance.

superstar54 (changes requested):
@PNOGillespie Thanks for the PR! Nice work. I added a few comments.

src/aiida_quantumespresso/workflows/protocols/xspectra/crystal.yaml
```diff
@@ -0,0 +1,14 @@
+default_inputs:
+    abs_atom_marker: X
+    clean_workdir: True
+    return_all_powder_spectra: False
+    core:
+        get_powder_spectrum: True
+default_protocol: moderate
+protocols:
```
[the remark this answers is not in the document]
PNOGillespie: Increasing `supercell_min_parameter` beyond the required amount has almost no effect on the accuracy of the calculation. The reason for the supercell is to prevent a spurious interaction between neighbouring core-hole images, since we are localising a core-hole in a periodic structure, so the periodic cell parameter will either be too small to separate the images or large enough that the spurious interaction becomes negligible.

In testing, we find that 8.0 ang is sufficient for this - both for metals and insulators, in agreement with the [original XSpectra publication](dx.doi.org/10.1103/PhysRevB.66.195107) - though this can be higher depending on both the system type and the choice of core-hole approximation. For instance, when studying a carbon anode material more recently, we found that for a system of graphene + Li the cell size convergence is very slow using a full-core-hole approach but agrees with the minimum parameter of 8.0 using an excited-core-hole approach.

The only other parameters which affect the accuracy of the XAS spectrum are already covered by the various other `WorkChain`s called in the workflow, so the protocol options in the `CrystalWorkChain` are really just there to set the level of accuracy in the sub-processes.

src/aiida_quantumespresso/workflows/xspectra/crystal.py
```diff
@@ -0,0 +1,676 @@
...
+        abs_atom_marker = inputs['abs_atom_marker'].value
+        if abs_atom_marker in kinds_present:
+            raise ValidationError(
+                f'The marker given for the absorbing atom ("{abs_atom_marker}") matches an existing Kind in the '
+                f'input structure ({kinds_present}).'
+            )
+
+        if not inputs['core']['get_powder_spectrum'].value:
+            raise ValidationError(
+                'The ``get_powder_spectrum`` input for the XspectraCoreWorkChain namespace must be ``True``.'
```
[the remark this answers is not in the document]
PNOGillespie: This is to ensure that the WorkChain returns the expected `powder_spectrum` output - otherwise the `CoreWorkChain` sub-processes won't return it, since `get_powder_spectrum` is `False` by default. Getting the powder spectrum is required, since the intention of the `CrystalWorkChain` is to return the complete powder XAS spectrum for each chosen element in the structure - thus it will need to compute this for each symetrically-inequivalent atom of each element.

To generate a reasonable approximation to the powder spectrum, we compute the XAS along each symetrically-inequivalent periodic vector of the cell and take the average (as detailed in [Brouder *et al* (1990)](dx.doi.org/10.1088/0953-8984/2/3/018)) - thus we need to calculate for 1, 2, or 3 different vectors depending on the spacegroup of the cell.

src/aiida_quantumespresso/workflows/xspectra/crystal.py
PNOGillespie: Bear in mind that this is used *instead* of the code used in the `CoreWorkChain`, thus only running upf2plotcore.sh once for each element before continuing to the `CoreWorkChain`s - thus saving a bit of computing effort.

src/aiida_quantumespresso/workflows/xspectra/crystal.py
PNOGillespie: The issue here is that AiiDA (as far as I know) doesn't allow the `expose_inputs` port type to be dynamic. Since we need to change the xspectra.x and pw.x input parameters to fit differences in the absorbing atom position, this will technically break data provenance. This is just a consequence of having a `WorkChain` process which branches into multiple sub-processes in order to compute one final property - the same issue exists with the `CoreWorkChain` since we run multiple XAS calculations using one set of "parent" input parameters and thus break provenance for the `parameters`.

If it is possible to have a dynamic set of namespaces for each copy of the input parameters, then we would need to refactor both the `CrystalWorkChain` and `CoreWorkChain` to have a different namespace for each element and each polarisation vector before launching. This could be possible for the `CrystalWorkChain`, since we can know the list of elements to run calculations for beforehand, but not for the `CoreWorkChain` since we won't know the symmetry properties beforehand.

src/aiida_quantumespresso/workflows/xspectra/crystal.py
```diff
@@ -0,0 +1,676 @@
...
+            for element in absorbing_elements_list:
+                if element not in elements_present:
+                    extra_elements.append(element)
+            if len(extra_elements) > 0:
+                raise ValidationError(
+                    f'The some elements in ``elements_list`` {extra_elements} do not exist in the'
+                    f' structure provided {elements_present}.'
+                )
+        else:
+            absorbing_elements_list = sorted([kind.symbol for kind in structure.kinds])
```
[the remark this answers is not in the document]
PNOGillespie: We can actually just get rid of the line entirely and change `elements_present` on line 374 to be `sorted([kind.symbol for kind in structure.kinds])`, since it can then be re-used in the other checks if needed.

src/aiida_quantumespresso/workflows/xspectra/crystal.py
PNOGillespie: Bear in mind that `structure_preparation_settings`, which we are unpacking here, is already an input node to the WorkChain. We're only losing provenance in the technical sense that the node identity is being lost, not the actual information itself.

I will see if I can refactor `structure_preparation_settings` into an `input_namespace`, which would then preserve the IDs of these parameters, but I'm not sure how well this works for different data types.

src/aiida_quantumespresso/workflows/xspectra/crystal.py
PNOGillespie: I will test and add a method to catch this, thus only setting `core_hole_pseudo` and not `gipaw_pseudo` for the element if there is only one instance of it in the structure.

src/aiida_quantumespresso/workflows/xspectra/crystal.py
PNOGillespie: I will add something to pass the `elements_list` given in the `WorkChain` inputs to this variable here instead of re-defining it. In order for this to work, we will have to change `elements_list` to be a required input so that it is always defined as a `List` node in the inputs.

Since setting a default value is impossible in this case, we can instead change the `get_builder_from_protocol()` method to define this automatically and let the built-in checks confirm that a list of elements has been provided.

src/aiida_quantumespresso/workflows/xspectra/crystal.py
sphuber: The argument to use `append_` applies when you don't need the keys. In that case, you would often end up creating keys like `f'key_index_{i}` just to have different keys, but if you just care about the values, then a list is more appropriate. Consider this:
```python
def submit(self):
    for i in range(n):
        node = submit(...)
        self.to_context(f'namespace.node_{i}'=node)

def analyse(self):
    for node in self.ctx.namespace:
        ...
```
versus
```python
def submit(self):
    for i in range(n):
        node = submit(...)
        self.to_context(namespace=append_(node))

def analyse(self):
    for node in self.ctx.namespace:
        ...
```
In the end the difference is small and subjective, but I think the second is cleaner since if you don't need the keys, there is no point in creating them.

PNOGillespie: Thanks for the clarification @sphuber, much appreciated!

On reflection, I think for this WorkChain we should convert it to use keys. In the case of the `CoreWorkChain` the important detail in that case was the polarisation vector used for the `XspectraCalculation`, though this information can be retrieved from the input parameters of the node if we need to know for certain - which is what is done in post-processing automatically.

For the `CrystalWorkChain`, the important detail is which site the core-hole is localised at. This can in principle be obtained from the structure used for the calculation, but would be much more laborious than just assigning the key beforehand - and also more useful for the user if they wish to retrieve the information later on - so I will just convert this to use explicit keys.

src/aiida_quantumespresso/workflows/protocols/xspectra/crystal.yaml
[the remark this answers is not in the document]
superstar54: The parameter in the protocol should be well-tested so that the user can use it for safety. I suggest that we do some testing on the supercell size. And show the results in one of our meetings with others. Of course, no need to do this for merging this PR.

superstar54 (approved):

superstar54 (approved):

PR #891, opened 2023-02-24
from_scratch for bands calculation

### LINKED ISSUE #877: `restart_mode` should be set to `from_scratch` for a `bands` calculation.
After I update to `4.1.0` I get the warning "`restart_mode` should be set to `from_scratch` for a `bands` calculation.". I launch the band structure calculation from `PwBandsWorkChain` and I didn't override the inputs so I think it should be set. My question is where is this `restart_mode` set? I find from https://github.com/aiidateam/aiida-quantumespresso/pull/475 that the `restart_mode` is removed there.

@mbercx is there any new changes in 4.1.0 that cause the issue? Let me know if more information needed.

### LINKED ISSUE #887: PDOS workchain failed with "bad Fermi energy"
This can be reproduce by submit a calculation on SiO2 from example structure importer, band structure calculation in AiiDAlab-THEOS (settings: `insulator`, `moderate`).
The error is:

```
...
   Computing kpt #:  2701  of  2701
     total cpu time spent up to now is      360.2 secs
 
     ethr =  3.75E-12,  avg # of iterations = 22.7
     Message from routine efermit:
     internal error, cannot bracket Ef
 
     Error in routine tetra_weights (1):
     bad Fermi energy
 
     stopping ...
```

The pw input and output is attached
[pw-303-pdos-error.in.txt](https://github.com/aiidateam/aiida-quantumespresso/files/10697151/pw-303-pdos-error.in.txt)
[pw-303-pdos-error.out.txt](https://github.com/aiidateam/aiida-quantumespresso/files/10697152/pw-303-pdos-error.out.txt)

unkcpz:
fixes #877

### DIFF
--- a/src/aiida_quantumespresso/workflows/pw/bands.py
+++ b/src/aiida_quantumespresso/workflows/pw/bands.py
@@ -270,6 +270,9 @@ class PwBandsWorkChain(ProtocolMixin, WorkChain):
         # The following flags always have to be set in the parameters, regardless of what caller specified in the inputs
         inputs.pw.parameters['CONTROL']['calculation'] = 'bands'
 
+        # Always set restart mode to 'from_scratch' for the bands calculation
+        inputs.pw.parameters['CONTROL']['restart_mode'] = 'from_scratch'
+
         # Only set the following parameters if not directly explicitly defined in the inputs
         inputs.pw.parameters['ELECTRONS'].setdefault('diagonalization', 'cg')
         inputs.pw.parameters['ELECTRONS'].setdefault('diago_full_acc', True)


src/aiida_quantumespresso/calculations/pw.py
```diff
@@ -168,6 +168,7 @@ def validate_inputs_base(value, _):
             if parameters.get('ELECTRONS', {}).get('startingpot', 'file') != 'file':
                 return f'`startingpot` should be set to `file` for a `{calculation_type}` calculation.'
             if parameters.get('CONTROL', {}).get('restart_mode', 'from_scratch') != 'from_scratch':
+                print(parameters)
```
sphuber, line 171:
```suggestion
```

sphuber (changes requested):
Thanks @unkcpz . I agree that this change is correct (QE docs say not to use `restart` for non-scf calculations https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm65). Just please remove the print statement

src/aiida_quantumespresso/calculations/pw.py
unkcpz: Removed. 
I also make a mistake by pushing directly to `main`. Hopefully, (and please pretend) you didn't notice @sphuber :crying_cat_face:

sphuber: I know nothing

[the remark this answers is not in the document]
sphuber: Is this still true? Or is the value now actually correctly getting set to `from_scratch`? I am confused whether this is now working as intended and fixes the issue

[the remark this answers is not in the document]
unkcpz: Yes, it is now working. I test locally and the warning disappeared. More importantly, the `restart_mode` of band calcualtion is `from_scratch` if check from the remote folder.

mbercx:
I remain somewhat opposed to the approach of setting/overriding inputs inside the steps of the work chain in case it can be avoided. A couple of thoughts:

1. Overriding inputs inside the work chain can lead to confusion in case the user has explicitly set a different value, and will prohibit the user from changing it unless logic is implemented to respect explicitly provided inputs. In some cases it's true that it seems you _have_ to set a certain value (e.g. `from_scratch` for a `bands` calculation), but perhaps we don't know the use case (this was also the reason for removing the line that sets the `restart_mode` in https://github.com/aiidateam/aiida-quantumespresso/pull/475).
2. If you set/override inputs inside the work chain, it's more difficult for the user to find where this value comes from when he/she is checking the input files. We already have a place for setting the inputs of a work chain: the protocol files. So I think it makes sense to _explicitly_ set default values there. This also respects the original "separation of church and state" principle which lead to the concept of defining a protocol.

@sphuber @unkcpz if you agree, I can work on a PR that implements these concepts more broadly for the `PwBandsWorkChain`.

mbercx:
Another reason to put these inputs in the protocol instead: the builder repr now looks like this:

```python
Process class: PwBandsWorkChain
Inputs:
bands:
  metadata: {}
  pw:
    code: Plane wave Quantum ESPRESSO v7.2 - local compilation
    metadata:
      options:
        max_wallclock_seconds: 43200
        resources:
          num_machines: 1
        stash: {}
        withmpi: true
    monitors: {}
    parameters:
      CONTROL:
        calculation: scf
<...>
```

So it looks like we'd actually be running an `scf` for the `bands` step, but this is updated inside the work chain...

mbercx:
See https://github.com/aiidateam/aiida-quantumespresso/pull/902

mbercx:
Superseded by https://github.com/aiidateam/aiida-quantumespresso/pull/902

PR #892, opened 2023-03-22
Added useful error message when the schema for pw.x or cp.x is missing

rikigigi:
It is a lot easier to get the correct file from https://github.com/QEF/qeschemas and put in the correct folder without each time going into the generated xml file.

example error message:
`XMLUnsupportedFormatError: unrecognized XML file version: cannot find schema qes_230310.xsd in /scratch/rbertoss/aiida-quantumespresso/src/aiida_quantumespresso/parsers/parse_xml/schemas. You can look for it in https://github.com/QEF/qeschemas`


sphuber (approved):
Thanks @rikigigi even though this should not really occur, but only when a new release of QE comes out and we haven't added the schema yet, I guess it can't hurt to add it. Btw, we just added the schema for 7.2 which should be released soon

PR #894, opened 2023-03-27
Computer node don't have get_name attribute

unkcpz:


unkcpz:
This happened when specify different computers for pw.x and ph.x, maybe it is not tested, I'll check and add the test if it is not there.

sphuber:
> This happened when specify different computers for pw.x and ph.x, maybe it is not tested, I'll check and add the test if it is not there.

Ah yeah, true. That will be the reason. Think the check should not use `not ... == ...` but `... != ...`. If you can add that change, we can just merge this. No need to add a test

unkcpz:
Okay, done.
Besides the check itself, I think it is not always true to just check the label difference. I have two computers set with a bit different SBATCH and slurm account, the `workdir` folder is shared and the same, so the parent calculation in principle can be used for PH calculation. But that requires adding another attribute like `storage_computer` or `workdir_computer` for the computer node.

PR #898, opened 2023-04-05
QE-Support: add latest pw qeschema 

bastonero:
We add the new xml schema for `pw.x` released along QuantumESPRESSO `v7.2`.


mbercx:
Thanks @bastonero! Could you also do a minimal run with `pw.x` v7.2 and add a test for the latest schema, see:

https://github.com/aiidateam/aiida-quantumespresso/wiki/Quantum-ESPRESSO-versions#adding-support

sphuber (approved):
Thanks @bastonero

PR #899, opened 2023-04-06
HotFix: adding and parsing the ``CRASH`` file

### LINKED ISSUE #890: Retrieve and parse the `CRASH` file for `pw.x`
If a crash occurs, `pw.x` will write the error to the `CRASH` file. Currently we don't retrieve nor parse this and rely solely on the `stdout`. However, if the error does not occur in the main thread it won't show up in `stdout` but it is guaranteed to end up in `CRASH`. So we should rely on this file for the error instead.

See original report on [`aiidalab/aiidalab-qe`](https://github.com/aiidalab/aiidalab-qe/issues/355).

bastonero:
Fixes #890 

QuantumESPRESSO ``v7.2`` now moves the error messages entirely in the ``CRASH`` file (see issue #890).

Previous versions already had this file, but in this PR we make sure that if the ``CRASH`` file is not printed, while the error is printed in the ``stdout``, we still detect the error.

The ``CRASH`` file is added to the retrieved list.


src/aiida_quantumespresso/parsers/parse_raw/pw.py
```diff
@@ -359,7 +361,10 @@ def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None)
         except NameError:  # nat or other variables where not found, and thus not initialized
 
             # Try to get some error messages
-            lines = stdout.split('\n')
+            if crash is None:
+                lines = stdout.split('\n')
+            else:
+                lines = crash.split('\n')
```
sphuber, line 367:
```suggestion
            lines = crash.split('\n') if crash is None else stdout.split('\n')
```

src/aiida_quantumespresso/parsers/parse_raw/pw.py
```diff
@@ -313,7 +314,8 @@ def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None)
         if 'JOB DONE' in line:
             break
     else:
-        logs.error.append('ERROR_OUTPUT_STDOUT_INCOMPLETE')
+        if crash is None:
+            logs.error.append('ERROR_OUTPUT_STDOUT_INCOMPLETE')
```
sphuber, line 318: Could merge the separate else and if
```suggestion
    elif crash is None:
        logs.error.append('ERROR_OUTPUT_STDOUT_INCOMPLETE')
```

src/aiida_quantumespresso/parsers/parse_raw/pw.py
```diff
@@ -279,11 +279,12 @@ def detect_important_message(logs, line):
             logs.warning.append(message)
 
 
-def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None):
+def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None, crash=None):
     """Parses the stdout content of a Quantum ESPRESSO `pw.x` calculation.
 
     :param stdout: the stdout content as a string
     :param input_parameters: dictionary with the input parameters
+    :param crash: the ``CRASH`` content as a string
```
sphuber, line 287:
```suggestion
    :param crash_file: the content of the ``CRASH`` file as a string if it was written, ``None`` otherwise.
```

src/aiida_quantumespresso/parsers/parse_raw/pw.py
```diff
@@ -279,11 +279,12 @@ def detect_important_message(logs, line):
             logs.warning.append(message)
 
 
-def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None):
+def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None, crash=None):
```
sphuber, line 282: Think `crash_file` or `crash_content` would be a bit clearer.
```suggestion
def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None, crash_file=None):
```

src/aiida_quantumespresso/parsers/pw.py
```diff
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 """`Parser` implementation for the `PwCalculation` calculation job class."""
+import os
```
sphuber, line 3: Not used I believe
```suggestion
```

tests/parsers/test_pw.py
```diff
@@ -210,6 +210,27 @@ def test_pw_failed_computing_cholesky(fixture_localhost, generate_calc_job_node,
     assert calcfunction.exit_status == node.process_class.exit_codes.ERROR_COMPUTING_CHOLESKY.status
 
 
+def test_pw_failed_computing_cholesky_stdout(
```
sphuber, line 213: I see that you only added this for the failed cholesky test, but not for the other tests where you moved the error from the stdout to the crash file. What is your reasoning here? Couldn't we have just kept the old tests, parsing from stdout, and _add_ a test where the error is (only) in the crash file to test cases for QE 7.2?

sphuber (changes requested):
Thanks @bastonero , left some comments/suggestions

src/aiida_quantumespresso/parsers/pw.py
bastonero: Yes, I imported it cause at the beginning I put ``CRASH`` in the temporary folder. Don't know why the linter did not detect it.

src/aiida_quantumespresso/parsers/pw.py
```diff
@@ -322,14 +329,14 @@ def parse_stdout(self, parameters, parser_options=None, parsed_xml=None):
...
-            parsed_data, logs = parse_stdout(stdout, parameters, parser_options, parsed_xml)
+            parsed_data, logs = parse_stdout(stdout, parameters, parser_options, parsed_xml, crash)
         except Exception as exc:
             logs.critical.append(traceback.format_exc())
             self.exit_code_stdout = self.exit_codes.ERROR_UNEXPECTED_PARSER_EXCEPTION.format(exception=exc)
 
         # If the stdout was incomplete, most likely the job was interrupted before it could cleanly finish, so the
         # output files are most likely corrupt and cannot be restarted from
-        if 'ERROR_OUTPUT_STDOUT_INCOMPLETE' in logs['error']:
+        if 'ERROR_OUTPUT_STDOUT_INCOMPLETE' in logs['error'] and crash is None:
```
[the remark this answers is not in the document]
bastonero: Correct, thanks!

tests/parsers/test_pw.py
bastonero: I thought, as this will become the standard, to put all the errors in the CRASH, which seemed cleaner to me - no particular reasoning though. 

I replicated only one error as I would assume that if it works for one, it will for all the others - thus no need to have more tests. Or do you think we should have duplicate tests both for CRASH and stdout?

bastonero: Ok, because I also wanted to add more errors in a next PR, which are belonging to diagonalization errors (for `ppcg` and `paro` mainly).

So, should we leave it like this for now?

src/aiida_quantumespresso/parsers/parse_raw/pw.py
bastonero: I actually get an identation error when using `elif` instead of `else`. Moving it inside the for loop does not give the same results.

tests/parsers/test_pw.py
bastonero: I didn't notice that the `vcrelax` errors had different folder names. While moving their error in CRASH, the tests are not passing. So, I think at this point it is convinient to have duplicated tests?

src/aiida_quantumespresso/parsers/parse_raw/pw.py
sphuber: Well, you also have to dedent the `logs.error.append` statement properly. It is a 100% fine to do 
```python
if ..:
    something
elif ...:
    something
```

bastonero: I agree, but there is a `break` statement before, so placing `elif` inside the main loop (indenting it) produces a different result (i.e. tests are not passing).

bastonero: Ok, got the fix for it w/o `elif`

sphuber: Ah, I see now. The `else` was not the secondary clause to an `if` but a `for` loop. Indeed, then `elif` is not valid syntax. Fine to leave as is then

src/aiida_quantumespresso/parsers/parse_raw/pw.py
```diff
@@ -921,6 +923,23 @@ def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None)
...
+        lines = crash_file.split('\n')
+
+        for line_number, line in enumerate(lines):
+            # Compare the line to the known set of error and warning messages and add them to the log container
+            detect_important_message(logs, line)
+
+        if len(logs.error) or len(logs.warning) > 0:
+            parsed_data['bands'] = bands_data
+            parsed_data['structure'] = structure_data
+            parsed_data['trajectory'] = trajectory_data
```
sphuber, line 937: These are already set in lines 922 - 924 and since you don't change them, no need to set again., right?

src/aiida_quantumespresso/parsers/parse_raw/pw.py
```diff
@@ -921,6 +923,23 @@ def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None)
     parsed_data['structure'] = structure_data
     parsed_data['trajectory'] = trajectory_data
 
+    # Double check to detect error messages if CRASH is found
+    if crash_file is not None:
+        lines = crash_file.split('\n')
+
+        for line_number, line in enumerate(lines):
+            # Compare the line to the known set of error and warning messages and add them to the log container
+            detect_important_message(logs, line)
```
sphuber, line 932: I think in older versions of QE, errors are written to _both_ the CRASH file and stdout. In those cases, you are duplicating the error messages in the container. This duplication was fixed in lines 919 and 920, but you are undoing that here.

src/aiida_quantumespresso/parsers/parse_raw/pw.py
```diff
@@ -921,6 +923,23 @@ def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None)
...
+
+        if len(logs.error) or len(logs.warning) > 0:
+            parsed_data['bands'] = bands_data
+            parsed_data['structure'] = structure_data
+            parsed_data['trajectory'] = trajectory_data
+            return parsed_data, logs
+
+        # did not find any error message -> raise an Error and do not return anything
+        raise QEOutputParsingError('Parser cannot load basic info.')
```
sphuber, line 941: With the other comments in mind, I would change this maybe to the following. Note that I added the lines to deduplicate the log messages. These should be removed from lines 919 - 920
```suggestion

    # Double check to detect error messages if CRASH is found
    if crash_file is not None:
        for line_number, line in enumerate(crash_file.split('\n')):
            # Compare the line to the known set of error and warning messages and add them to the log container
            detect_important_message(logs, line)

        if len(logs.error) == 0 and len(logs.warning) == 0:
            raise QEOutputParsingError('The calculation crashed but the relevant error could not be parsed.')

    # Remove duplicate log messages by turning it into a set. Then convert back to list as that is what is expected
    logs.error = list(set(logs.error))
    logs.warning = list(set(logs.warning))
```

sphuber (changes requested):

src/aiida_quantumespresso/parsers/parse_raw/pw.py
bastonero: Correct, thanks!

tests/parsers/test_pw.py
bastonero: In the last commit, there are tests for both ``CRASH`` and ``stdout`` cases

sphuber (approved):

PR #901, opened 2023-04-06
Add more `exit_codes` to `PwCalculation`

bastonero:
As from QuantumESPRESSO `v6.6`, new diagonalization algorithms are implemented in `pw.x`, among which `ppcg` and `paro` (a short summary [here](https://wordpress.elsi-interchange.org/wp-content/uploads/2018/08/de_Gironcoli_Stefano-MolSSI_ELSI_2018-Richmond.pdf)).

All these algorithms come with some error messages in the routines:

- `davidson`: overlap S matrix not positive definite
- `ppcg`: `[Q, R]` or `zhegvd` failed

In this PR we add exit codes for detecting such error messages, plus other error messages.

Diagonalization errors have been put in the [400,500] number region, as if they failed one can switch to a different diagonalization algorithm (e.g. david -> ppcg -> paro -> cg).

Other error messages (which I encountered) do not have at the moment a possible handler (actually trying to use `cg` diagonalization might be helpful). They are therefore put in the [300,400] number region.

I attach here a [qe_error_messages.tar.gz](https://github.com/aiidateam/aiida-quantumespresso/files/11172671/qe_error_messages.tar.gz) containing the actual full output of such errors, so that one can cross-verify the full error, since in the tests I did not report the entire output.


src/aiida_quantumespresso/parsers/parse_raw/pw.py
```diff
@@ -250,6 +250,12 @@ def detect_important_message(logs, line):
             'problems computing cholesky': 'ERROR_COMPUTING_CHOLESKY',
             'dexx is negative': 'ERROR_DEXX_IS_NEGATIVE',
             'too many bands are not converged': 'ERROR_DIAGONALIZATION_TOO_MANY_BANDS_NOT_CONVERGED',
+            'S matrix not positive definite': 'ERROR_S_MATRIX_NOT_POSITIVE_DEFINITE',
+            'zhegvd failed': 'ERROR_ZHEGVD_FAILED',
+            '[Q, R] = qr(X, 0) failed': 'ERROR_QR_FAILED',
+            'probably because G_par is NOT a reciprocal lattice vector': 'ERROR_G_PAR',
+            'eigenvectors failed to converge': 'ERROR_EIGENVECTOR_CONVERGENCE',
+            'factorization': 'ERROR_BROYDEN_FACTORIZATION',
```
bastonero, line 258: I am just wondering here if the regex would detect something else (e.g. a line containing this word). Wondering if there are lines having this word. From QE v7.2 this is eventually not a problem, as it is put nevertheless in the CRASH.

mbercx: Not 100% sure of this, but isn't the `Error in routine broyden (1):` error always the `factorization` one? :)

sphuber:
> Diagonalization errors have been put in the [400,500] number region, as if they failed one can switch to a different diagonalization algorithm (e.g. david -> ppcg -> paro -> cg).

If the calculation fails with this error, it is "useless", is it not? Useless in the sense that you cannot restart from it and essentially have to restart from scratch with another diagonlization. If that is the case, a 300 would be more apt.

src/aiida_quantumespresso/parsers/parse_raw/pw.py
bastonero: Yes, so technically not a problem, correct?

mbercx: I mean you can look in the `stdout` for the `Error in routine broyden (1):` string. ^^ That's much less likely to appear elsewhere.

bastonero:
> If the calculation fails with this error, it is "useless", is it not? Useless in the sense that you cannot restart from it and essentially have to restart from scratch with another diagonlization. If that is the case, a 300 would be more apt.

True, but currently if the exit number is <400 it is considered as "not solvable". As e.g. CHOLESKY_ERROR, or TOO_MANY_BANDS, which are also useless, but one restart with a different diagonalization algorithm (currently ``cg``). What do you think? Shall we leave it above 400 or change the logic?

sphuber:
> True, but currently if the exit number is <400 it is considered as "not solvable". As e.g. CHOLESKY_ERROR, or TOO_MANY_BANDS, which are also useless, but one restart with a different diagonalization algorithm (currently cg). What do you think? Shall we leave it above 400 or change the logic?

Yeah, you are right, ok to keep in the 400s.

bastonero:
There seems to be an error with the intersphinx with aiida-core. Any suggestion how to solve?

tests/parsers/fixtures/pw/failed_g_par/CRASH
sphuber, line 1: Can you please reduce the size of the `failed_g_par` reference files to the bare minimum as you did for the other failure test cases?

src/aiida_quantumespresso/calculations/pw.py
```diff
@@ -111,6 +111,8 @@ def define(cls, spec):
                     'scheduler before the files were safely written to disk for a potential restart.')
         spec.exit_code(350, 'ERROR_UNEXPECTED_PARSER_EXCEPTION',
             message='The parser raised an unexpected exception: {exception}')
+        spec.exit_code(360, 'ERROR_G_PAR',
+            message='The code failed in finding a valid reciprocal lattice vector.')
```
sphuber, line 115: Can you please explain when/why this error occurs and why this is in 3xx instead of 46x like the others?

sphuber (commented):
Just one final requested change and a question @bastonero . After that, fine for me to merge.

sphuber:
> There seems to be an error with the intersphinx with aiida-core. Any suggestion how to solve?

I opened #910 which I think should solve the problem. Once that is merged, I will update this branch and merge.

sphuber (approved):
Thanks @bastonero

PR #903, opened 2023-04-06
Devops: Update `pylint` version

mbercx:
Besides bumping the version of `pylint` in `pyproject.toml`:

- Set the `pylint` `repo` to `local` in the pre-commit config.
- Remove the deprecated `bad-continuation` check from the disabled checks.
- Disable the `fixme` check to avoid complaints about `TODO` statements.
- Disable the `use-dict-literal` check; using literal dictionaries is slightly faster, but I find the `dict()` calls easier to read in some cases.

### DIFF
--- a/src/aiida_quantumespresso/cli/utils/validate.py
+++ b/src/aiida_quantumespresso/cli/utils/validate.py
@@ -60,11 +60,11 @@ def validate_hubbard_parameters(structure, parameters, hubbard_u=None, hubbard_v
 
         try:
             hubbard_file = load_node(pk=hubbard_file_pk)
-        except exceptions.NotExistent:
-            ValueError(f'{hubbard_file_pk} is not a valid pk')
+        except exceptions.NotExistent as exc:
+            raise ValueError(f'{hubbard_file_pk} is not a valid pk') from exc
         else:
             if not isinstance(hubbard_file, SinglefileData):
-                ValueError(f'Node<{hubbard_file_pk}> is not a SinglefileData but {type(hubbard_file)}')
+                raise ValueError(f'Node<{hubbard_file_pk}> is not a SinglefileData but {type(hubbard_file)}')
 
         parameters['SYSTEM']['lda_plus_u'] = True
         parameters['SYSTEM']['lda_plus_u_kind'] = 2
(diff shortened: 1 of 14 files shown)


src/aiida_quantumespresso/cli/utils/validate.py
```diff
@@ -60,11 +60,11 @@ def validate_hubbard_parameters(structure, parameters, hubbard_u=None, hubbard_v
...
         try:
             hubbard_file = load_node(pk=hubbard_file_pk)
-        except exceptions.NotExistent:
-            ValueError(f'{hubbard_file_pk} is not a valid pk')
+        except exceptions.NotExistent as exc:
+            raise ValueError(f'{hubbard_file_pk} is not a valid pk') from exc
         else:
             if not isinstance(hubbard_file, SinglefileData):
-                ValueError(f'Node<{hubbard_file_pk}> is not a SinglefileData but {type(hubbard_file)}')
+                raise ValueError(f'Node<{hubbard_file_pk}> is not a SinglefileData but {type(hubbard_file)}')
```
sphuber, line 67: These are actually bugs, so we should probably update the tests for these. Probably in a separate PR. @bastonero could you take care of this please?

tests/parsers/test_pp.py
```diff
@@ -141,7 +141,10 @@ def test_pp_default_1d(
         'data_array': data_array,
         'coords_array': coords_array
     },
-                         default_tolerance=dict(atol=0, rtol=1e-18))
+                         default_tolerance={
+                             'atol': 0,
+                             'rtol': 1e-18
+                         })
```
sphuber, line 147: Holy indentation batman! I think if you manually indent this, `yapf` should accept it just fine. This is craziness. Same for the other occurrences of this keyword in the tests.

sphuber (changes requested):
All good, just please fix the indentation of the `defaul_tolerances` in the parser test.

tests/parsers/test_pp.py
mbercx: `yapf` was being very stubborn, but it seems adding the first input argument as a keyword did the trick. 😎

src/aiida_quantumespresso/cli/utils/validate.py
bastonero: Sure! Next days I am on holidays, so probably won't be able to prepare the PR. I will look at this next Tuesday. So just adding tests for the `hubbard_file` and these lines?

sphuber (approved):

PR #906, opened 2023-04-09
`PwCalculation`: Fix restart validation for `nscf`/`bands`

### LINKED ISSUE #905: `PwCalculation`: Fix restart validation for `nscf`/`bands`
The `PwCalculation` input validation checks that for `nscf` and `bands` calculations the `restart_mode` is not explicitly set to `restart`:

https://github.com/aiidateam/aiida-quantumespresso/blob/5cae75ff671268dc1e5684e1c84f801a10839e0b/src/aiida_quantumespresso/calculations/pw.py#L160-L171

However, setting `restart_mode` to `restart` is a valid use case for interrupted `nscf` and `bands` runs (see https://github.com/aiidateam/aiida-quantumespresso/pull/902#issuecomment-1501149498). So this validation should be removed.

mbercx:
Fixes #905

For interrupted runs, restarting from a previous `nscf`/`bands` run with `CONTROL.restart_mode` set to `restart` is a valid use case. Here we remove the validation that raised a warning when the user tried to restart an interrupted run with this input.

### DIFF
--- a/src/aiida_quantumespresso/calculations/pw.py
+++ b/src/aiida_quantumespresso/calculations/pw.py
@@ -175,21 +175,16 @@ class PwCalculation(BasePwCpInputGenerator):
         parameters = value['parameters'].get_dict()
         calculation_type = parameters.get('CONTROL', {}).get('calculation', 'scf')
 
-        # Check that the restart input parameters are set correctly
-        if calculation_type in ('nscf', 'bands'):
-            if parameters.get('ELECTRONS', {}).get('startingpot', 'file') != 'file':
-                return f'`startingpot` should be set to `file` for a `{calculation_type}` calculation.'
-            if parameters.get('CONTROL', {}).get('restart_mode', 'from_scratch') != 'from_scratch':
-                warnings.warn(f'`restart_mode` should be set to `from_scratch` for a `{calculation_type}` calculation.')
-        elif 'parent_folder' in value:
+        # If a `parent_folder` input is provided, make sure the inputs are set to restart
+        if 'parent_folder' in value and calculation_type not in ('nscf', 'bands'):
             if not any([
                 parameters.get('CONTROL', {}).get('restart_mode', None) == 'restart',
                 parameters.get('ELECTRONS', {}).get('startingpot', None) == 'file',
                 parameters.get('ELECTRONS', {}).get('startingwfc', None) == 'file'
             ]):
                 warnings.warn(
-                    '`parent_folder` input was provided for the `PwCalculation`, but no '
-                    'input parameters are set to restart from these files.'
+                    f'`parent_folder` input was provided for the `{calculation_type}` `PwCalculation`, but no input'
+                    'parameters (e.g. `restart_mode`, `startingpot`, ...) are set to restart from this folder.'
                 )
 
     @classmethod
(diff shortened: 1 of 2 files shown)


src/aiida_quantumespresso/calculations/pw.py
```diff
@@ -163,31 +163,25 @@ def validate_inputs_base(value, _):
...
             if not any([
                 parameters.get('CONTROL', {}).get('restart_mode', None) == 'restart',
                 parameters.get('ELECTRONS', {}).get('startingpot', None) == 'file',
                 parameters.get('ELECTRONS', {}).get('startingwfc', None) == 'file'
             ]):
                 warnings.warn(
-                    '`parent_folder` input was provided for the `PwCalculation`, but no '
-                    'input parameters are set to restart from these files.'
+                    f'`parent_folder` input was provided for the `{calculation_type}` `PwCalculation`, but no input'
+                    'parameters (e.g. `restart_mode`, `startingpot`, ...) are set to restart from this folder.'
```
sphuber, line 175: Maybe we should be 100% clear here and actually spell it out.
```
Please set one of the following in the input parameters:
    parameters['CONTROL']['restart_mode'] = 'restart'
    parameters['ELECTRONS']['startingpot'] = 'file'
    parameters['ELECTRONS']['startingwfc'] = 'file'
```
Or, maybe we create a dedicated page in the documentation for dealing with all these restart minutia and simply link there. I think that will be best, because for one, that will start concentrating all knowledge in one place (you have done great work of elucidating all functionality, but now it is spread across various PRs and issues), and two, there will be other places where we will have to explain, so it is easier to be able to refer to one place.

src/aiida_quantumespresso/calculations/pw.py
```diff
@@ -163,31 +163,25 @@ def validate_inputs_base(value, _):
...
 
-        # Check that the restart input parameters are set correctly
-        if calculation_type in ('nscf', 'bands'):
-            if parameters.get('ELECTRONS', {}).get('startingpot', 'file') != 'file':
-                return f'`startingpot` should be set to `file` for a `{calculation_type}` calculation.'
-            if parameters.get('CONTROL', {}).get('restart_mode', 'from_scratch') != 'from_scratch':
-                warnings.warn(f'`restart_mode` should be set to `from_scratch` for a `{calculation_type}` calculation.')
-        elif 'parent_folder' in value:
+        # If a `parent_folder` input is provided, make sure the inputs are set to restart
+        if 'parent_folder' in value and calculation_type not in ('nscf', 'bands'):
```
sphuber, line 167: Shouldn't an `scf` calculation also not have "some" restart parameters if a `parent_folder` is specified? Why only `nscf` and `bands`? What is the use case for restarting an `scf` calc `from_scratch` and without reading the CD from file?

sphuber (changes requested):

src/aiida_quantumespresso/calculations/pw.py
mbercx: Got confused myself here again. Note the `not in ('nscf', 'bands')`, i.e. this check is only done for _other_ calculation types. The reason is that for `nscf` and `bands`, the default of `ELECTRONS.startingpot` changes to `'file'`, see:

https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm899

src/aiida_quantumespresso/calculations/pw.py
mbercx: I like the idea of being more explicit. I also agree we should gather all restarting concepts in one documentation page, but as this is quite an extensive exercise I've opened an issue for this instead:

https://github.com/aiidateam/aiida-quantumespresso/issues/916

sphuber (dismissed):

src/aiida_quantumespresso/calculations/pw.py
sphuber: Still not sure about the logic though. Why would there only have to be validation for anything but `nscf` and `bands`? For example, if the user specifies `bands` but then `ELECTRONS.startingpot = 'atomic'`, surely that should also emit a warning then?

mbercx: Sure, that'd be another validation we can do. Much less likely use case though, someone what assigns a parent folder and then specifically sets `ELECTRONS.startingpot = 'atomic'` might be up to some weird business that actually requires this input.

PR #909, opened 2023-04-12
Fix `cli` for hubbard inputs

bastonero:
In this PR, we add test to the `cli.calculations.pw` for hubbard input, which remained untested till now (as reported in #903 ).


src/aiida_quantumespresso/cli/utils/validate.py
```diff
@@ -59,9 +59,9 @@ def validate_hubbard_parameters(structure, parameters, hubbard_u=None, hubbard_v
     if hubbard_file_pk:
 
         try:
-            hubbard_file = load_node(pk=hubbard_file_pk)
-        except exceptions.NotExistent as exc:
-            raise ValueError(f'{hubbard_file_pk} is not a valid pk') from exc
+            hubbard_file = load_node(pk=hubbard_file_pk.pk)
```
bastonero, line 62: Actually, I am not sure why here the `hubbard_file_pk` is loaded instead of being passed as an integer. At least, that what it seems happening in the tests.

sphuber: Because that value comes from the `hubbard_file_pk` argument in the CLI command which is provided by the `HUBBARD_FILE` reusable option. That option takes an identifier on the command line (pk, uuid or label), but it _returns_ the corresponding node, already loaded. I would recommend changing the variable name of that `HUBBARD_FILE` option from `hubbard_file_pk` to `hubbard_file` and in the `validate_hubbard_parameters` as well. The whole conditional `if hubbard_file_pk` can be removed, because it will already pass a node in that is guaranteed to be a `SinglelfileNode. Also update the docs of this method that `hubbard_file` is a `SinglefileData` with Hubbard parameters and not the pk of one.

tests/cli/calculations/test_pw.py
```diff
@@ -8,3 +12,42 @@ def test_command_base(run_cli_process_launch_command, fixture_code, sssp):
...
+    options = ['-X', code.full_label, '-F', sssp.label, '--hubbard-u', 'Si', '5.0']
+    run_cli_process_launch_command(launch_calculation, options=options)
+
+    content_original = 'a very creative content'
+
+    with tempfile.NamedTemporaryFile(mode='w+') as handle:
+        filepath = pathlib.Path(handle.name).resolve()
+        handle.write(content_original)
+        handle.flush()
+        filepk = SinglefileData(file=filepath).store().pk
```
sphuber, line 50: This paradigm can be accomplished more succinctly using streams
```suggestion
    import io
    filepk = SinglefileData(io.StringIO(content_original).store().pk
```

tests/cli/calculations/test_pw.py
```diff
@@ -8,3 +12,42 @@ def test_command_base(run_cli_process_launch_command, fixture_code, sssp):
...
+    """Test invoking the calculation launch command with only required inputs."""
+    code = fixture_code('quantumespresso.pw').store()
+    options = ['-X', code.full_label, '-F', sssp.label] + cmd_options
+    result = run_cli_process_launch_command(launch_calculation, options=options, raises=ValueError)
+    assert re.match(match, ' '.join(result.output_lines))
+
+
+@pytest.mark.usefixtures('aiida_profile')
+def test_valid_hubbard_parameters(run_cli_process_launch_command, fixture_code, sssp):
+    """Test invoking the calculation launch command with only required inputs."""
```
sphuber, line 33: Please update docstring

tests/cli/calculations/test_pw.py
```diff
@@ -8,3 +12,42 @@ def test_command_base(run_cli_process_launch_command, fixture_code, sssp):
...
+
+
+@pytest.mark.parametrize(('cmd_options', 'match'), (
+    (['--hubbard-u', 'Mg'], ".*Option '--hubbard-u' requires 2 arguments.*"),
+    (['--hubbard-u', 'Mg', '5.0'
+      ], '.*kinds in the specified Hubbard U is not a strict subset of the structure kinds.*'),
+    (['--hubbard-file', '1000000'], '.*no SinglefileData found with ID*'),
+))
+def test_invalid_hubbard_parameters(run_cli_process_launch_command, fixture_code, sssp, cmd_options, match):
+    """Test invoking the calculation launch command with only required inputs."""
```
sphuber, line 24: Same here

src/aiida_quantumespresso/cli/utils/validate.py
```diff
@@ -59,12 +59,12 @@ def validate_hubbard_parameters(structure, parameters, hubbard_u=None, hubbard_v
     if hubbard_file_pk:
 
         try:
-            hubbard_file = load_node(pk=hubbard_file_pk)
-        except exceptions.NotExistent:
-            ValueError(f'{hubbard_file_pk} is not a valid pk')
+            hubbard_file = load_node(pk=hubbard_file_pk.pk)
```
sphuber, line 62: See my comment to your question. It is best to just have the function accept a node instead of a pk. Now you are accepting the node, taking its pk, and loading the corresponding node again 😅 That is a round trip for nothing

sphuber (changes requested):

src/aiida_quantumespresso/cli/utils/validate.py
```diff
@@ -36,35 +35,30 @@ def validate_kpoints_mesh(ctx, param, value):
...
+    if hubbard_file:
 
-        try:
-            hubbard_file = load_node(pk=hubbard_file_pk)
-        except exceptions.NotExistent as exc:
-            raise ValueError(f'{hubbard_file_pk} is not a valid pk') from exc
-        else:
-            if not isinstance(hubbard_file, SinglefileData):
-                raise ValueError(f'Node<{hubbard_file_pk}> is not a SinglefileData but {type(hubbard_file)}')
+        if not isinstance(hubbard_file, SinglefileData):
```
bastonero, line 60: Last question: I still kept this one as we should make sure the user has loaded the correct node (i.e. a ``SingleFileData`` node). I will be check also in the ``PwCalculation`` validator, although I thought it would be nicer to be blocked beforehand, as it costs only two code lines.

sphuber: That check is also already done by the paramtype of the option. If you look at its declaration, it specifies the entry point of the singlefiledata type and so that is the type that is checked.

bastonero:
Thanks @sphuber !

PR #911, opened 2023-04-12
Having pw schema XML support all valide disk_io options

### LINKED ISSUE #908: The schema needs to updated to support more disk_io settings
When I used the `nowf` for pw calculation, which is now supported in pw.x https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm128 I got:

```
Report: [357392|PwBaseWorkChain|run_process]: launching PwCalculation<357395> iteration #1
Error: 1 XML schema validation error(s) schema: /home/jyu/micromamba/envs/aiida-sssp-unstable/lib/python3.9/site-packages/aiida_quantumespresso/parsers/parse_xml/schemas/qes_220603.xsd:
Error: failed validating 'nowf' with XsdEnumerationFacets(['low', 'high']):

Reason: value must be one of ['low', 'high']

Schema:

  <enumeration xmlns="http://www.w3.org/2001/XMLSchema" value="low" />

Instance:

  <disk_io>nowf</disk_io>

Path: /{http://www.quantum-espresso.org/ns/qes/qes-1.0}espresso/input/control_variables/disk_io
```

It seems the options are not included in the latest schema at https://github.com/QEF/qeschemas/blob/master/PW_CPV/previous_schemas/qes_230310.xsd

The QE start support this from `6.6` https://gitlab.com/QEF/q-e/-/blob/qe-6.6/PW/Doc/INPUT_PW.html#L620

unkcpz:
fixes #908

### DIFF
--- a/tests/parsers/fixtures/pw/default_xml_220603/data-file-schema.xml
+++ b/tests/parsers/fixtures/pw/default_xml_220603/data-file-schema.xml
@@ -26,7 +26,7 @@
       <stress>true</stress>
       <forces>true</forces>
       <wf_collect>true</wf_collect>
-      <disk_io>nowf</disk_io>
+      <disk_io>low</disk_io>
       <max_seconds>41040</max_seconds>
       <nstep>1</nstep>
       <etot_conv_thr>1.000000000000000e-5</etot_conv_thr>


tests/parsers/fixtures/pw/default_xml_220603/data-file-schema.xml
```diff
@@ -26,7 +26,7 @@
       <stress>true</stress>
       <forces>true</forces>
       <wf_collect>true</wf_collect>
-      <disk_io>low</disk_io>
+      <disk_io>nowf</disk_io>
```
sphuber, line 29: As you mentioned in the corresponding issue, this change as a means to test the changes in the schema is perhaps not the best. If we really feel we need to test this, we add a dedicated test that takes an example output XML for each schema version, then parametrizes it over the valid values of `disk_io` and just calls the schema validation. So completely decouple it from the parser tests.

But I am not sure this is even needed as now we are testing the validity of the schema, which should be static. So I think that if you run an example calculation locally with these changes and disk_io set to nowf and it works, then we accept that without specific unit tests.
```suggestion
      <disk_io>low</disk_io>
```

sphuber (changes requested):

unkcpz:
Okay, thanks change it back. I test locally and works all good.

bastonero:
Passing by I'd suggest putting a reminder in the Wiki about this for future scheme releases. And maybe a test with a xml produced with a calculation specifying the ``disc_io`` with one of these options? (probably overtesting though)

PR #912, opened 2023-04-12
``PwBaseWorkChain``: improve diagonalization handler

### LINKED ISSUE #880: Alternative `process_handler` for Cholesky error
The "Choelsky error" is probably one of the most strange errors to handle in QuantumESRPESSO, since apparently there is not a straightforward solution. Currently, this error is handled by changing the `diagonalization` from `davidson` (default) to `cg`. Nevertheless, the conjugate gradient algorithm is VERY slow, and an other solution is desirable.

I tested for quite a while now a possible general alternative. Since QE exploits parallel diagonalization over N^2 cores, this might be affected by quite some numerical noise. Therfore, I tried to force this to be performed over 1 only core, ending up all the time to a successful run of the `pw.x` code. This alternative is still MUCH faster than the current CG solution, which can still be employed as a "last resort". 

This would mean  adding `['-ndiag','1']` to the `pw.settings` `cmdline` key in a consistent way.

Let me know your thoughts, and I hope this could help to have a more effective handler.

bastonero:
Fixes #880 

The current handler for diagonalization errors is switching 
from the default value (``david``) to conjugate gradient (``cg``). 
While being a safe solution, the ``cg`` algorithm is rather slow
compared to other options, now fully supported by the plugin (#901).
Therefore, we implement a new strategy, which simply switches to 
different algorithms before resorting to ``cg``.

### DIFF
--- a/src/aiida_quantumespresso/workflows/pw/base.py
+++ b/src/aiida_quantumespresso/workflows/pw/base.py
@@ -407,29 +407,24 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         still not used. Conjugate gradient (CG) is kept as last option, as it is the slowest among the
         available ones, but on the contrary it is the most stable as well, thus kept as `last resort`.
 
-        If also CG is failing, abort.
+        Once the error handler has tried all ``diagonalization`` options, abort.
         """
-        if 'diagonalizations' not in self.ctx:
-            # Initialize a diagonalizations list to keep track of inputs that haven't been tried in order or preference
-            self.ctx.diagonalizations = ['david', 'ppcg', 'paro', 'cg']
-
         current = self.ctx.inputs.parameters['ELECTRONS'].get('diagonalization', 'david')
-        try:
-            self.ctx.diagonalizations.remove(current)
-        except ValueError:  # if it is different from the one we support in the handler
-            pass
 
-        if not self.ctx.diagonalizations:
-            action = 'found diagonalization issues but already exploited different algorithms, aborting...'
+        if 'diagonalizations' not in self.ctx:
+            # Initialize a list to track diagonalisations that haven't been tried in reverse order or preference
+            self.ctx.diagonalizations = [value for value in ['cg', 'paro', 'ppcg', 'david'] if value != current.lower()]
+
+        try:
+            new = self.ctx.diagonalizations.pop()
+            self.ctx.inputs.parameters['ELECTRONS']['diagonalization'] = new
+            action = f'found diagonalization issues for ``{current}``, switching to ``{new}`` diagonalization.'
+            self.report_error_handled(calculation, action)
+            return ProcessHandlerReport(True)
+        except IndexError:
+            action = 'found diagonalization issues but already exploited all supported algorithms, aborting...'
             self.report_error_handled(calculation, action)
             return ProcessHandlerReport(True, self.exit_codes.ERROR_KNOWN_UNRECOVERABLE_FAILURE)
-
-        new = self.ctx.diagonalizations[0]
-        self.ctx.inputs.parameters['ELECTRONS']['diagonalization'] = new
-
-        action = f'found diagonalization issues for ``{current}``, switching to ``{new}`` diagonalization.'
-        self.report_error_handled(calculation, action)
-        return ProcessHandlerReport(True)
 
     @process_handler(priority=580, exit_codes=[
         PwCalculation.exit_codes.ERROR_OUT_OF_WALLTIME,
--- a/src/aiida_quantumespresso/workflows/pw/base.py
+++ b/src/aiida_quantumespresso/workflows/pw/base.py
@@ -238,7 +238,6 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         default namelists for the ``parameters`` are set to empty dictionaries if not specified.
         """
         super().setup()
-        self.ctx.diagonalizations = ['david', 'ppcg', 'paro', 'cg']  # ``cg`` as last resort
         self.ctx.inputs = AttributeDict(self.exposed_inputs(PwCalculation, 'pw'))
 
         self.ctx.inputs.parameters = self.ctx.inputs.parameters.get_dict()
@@ -410,14 +409,18 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
 
         If also CG is failing, abort.
         """
-        current = self.ctx.inputs.parameters['ELECTRONS'].get('diagonalization', 'david')
+        if 'diagonalizations' not in self.ctx:
+            # Initialize a diagonalizations list to keep track of inputs that haven't been tried in order or preference
+            self.ctx.diagonalizations = ['david', 'ppcg', 'paro', 'cg']
 
-        if current == 'cg':
-            action = 'found diagonalization issues but already running with conjugate gradient algorithm, aborting...'
+        current = self.ctx.inputs.parameters['ELECTRONS'].get('diagonalization', 'david')
+        self.ctx.diagonalizations.remove(current)
+
+        if not self.ctx.diagonalizations:
+            action = 'found diagonalization issues but already exploited different algorithms, aborting...'
             self.report_error_handled(calculation, action)
             return ProcessHandlerReport(True, self.exit_codes.ERROR_KNOWN_UNRECOVERABLE_FAILURE)
 
-        self.ctx.diagonalizations.pop(self.ctx.diagonalizations.index(current))
         new = self.ctx.diagonalizations[0]
         self.ctx.inputs.parameters['ELECTRONS']['diagonalization'] = new
 
--- a/src/aiida_quantumespresso/workflows/pw/base.py
+++ b/src/aiida_quantumespresso/workflows/pw/base.py
@@ -7,7 +7,7 @@ from aiida.engine import BaseRestartWorkChain, ExitCode, ProcessHandlerReport, p
 from aiida.plugins import CalculationFactory, GroupFactory
 
 from aiida_quantumespresso.calculations.functions.create_kpoints_from_distance import create_kpoints_from_distance
-from aiida_quantumespresso.common.types import ElectronicType, RestartType, SpinType
+from aiida_quantumespresso.common.types import DiagonalizationType, ElectronicType, RestartType, SpinType
 from aiida_quantumespresso.utils.defaults.calculation import pw as qe_defaults
 
 from ..protocols.utils import ProtocolMixin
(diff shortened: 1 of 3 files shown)


src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -405,21 +440,35 @@ def handle_known_unrecoverable_failure(self, calculation):
...
+        current_diagonalization = self.get_current_diagonalization()
+        if current_diagonalization == DiagonalizationType.CG:
             action = 'found diagonalization issues but already running with conjugate gradient algorithm, aborting...'
             self.report_error_handled(calculation, action)
             return ProcessHandlerReport(True, self.exit_codes.ERROR_KNOWN_UNRECOVERABLE_FAILURE)
 
-        self.ctx.inputs.parameters['ELECTRONS']['diagonalization'] = 'cg'
-        action = 'found diagonalization issues, switching to conjugate gradient diagonalization.'
+        self.ctx.diagonalizations.pop(self.ctx.diagonalizations.index(current_diagonalization))
+        if self.ctx.diagonalizations:
```
bastonero, line 466: Actually, I think we could simplify the handler logic just setting `CG` to be by default the last one in ``DiagonalizationType``.

mbercx:
Thanks @bastonero! Just gave the code a quick look, and I'm wondering if it's not a bit over-engineered. Can't we just define e.g. a dictionary that maps the current diagonalization setting to the one we want to use next, where `cg` is mapped to `None` that then can be caught to return the `ERROR_KNOWN_UNRECOVERABLE_FAILURE` exit status? I'm also not sure if the `DiagonalizationType` `Enum` adds a lot of value here, because it just corresponds directly to a single input tag for QE.

bastonero:
>  I'm wondering if it's not a bit over-engineered. 

Yes, I wondered that too, so thanks for the feedback.

> Can't we just define e.g. a dictionary that maps the current diagonalization setting to the one we want to use next, where `cg` is mapped to `None` that then can be caught to return the `ERROR_KNOWN_UNRECOVERABLE_FAILURE` exit status? 

Yes and no. At the beginning I also thought about this option, but then I wondered - what if one starts directly from e.g. ``paro`` and not ``david``? Then one might want to try ``david`` as well, but if we set a dictionary with a cyclic call, this won't happen. 

Let's imagine we have the following: david -> ppcg -> paro -> cg . Then if paro fails, it will call directly cg, and ppcg and david are not exploited. That's why the logic is more complicated.

> I'm also not sure if the `DiagonalizationType` `Enum` adds a lot of value here, because it just corresponds directly to a single input tag for QE.

That's true, it might be overcrowded. I just thought maybe we could use it in the future for something, and also it was a nice way to keep track of all the diagonalization options. Moreover, now in QE there are other two diagonalization options (namely 'rmm-davidson' and 'rmm-paro'). So having a `DiagonalizationType` seemed to me a nice way to have all of them listed. But  maybe it is not as necessary.

src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -405,21 +406,34 @@ def handle_known_unrecoverable_failure(self, calculation):
...
+        We use the following strategy. When a diagonalization algorithm fails, we try using an other one
+        still not used. Conjugate gradient (CG) is kept as last option, as it is the slowest among the
+        available ones, but on the contrary it is the most stable as well, thus kept as `last resort`.
+
+        If also CG is failing, abort.
         """
-        if self.ctx.inputs.parameters['ELECTRONS'].get('diagonalization', None) == 'cg':
+        current = self.ctx.inputs.parameters['ELECTRONS'].get('diagonalization', 'david')
+
+        if current == 'cg':
```
mbercx, line 427: Is a failure with `cg` truly hopeless? No chance that e.g. running with `paro` might work? We could just treat `cg` on an equal footing as the other diagonalization settings, keeping it last in the order of preference.

mbercx (changes requested):
Thanks @bastonero! Much cleaner in my view. I left some comments/suggestions.

EDIT: some more notes:

> Let's imagine we have the following: david -> ppcg -> paro -> cg . Then if paro fails, it will call directly cg, and ppcg and david are not exploited. That's why the logic is more complicated.

I see, thanks! I was thinking of a strict order of preference, didn't think of the use case of switching back to `david` if `paro` failed.

> That's true, it might be overcrowded. I just thought maybe we could use it in the future for something, and also it was a nice way to keep track of all the diagonalization options. Moreover, now in QE there are other two diagonalization options (namely 'rmm-davidson' and 'rmm-paro'). So having a DiagonalizationType seemed to me a nice way to have all of them listed. But maybe it is not as necessary.

Having worked my way through a lot of somewhat over-engineered and difficult to read AiiDA code, I've become rather apprehensive towards trying to make code more extensible for some hypothetical use case in the future if it makes the code more difficult to understand and maintain. 🙃 

Re the `DiagonalizationType`: I think it differs from e.g. `RestartType` in that the latter tries to convert _multiple_ QE inputs into something that is easier to grasp and work with (restarting in QE is still confusing for me :joy:). If `DiagonalizationType` would touch multiple QE inputs I'd be more inclined to use it.

src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -405,21 +406,34 @@ def handle_known_unrecoverable_failure(self, calculation):
...
-        Switch to ``diagonalization = 'cg'`` if not already running with this setting, and restart from the charge
-        density. In case the run already used conjugate gradient diagonalization, abort.
+        We use the following strategy. When a diagonalization algorithm fails, we try using an other one
+        still not used. Conjugate gradient (CG) is kept as last option, as it is the slowest among the
+        available ones, but on the contrary it is the most stable as well, thus kept as `last resort`.
+
+        If also CG is failing, abort.
         """
-        if self.ctx.inputs.parameters['ELECTRONS'].get('diagonalization', None) == 'cg':
+        current = self.ctx.inputs.parameters['ELECTRONS'].get('diagonalization', 'david')
```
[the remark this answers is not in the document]
sphuber: Is the reason for choosing `david` as the default in line 429 because it is the default in QE? Has that always been the case for all versions? Do we parse this from the output? If so, it might be safer to get the current value from the parsed outputs if not explicitly defined in the input parameters

src/aiida_quantumespresso/workflows/pw/base.py
bastonero: Yes, one can definitely try and see if any of the available ones is working

src/aiida_quantumespresso/workflows/pw/base.py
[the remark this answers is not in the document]
bastonero: I am rather confident `david` is the default for all the versions, or for sure the versions the plugin is supporting right now. But I'll check the output, and take that one in case

src/aiida_quantumespresso/workflows/pw/base.py
mbercx: Simple trial and error, love it. :D Frankly I don't remember a case where `david` failed and `paro` didn't fix the issue, so it's probably a bit overkill, but 🤷‍♂️

bastonero: Right, forgot about reason for the ordering - no one in particular, but for some cases (magnetic ones), I found `ppcg` to be more stable. But again, anything better than `cg`

src/aiida_quantumespresso/workflows/pw/base.py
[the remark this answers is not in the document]
sphuber: If it is not in the outputs, fine to just hard-code it then for now

bastonero:
@mbercx @sphuber do you have other comments/suggestions? Otherwise, can this be merged for the next release?

src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -405,21 +405,41 @@ def handle_known_unrecoverable_failure(self, calculation):
...
-        if self.ctx.inputs.parameters['ELECTRONS'].get('diagonalization', None) == 'cg':
-            action = 'found diagonalization issues but already running with conjugate gradient algorithm, aborting...'
+        if 'diagonalizations' not in self.ctx:
+            # Initialize a diagonalizations list to keep track of inputs that haven't been tried in order or preference
+            self.ctx.diagonalizations = ['david', 'ppcg', 'paro', 'cg']
+
+        current = self.ctx.inputs.parameters['ELECTRONS'].get('diagonalization', 'david')
+        try:
+            self.ctx.diagonalizations.remove(current)
+        except ValueError:  # if it is different from the one we support in the handler
```
mbercx, line 431: Right, `remove` _always_ raises when the value isn't present, and we don't support some newer QE options yet. I was tempted to use a `set` instead, but of course we care about the order.

I know that the `try`-`except` implementation is "thread-safe" and respects one of my favorite rules in life (Easier to ask forgiveness than permission), but I'm still _somewhat_ in favor of the `if` approach:

```python
        if current in self.ctx.diagonalizations:
            self.ctx.diagonalizations.remove(current)
```

It's just more readable IMO, which is also part of the Zen of Python. Not critical though, so either is fine.

More importantly, this reminds me that there is also the possibility to set `'ParO'`, so we probably should add a `.lower()` to the `current`.

sphuber (dismissed):
Thanks @bastonero . I will approve, but will leave the final approval and merging to @mbercx

src/aiida_quantumespresso/workflows/pw/base.py
mbercx: Ok, I still wasn't happy with the remove thing, so how about this: 😅 

```python
        current = self.ctx.inputs.parameters['ELECTRONS'].get('diagonalization', 'david')

        if 'diagonalizations' not in self.ctx:
            # Initialize a list to track diagonalisations that haven't been tried in reverse order or preference
            self.ctx.diagonalizations = [value for value in ['cg', 'paro', 'ppcg', 'david'] if value != current.lower()]

        try:
            new = self.ctx.diagonalizations.pop()
            self.ctx.inputs.parameters['ELECTRONS']['diagonalization'] = new
        except IndexError:
            action = 'found diagonalization issues but already exploited all supported algorithms, aborting...'
            return ProcessHandlerReport(True, self.exit_codes.ERROR_KNOWN_UNRECOVERABLE_FAILURE)
```

To me this is more readable and Pythonic, but this may be a matter of preference. ^^

mbercx (changes requested):
I admit I've gone into nitpick mode at this point, @bastonero. Have a look and let me know what you think!

src/aiida_quantumespresso/workflows/pw/base.py
bastonero: Looks much cleaner, thanks for the review!

mbercx (approved):
beautiful

mbercx:
Seems you didn't have the permissions after all, but thanks for updating the OP to the desired commit message, @bastonero!

Issue #752, opened 2021-10-12
`PwCalculation`: Add exit code + error handling for `scale_h` error


mbercx:
If during a `vc-relax` calculation, the volume contracts significantly:

```
$ verdi calcjob outputcat 392190 | grep volume
     unit-cell volume          =    2126.3863 (a.u.)^3
unit cell volume :      2126.38625529 [a.u.^3]
MT volume        : 0.000000 ( 0.00%)
IT volume        : 2126.386255 (100.00%)
     new unit-cell volume =   2100.60741 a.u.^3 (   311.27790 Ang^3 )
     new unit-cell volume =   2062.13363 a.u.^3 (   305.57668 Ang^3 )
     new unit-cell volume =   2004.26640 a.u.^3 (   297.00164 Ang^3 )
     new unit-cell volume =   1917.86434 a.u.^3 (   284.19817 Ang^3 )
     new unit-cell volume =   1792.43954 a.u.^3 (   265.61214 Ang^3 )
     new unit-cell volume =   1613.65390 a.u.^3 (   239.11884 Ang^3 )
     new unit-cell volume =   1460.74585 a.u.^3 (   216.46020 Ang^3 )
     new unit-cell volume =   1474.38804 a.u.^3 (   218.48177 Ang^3 )
     new unit-cell volume =   1472.33596 a.u.^3 (   218.17768 Ang^3 )
     new unit-cell volume =   1461.82518 a.u.^3 (   216.62014 Ang^3 )
     new unit-cell volume =   1445.38252 a.u.^3 (   214.18359 Ang^3 )
     new unit-cell volume =   1419.31352 a.u.^3 (   210.32056 Ang^3 )
     new unit-cell volume =   1377.85011 a.u.^3 (   204.17632 Ang^3 )
     new unit-cell volume =   1311.32208 a.u.^3 (   194.31788 Ang^3 )
     new unit-cell volume =   1203.11997 a.u.^3 (   178.28399 Ang^3 )
     new unit-cell volume =   1102.12458 a.u.^3 (   163.31801 Ang^3 )
     new unit-cell volume =    993.34300 a.u.^3 (   147.19825 Ang^3 )
     new unit-cell volume =    886.18203 a.u.^3 (   131.31863 Ang^3 )
```

this results in the following error:

```
[...]

PR #913, opened 2023-04-13
``PwCalculation``: add exit code and handler for ``scale_h`` error

bastonero:
This PR adds the exit code and handler for the ``scale_h`` error presented in #752.


src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -468,6 +468,7 @@ def handle_vcrelax_converged_except_final_scf(self, calculation):
             PwCalculation.exit_codes.ERROR_IONIC_CYCLE_EXCEEDED_NSTEP,
             PwCalculation.exit_codes.ERROR_IONIC_CYCLE_BFGS_HISTORY_FAILURE,
             PwCalculation.exit_codes.ERROR_IONIC_CYCLE_BFGS_HISTORY_AND_FINAL_SCF_FAILURE,
+            PwCalculation.exit_codes.ERROR_SCALE_H,
```
bastonero, line 471: @sphuber as this handler is already doing the job, I simply added the exit code of `scale_h` here. Shall we have a separate one for which in case the `cell_factor` might be exploited?

bastonero:
> Also, another thing I just realized. With your recent changes to parsing the `CRASH` file and each error test having two versions (one with CRASH file, and one with stdout) the reference files are being duplicated, adding a lot of extra weight to the repo. I wonder if we can simply use relative symlinks to avoid the content duplication. Could you try to replace the `stdout` copy of the stdout and XML files with symlinks to the defualt one?

Sure. Nevertheless, I think we can do it only for XML, as the stdout will differ depending whether the error output is in the stdout or in CRASH. Therefore I guess having duplicated stdout is unavoidable. What do you think?

src/aiida_quantumespresso/workflows/pw/base.py
bastonero: Yes, it is for pw.x ([here](https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm1152) the definition). It seems indeed that restarting from the output structure is fine. Alternatively, we might consider increasing its value (e.g. double it to 4).

tests/parsers/test_pw.py
```diff
@@ -1093,7 +1093,27 @@ def test_pw_vcrelax_failed_not_converged_nstep(
...
+    inputs = generate_inputs(calculation_type='vc-relax')
+    node = generate_calc_job_node(entry_point_calc_job, fixture_localhost, name, inputs)
+    parser = generate_parser(entry_point_parser)
+    results, calcfunction = parser.parse_from_node(node, store_provenance=False)
+    expected_exit_status = node.process_class.exit_codes.ERROR_RADIAL_FFT_SIGNIFICANT_VOLUME_CONTRACTION.status
+
+    assert calcfunction.is_failed
+    assert calcfunction.exit_status == expected_exit_status
+    assert orm.Log.collection.get_logs_for(node)
+    assert 'output_structure' in results
```
bastonero, line 1116: Not sure why here the test is passing, while the xml was removed from the fixture folder.

sphuber:
> stdout will differ depending whether the error output is in the stdout or in CRASH.

Is this because in the newer versions of QE, the error information is _only_ in `CRASH` and not the `stdout`? At least before, it used to be in both, correct? But if so, and the stdout no longer contains the error info in latest QE versions, then yes, you will have to duplicate, but that is not too bad since anyway you probably just want to include the minimal output in the stdout since we are just testing the error is parsed correctly. Don't really care that all other parameters are correctly parsed.

tests/parsers/test_pw.py
bastonero: I see, but as there's no __structure__ information in the stdout, how is it that there is an ``output_structure`` in the results?

src/aiida_quantumespresso/calculations/pw.py
```diff
@@ -146,6 +146,9 @@ def define(cls, spec):
             message='Then ionic minimization cycle converged but the thresholds are exceeded in the final SCF.')
         spec.exit_code(502, 'ERROR_IONIC_CYCLE_EXCEEDED_NSTEP',
             message='The ionic minimization cycle did not converge after the maximum number of steps.')
+        spec.exit_code(503, 'ERROR_SCALE_H',
+            message=('The cell relaxation caused a significant volume contraction '
+                     'and there is not enoguh space allocated for radial FFT.'))
```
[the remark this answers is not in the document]
bastonero: Open it again. Actually, this can also be a __significant expansion__. Shall we change the name to e.g. __VOLUME_SCALING__ or similar?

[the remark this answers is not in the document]
sphuber: If it can also relate to expansion, then I agree that is a better name. Currently you used `_CONTRACTION` still though? Did you still need to adapt this?

src/aiida_quantumespresso/workflows/pw/base.py
sphuber: Any update on feedback from the developers? If there are no "negative" side effects to just increasing the `cell_factor`, why not do that. But also, if the output structure is parsable, if we restart the volume change should be a lot different and so the default may be sufficient, or am I misunderstanding?

tests/parsers/test_pw.py
sphuber: I think in this case it may simply return the input structure. Should be easy to check

src/aiida_quantumespresso/workflows/pw/base.py
bastonero: Here the reply from Paolo Giannozzi:
"""
The reason for frequent "not enough space allocated" errors was that the interpolation tables for pseudopotential matrix elements were computed for wave-vectors in units of 2\pi/a, so smaller volumes resulted in larger grids This behavior was changed in this merge request: https://gitlab.com/QEF/q-e/-/merge_requests/822, so I don't expect this error to show up any longer except in some corner cases. If it does, please provide a test case.
"""
So I guess it is best to increase eventually `cell_factor` and restart from the last structure, if this happens. It would be great if @mbercx could verify if the newer QE is crashing when using the same calculation where this happened. In any case, I would vote for increasing `cell_factor` in the handler for such corner cases.

src/aiida_quantumespresso/calculations/pw.py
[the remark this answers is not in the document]
bastonero: In case yes, but from the reply of Paolo Giannozzi, this error seems to appear indeed only for large contraction, and not for expansions.

src/aiida_quantumespresso/workflows/pw/base.py
sphuber: Thanks for reaching out to the devs. Fine for me as well then to increase the cell_factor in the handler.

sphuber (approved):

sphuber:
Thanks @bastonero

PR #918, opened 2023-04-16
`EpwCalculation`: Allow restart from previous `epw.x` run

### LINKED ISSUE #917: `EpwCalculation`: Allow for restarts from previous `epw.x` runs
Currently the `EpwCalculation` expects both the `parent_folder_nscf` and `parent_folder_ph` inputs to start from. However, an `epw.x` calculation can also be restarted from a previous one, and this is an often occurring use case in typical EPW-based workflows.

mbercx:
Fixes #917


mbercx:
After a discussion with @sponce24 and @ymzhang0, the plan is to move the `EpwCalculation` to a separate `aiida-epw` package.

PR #921, opened 2023-04-21
Devops: Add script to update `CHANGELOG.md`

mbercx:
Add a GitHub workflow that automatically gathers the commits since the last release tag and prepends the list to the `CHANGELOG.md`. The pull request hashes are also converted into links to the corresponding pull request.

### DIFF
--- a/dev/update_changelog.py
+++ b/dev/update_changelog.py
@@ -1,0 +1,100 @@
+#!/bin/bash
+# -*- coding: utf-8 -*-
+"""Script for automatically updating the `CHANGELOG.md` based on the commits since the latest release tag."""
+from pathlib import Path
+import re
+import subprocess
+
+DEFAULT_CHANGELOG_SECTIONS = """
+### ‼️ Breaking changes
+
+
+### ✨ New features
+
+
+### 🗑️ Deprecation
+
+
+### 👌 Improvements
+
+
+### 🐛 Bug Fixes
+
+
+### 📚 Documentation
+
+
+### 🔧 Maintenance
+
+
+### ⬆️ Update dependencies
+
+
+### ♻️ Refactor
+
+"""
+
+
+def update_changelog():
+    """Update the `CHANGELOG.md` for a first draft of the release."""
+
+    print('🔍 Checking the current version number')
+    with Path('CHANGELOG.md').open('r', encoding='utf8') as handle:
+        current_changelog = handle.read()
+
+    with Path('src/aiida_quantumespresso/__init__.py').open('r', encoding='utf8') as handle:
+        version = re.search(r"__version__ = '(?P<version_number>\d+.+)'", handle.read()).groupdict()['version_number']
+
+    if str(version) in current_changelog:
+        print('🛑 Current version already in `CHANGELOG.md`. Skipping...')
+        return
+
+    print('⬆️ Found updated version number, adapting `CHANGELOG.md`.')
+    tags = subprocess.run(['git', 'tag'], capture_output=True, check=False).stdout
+
+    tag_pattern = re.compile(r'(v\d\.\d\.\d)\n')
+    tags = tags.decode()
+
+    latest_tag = tag_pattern.findall(tags)[-1]
+
+    print(f'🔄 Comparing with latest tag `{latest_tag}`.')
+    commits = subprocess.run(['git', 'log', "--pretty=format:'%s %h %H'", f'{latest_tag}..origin/main'],
+                             capture_output=True,
+                             check=False).stdout
+    commits = commits.decode()
+
+    pr_pattern = re.compile(r'\(\S(?P<pr_number>\d+)\)')
+
+    changelog_commits = []
+
+    for commit in commits.splitlines():
+
+        # Remove the PR number from the commit message
+        pr_match = pr_pattern.search(commit)
+
+        if pr_match is not None:
+            pr_number = pr_match.groupdict()['pr_number']
+            commit = commit.replace(fr'(#{pr_number})', '')
+
+        # Add the commit hash (short) to link to the
+        # 0aba276f7042d51b91d5699530de7336cd62e2c2
+        commit = commit.split()
+        hash_long = commit.pop()
+        hash_short = commit.pop()
+        commit.append(f'[[{hash_short}](https://github.com/aiidateam/aiida-quantumespresso/commit/{hash_long})]')
+        commit = commit.strip("'")
+        changelog_commits.append(' '.join(commit))
+
+    changelog_message = f'## v{version}\n' + DEFAULT_CHANGELOG_SECTIONS
+
+    for commit_title in changelog_commits:
+        changelog_message += f'\n* {commit_title}'
+
+    with Path('CHANGELOG.md').open('w', encoding='utf8') as handle:
+        handle.write(changelog_message + '\n\n' + current_changelog)
+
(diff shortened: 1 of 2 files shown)


mbercx:
Note: even though I think the GitHub workflows are fun to play around with, in hindsight it may be more sensible to have a `pre-commit` hook for this. Consider only having to do:

1. `git checkout -b release/v4.3.0`
2. Change version in `__version__`.
3. `git commit -am 'Release v4.3.0'` -> `pre-commit` automatically does first draft update of `CHANGELOG.md`
4. Finalise `CHANGELOG.md`
5. `git commit -a`
6. `git push -u origin`
7. Open PR

The current workflow is slightly more complicated. After the first two steps:

3. `git commit -am 'Release v4.3.0'`
4. `git push -u origin` -> GitHub action automatically does first draft update of `CHANGELOG.md`
5. `git pull` changes locally
6. Finalise `CHANGELOG.md`
7. `git commit -a`
8. `git push origin`
9. Open PR

I mean, it's _two_ steps less 😁. That said, the GitHub action allows the check to be more specific to only `release/**` branches.

In any case, I'd still have to learn how to actually write a `pre-commit` hook, so...

... Of course, we could also just run the script manually now that I think of it. 😅 Not nearly as badass though, but:

3. python `.github/scripts/update_changelog.py` -> automatically does first draft update of `CHANGELOG.md`
4. Finalise `CHANGELOG.md`
5. `git commit -am 'Release v4.3.0'`
6. `git push -u origin`
7.  Open PR

mbercx:
After some discussion it was decided to just convert this into a script the maintainer can run locally. Much more sensible since they can just adapt the `CHANGELOG.md` afterwards and then continue with the release.

@sphuber I put the scripts in a `dev` folder in the root folder of the repo. We can also start adding scripts there to e.g. trim output files for the text fixtures. What do you think?

dev/update_changelog.py
```diff
@@ -0,0 +1,100 @@
...
+
+"""
+
+
+def update_changelog():
+    """Update the `CHANGELOG.md` for a first draft of the release."""
+
+    print('🔍 Checking the current version number')
+    with Path('CHANGELOG.md').open('r', encoding='utf8') as handle:
+        current_changelog = handle.read()
```
sphuber, line 43:
```suggestion
    current_changelog = Path('CHANGELOG.md').read_text()
```

dev/update_changelog.py
```diff
@@ -0,0 +1,100 @@
...
+
+    with Path('src/aiida_quantumespresso/__init__.py').open('r', encoding='utf8') as handle:
+        version = re.search(r"__version__ = '(?P<version_number>\d+.+)'", handle.read()).groupdict()['version_number']
+
+    if str(version) in current_changelog:
```
sphuber, line 48: This is a bit fragile I guess. It is possible that a changelog will contain a future version in its notes. For example, it could include "Deprecated `PwBaseWorkChain`, will be removed in v15.0.0`. Guess it might be unlikely but maybe can be easily made more robust by checking that the version appears in a line starting with `##` as in a markdown section header?

dev/update_changelog.py
```diff
@@ -0,0 +1,100 @@
...
+    print('⬆️ Found updated version number, adapting `CHANGELOG.md`.')
+    tags = subprocess.run(['git', 'tag'], capture_output=True, check=False).stdout
+
+    tag_pattern = re.compile(r'(v\d\.\d\.\d)\n')
+    tags = tags.decode()
+
+    latest_tag = tag_pattern.findall(tags)[-1]
+
+    print(f'🔄 Comparing with latest tag `{latest_tag}`.')
+    commits = subprocess.run(['git', 'log', "--pretty=format:'%s %h %H'", f'{latest_tag}..origin/main'],
```
sphuber, line 61: How about you use `--pretty-format:'%h|%H|%s'`. You can then later use:
```
hash_short, hash_long, message = commit.split('|', maxsplit=2)
```
Think that is a bit cleaner and faster as you don't split and re-join the message.

dev/update_changelog.py
```diff
@@ -0,0 +1,100 @@
...
+        commit = commit.strip("'")
+        changelog_commits.append(' '.join(commit))
+
+    changelog_message = f'## v{version}\n' + DEFAULT_CHANGELOG_SECTIONS
+
+    for commit_title in changelog_commits:
+        changelog_message += f'\n* {commit_title}'
+
+    with Path('CHANGELOG.md').open('w', encoding='utf8') as handle:
+        handle.write(changelog_message + '\n\n' + current_changelog)
```
sphuber, line 94:
```suggestion
    Path('CHANGELOG.md').write_text(f'{changelog_message}\n\n{current_changelog}')
```

sphuber: Normally this would be a dangerous operation, writing directly to the source file, as you can lose it if something fails. Would be better to write to temp file and then `os.rename` to target file which is atomic. But since it is versioned by git, I guess we can always recover, so can keep this

sphuber (changes requested):
Thanks @mbercx . Useful addition, just left some suggestions. Also was wondering if we need a new top-level directory for this. Maybe we just put it in `.github/workflows`. Even though it is not used by an automated workflow, I think it makes some sense there. The advantage is that you can also easily reuse the `get_version_from_module` function from the `validate_release_tag.py`. This solves the question about your current approach being maybe not the most robust.

dev/update_changelog.py
```diff
@@ -0,0 +1,100 @@
...
+
+    with Path('src/aiida_quantumespresso/__init__.py').open('r', encoding='utf8') as handle:
+        version = re.search(r"__version__ = '(?P<version_number>\d+.+)'", handle.read()).groupdict()['version_number']
+
+    if str(version) in current_changelog:
+        print('🛑 Current version already in `CHANGELOG.md`. Skipping...')
+        return
+
+    print('⬆️ Found updated version number, adapting `CHANGELOG.md`.')
+    tags = subprocess.run(['git', 'tag'], capture_output=True, check=False).stdout
```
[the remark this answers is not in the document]
mbercx: `pylint` wants you to explicitly set `check`:

https://pylint.pycqa.org/en/latest/user_guide/messages/warning/subprocess-run-check.html

So I added it quickly, since I wasn't really concerned with it, but happy to change it to `True`. This is actually exactly why `pylint` complains, so you think about it. I was just stubborn in my ignorance because I wanted to test the GitHub action ^^

dev/update_changelog.py
mbercx: I think now we can also just really grab the version from the installed `aiida_quantumespresso` instance and convert it into a string? I didn't do that here because I wanted the check on GitHub to be fast and so I wanted to avoid installing the package.

dev/update_changelog.py
```diff
@@ -0,0 +1,100 @@
...
+
+    tag_pattern = re.compile(r'(v\d\.\d\.\d)\n')
+    tags = tags.decode()
+
+    latest_tag = tag_pattern.findall(tags)[-1]
```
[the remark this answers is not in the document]
mbercx: Nice! True about the `re.compile`, I think I was fighting with getting the correct regex at night and somehow got desperate. 😅 Didn't know about the `encoding` input argument of `subprocess.run()`, nor the nice sort option of `tag`, thanks!

dev/update_changelog.py
mbercx: Thanks! Good to know.

sphuber (approved):

Issue #705, opened 2021-07-20
`PwRelaxWorkChain`: Revisit meta iterations and final SCF


mbercx:
Currently, the `PwRelaxWorkChain` performs several relaxations (meta iterations) in case the change in volume of the relaxation is above 1%:

https://github.com/aiidateam/aiida-quantumespresso/blob/da941610af7a45993c54af04f986c69d275d26f8/aiida_quantumespresso/workflows/pw/relax.py#L305-L330

One possible motivation for this could be the (isotropic?) Pulay stress: checking the volume change and repeating the calculation in case it is above a certain threshold would be a viable fix for this problem. There possible other reasons though (meta states? I don't fully understand this yet), so we have to carefully check for some runs that this is indeed the case.

If the only motivation is the Pulay stress, I think we can remove this meta-convergence and simply rely on the error handler in the `PwBaseWorkChain` that checks if the pressure obtained from the final SCF with a regenerated basis set is still below the threshold. In case the pressure is converged to the threshold in the `vc-relax`, the resulting pressure from this final SCF should be exactly the Pulay stress (plus-minus the threshold, of course).

Regarding the final SCF: an important historical use case for this was the fact that some structures came from databases where the license has restrictions on making the structures publicly available. Since the original structures are stored in the provenance, the extra SCF is important to cut off the published provenance at this point.

giovannipizzi:
Additional (historical) note: the final SCF (with printing the stress and forces turned on) would allow to have an indication, in the provenance, of the forces and stresses (if they are below a certain threshold, we are allowed by the license to publish the data).
Without the final SCF, one could still cut the whole provenance, and use the final structure as the "initial" node, but then one would not know if the stress and forces are below the threshold that was decided.

mbercx:
Some more investigations on this, looking at 2510 work chains from the MC3D that ran 3 meta-convergence `vc-relax`:

First some statistics. Out of 2510 work chains:
* 1886 have 3 meta-convergence steps
* 456 have 4 meta-convergence steps
* 168 have 5 meta-convergence steps

I'll focus on looking at the ones with 3 meta-convergence steps, then we can simply look at step 2. A couple of reasons why a third step could be necessary:

1. Change in k-points grid: If the unit cell changes significantly, it's possible that the original k-points mesh no longer represents the kpoints density defined by `kpoints_distance`. This happened on the second step in 1227 cases, and might be a legitimate reason for keeping the meta-convergence.
[...]

PR #927, opened 2023-05-05
`PwCalculation`: refactor `parent_folder` validation

mbercx:
In 3fea26b we added validation for restarting from a `parent_folder` to the 
`PwCalculation`. Here we ran into the following problem:

1. We want to check that a `parent_folder` is present when `CONTROL.calculation`
   is `nscf` or `bands`
2. The `parent_folder` will not be present in the initial inputs of e.g. the
   `PwBandsWorkChain` `bands` name space, since it is only created on runtime.
3. The validation will hence always fail, raising a warning when the builder is put
   together using the `get_builder_from_protocol()` method.

Our solution at the time was to split up the validation in a base and full one,
and override the validator. However, a more elegant solution is to check if the 
`parent_folder` is in the name space, and only do the restart validation in this case.
Wrapping work chains that dynamically assign this input should exclude the
`parent_folder` port when exposing the inputs of the `PwBaseWorkChain`.

### DIFF
--- a/src/aiida_quantumespresso/calculations/pw.py
+++ b/src/aiida_quantumespresso/calculations/pw.py
@@ -176,11 +176,29 @@ class PwCalculation(BasePwCpInputGenerator):
                     'is `False` and/or `electron_maxstep` is 0.')
         # yapf: enable
 
-    @staticmethod
-    def validate_inputs_base(value, _):
-        """Validate the top level namespace."""
+    @classmethod
+    def validate_inputs(cls, value, port_namespace):
+        """Validate the top level namespace.
+
+        Check that the restart input parameters are set correctly. In case of 'nscf' and 'bands' calculations, this
+        means ``parent_folder`` is provided. For other calculations, if the ``parent_folder`` is provided, the restart
+        settings must be set to use some of the outputs.
+
+        Note that the validator will only check the logic in case the ``parent_folder`` is a port in the
+        ``port_namespace``. This is because the ``PwCalculation`` can be wrapped inside a work chain that only provides
+        the ``parent_folder`` input at a later step in the outline. To avoid raising any warnings, such a work chain
+        must exclude the ``parent_folder`` port when exposing the inputs of the ``PwCalculation``.
+        """
+        result = super().validate_inputs(value, port_namespace)
+
+        if result is not None:
+            return result
+
         parameters = value['parameters'].get_dict()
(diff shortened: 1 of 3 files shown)


mbercx:
Note: Docs failing because of https://github.com/aiidateam/aiida-quantumespresso/pull/928

mbercx:
@sphuber this change is unfortunately breaking backwards compatibility. I'm fine with leaving this PR blocked until v5 if you think it's necessary, but I would be inclined to just merge it and communicate these "small" breaking changes clearly in the change log and perhaps even the AiiDA Slack/mailing list.

sphuber:
> @sphuber this change is unfortunately breaking backwards compatibility. I'm fine with leaving this PR blocked until v5 if you think it's necessary, but I would be inclined to just merge it and communicate these "small" breaking changes clearly in the change log and perhaps even the AiiDA Slack/mailing list.

Do you mean breaking as in wrapping workchains using `validate_inputs_base`? I think we introduced this not so long ago right? Are there even any workchains out there calling that, except for the ones in here, that you update in the PR anyway?

mbercx:
> Are there even any workchains out there calling that, except for the ones in here, that you update in the PR anyway?

Well, I know of some in my other packages, but these are of course not an issue. Maybe @bastonero or @unkcpz might also have run into this validation. I'm fine with these small and easy to fix backwards-incompatible changes though. I favor dealing with the scorn instead of the red tape, and generally think "it's better to ask forgiveness than permission" ^^

sphuber:
> I'm fine with these small and easy to fix backwards-incompatible changes though. I favor dealing with the scorn instead of the red tape, and generally think "it's better to ask forgiveness than permission" ^^

Think it really depends. If the breaking change is going to affect few users and it can be relatively easily addressed, then it makes it more acceptable. I would be happy to accept this change so I will approve

sphuber (approved):
Thanks. I agree with the approach. It makes sense for the parent workchains to exclude the `parent_folder`

mbercx:
> If the breaking change is going to affect few users and it can be relatively easily addressed, then it makes it more acceptable.

For sure. Anything more difficult would require a major release. That said, I'm also not shy to push out more major releases if we can bundle together some solid backwards-incompatible changes in each. An example is https://github.com/aiidateam/aiida-quantumespresso/issues/705, but we'll discuss that when I open the PR for it.

PR #933, opened 2023-05-16
Remove useless fixture pseudos folder

unkcpz:


unkcpz:
maybe I miss something, I use the same fixture in aiidalab-qe, and the `tests/fixtures/pseudos` is not needed.

unkcpz:
I am an idiot, it is not useless clearly. Close the PR.

PR #934, opened 2023-05-20
📚 Docs: Add first version of `PwRelaxWorkChain` how to

mbercx:
First version of the How-to section on the `PwRelaxWorkChain`. Describes basic usage via the `get_builder_from_protocol()` method, as well as determining the basic degrees of freedom using the `RelaxType` enum.

### DIFF
--- a/docs/source/howto/workflows/pw/relax.md
+++ b/docs/source/howto/workflows/pw/relax.md
@@ -2,4 +2,68 @@
 
 # `PwRelaxWorkChain`
 
-*To be added.*
+First, load the AiiDA orm and activate your profile:
+
+```python
+from aiida import orm, load_profile
+
+load_profile()
+```
+
+Import the `PwRelaxWorkChain` and use the protocols to run a `structure` with the `code` you have set up:
+
+```python
+from ase.build import bulk
+from aiida_quantumespresso.workflows.pw.relax import PwRelaxWorkChain
+
+code = orm.load_code('pw@localhost')
+structure = orm.StructureData(ase=bulk('Si', 'diamond', 5.4))
+
+builder = PwRelaxWorkChain.get_builder_from_protocol(
+    code=code,
+    structure=structure,
+    protocol='fast'
+)
+```
+
+The calculation can then be executed using for example the `run` function of the engine:
+
+```python
+from aiida.engine import run
+
+results = run(builder)
+```
+
+The `results` will contain the output nodes, for example the `output_structure`:
+
+```python
+results['output_structure'].get_pymatgen()
+```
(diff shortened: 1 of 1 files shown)


docs/source/howto/workflows/pw/relax.md
```diff
@@ -2,4 +2,101 @@
...
+load_profile()
+```
+
+Next, load the `PwRelaxWorkChain` and the structure/code you want to run:
+
+```python
+from aiida_quantumespresso.workflows.pw.relax import PwRelaxWorkChain
+
+code = orm.load_code('qe-7.2-pw@localhost')
+structure = orm.load_node(1)  # FCC Silicon
```
sphuber, line 21: Why not make it actually runnable?
```suggestion
```python
from ase.build import bulk
from aiida_quantumespresso.workflows.pw.relax import PwRelaxWorkChain

code = orm.load_code('qe-7.2-pw@localhost')
structure = orm.StructureData(ase=bulk('Si', 'diamond', 5.43))  # FCC Silicon
```

sphuber (commented):
Thanks @mbercx . Think this is great to have. I just have two questions now that we restarted the docs from scratch and will start adding pieces by different authors.

I feel the tone of this addition is more that of a tutorial, rather than a how-to guide. Would it make sense to add this rather to the tutorials section? As you can see, the entry I added for `PwCalculation` is in similar spirit to what you wrote here, in the sense that it is a complete and step-by-step instruction. The changing of relax type would be a great section for the how-to guide though.

Think we should also make some agreements up front on the language. Especially for how-to's, I think we shouldn't use the second person singular, but a more neutral language. If you prefer, for tutorial content, I think it could make sense the way you wrote it. This is always going to be heavily influenced by personal preference, but I think that if we want to have the docs be readable, they should be as consistent as possible, despite being written by different authors.

mbercx:
Thanks for the comments, @sphuber! This is exactly why I still left the PR as a WIP, to discuss how to properly structure the docs. 👍 📚 

I completely agree that it is good to have a clear separation between tutorials and how-to's, although there will always be a bit of overlap. Also fine with rethinking the language. Let me have a crack at adapting this documentation, but also write down our policies on this so we can crystallise our approach.

docs/source/howto/workflows/pw/relax.md
mbercx: Although having the structure is a bonus, in the end they probably still need to adapt the code ID. ^^

mbercx:
Superseded by https://github.com/aiidateam/aiida-quantumespresso/pull/1125

PR #936, opened 2023-05-21
fix lda_plus_u_calculation in output_parameters

### LINKED ISSUE #935: `lda_plus_u_calculation` in output_parameters is wrong
I did a `DFT+U` calculation, but the `lda_plus_u_calculation` is `False` in the `output_parameters`.

superstar54:
fix #935 

The `dftU` is in the `dft` section of the `output`.

### DIFF
--- a/src/aiida_quantumespresso/parsers/parse_xml/parse.py
+++ b/src/aiida_quantumespresso/parsers/parse_xml/parse.py
@@ -233,7 +233,7 @@ def parse_xml_post_6_2(xml):
         'starting_magnetization': starting_magnetization,
         'has_electric_field': has_electric_field,
         'has_dipole_correction': has_dipole_correction,
-        'lda_plus_u_calculation': 'dftU' in outputs,
+        'lda_plus_u_calculation': 'dftU' in outputs['dft'],
         'format_name': xml_dictionary['general_info']['xml_format']['@NAME'],
         'format_version': xml_dictionary['general_info']['xml_format']['@VERSION'],
         # TODO: check that format version: a) matches the XSD schema version; b) is updated as well


mbercx:
Thanks @superstar54! What version of QE are you running here? Could you maybe also provide the corresponding output files (`aiida.out`, `data-file-schema.xml`) so I can set up a test for this?

superstar54:
`PWSCF v.7.0`.
Here are the files
[dftU.zip](https://github.com/aiidateam/aiida-quantumespresso/files/11523819/dftU.zip)

Thanks for testing!

sphuber:
Are we only supporting v7.0 and up now @mbercx ? And is the location of this keyword consistent for all other versions > v7.0?

mbercx:
> Are we only supporting v7.0 and up now @mbercx ? And is the location of this keyword consistent for all other versions > v7.0?

Both good questions. Indeed, according to our compatibility policy 

> Starting from aiida-quantumespresso==4.0, the last three minor versions of Quantum ESPRESSO are supported. Older versions are supported up to a maximum of two years.

We should support v7.0-v7.2. As for the hubbard key word, I'm assuming that things might have changed with the new hubbard inputs in v7.1. However, perhaps we should let someone answer that has actually been running `pw.x` calculations with Hubbard corrections, @bastonero?

bastonero:
Indeed, from v7.1 all the Hubbard related keywords in `SYSTEM` are dropped. Only the `HUBBARD` card should be specified, which is done via the new `HubbardStructureData`, and not directly from `parameters` of the builder.

Nevertheless, I know that QE did not really change internally the logic, and many parameters are still the same. 

Conclusion: it might be that the xml still has these keywords (😅). I will check the xml of a recent calculation of mine and update you

bastonero:
Here you have an example of dataschema using the latest (v7.2) QE version. Indeed, the `output_parameters` has incorrectly `lda_plus_u_correction: False` in my PwCalculation node. 

Shall we also think of dropping the VERY old convention of `lda_plus_u` naming? A boolean `hubbard_correction` would do the job.

[data.tar.gz](https://github.com/aiidateam/aiida-quantumespresso/files/11527476/data.tar.gz)

mbercx:
> Shall we also think of dropping the VERY old convention of lda_plus_u naming? A boolean hubbard_correction would do the job.

Yeah, the `output_parameters` could do with a good scrubbing. This will be part of the work that @sphuber [has been doing on moving all possible parsing to the XML file](https://github.com/aiidateam/aiida-quantumespresso/tree/fix/353/pw-parser-xml-only). @sphuber perhaps you can open a PR with this work so we can start collaborating on it there? I remember I was still supposed to give you some files for additional testing, but forgot which ones. 😅

sphuber:
https://github.com/aiidateam/aiida-quantumespresso/pull/940

Let's pull that over the line first indeed, before optimizing other changes in the `output_parameters` output. @mbercx is there anything we still want to release in the current major version? Or can we start merging stuff that can be considered breaking and that we will release with v5.0?

What about the ACWF stuff? Is that complete?

PR #940, opened 2023-05-23
`PwParser`: Refactor to primarily parse from XML and only parse from stdout what is necessary

### LINKED ISSUE #353: Parsing only the XML output (of PW)
This issue is open-ended and meant for tracking our discussion on the topic.

Now we parse both XML and textual output of PW, and then we merge the two sources of information. Textual output, being less structured, is harder to parse and to validate, so we would like to only use the XML output. However, removing the ability to parse text creates a backwards-incompatibility problem, because:
* versions between 6.0 and 6.1 (included) only enable the new XML format on request (when configured with `--enable-xml` or compiled `-D__XSD`), otherwise they use the old XML format;
* versions before 6.0 only output in the old XML format;
* in general, the new XML format/schema has improved with time, so older versions may have missing or incorrect information (which we must parse from the text).

One approach could be to start ignoring the text output from a certain version of QE onwards, and keep the rest of the code for backwards compatibility. After some time, old versions of QE can be deprecated and the text-parsing code removed completely.

One concern to keep in mind is our desire to keep consistent the results of the parser between different versions _of the plugin_; this may further constrain our choices.

sphuber:
Fixes #353

### DIFF
--- a/src/aiida_quantumespresso/parsers/cp.py
+++ b/src/aiida_quantumespresso/parsers/cp.py
@@ -72,6 +72,7 @@ class CpParser(BaseParser):
         out_dict, _raw_successful = parse_cp_raw_output(
             stdout, output_xml, output_xml_counter, print_counter_xml
         )
+        out_dict.pop('trajectory', None)
 
         if not no_trajectory_output:
             # parse the trajectory. Units in Angstrom, picoseconds and eV.
--- a/src/aiida_quantumespresso/parsers/neb.py
+++ b/src/aiida_quantumespresso/parsers/neb.py
@@ -113,8 +113,9 @@ class NebParser(BaseParser):
                 return self.exit(self.exit_codes.ERROR_UNEXPECTED_PARSER_EXCEPTION.format(exception=exc))
 
             parsed_structure = parsed_data_stdout.pop('structure', {})
-            parsed_trajectory = parsed_data_stdout.pop('trajectory', {})
-            parsed_parameters = PwParser.build_output_parameters(parsed_data_xml, parsed_data_stdout)
+            parsed_trajectory = parsed_data_xml.pop('trajectory', {})
+            parsed_parameters = parsed_data_xml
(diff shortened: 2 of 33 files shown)


mbercx:
@sphuber: @t-reents and I have been working on getting the tests to pass in:

https://github.com/mbercx/aiida-quantumespresso/tree/fix/353/pw-parser-xml-only

Note that in https://github.com/mbercx/aiida-quantumespresso/commit/45b9183d4402e117b846e516dc9b49394cacf58a we decided that for the interrupted runs, we accept that we no longer parse as much information from the `stdout` as before. If we want to keep `stdout` parsing as a backup just in case the XML is not written properly for everything, then we have to keep maintaining the `stdout` parsing and I think this is exactly what we want to avoid here.

Now all tests are passing:

https://github.com/mbercx/aiida-quantumespresso/actions/runs/7279821777

Do you mind if I rebase this PR, add my changes here? I think we still need to do some final polishing of parts of the code, and a good ol' review, but we're close to finalising this!

sphuber:
Thanks for the help guys. Regarding the `stdout` parsing: I am afraid we unfortunately anyway have to keep it because all the error parsing comes from there. The XML reports nothing in terms of warnings or errors I believe, so all our error codes rely on the stdout. Or am I missing something?

t-reents:
No, that's correct. Moreover, something we were also discussing, a bunch of parameters that are present in the trajectory are only available in the stdout, e.g. magnetization in each step of a relaxation. The same is true for charges etc.

On the other hand, the final parameters are available in the XML as well.

sphuber:
Ideally, we would collect all data that we parse from the stdout that is not part of the XML and make a comprehensive list. We can then present this to the QE developers and ask whether (part of) that data can be made available in the XML. Especially parsing the various errors from XML would greatly robustify things I think. But this would be a lot of work and would only really bear fruits in quite a long time :D

PR #942, opened 2023-05-24
Tests: Catch warning in `restart_mode` test for `PwBaseWorkChain`

mbercx:
The `test_restart_mode` test was raising a warning related to the presence of the `parent_folder` input for `restart_mode = restart` (which is expected).

Here the test is adapted to catch the warning for this this input and make sure the message is correct.

### DIFF
--- a/tests/workflows/pw/test_base.py
+++ b/tests/workflows/pw/test_base.py
@@ -315,9 +315,8 @@ def test_restart_mode(generate_workchain_pw, generate_calc_job_node, restart_mod
     if restart_mode == 'restart':
         process = generate_workchain_pw(inputs=inputs)
     else:
-        with pytest.warns(UserWarning) as warning:
+        with pytest.warns(UserWarning, match='but no input parameters were'):
             process = generate_workchain_pw(inputs=inputs)
-        assert 'but no input parameters were' in str(warning[0])
 
     process.setup()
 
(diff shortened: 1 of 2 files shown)


tests/workflows/pw/test_base.py
```diff
@@ -312,7 +312,13 @@ def test_restart_mode(generate_workchain_pw, generate_calc_job_node, restart_mod
     inputs['pw']['parent_folder'] = node.outputs.remote_folder
     inputs['pw']['parameters'] = Dict({'CONTROL': {'restart_mode': restart_mode}})
 
-    process = generate_workchain_pw(inputs=inputs)
+    if restart_mode == 'restart':
+        process = generate_workchain_pw(inputs=inputs)
+    else:
+        with pytest.warns(UserWarning) as warning:
+            process = generate_workchain_pw(inputs=inputs)
+        assert 'but no input parameters were' in str(warning[0])
```
sphuber, line 320: you can use the `match` argument in `pytest.warns` to compare the warning to a regex: https://docs.pytest.org/en/6.2.x/warnings.html#warns
Also the typical approach for `pytest.raises`

sphuber (changes requested):
Thanks, just minor suggestion

tests/workflows/pw/test_base.py
mbercx: This is what I get for using the free version of ChatGPT. Will fix 👍

sphuber (approved):

PR #943, opened 2023-05-24
Devops: Update `tox` configuration

mbercx:
The `tox` configuration was still using outdated Python versions. Here the Python versions are updated to those supported by the package, the environments are cleaned up a bit, and the `pre-commit` environment is added to those run by default.

The docs environment are removed, since these aren't usually used and we still have an issue with the `intersphinx_mapping` to `aiida-core`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -214,12 +214,14 @@ split_arguments_when_comma_terminated = true
 [tool.tox]
 legacy_tox_ini = """
 [tox]
-envlist = py37
+envlist =
+    py{38,38,310,311}
+    py38-pre-commit
+    py38-docs-{clean,update}
 
 [testenv]
+description = Run the pytest tests
 usedevelop=True
-
-[testenv:py{36,37,38,39}]
 extras = tests
 commands = pytest {posargs}
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -215,9 +215,8 @@ split_arguments_when_comma_terminated = true
 legacy_tox_ini = """
 [tox]
 envlist =
-    py{38,38,310,311}
+    py{38,39,310,311}
     py38-pre-commit
-    py38-docs-{clean,update}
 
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -214,12 +214,14 @@ split_arguments_when_comma_terminated = true
 [tool.tox]
 legacy_tox_ini = """
 [tox]
-envlist = py37
+envlist =
+    py{38,38,310,311}
```
sphuber, line 218:
```suggestion
    py{38,39,310,311}
```

chrisjsewell:
> add the pre-commit and docs environments to those run by default

Heya, personally I don't use tox like this. I leave the default as a singular environment, running pytest, so that I can simply run `tox` on any project and have it run the unit tests.
The other things are optional.

In fact, I usually run pre-commit directly, because I don't allow for any hooks that run locally, e.g. by using ruff and not pylint

pyproject.toml
mbercx: Your eagle-eyed Dutchman eyes caught my typo again. This was a test, of course. 😇

mbercx:
Cheers for the comment @chrisjsewell. I typically use the built-in testing feature of VSCode for running tests in a single environment, so `tox` is really more for running in all supported Python versions for me. I've removed the docs environments though, I don't really use those and they are failing anyways because we still have an `intersphinx_mapping` issue to fix.

pyproject.toml
sphuber: Well, since your recent barrage of slights regarding my supposed sloppiness, I have been just scouring your changes for any hint of a fuck-up, merely as a sordid form of retribution

sphuber (approved):

pyproject.toml
mbercx: I think we have a very healthy working relationship. ❤️

PR #949, opened 2023-06-02
Docs: Small improvements to "Installation" page

mbercx:


docs/source/installation/index.md
```diff
@@ -108,10 +109,10 @@ In this example, we will set up the `localhost`, the computer where AiiDA itself
 To set up a computer, use the ``verdi`` CLI of ``aiida-core``.
 
 ```console
-$ verdi computer setup -n -L localhost -H localhost -T core.local -S core.direct
+$ verdi computer setup -n -L localhost -H localhost -T core.local -S core.direct -w $AIIDA_PATH/workdir
```
sphuber, line 112: I don't think it is guaranteed that `$AIIDA_PATH` is defined, and so this will resolve to `-w /workdir` which is not what we want. The computer setup will succeed just fine but then the first calc will fail horribly. Or in this case, I guess when `verdi computer test` is run actually, since we suggest users run that further down in the test.  Why not use `/tmp`? Isn't that the default tmp folder for both Linux and MacOS?

docs/source/installation/index.md
```diff
@@ -186,13 +187,18 @@ default_calc_job_plugin='quantumespresso.pw',
 
 :::{important}
 Make sure to replace `/path/to/pw.x` with the actual absolute path to the `pw.x` binary.
+If the binary is in your `PATH`, you can find the first `pw.x` binary in your `PATH` directories using the `which` command:
+
+```console
+which pw.x
+```
```
sphuber, line 194: The path actually no longer has to be absolute (since v2.3) so I think we should just recommend to specify `pw.x`. The current version of `aiida-quantumespresso` requires `aiida-core~=2.3`.

sphuber (changes requested):

docs/source/installation/index.md
mbercx: Fair point, I've gotten so used to _always_ defining an `AIIDA_PATH` that I didn't consider this.

I'm not sure about `/tmp` though. How persistent is the data there? Won't this cause trouble for restarts?

docs/source/installation/index.md
mbercx: Good to know, but maybe it's still good policy to set the _full_ path? A user might have multiple `pw.x` binaries lying around, and this way they put some thought into using the one they actually want to run.

docs/source/installation/index.md
sphuber: Yeah, sure, for production profiles this is not what you want. But I guess that if you are setting up the localhost to run QE calculations, it is mostly for testing and smaller things. But fine to also just put it to a folder in home, like `~/.scratch`. Just not sure if that will automatically be created, or we have to then also instruct them to be sure to create the folder manually.

docs/source/installation/index.md
sphuber: Maybe we can just leave the relative path as the first suggestion and then in a note or warning box below it, give more info about how to pass an absolute path if you want to be sure which executable to call.

docs/source/installation/index.md
mbercx: How about using `${AIIDA_PATH:-$HOME}/aiida_scratch`?[^1] I'm quite sure the folder will be created, but happy to double check. Regarding the actual folder name:

1. I know using a hidden folder is in the spirit of AiiDA abstracting away the actual running of calculations, but I would prefer a regular folder since it should be easy for users to find and clean up.
2. I like using `scratch` as the name, since it's in line with what a lot of users will know for "the place where you run calculations". Although I also like having the folder name match the option for clarity. So I'm a bit conflicted.
3. I was also considering prepending the folder name with `aiida_`, i.e. `aiida_scratch`. This is probably better in case the `AIIDA_PATH` variable doesn't exist, since the folder is less likely to exist and it's more clear which "scratch it is, but for separated environments it's a bit redundant. Separated environments typically means that the user knows what they are doing, or they are using `aiida-project`.

So for these instructions, I'd pick `aiida_scratch`.

Note that this discussion is also very relevant to https://github.com/aiidateam/aiida-core/issues/749.

[^1]: Default variable code courtesy of ChatGPT. ^^

sphuber: Happy to accept that suggestion, but aren't we just overcomplicating this? Why don't we do a simple `~/aiida_scratch` and then follow with an info box explaining what the value is for, and what other values one could consider, for example `$AIIDA_PATH/work_dir` or whatever. These instructions are going to be for the novice user that is most likely going to install this for the first time and just test things out, and so `~/aiida_scratch` should be good enough. I think that `${AIIDA_PATH:-$HOME}/aiida_scratch` will look scary and raise a bunch of questions

mbercx: Fair point! I think perhaps the more intimidating solution might be one for the automated localhost computer setup described in the issue on `aiida-core`. Will use the simplified solution here, but then I'm in favor of aiida_work_dir for the instructions.

docs/source/installation/index.md
```diff
@@ -126,17 +127,18 @@ $ verdi computer test localhost
...
 from aiida.orm import Computer
+from pathlib import Path
 
 computer = Computer(
     label='localhost',
     hostname='localhost',
     transport_type='core.local',
-    scheduler_type='core.direct'
+    scheduler_type='core.direct',
+    workdir=Path('$AIIDA_PATH/workdir').absolute().as_posix()
```
sphuber, line 141:
```suggestion
    workdir=Path('~/aiida_work_dir').resolve()
```

docs/source/installation/index.md
```diff
@@ -108,10 +109,10 @@ In this example, we will set up the `localhost`, the computer where AiiDA itself
 To set up a computer, use the ``verdi`` CLI of ``aiida-core``.
 
 ```console
-$ verdi computer setup -n -L localhost -H localhost -T core.local -S core.direct
+$ verdi computer setup -n -L localhost -H localhost -T core.local -S core.direct -w ${AIIDA_PATH:-$HOME}/aiida_scratch
```
sphuber, line 112:
```suggestion
$ verdi computer setup -n -L localhost -H localhost -T core.local -S core.direct -w ~/aiida_work_dir
```

sphuber (changes requested):
Ok, let's go with `~/aiida_work_dir` then

sphuber (approved):
All good, thanks @mbercx

PR #952, opened 2023-06-05
`PwRelaxWorkChain`: Revisit work chain logic

### LINKED ISSUE #705: `PwRelaxWorkChain`: Revisit meta iterations and final SCF

### LINKED ISSUE #733: `PwRelaxWorkChain`: Use multiple relaxations with increasing precision
## Problem description

In the current approach for the `PwRelaxWorkChain`, the initial relaxation run immediately uses the desired electronic precision settings provided by the user (or protocol). Typically these settings are very strict, and potentially difficult to obtain with a structure that is still far from the geometric minimum. Moreover, having a very precise calculation of the forces is not so critical when still far from the minimum.

## Desired feature

I've already discussed with both @mkotiuga and @simonpintarelli that it might be more interesting to start with a first run with reduced precision settings, to then move to the ones specified (either immediately or in multiple runs). This might be something worthwhile to explore, both to:

1. Improve convergence: especially when starting from less-sensible initial geometries, very precise electronic convergence can be difficult to obtain.
2. Improve efficiency: obtaining such strict electronic convergence will not influence the initial relaxation steps much, and running with lower k-point densities and less strict energy thresholds can have a large effect on the runtime.

## Potential issues

Of course, it will require some testing to figure out if there is truly a benefit from implementing such a logic in the work chain. There is also the issue that this might lead to different results, so we'll have to carefully check this.

mbercx:
Fixes #733
Fixes #705 

Will write full commit message when all changes are agreed upon.

### DIFF
--- a/src/aiida_quantumespresso/workflows/protocols/pw/relax.yaml
+++ b/src/aiida_quantumespresso/workflows/protocols/pw/relax.yaml
@@ -2,24 +2,28 @@ default_inputs:
     clean_workdir: False
     max_meta_convergence_iterations: 5
     meta_convergence: True
-    volume_convergence: 0.02
-    base:
+    base_relax:
         pw:
             parameters:
                 CELL:
                     press_conv_thr: 0.5
-    base_final_scf:
+    base_init_relax:
+        kpoints_distance: 0.3
+        meta_parameters:
+            conv_thr_per_atom: 1.e-6
+            etot_conv_thr_per_atom: 1.e-3
         pw:
             parameters:
                 CONTROL:
                     calculation: scf
+                    forc_conv_thr: 0.5e-2
+                CELL:
+                    press_conv_thr: 2
 default_protocol: balanced
 protocols:
     balanced:
         description: 'Protocol to perform a relaxation at normal precision at moderate computational cost.'
     stringent:
         description: 'Protocol to perform a relaxation at high precision at higher computational cost.'
-        volume_convergence: 0.01
     fast:
         description: 'Protocol to perform a relaxation at low precision at minimal computational cost for testing purposes.'
-        volume_convergence: 0.05
--- a/src/aiida_quantumespresso/workflows/pw/relax.py
+++ b/src/aiida_quantumespresso/workflows/pw/relax.py
@@ -1,26 +1,18 @@
 # -*- coding: utf-8 -*-
 """Workchain to relax a structure using Quantum ESPRESSO pw.x."""
+# pylint: disable=no-member
 from aiida import orm
 from aiida.common import AttributeDict, exceptions
 from aiida.common.lang import type_check
 from aiida.engine import ToContext, WorkChain, append_, if_, while_
-from aiida.plugins import CalculationFactory, WorkflowFactory
 
+from aiida_quantumespresso.calculations.functions.create_kpoints_from_distance import create_kpoints_from_distance
+from aiida_quantumespresso.calculations.pw import PwCalculation
 from aiida_quantumespresso.common.types import RelaxType
 from aiida_quantumespresso.utils.mapping import prepare_process_inputs
+from aiida_quantumespresso.workflows.pw.base import PwBaseWorkChain
 
 from ..protocols.utils import ProtocolMixin
-
-PwCalculation = CalculationFactory('quantumespresso.pw')
-PwBaseWorkChain = WorkflowFactory('quantumespresso.pw.base')
-
-
-def validate_inputs(inputs, _):
-    """Validate the top level namespace."""
-    parameters = inputs['base']['pw']['parameters'].get_dict()
-
-    if 'calculation' not in parameters.get('CONTROL', {}):
-        return 'The parameters in `base.pw.parameters` do not specify the required key `CONTROL.calculation`.'
 
 
 class PwRelaxWorkChain(ProtocolMixin, WorkChain):
@@ -31,13 +23,16 @@ class PwRelaxWorkChain(ProtocolMixin, WorkChain):
         """Define the process specification."""
         # yapf: disable
         super().define(spec)
-        spec.expose_inputs(PwBaseWorkChain, namespace='base',
-            exclude=('clean_workdir', 'pw.structure', 'pw.parent_folder'),
-            namespace_options={'help': 'Inputs for the `PwBaseWorkChain` for the main relax loop.'})
-        spec.expose_inputs(PwBaseWorkChain, namespace='base_final_scf',
+        spec.expose_inputs(PwBaseWorkChain, namespace='base_init_relax',
             exclude=('clean_workdir', 'pw.structure', 'pw.parent_folder'),
             namespace_options={'required': False, 'populate_defaults': False,
-                'help': 'Inputs for the `PwBaseWorkChain` for the final scf.'})
+                'help': (
+                    'Inputs for the `PwBaseWorkChain` that runs an initial geometry optimization, typically with looser'
+                    'precision settings to find a geometry that is already closer to the final one quickly.'
+                )})
+        spec.expose_inputs(PwBaseWorkChain, namespace='base_relax',
+            exclude=('clean_workdir', 'pw.structure', 'pw.parent_folder'),
+            namespace_options={'help': 'Inputs for the `PwBaseWorkChain` for the main relax loop.'})
         spec.input('structure', valid_type=orm.StructureData, help='The inputs structure.')
         spec.input('meta_convergence', valid_type=orm.Bool, default=lambda: orm.Bool(True),
             help='If `True` the workchain will perform a meta-convergence on the cell volume.')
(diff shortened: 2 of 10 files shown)


mbercx:
@sphuber some initial notes/questions/decisions:

1. Regarding the meta-convergence: 

   * In #705, I forgot to describe one of the most important reasons to more away from using the volume for the meta iterations: although it's rare that the volume condition rarely doesn't catch an issue with Pulay stresses, it _will_ often run another geometry optimization unnecessarily.
   * Regarding the condition on the kpoints mesh not changing: should we be very strict, and expect _the same_ mesh, or is it fine in case the previous run was done with a denser mesh than the output structure would require? Of course kpoints convergence will be achieved in the latter case, but there might be use cases where the user wants the kpoints mesh to approach the desired density as close as possible. For now I've decided to be strict.
 
2. Should the initial relaxation be run by default? Currently it is, and if we don't make it the default I'm wondering how to allow the user to enable it while also being able to provide a protocol for this step.
3. Re the documenation: I was planning to describe the work chain Purpose and Logic in the Topics section. What do you think?

src/aiida_quantumespresso/workflows/pw/relax.py
```diff
@@ -192,74 +198,134 @@ def setup(self):
...
+            inputs_create_kpoints = {
+                'structure': base_relax_workchain.outputs.output_structure,
+                'distance': self.ctx.relax_inputs.kpoints_distance,
+                'force_parity': self.ctx.relax_inputs.get('kpoints_force_parity', orm.Bool(False)),
+                'metadata': {
+                    'store_provenance': False
+                }
+            }
+            new_kpts_mesh, _ = create_kpoints_from_distance(**inputs_create_kpoints).get_kpoints_mesh()
+            if tuple(input_kpts_mesh) != tuple(new_kpts_mesh):
```
sphuber, line 281: _EDIT: just saw your point in the PR description. Will leave this here because it contains some more thoughts on the exact criterion_.

Should we maybe here check only that the new grid is not denser? If the new required grid is _less_ dense, then there is no point in rerunning is there? The criterion for `kpoints_distance` that a _minimum_ density is guaranteed in order to guarantee a minimum precision.

One tricky thing I guess might be if you have non isometric meshes, such as `(2, 3, 1)` and `(3, 2, 1)`. Which one is denser? I guess there you would have to compute the actual distance between kpoints in 1/A for both both meshes in each direction and as soon as just one is larger for the new structure, we conclude we have to rerun.

src/aiida_quantumespresso/workflows/pw/relax.py
```diff
@@ -192,74 +198,134 @@ def setup(self):
...
         self.ctx.iteration += 1
-
         inputs = self.ctx.relax_inputs
-        inputs.pw.structure = self.ctx.current_structure
 
-        # If one of the nested `PwBaseWorkChains` changed the number of bands, apply it here
+        if 'base_init_relax_workchain' in self.ctx:
+            inputs.pw.structure = self.ctx.base_init_relax_workchain.outputs.output_structure
+        else:
+            inputs.pw.structure = self.ctx.current_structure
```
sphuber, line 299: This is not necessary because `self.ctx.current_structure` is correctly set in `inspect_init_relax`

src/aiida_quantumespresso/workflows/pw/relax.py
```diff
@@ -269,80 +335,19 @@ def inspect_relax(self):
...
     def results(self):
         """Attach the output parameters and structure of the last workchain to the outputs."""
-        if self.ctx.is_converged and self.ctx.iteration <= self.inputs.max_meta_convergence_iterations.value:
-            self.report(f'workchain completed after {self.ctx.iteration} iterations')
-        else:
-            self.report('maximum number of meta convergence iterations exceeded')
-
         # Get the latest relax workchain and pass the outputs
-        final_relax_workchain = self.ctx.workchains[-1]
+        final_relax_workchain = self.ctx.base_relax_workchains[-1]
```
sphuber, line 350: You can remove the try-except on line 355. The `self.ctx.workchain_scf` is never set as that used to be done in the final scf.

src/aiida_quantumespresso/workflows/pw/relax.py
```diff
@@ -192,74 +198,134 @@ def setup(self):
...
+
+        # If meta convergence is switched off we are done
+        if not self.ctx.meta_convergence:
+            self.report('Meta-convergence disabled: workchain completed after a single iteration.')
+            return False
+
+        # Stop if the maximum number of meta iterations has been reached
+        if self.ctx.iteration == self.inputs.max_meta_convergence_iterations.value:
+            self.report('Maximum number of meta convergence iterations reached.')
+            return False
```
sphuber, line 257: I was looking for the logic where this would set an exit code to say the workchain didn't converge, but to my surprise it is not there, and also didn't use to be there. Think we'd better add one, right?

src/aiida_quantumespresso/workflows/pw/relax.py
```diff
@@ -31,13 +23,16 @@ def define(cls, spec):
         """Define the process specification."""
         # yapf: disable
         super().define(spec)
-        spec.expose_inputs(PwBaseWorkChain, namespace='base',
+        spec.expose_inputs(PwBaseWorkChain, namespace='base_relax',
             exclude=('clean_workdir', 'pw.structure', 'pw.parent_folder'),
             namespace_options={'help': 'Inputs for the `PwBaseWorkChain` for the main relax loop.'})
-        spec.expose_inputs(PwBaseWorkChain, namespace='base_final_scf',
+        spec.expose_inputs(PwBaseWorkChain, namespace='base_init_relax',
```
sphuber, line 29: Nitpick: Maybe move this before the `base_relax` namespace since chronologically it will be used before.

sphuber (commented):
Thanks @mbercx . I think this is looking good. Left some comments.

> Should the initial relaxation be run by default? Currently it is, and if we don't make it the default I'm wondering how to allow the user to enable it while also being able to provide a protocol for this step.

_Is_ it the default? If the `base_init_relax` namespace is not defined at all, it won't be run. Do you maybe mean the default when using `get_builder_from_protocol`? If it works really well for most real use-cases, I think it makes sense to actually use it by default. We can add a note in the documentation how to skip it, which should not be complicated.

> Re the documenation: I was planning to describe the work chain Purpose and Logic in the Topics section. What do you think?

Yeah, makes perfect sense.

src/aiida_quantumespresso/workflows/pw/relax.py
mbercx: I'm not sure how common that case last case is, but it's a fair point. Maybe we can accept when the new mesh is denser along every direction. This would _not_ accept your last use case, but I think that's fine, and simpler than _really_ calculating the density.

src/aiida_quantumespresso/workflows/protocols/pw/relax.yaml
```diff
@@ -2,24 +2,27 @@ default_inputs:
...
+        meta_parameters:
+            conv_thr_per_atom: 1.e-6
+            etot_conv_thr_per_atom: 1.e-3
         pw:
             parameters:
                 CONTROL:
-                    calculation: scf
+                    forc_conv_thr: 0.5e-2
+                CELL:
+                    press_conv_thr: 2
```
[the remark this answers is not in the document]
mbercx: side-eye-monkey

Minotakm:
Probably I have done some changes while I was running it. I will try to find some time next week to check.  Thanks for letting me know.

mbercx:
Thanks @Minotakm! As mentioned, I've merged the PR for now to do an alpha release, but comments or changes are very welcome! I'll be working on some other parts of the code and gathering feedback on any breaking changes, so we can get the best API possible before doing a stable major release.

mbercx:
@bastonero I remember you also mentioned that you liked the previous work chain version, and wanted to keep it? I'm not keen on maintaining two versions of the `PwRelaxWorkChain`, but I would consider moving it to a separate package for "unsupported" work chains. Thoughts welcome!

bastonero:
@mbercx thanks for pinging me! Honestly, i don't think it is so much effort to keeping both. Moreover, they do two different things, so I think with a correct naming they could be both of interests.

Issue #954, opened 2023-06-12
Monkhorst-Pack grid for 1D/2D structures in ph.x


AndresOrtegaGuerrero:
Due to  #907 ,  `PhBaseWorkChain` should include a logic to determine the grid based on the periodicity of the structure. Ideally such grid should be obtained with `create_kpoints_from_distance` like in `PwBaseWorkChain` since it takes into account the periodicity.

PR #956, opened 2023-06-12
Adding logic to PhBaseWorkChain to generate mesh from structure and qpoints_distance

AndresOrtegaGuerrero:
This PR addresses #954


src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -103,7 +102,23 @@ def get_builder_from_protocol(cls, code, parent_folder=None, protocol=None, over
...
-        builder.ph['qpoints'] = qpoints
+
+        if 'qpoints' in inputs['ph']:
+            qpoints_mesh = inputs['ph'].pop('qpoints')
+            qpoints = orm.KpointsData()
+            qpoints.set_kpoints_mesh(qpoints_mesh)
+            builder.ph['qpoints'] = qpoints
+        else:
+            inputs = {
+                'structure': structure,
```
sphuber, line 113: Just wondering if we should get the `structure` from the `parent_folder`? It is a bit surprising that the `parent_folder` is optional in this method, because it is required for the `PhCalculation`, which makes sense. You cannot run a `ph.x` calculation from scratch but always need a completed `pw.x` calculation.

I am wondering then if we should not make `parent_folder` a required input and remove the `structure` input. We can then obtain the `structure` from the `parent_folder` by doing:
```python
try:
    # Use the output structure by default
    structure = parent_folder.creator.outputs.output_structure
except AttributeError:
    # Fall back on the input structure which should always exist
    structure = parent_folder.creator.inputs.structure
```

src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -103,7 +102,23 @@ def get_builder_from_protocol(cls, code, parent_folder=None, protocol=None, over
...
+                'distance': orm.Float(inputs['qpoints_distance']),
```
sphuber, line 114: Does this actually work? I think the inputs are not guaranteed to have a `qpoints_distance` do they? I think a default value should be added to the protocol file.

src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -103,7 +102,23 @@ def get_builder_from_protocol(cls, code, parent_folder=None, protocol=None, over
...
+                'force_parity': orm.Bool(False),
```
sphuber, line 115: Same for this. Would add a default to the protocol, and then get the value from `inputs['force_parity']` allowing a user to override the default value with the `overrides` argument.

sphuber (changes requested):
Thanks @AndresOrtegaGuerrero have some questions/suggestions

src/aiida_quantumespresso/workflows/ph/base.py
AndresOrtegaGuerrero: Yes, I think it would be the best way to add `qpoints_distance` in the protocol, and also to create a dedicated input in the workchain. Maybe to do a implementation similar that kpoints in `PwBaseWorkChain`. However i dont know if I should also add qpoints in PhWorkChain

src/aiida_quantumespresso/workflows/ph/base.py
AndresOrtegaGuerrero: @sphuber  Thank you! I am trying to think in the best way to organize this PR (design-wise) , because it should be linked or beneficial for `aiida-quantumespresso-ph`. `PhCalculation` has the input `parent_folder` and these are expose in `PhBaseWorkChain` so we could include this logic

src/aiida_quantumespresso/workflows/ph/base.py
AndresOrtegaGuerrero: Then do you think we should also add force_parity and qpoints_distance as required inputs for the workchain?

src/aiida_quantumespresso/workflows/ph/base.py
AndresOrtegaGuerrero: In this case, parent_folder is a expose_input from `PhCalculation`. In the `get_builder_from_protocol` there is the parent_folder so I should instead call the structure from it.

AndresOrtegaGuerrero:
@sphuber  Thank you for your suggestions, I will organize this PR, to include all of them. Let me know your opinion of the creation of a function in the workchain validate_qpoints in the workchain. I am a lil skeptical to include qpoints as input of the workchain (Like in PwBaseWorkchain). I guess if these new inputs we discussed are added, I should also consider a test for it as well.

src/aiida_quantumespresso/workflows/ph/base.py
sphuber: No, I don't think they should be required because it should still be possible to pass explicit `qpoints` directly. I would literally mirror what the `PwBaseWorkChain` does and add them as optional inputs, and make the `qpoints` also optional. Then you add the same validation that either `qpoints` or `qpoints_distance` has to be defined.

src/aiida_quantumespresso/workflows/ph/base.py
sphuber: Ah, I now see this comment. This is exactly what I described in my response to the other comment, so yes, I think this should be the correct approach.

src/aiida_quantumespresso/workflows/ph/base.py
AndresOrtegaGuerrero: Yes, that what I mean, thank you! I will do it like that

AndresOrtegaGuerrero:
For a new protocol using distance I have done some test with the function` create_kpoints_from_distance` , we could use `qpoints_distance = 0.6` for fast, `qpoints_distance = 0.3` for moderate (Default), and `qpoints_distance = 0.2` precise

src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -45,6 +56,8 @@ def define(cls, spec):
             cls.results,
         )
         spec.expose_outputs(PhCalculation, exclude=('retrieved_folder',))
+        spec.exit_code(400, 'ERROR_INVALID_INPUT_QPOINTS',
```
bastonero, line 59: Why can't we already raise the error before the process is started? E.g. via a `validator`?

bastonero (commented):
Thanks a lot @AndresOrtegaGuerrero ! I was looking for doing this, but you already did it. Is there something important on why this is not yet merged? @sphuber @mbercx ?
I would love to see this merges 🚀

mbercx:
>  I was about to implement this, but you already did it. Is there something important on why this is not yet merged? @sphuber @mbercx ?

I must admit when I see @sphuber started reviewing a PR I stop paying attention myself. ^^

Would be good to have this feature indeed. @sphuber @AndresOrtegaGuerrero happy to help if needed.

src/aiida_quantumespresso/workflows/ph/base.py
AndresOrtegaGuerrero: Hi @bastonero , could you put an example ?

src/aiida_quantumespresso/workflows/ph/base.py
bastonero: You can add after the input specifications
```python
spec.inputs.validator = validate_inputs
```
and then outside the class something like this:
```python
def validate_inputs(inputs, _):
    """Validate the top level namespace."""
    if 'qpoints_distance' not in inputs and 'qpoints' not in inputs:
            return 'Either `qpoints_distance` or `qpoints` must be given`
```
And then it would be great if you can write a simple test (you can look into the tests folder and look for something similar in the validations of the inputs)

src/aiida_quantumespresso/workflows/ph/base.py
bastonero: You can add after the input specifications
```python
spec.inputs.validator = validate_inputs
```
and then outside the class something like this:
```python
def validate_inputs(inputs, _):
    """Validate the top level namespace."""
    if 'qpoints_distance' not in inputs and 'qpoints' not in inputs:
            return 'Either `qpoints_distance` or `qpoints` must be given`
```
And then it would be great if you can write a simple test (you can look into the tests folder and look for something similar in the validations of the inputs)

bastonero: This way you don't start the process from the very beginning avoiding to store a node that would fail due to "bad inputs"

sphuber:
I was simply waiting for an update after the review. @AndresOrtegaGuerrero please let me know if you are done with the changes and I will review it again.

AndresOrtegaGuerrero:
@sphuber @mbercx @bastonero . Hello , I actually stoped working on it since other things came first. I included the validator. I saw a similar one for the PwBandsWorkChain, yet i didnt see the test? any suggestions ?. On the other hand, @bastonero could you test if the branch is behaving as expected ?

bastonero:
Hello @AndresOrtegaGuerrero, thanks a lot for prioritizing this!

> On the other hand, @bastonero could you test if the branch is behaving as expected ?

I don't know if I can push commits to your empa branch. Nevertheless, you can proceed as follows:
```pip install -e .[tests]``` from the aiida-quantumespresso folder
Then add something like this into `tests/workflows/ph/test_base.py`
```python
@pytest.mark.usefixtures('aiida_profile')
def test_invalid_inputs(generate_workchain_ph, generate_inputs_ph):
    """Test `PhBaseWorkChain` validation methods."""
    inputs ={'ph': generate_inputs_ph()}
    # modify inputs to make them invalid - e.g. inputs['qpoints_distance'] = orm.Float(0.2) ?
    message = 'WRITE HERE THE ERROR MESSAGE YOU EXPECT'
    with pytest.raises(ValueError, match=message):
        generate_workchain_ph(inputs=inputs)
```
Then do
```
cd tests/workflows/ph/
pytest -sv test_base.py
```
And check everything is working. If not, you can change the code where it complains and run the tests again, till they succeed. Then you should probably pull from the main branch, as I added new tests for the ph protocols. Since now you changed it to have the `qpoints_distance`, you will have to just change the files for the regression tests (i.e. `tests/workflows/protocols/ph/test_base/test_default.yaml`). See also how I have done it (https://github.com/aiidateam/aiida-quantumespresso/commit/39287e03cb6bbf1915662685a5c441e9c7c36030). I think it's useful for you to learn how to do tests, as it is the best and fastest way to develop in aiida, and avoids to run actual jobs.
So check also the tests for the protocol, i.e. test everything in `tests/workflows/protocols/ph`.

Let me know if you have any trouble, and I'll try to help out!

tests/workflows/ph/test_base.py
```diff
@@ -9,6 +9,14 @@
 from aiida_quantumespresso.calculations.ph import PhCalculation
 from aiida_quantumespresso.workflows.ph.base import PhBaseWorkChain
 
+@pytest.mark.usefixtures('aiida_profile')
+def test_invalid_inputs(generate_workchain_ph, generate_inputs_ph):
+    """Test `PhBaseWorkChain` validation methods."""
+    inputs ={'ph': generate_inputs_ph()}
+    message = r'Neither `qpoints` nor `qpoints_distance` was specified.'
```
bastonero, line 16: "_were_ specified"

src/aiida_quantumespresso/workflows/protocols/ph/base.yaml
```diff
@@ -1,4 +1,5 @@
 default_inputs:
+    qpoinsts_distance: 0.3
```
bastonero, line 2:
```suggestion
    qpoints_distance: 0.3
```

src/aiida_quantumespresso/workflows/protocols/ph/base.yaml
```diff
@@ -11,27 +12,22 @@ default_inputs:
...
-          - 3
-          - 3
+
 default_protocol: moderate
 protocols:
     moderate:
         description: 'Protocol to perform the computation at normal precision at moderate computational cost.'
     precise:
         description: 'Protocol to perform the computation at high precision at higher computational cost.'
+        qpoinsts_distance: 0.2
```
bastonero, line 22:
```suggestion
        qpoints_distance: 0.2
```

src/aiida_quantumespresso/workflows/protocols/ph/base.yaml
```diff
@@ -11,27 +12,22 @@ default_inputs:
...
         ph:
             parameters:
                 INPUTPH:
                     tr2_ph: 1.0e-20
     fast:
         description: 'Protocol to perform the computation at low precision at minimal computational cost for testing purposes.'
+        qpoinsts_distance: 0.6
```
bastonero, line 29:
```suggestion
        qpoints_distance: 0.6
```

src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -34,9 +41,20 @@ def define(cls, spec):
         super().define(spec)
         spec.expose_inputs(PhCalculation, namespace='ph')
         spec.input('only_initialization', valid_type=orm.Bool, default=lambda: orm.Bool(False))
+        spec.input('qpoints', valid_type=orm.KpointsData, required=False,
+            help='An explicit qpoints list or mesh. Either this or `qpoints_distance` has to be provided.')
```
bastonero, line 45:
```suggestion
            help='An explicit qpoints list or mesh. Either this or `qpoints_distance` should be provided.')
```

src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -46,6 +64,8 @@ def define(cls, spec):
             cls.results,
         )
         spec.expose_outputs(PhCalculation, exclude=('retrieved_folder',))
+        spec.exit_code(400, 'ERROR_INVALID_INPUT_QPOINTS',
+                       message='Neither `qpoints` nor `qpoints_distance` was specified.')
```
bastonero, line 68:
```suggestion
                       message='Neither `qpoints` nor `qpoints_distance` were specified.')
```

src/aiida_quantumespresso/workflows/ph/base.py
bastonero: Do we really need the exit code here?

AndresOrtegaGuerrero:
I am still addressing the tests, i will request review once is ready. I did push but is not ready sorry about that

src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -121,7 +138,16 @@ def get_builder_from_protocol(
         if 'settings' in inputs['ph']:
             builder.ph['settings'] = orm.Dict(inputs['ph']['settings'])
         builder.clean_workdir = orm.Bool(inputs['clean_workdir'])
-        builder.ph['qpoints'] = qpoints
+
+        if 'qpoints' in inputs:
```
AndresOrtegaGuerrero, line 142: @sphuber , @mbercx I wonder how to address something. I added this logic about qpoints, but i declareded the input as `spec.input('qpoints', valid_type=orm.KpointsData, required=False,
            help='An explicit qpoints list or mesh. Either this or `qpoints_distance` should to be provided.')` However I think it should be a list in order to not interfere with the logic from aiid-quantumespresso-ph ? Any opinions about this? Because i guess i should add test for this as well?

src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -16,6 +17,12 @@
 PwCalculation = CalculationFactory('quantumespresso.pw')
 
 
+def validate_inputs(inputs):
+    """Validate the top level namespace."""
+    if 'qpoints_distance' not in inputs and 'qpoints' not in inputs:
+        return PhBaseWorkChain.exit_codes.ERROR_INVALID_INPUT_QPOINTS.message
```
bastonero, line 23: I would probably avoid to define the exit code, since it's a validation error, hence it's not a proper exit code.

src/aiida_quantumespresso/workflows/ph/base.py
bastonero: See also the validation function comment, but essentially I would remove this exit code. It's not useful as it is only a validation error

src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -148,6 +172,38 @@ def validate_parameters(self):
...
+        """Validate the inputs related to qpoints.
+
+        Either an explicit `KpointsData` with given mesh/path, or a desired qpoints distance should be specified. In
+        the case of the latter, the `KpointsData` will be constructed for the input `StructureData`
+        from the parent_folder using the `create_kpoints_from_distance` calculation function.
+        """
+
+        if all(key not in self.inputs for key in ['qpoints', 'qpoints_distance']):
+            return self.exit_codes.ERROR_INVALID_INPUT_QPOINTS
+
```
bastonero, line 185: I guess we can drop these lines as the validation is done at the beginning

bastonero: And I would probably rename the function `set_qpoints`

tests/workflows/protocols/ph/test_base/test_default.yml
```diff
@@ -11,10 +11,5 @@ ph:
   parameters:
     INPUTPH:
       tr2_ph: 1.0e-18
-  qpoints:
```
bastonero, line 14: Could you also add a test where you test the qpoint when given as a list?

src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -121,7 +136,16 @@ def get_builder_from_protocol(
         if 'settings' in inputs['ph']:
             builder.ph['settings'] = orm.Dict(inputs['ph']['settings'])
         builder.clean_workdir = orm.Bool(inputs['clean_workdir'])
-        builder.ph['qpoints'] = qpoints
+
+        if 'qpoints' in inputs:
+            qpoints_mesh = inputs['qpoints']
+            qpoints = orm.KpointsData()
+            qpoints.set_kpoints_mesh(qpoints_mesh)
+            builder.qpoints = qpoints
```
bastonero, line 144: I would add a test for this piece of code

bastonero (changes requested):
Thanks a lot @AndresOrtegaGuerrero ! A part of minor things, everything in good shape!

AndresOrtegaGuerrero:
@bastonero  thank you i will address your comments!

bastonero (dismissed):
Thanks @AndresOrtegaGuerrero, looks great!

src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -34,9 +41,20 @@ def define(cls, spec):
         super().define(spec)
         spec.expose_inputs(PhCalculation, namespace='ph')
```
mbercx, line 42: We should also exclude the `qpoints` input of the wrapped `PhCalculation` here:

```suggestion
        spec.expose_inputs(PhCalculation, namespace='ph', exclude=('qpoints', ))
```

src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -78,7 +96,6 @@ def get_builder_from_protocol(
         """Return a builder prepopulated with inputs selected according to the chosen protocol.
 
         :param code: the ``Code`` instance configured for the ``quantumespresso.ph`` plugin.
-        :param structure: the ``StructureData`` instance to use.
```
mbercx, line 81: Thanks for fixing this copy-pasta. :)

src/aiida_quantumespresso/workflows/ph/base.py
mbercx: Sorry for the slow response @AndresOrtegaGuerrero! I guess this comment is outdated? I realise now that this same logic is actually broken in the `PwBaseWorkChain`, will open a fix. 

https://github.com/aiidateam/aiida-quantumespresso/blob/bec6dd6b56b4cd3bbed3f3ab8fb97c7f7bdc0214/src/aiida_quantumespresso/workflows/pw/base.py#L225-L226

src/aiida_quantumespresso/workflows/ph/base.py
mbercx: One thought, not to be implemented here but to consider: What is more user friendly:

1. Having both the `qpoints` and `qpoints_distance` at the `PhBaseWorkChain` level, as is the case now?
2. Only adding an (optional) `qpoints_distance` input at the `PhBaseWorkChain` level, and leaving the `qpoints` input at the `PhCalculation` level?

mbercx (changes requested):
Thanks a lot @AndresOrtegaGuerrero! Changes are looking good, just had a few final comments.

src/aiida_quantumespresso/workflows/ph/base.py
AndresOrtegaGuerrero: This i am not sure :D , i didnt want to change the logic thinking of how it could affect the workflows of aiida_quantumespresso_ph , @bastonero do you have an opinion on this ?

bastonero: Well honestly since the `PwBaseWorkChain` and other workchains has `qpoints` at the same level of `qpoints_distance`, I would put `qpoints` higher up. We should push the idea of using only `WorkChains` and not `CalcJobs`. But above of all, documentation is lacking ^^

mbercx: True true, my thinking was that we still specify other direct inputs at the calcjob level, so maybe it's more reasonable keeping it there, instead of pushing it higher. But I have no strong opinion.

Except the documentation part. That I fully agree 😓

bastonero: Yeah I understand the point. I just don't know what's the amount of people we annoy when changing patterns. My strong opinion is to have the same "patterns" over the different WorkChains, so one expects `qpoints` or `kpoints` at a certain level. Which one I don't know. But probably i would stick to moving it higher just for historical reasons, and then maybe having an other related issue where to discuss this, and put it e.g. in `5.0`?

mbercx: Fully agree, I wouldn't have suggested implementing it in this PR. Will give it some more thought and potentially open an issue. :)

mbercx (approved):
LGTM! Thanks again @AndresOrtegaGuerrero.

EDIT: @sphuber merging is still blocked since you have requested changes. Would you still like to have a look?

sphuber (dismissed):

PR #959, opened 2023-06-20
PP parser: attach context to the `ERROR_OUTPUT_DATAFILE_PARSE` exit code.

yakutovicha:
It would have been much faster to figure out the origin of the problem with the PP parser if the exit code could have told me a little bit more about it.

### DIFF
--- a/src/aiida_quantumespresso/calculations/pp.py
+++ b/src/aiida_quantumespresso/calculations/pp.py
@@ -116,7 +116,7 @@ class PpCalculation(CalcJob):
         spec.exit_code(332, 'ERROR_UNSUPPORTED_DATAFILE_FORMAT',
             message='The data file format is not supported by the parser')
         spec.exit_code(333, 'ERROR_OUTPUT_DATAFILE_PARSE',
-            message='The formatted data output file `{filename}` could not be parsed.\n{context}')
+            message='The formatted data output file `{filename}` could not be parsed: {exception}')
         # yapf: enable
 
     def prepare_for_submission(self, folder):  # pylint: disable=too-many-branches,too-many-statements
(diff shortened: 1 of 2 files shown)


src/aiida_quantumespresso/calculations/pp.py
```diff
@@ -116,7 +116,7 @@ def define(cls, spec):
         spec.exit_code(332, 'ERROR_UNSUPPORTED_DATAFILE_FORMAT',
             message='The data file format is not supported by the parser')
         spec.exit_code(333, 'ERROR_OUTPUT_DATAFILE_PARSE',
-            message='The formatted data output file `{filename}` could not be parsed')
+            message='The formatted data output file `{filename}` could not be parsed.\n{context}')
```
sphuber, line 119:
```suggestion
            message='The formatted data output file `{filename}` could not be parsed: {exception')
```

src/aiida_quantumespresso/parsers/pp.py
```diff
@@ -132,8 +132,8 @@ def get_key_from_filename(filename):
                     key = get_key_from_filename(filename)
                     data_parsed.append((key, parsers[iflag](data_raw, self.units_dict[parsed_data['plot_num']])))
                     del data_raw
-                except Exception:  # pylint: disable=broad-except
-                    return self.exit_codes.ERROR_OUTPUT_DATAFILE_PARSE.format(filename=filename)
+                except Exception as exc:  # pylint: disable=broad-except
+                    return self.exit_codes.ERROR_OUTPUT_DATAFILE_PARSE.format(filename=filename, context=exc)
```
sphuber, line 136:
```suggestion
                except Exception as exception:  # pylint: disable=broad-except
                    return self.exit_codes.ERROR_OUTPUT_DATAFILE_PARSE.format(filename=filename, exception=exception)
```

sphuber (changes requested):
Makes sense to me, thanks @yakutovicha . Just slight changes to keep in line with other exit codes

PR #969, opened 2023-09-21
XAS: Enable Correct Parsing of Hubbard and Magnetic Data

PNOGillespie:
## Overview

In this PR, we update the various `WorkChains` and `CalcFunctions` relevant for calculations with XSpectra to work with the new `HubbardStructureData` data type and to handle system with magnetic data reliably, thus making calculations of systems with known magnetic and Hubbard parameter configurations possible with each `WorkChain`.

## Changes
- `get_xspectra_structures`:
     - Passes Hubbard parameters from the input structure to the structures returned by the `CalcFunction`. If a supercell is created from the input structure, then the parameters are scaled to the new supercell using `HubbardUtils.get_hubbard_for_supercell()`.
     - Fixes a previous issue with the `CalcFunction` where the information on equivalent atomic sites was taken from the standardized structure even if the input structure was not standardized.
     - Introduces a new `Bool` input `use_element_types` which instructs the `CalcFunction` to treat `Kind`s of the same element as the same for the purposes of symmetry analysis even if their `kind_name`s differ.
- `XspectraCrystalWorkChain`:
     - Updates the `WorkChain` to set `standardize_structure = False` if the input structure is `HubbardStructureData` in order to preserve the Hubbard data during structure creation.
     - Automatically sets the `starting_magnetization` of the absorbing atom in each subsequent `XspectraCoreWorkChain` based on the magnetic moment given for the `Kind` that it replaces. Note that this only works if `starting_magnetization` is provided as a `dict` instead of separate entries of `starting_magnetization(n)`.
- `XspectraCoreWorkChain`:
     - Updates the `WorkChain` so that it applies the GIPAW pseudopotential provided for the calculation to all `Kinds` corresponding to the absorbing element even if multiple of such exist.


superstar54 (changes requested):
Hi @PNOGillespie , thanks for the work. 
Could you add a test for the `get_xspectra_structure` using `HubbardStructureData`?

src/aiida_quantumespresso/workflows/functions/get_xspectra_structures.py
```diff
@@ -203,38 +229,83 @@ def get_xspectra_structures(structure, **kwargs):  # pylint: disable=too-many-st
...
+        # then we must briefly operate on a "cleaned" version of the
+        # structure tuple.
+        if use_element_types:
+            clean_structure_tuple = (spglib_tuple[0], spglib_tuple[1], [])
+            for i in spglib_tuple[2]:
+                if i >= 1000:
+                    new_i = int(i / 1000)
+                else:
+                    new_i = i
+                clean_structure_tuple[2].append(new_i)
```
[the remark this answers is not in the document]
PNOGillespie: There should now be a more detailed comment above this block of code to explain what's going on, let me know if this is clear enough.

PNOGillespie:
Hi @superstar54. I noted (and fixed) a bug that I spotted with the function which will be relevant for cases where different `Kind` names applied to the structure would break the symmetry (e.g. in the case of antiferromagnetic systems). This should now be ready for final review.

If you could give the PR another review (or ping someone with write access) that would be much appreciated.

src/aiida_quantumespresso/workflows/functions/get_xspectra_structures.py
```diff
@@ -203,58 +229,116 @@ def get_xspectra_structures(structure, **kwargs):  # pylint: disable=too-many-st
...
+        # new label is reduced to its normal element number.
+        if use_element_types:
+            cleaned_structure_tuple = (spglib_tuple[0], spglib_tuple[1], [])
+            for i in spglib_tuple[2]:
+                if i >= 1000:
+                    new_i = int(np.trunc(i / 1000))
+                else:
+                    new_i = i
+                cleaned_structure_tuple[2].append(new_i)
+            cleaned_symmetry_dataset = spglib.get_symmetry_dataset(cleaned_structure_tuple, **spglib_kwargs)
```
superstar54, line 264: I think there is no need to create a new `clearned_symmetry_dataset` here. For the following reasons:
- Since the `symmetry_dataset` is used to check if standardization is needed or not. `use_element_types` will affect the check.
- There are too many `if use_element_types:` check, we can avoid it if use the same name.

```suggestion
        if use_element_types:
            cleaned_structure_tuple = (spglib_tuple[0], spglib_tuple[1], [])
            for i in spglib_tuple[2]:
                if i >= 1000:
                    new_i = int(np.trunc(i / 1000))
                else:
                    new_i = i
                cleaned_structure_tuple[2].append(new_i)
            symmetry_dataset = spglib.get_symmetry_dataset(cleaned_structure_tuple, **spglib_kwargs)
        else:
            symmetry_dataset = spglib.get_symmetry_dataset(spglib_tuple, **spglib_kwargs)
```

superstar54 (changes requested):
@PNOGillespie Thanks for the update!
I have a feeling that the `get_xspectra_structures` function is too long and complex with many nested `for` and `if` block. Thus, it's hard for people to follow the logic. I would suggest simplifying its logic through modularization. You can create sub-functions for some key tasks, such as the `generate_hubbard_supercell` function, etc.

PNOGillespie:
Hi @superstar54. I reviewed the source code to see if I could de-convolute and consolidate some of the process logic. I have managed to reduce the number of `if` statements somewhat in a few different ways, but modularisation won't simplify things here since there are very few cases where code is really re-used.

In regards to your suggestion here:
>  think there is no need to create a new `clearned_symmetry_dataset` here.

The `CalcFunction` needs to know which `Kind` *should* be present at each site, so *both* "cleaned" and "non-cleaned" datasets are needed in order to map the correct `Kind` to each chosen absorbing atom while also determining the symmetry properties correctly (i.e. where we assume that all `Kind`s of the same element are the same). This information is then needed by `XspectraCrystalWorkChain` in order to correctly set the magnetic moment (particularly for an anti-ferromagnetic system).

I reduced the number of `if use_element_types` statements by taking your suggestion and then generating the arrays needed to determine which `Kind`s correspond to which site on-the-fly, since the alternative is to constantly check if `use_element_types == True`.

All this behaviour should be covered by the tests, since there are test cases for c-Si using `[Si, Si]` and `[Si0, Si1]`

Let me know what you think of the changes.

src/aiida_quantumespresso/workflows/functions/get_xspectra_structures.py
```diff
@@ -282,42 +280,32 @@ def get_xspectra_structures(structure, **kwargs):  # pylint: disable=too-many-st
...
         else:  # convert the 'std_types' from standardized to primitive cell
-            spglib_std_types = symmetry_dataset['std_types']
-            spglib_std_map_to_prim = symmetry_dataset['std_mapping_to_primitive']
+            # we generate the type-specific data on-the-fly since we need to
+            # know which type (and thus kind) *should* be at each site
+            # even if we "cleaned" the structure previously
+            spglib_std_types = spglib.get_symmetry_dataset(spglib_tuple, **spglib_kwargs)['std_types']
+            spglib_map_to_prim = spglib.get_symmetry_dataset(spglib_tuple, **spglib_kwargs)['mapping_to_primitive']
+            spglib_std_map_to_prim = spglib.get_symmetry_dataset(spglib_tuple,
+                                                                 **spglib_kwargs)['std_mapping_to_primitive']
```
superstar54, line 294: please change it to run `get_symmetry_dataset` only one time.

superstar54 (changes requested):
A small change is needed.

superstar54 (approved):
LGTM! Thanks for the work! @PNOGillespie

PR #972, opened 2023-10-17
Fix tot magnetization base

AndresOrtegaGuerrero:
If tot_magnetization is used in overrided , this logic should remove starting_magnetization


src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -198,6 +198,10 @@ def get_builder_from_protocol(
             parameter_overrides = overrides.get('pw', {}).get('parameters', {})
             parameters = recursive_merge(parameters, parameter_overrides)
 
+            # if tot_magnetization in overrides , remove starting_magnetization from parameters
+            if overrides['pw']['parameters']['SYSTEM'].get('tot_magnetization') is not None:
+                parameters['SYSTEM'].pop('starting_magnetization', None)
```
sphuber, line 203: You need to guard against `SYSTEM` not yet being defined.
```suggestion
            # if tot_magnetization in overrides , remove starting_magnetization from parameters
            if parameters.get('SYSTEM', {}).get('tot_magnetization') is not None:
                parameters.setdefault('SYSTEM', {}).pop('starting_magnetization', None)
```

sphuber (changes requested):
Thanks @AndresOrtegaGuerrero . Added suggestion that I think should fix the failing tests

src/aiida_quantumespresso/workflows/pw/base.py
AndresOrtegaGuerrero: @sphuber Thank you so much for the help! , I will also continue checking the PdosWorkChain and BandsWorkChain, since somehow it produces a problem of not having a fermi_energy

sphuber:
@AndresOrtegaGuerrero I fixed the docs. Can you please update your branch? Normally I can do it, but for some reason it is not letting me this time

AndresOrtegaGuerrero:
@AndresOrtegaGuerrero Thank you ! I just updated, maybe because i am doing the PR from the fork at nanotech-empa ?

sphuber (approved):
Thanks @AndresOrtegaGuerrero

PR #974, opened 2023-10-19
`Q2rCalculation`: Add the `output_parameters` output to the spec

### LINKED ISSUE #971: q2r "output_parameters"
Hello.

I am trying the q2r calculation and get the error
`Error: invalid value uuid: 2f4cb14b-ad18-4587-9613-7c75d1d6b4e7 (unstored) specified with label output_parameters: Error validating output 'uuid: 2f4cb14b-ad18-4587-9613-7c75d1d6b4e7 (unstored)' for port 'outputs': Unexpected ports {'output_parameters': <Dict: uuid: 2f4cb14b-ad18-4587-9613-7c75d1d6b4e7 (unstored)>}, for a non dynamic namespace
`
I found that in Q2rParser.parse(), there is a line
`self.out('output_parameters', Dict(parsed_data))`
but there is no definition of `output_parameters` in the q2r calculation.

Should the line in Q2rParser.parse() be deleted?

sphuber:
Fixes #971 

The `Q2rCalculation` did not declare the `output_parameters` output even though the `Q2rParser` attaches it always. This would lead to the calculation excepting because an unknown output would be attached. This was uncaught by the test because the `Parser.parse_from_node` method does not automatically validate the outputs. This is because the output spec that is checked is that of the `calcfunction` (which accepts anything) and not that of the `Q2rCalculation`.

### DIFF
--- a/src/aiida_quantumespresso/calculations/q2r.py
+++ b/src/aiida_quantumespresso/calculations/q2r.py
@@ -32,6 +32,7 @@ class Q2rCalculation(NamelistsCalculation):
         super().define(spec)
         spec.input('parent_folder', valid_type=(orm.RemoteData, orm.FolderData), required=True)
         spec.output('force_constants', valid_type=ForceConstantsData)
+        spec.output('output_parameters', valid_type=orm.Dict)
         spec.exit_code(330, 'ERROR_READING_FORCE_CONSTANTS_FILE',
             message='The force constants file could not be read.')
         # yapf: enable


mbercx (approved):
Thanks @sphuber. I find it somewhat strange I hadn't run into this before, but I probably didn't run any `q2r.x` calculations anymore after https://github.com/aiidateam/aiida-quantumespresso/commit/1c223c78f0a2eda38089a08d9f21626f49d2fd6b was merged.

PR #977, opened 2023-10-28
Dependencies: Update `pydantic~=2.4`

sphuber:
The requirement for `pydantic` was pinned to `v1` since `v2` has a lot of backwards incompatible changes and it is difficult to provide a version that is compatible with both versions.

As of `v2.5.0`, `aiida-core` also directly depends on `pydantic` and it requires `~=2.4`, so here we apply the same requirement. The deprecated code is replaced.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ dependencies = [
     'importlib_resources',
     'jsonschema',
     'numpy',
-    'pydantic~=1.10,>=1.10.8',
+    'pydantic~=2.0',
     'packaging',
     'qe-tools~=2.0',
     'xmlschema~=2.0'
--- a/src/aiida_quantumespresso/calculations/pw.py
+++ b/src/aiida_quantumespresso/calculations/pw.py
@@ -210,7 +210,7 @@ class PwCalculation(BasePwCpInputGenerator):
                     'Please set one of the following in the input parameters:\n'
                     "    parameters['CONTROL']['restart_mode'] = 'restart'\n"
                     "    parameters['ELECTRONS']['startingpot'] = 'file'\n"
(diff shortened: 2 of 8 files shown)


mbercx:
Thanks @sphuber! Quick question: if we merge this, the next `aiida-quantumespresso` release would only be compatible with `aiida-core~=2.5`, correct? Don't we have to update the `aiida-core` dependency then? And if we do, do we still need an explicit dependency on `pydantic` here?

sphuber:
> Quick question: if we merge this, the next aiida-quantumespresso release would only be compatible with aiida-core~=2.5, correct?

Indirectly, yes. But you can put it also the other way around. Without this fix, `aiida-quantumespresso` will be incompatible with `aiida-core>=2.5`. Given the big changes between pydantic v1 and v2, it is almost impossible to have compatibility for both versions at the same time, and so our hand is pretty much forced. But `aiida-core` is forward compatible in the minor versions, so it shouldn't be a problem. Everyone should be able to update without issues.

> Don't we have to update the aiida-core dependency then?

Technically that is not really our responsiblity. Package managers like `pip` will figure this out.

> And if we do, do we still need an explicit dependency on pydantic here?

I would say yes absolutely, because we directly use it. So we should specify what requirements `aiida-quantumespresso` has on that package

mbercx (approved):
No notes, thanks @sphuber!

unkcpz:
@sphuber @mbercx Plan to make a release with this commit? We want to give it a test on AiiDAlab.

sphuber:
I think we are long due for a release. Marnik had planned on preparing it but hasn't come around to it. I think I will start preparing it now and have him review it

PR #979, opened 2023-10-28
Devops: Address various deprecation warnings from `aiida-core`

sphuber:
AiiDA v2.5 upgraded to Sqlalchemy v2 which as of v2.0.19 emits a warning
that is entirely consequential but was too difficult to actually address.
As a workaround, `aiida-core` simply filters the warning. This works for
normal usage, but `pytest` eliminates the filter and so the warnings do
show up in the final warning report of `pytest`. Therefore the warning
has to be explicitly ignored in the `pytest` config as well.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -197,6 +197,7 @@ testpaths = [
     'tests',
 ]
 filterwarnings = [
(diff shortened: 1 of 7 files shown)


mbercx (approved):
LGTM 🚀

PR #981, opened 2023-11-04
Customised docker image for running calculation out-of-box

unkcpz:
supersede #980 since push to ghcr.io use `GITHUB_TOKEN` which has no write permission to package for forked PR.


mbercx:
Thanks @unkcpz! Just for some context: what is the goal here? To run full integration tests via a docker setup?

unkcpz:
Three goals from my point of view:
1. As an example to show how to set up a pre-installed computer/code in an image, the use case comes from https://aiida.discourse.group/t/building-docker-image-with-pre-configured-aiida-computer/132
2. An image for plugin developers that can easily run real calculation without burden of setup, this was mentioned by @bastonero in a slack channel if I remember correctly—pinning @bastonero for comment.
3. (Not important since there are already the integration tests run real calculations) an advantage I can notice is the image make it a bit easy to reproduce the integration test issue.

.docker/Dockerfile
```diff
@@ -0,0 +1,26 @@
+#FROM ghcr.io/aiidateam/aiida-core-with-services:edge
+FROM ghcr.io/aiidateam/aiida-core-with-services:pr-6170
```
unkcpz, line 2:
```suggestion
FROM ghcr.io/aiidateam/aiida-core-with-services:edge
```

.docker/tests/conftest.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@pytest.fixture(scope='session')
+def docker_compose(docker_services):
+    # pylint: disable=protected-access
+    return docker_services._docker_compose
+
+
+def is_container_ready(dodkec_compose):
+    output = dodkec_compose.execute('exec -T aiida verdi status').decode().strip()
```
unkcpz, line 21:
```suggestion
def is_container_ready(docker_compose):
    output = docker_compose.execute('exec -T aiida verdi status').decode().strip()
```

.docker/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM ghcr.io/aiidateam/aiida-core-with-services:edge
```
unkcpz, line 1:
```suggestion
FROM ghcr.io/aiidateam/aiida-core-with-services:pr-6176
#FROM ghcr.io/aiidateam/aiida-core-with-services:edge
```

.docker/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM ghcr.io/aiidateam/aiida-core-with-services:pr-6176
+#FROM ghcr.io/aiidateam/aiida-core-with-services:edge
```
unkcpz, line 2:
```suggestion
FROM ghcr.io/aiidateam/aiida-core-with-services:latest
```

bastonero:
Hi @unkcpz, indeed I once said it'd be great to have something like this. Re testing. I think it is extremely valuable and it can be used as an alternative to the virtual machine, which I think it has even too much stuff and it is (I guess) much more heavy to run.

PR #984, opened 2023-11-23
Fix `max_iterations` not overridden and add tests for `PhBaseWorkChain` protocols

bastonero:
All the base workchains were missing to override the `max_iterations` input.
The `PhBaseWorkChain.get_builder_from_protocol` was still untested. Tests are added and the `electronic_type` input is added as well to allow for the automated/user-friendly computation of dielectric and effective charges tensors when needed (i.e. for insulators).


sphuber (approved):
All good, thanks @bastonero

Issue #637, opened 2021-01-06
New handler for 520 pw errors (history already reset at previous step: stopping)


giovannipizzi:
While running many PwRelax workflows, I encountered a number of 520 errors from the PwCalculation (`handle_relax_recoverable_ionic_convergence_error_bfgs`, that arises when in the output we have `history already reset at previous step: stopping` from the BFGS algorithm).

From my understanding, this comes when we are already quite close to the energy minimum and the "noise" on the gradient is too large so the algorithm doesn't know where to move, and stops.
In practice, in most cases we have essentially already reached the minimum.

A practical solution in these cases would be to switch to damped dynamics, that while in general slower, are more robust and work fine in this case.

I've therefore implemented the following handler, whose diff is this:
```diff
     @process_handler(priority=560, exit_codes=[
         PwCalculation.exit_codes.ERROR_IONIC_CONVERGENCE_NOT_REACHED,
         PwCalculation.exit_codes.ERROR_IONIC_CYCLE_EXCEEDED_NSTEP,
+        ])
+    def handle_relax_recoverable_ionic_convergence_error(self, calculation):
+        """Handle various exit codes for recoverable `relax` calculations with failed ionic convergence.
+
+        These exit codes signify that the ionic convergence thresholds were not met, but the output structure is usable,
+        so the solution is to simply restart from scratch but from the output structure.
+        """
+        self.ctx.restart_calc = None
+        self.ctx.inputs.structure = calculation.outputs.output_structure
+        action = 'no ionic convergence but clean shutdown: restarting from scratch but using output structure.'
+        self.report_error_handled(calculation, action)
+        return ProcessHandlerReport(True)
+
+    @process_handler(priority=560, exit_codes=[
         PwCalculation.exit_codes.ERROR_IONIC_CYCLE_BFGS_HISTORY_FAILURE,
         PwCalculation.exit_codes.ERROR_IONIC_CYCLE_BFGS_HISTORY_AND_FINAL_SCF_FAILURE,
         ])
-    def handle_relax_recoverable_ionic_convergence_error(self, calculation):
+    def handle_relax_recoverable_ionic_convergence_error_bfgs(self, calculation):
         """Handle various exit codes for recoverable `relax` calculations with failed ionic convergence.
 
         These exit codes signify that the ionic convergence thresholds were not met, but the output structure is usable,
         so the solution is to simply restart from scratch but from the output structure.
[...]

PR #985, opened 2023-11-23
`PwBaseWorkChain`: handler for BFGS history failure

bastonero:
Sometimes the BFGS algorithm for ionic minimizaiton fails, and the current handler simply restart from scratch. This might work, but here we try to improve upon this simplistic solution, trying first to lower the trusted radius, and then trying to switch algorithm.

### DIFF
--- a/src/aiida_quantumespresso/workflows/pw/base.py
+++ b/src/aiida_quantumespresso/workflows/pw/base.py
@@ -505,9 +505,13 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         self.report_error_handled(calculation, action)
         return ProcessHandlerReport(True)
 
-    @process_handler(priority=559, exit_codes=[
-        PwCalculation.exit_codes.ERROR_IONIC_CYCLE_BFGS_HISTORY_FAILURE,
-    ])
+    @process_handler(
+        priority=559,
+        exit_codes=[
+            PwCalculation.exit_codes.ERROR_IONIC_CYCLE_BFGS_HISTORY_FAILURE,
+            PwCalculation.exit_codes.ERROR_IONIC_CYCLE_BFGS_HISTORY_AND_FINAL_SCF_FAILURE,
+        ]
+    )
     def handle_relax_recoverable_ionic_convergence_bfgs_history_error(self, calculation):
         """Handle failure of the ionic minimization algorithm (BFGS).
 
@@ -518,19 +522,28 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         algorithm, e.g. `damp` (and `damp-w` for vc-relax).
         """
         trust_radius_min = self.ctx.inputs.parameters.setdefault('IONS', {}).setdefault('trust_radius_min', 1.0e-3)
-        if trust_radius_min > 1.0e-4:
+        calculation_type = self.ctx.inputs.parameters.setdefault('CONTROL', {}).setdefault('calculation', 'relax')
+        ion_dynamics = self.ctx.inputs.parameters.setdefault('IONS', {}).setdefault('ion_dynamics', 'bfgs')
+        if calculation_type == 'vc-relax':
+            cell_dynamics = self.ctx.inputs.parameters.setdefault('CELL', {}).setdefault('cell_dynamics', 'bfgs')
+
+        if calculation_type == 'relax' and ion_dynamics == 'bfgs':
+            self.ctx.inputs.parameters.setdefault('IONS', {})['ion_dynamics'] = 'damp'
+            action = 'bfgs history (ionic only) failure: restarting with `damp` dynamics.'
+        elif calculation_type == 'relax' and ion_dynamics == 'damp':
+            self.ctx.inputs.parameters.setdefault('IONS', {})['ion_dynamics'] = 'fire'
+            action = 'bfgs history (ionic only) failure: restarting with `fire` dynamics.'
+        elif trust_radius_min > 1.0e-4 and ion_dynamics == 'bfgs' and calculation_type == 'vc-relax':
             self.ctx.inputs.parameters.setdefault('IONS', {})['trust_radius_ini'] = trust_radius_min  # start close
             new_trust_radius_min = trust_radius_min / 10.0
             self.ctx.inputs.parameters.setdefault('IONS', {})['trust_radius_min'] = new_trust_radius_min
-            action = f'bfgs history failure: restarting with trust_radius_min={new_trust_radius_min:.5f}.'
-        else:
+            action = f'bfgs history (vc-relax) failure: restarting with `trust_radius_min={new_trust_radius_min:.5f}`.'
+        elif calculation_type == 'vc-relax' and cell_dynamics == 'bfgs':
             self.ctx.inputs.parameters.setdefault('IONS', {})['ion_dynamics'] = 'damp'
-            action = (
-                f'bfgs history failure and `trust_radius_min < {trust_radius_min:.5f}`: '
-                'restarting with damp dynamics.'
-            )
-            if self.ctx.inputs.parameters['CONTROL']['calculation'] == 'vc-relax':
-                self.ctx.inputs.parameters.setdefault('CELL', {})['cell_dynamics'] = 'damp-w'
+            self.ctx.inputs.parameters.setdefault('CELL', {})['cell_dynamics'] = 'damp-w'
+            action = 'bfgs history (vc-relax) failure: restarting with `damp(-w)` dynamics.'
+        else:
+            return ProcessHandlerReport(False)
 
         self.ctx.inputs.structure = calculation.outputs.output_structure
 
--- a/src/aiida_quantumespresso/workflows/pw/base.py
+++ b/src/aiida_quantumespresso/workflows/pw/base.py
@@ -505,8 +505,41 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         self.report_error_handled(calculation, action)
         return ProcessHandlerReport(True)
 
+    @process_handler(priority=559, exit_codes=[
+        PwCalculation.exit_codes.ERROR_IONIC_CYCLE_BFGS_HISTORY_FAILURE,
+    ])
+    def handle_relax_recoverable_ionic_convergence_bfgs_history_error(self, calculation):
+        """Handle failure of the ionic minimization algorithm (BFGS).
+
+        When BFGS history fails, this can mean two things: the structure is close to the global minimum,
+        but the moves the algorithm wants to do are smaller than `trust_radius_min`, or the structure is
+        close to a local minimum (hard to detect). For the first, we restart with lowered trust_radius_min.
+        For the first case, one can lower the trust radius; for the second one, one can exploit a different
+        algorithm, e.g. `damp` (and `damp-w` for vc-relax).
+        """
+        trust_radius_min = self.ctx.inputs.parameters.setdefault('IONS', {}).setdefault('trust_radius_min', 1.0e-3)
+        if trust_radius_min > 1.0e-4:
+            self.ctx.inputs.parameters.setdefault('IONS', {})['trust_radius_ini'] = trust_radius_min  # start close
+            new_trust_radius_min = trust_radius_min / 10.0
+            self.ctx.inputs.parameters.setdefault('IONS', {})['trust_radius_min'] = new_trust_radius_min
+            action = f'bfgs history failure: restarting with trust_radius_min={new_trust_radius_min:.5f}.'
+        else:
+            self.ctx.inputs.parameters.setdefault('IONS', {})['ion_dynamics'] = 'damp'
+            action = (
+                f'bfgs history failure and `trust_radius_min < {trust_radius_min:.5f}`: '
+                'restarting with damp dynamics.'
+            )
+            if self.ctx.inputs.parameters['CONTROL']['calculation'] == 'vc-relax':
+                self.ctx.inputs.parameters.setdefault('CELL', {})['cell_dynamics'] = 'damp-w'
+
+        self.ctx.inputs.structure = calculation.outputs.output_structure
+
+        self.set_restart_type(RestartType.FROM_SCRATCH)
+        self.report_error_handled(calculation, action)
+        return ProcessHandlerReport(True)
+
     @process_handler(
-        priority=559, exit_codes=[
+        priority=555, exit_codes=[
             PwCalculation.exit_codes.ERROR_RADIAL_FFT_SIGNIFICANT_VOLUME_CONTRACTION,
         ]
     )
--- a/src/aiida_quantumespresso/workflows/pw/base.py
+++ b/src/aiida_quantumespresso/workflows/pw/base.py
@@ -250,6 +250,13 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         self.ctx.inputs.parameters.setdefault('ELECTRONS', {})
         self.ctx.inputs.parameters.setdefault('SYSTEM', {})
 
+        calculation_type = self.ctx.inputs.parameters['CONTROL'].get('calculation', None)
+        if calculation_type in ['relax', 'md']:
+            self.ctx.inputs.parameters.setdefault('IONS', {})
+        if calculation_type in ['vc-relax', 'vc-md']:
+            self.ctx.inputs.parameters.setdefault('IONS', {})
+            self.ctx.inputs.parameters.setdefault('CELL', {})
+
         self.ctx.inputs.settings = self.ctx.inputs.settings.get_dict() if 'settings' in self.ctx.inputs else {}
 
     def validate_kpoints(self):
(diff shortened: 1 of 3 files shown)


src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -504,8 +504,41 @@ def handle_relax_recoverable_ionic_convergence_error(self, calculation):
         self.report_error_handled(calculation, action)
         return ProcessHandlerReport(True)
 
+    @process_handler(priority=559, exit_codes=[
+        PwCalculation.exit_codes.ERROR_IONIC_CYCLE_BFGS_HISTORY_FAILURE,
```
sphuber, line 508: Should we remove this exit code from the handler above (with priority 560)? Otherwise it will be handled twice. Since the first handler will just set the restart type and report the error being handled, I don't think we need it. And do you maybe want to handle `ERROR_IONIC_CYCLE_BFGS_HISTORY_AND_FINAL_SCF_FAILURE` also in this new way?

bastonero: Why it will be handled twice? If the first handler fixes the issue, the second one should't be called, should it?
Re `ERROR_IONIC_CYCLE_BFGS_HISTORY_AND_FINAL_SCF_FAILURE` I wasn't unsure what the `FINAL_SCF_FAILURE` actually meant. But indeed we can make it to be handled too

bastonero:
Thanks @mbercx !
> Thanks @bastonero! Does this one fix #637?

Yes, that was the idea (I forgot the number of that issue, thanks for linking).

> I remember I was testing this when @giovannipizzi first opened the issue (and then got distracted, as usual), and the damped was _very_ effective for `relax`, but not-so-effective for `vc-relax`. The strategy here is already quite good, but we may consider immediately switching to the damped algorithm for `calculation == relax`.

Ok, amazing, then it's also tested! We can then proceed as you suggest.

bastonero:
@sphuber now working with the new logic suggested by @mbercx . I also added for the `relax` case the `fire` algorithm`. I think it's new, we should make sure from which version of QE it's implemented. But now we just support the latest 3 versions correct?

src/aiida_quantumespresso/workflows/pw/base.py
sphuber: You are right, it returns `ProcessHandlerReport(True)` where the `True` is for the `do_break` argument, which is `False` by default. If set to `True` all other handlers are skipped. But that makes the problem worse. This means that your new handler is never actually called, right? Because it first matches the handler with priority 560 and then breaks. So it will never get to the next one.

The `ERROR_IONIC_CYCLE_BFGS_HISTORY_AND_FINAL_SCF_FAILURE` exit code means that first BFGS failed due to the history and then _on top_ of that, the final scf also didn't converge. I think it makes sense to handle this as the normal BFGS history problem.

sphuber:
> But now we just support the latest 3 versions correct?

I think we only support v6.6 and newer, but maybe for the next release we already even planned to drop support for v6 completely. @mbercx ?

src/aiida_quantumespresso/workflows/pw/base.py
bastonero: Ok, interesting. But why it should stop first at the handler with priority `560`? I thought lower priority number meant it would be called sooner. Is it the opposite? If this is the case we can just change it to `561`

bastonero:
Looking into QE history it seems the `fire` algorithm is there since a while, so it should be a safe addition.

src/aiida_quantumespresso/workflows/pw/base.py
sphuber: Nope, they are called based on priority in reverse order:
https://github.com/aiidateam/aiida-core/blob/main/aiida/engine/processes/workchains/restart.py#L243
This means higher numbers first. See also the docs: https://aiida.readthedocs.io/projects/aiida-core/en/latest/howto/workchains_restart.html#multiple-process-handlers

> The process handlers with a higher priority will be called first.

src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -482,6 +482,52 @@ def handle_vcrelax_converged_except_final_scf(self, calculation):
...
+        elif calculation_type == 'vc-relax' and cell_dynamics == 'bfgs':
+            self.ctx.inputs.parameters.setdefault('IONS', {})['ion_dynamics'] = 'damp'
+            self.ctx.inputs.parameters.setdefault('CELL', {})['cell_dynamics'] = 'damp-w'
+            action = 'bfgs history (vc-relax) failure: restarting with `damp(-w)` dynamics.'
+        else:
+            return ProcessHandlerReport(False)
+
+        self.ctx.inputs.structure = calculation.outputs.output_structure
+
+        self.set_restart_type(RestartType.FROM_SCRATCH)
```
mbercx, line 527: Just a quick note here as I'm going through this: I think we've already discussed this before somewhere, but is there a reason why we wouldn't restart from the charge density for restarts where the structure has changed?

bastonero: Yeah, in principle we can do that. Shall I update?

bastonero: I did the same for the other handler with priority `560`

src/aiida_quantumespresso/workflows/pw/base.py
mbercx: To nitpick a bit further here ^^

If the `handle_relax_recoverable_ionic_convergence_bfgs_history_error` process handler is called _first_, and it switches away from `bfgs` at some point, can't we just remove the `ERROR_IONIC_CYCLE_BFGS_HISTORY_FAILURE` and `ERROR_IONIC_CYCLE_BFGS_HISTORY_AND_FINAL_SCF_FAILURE` exit codes from the `handle_relax_recoverable_ionic_convergence_error` process handler? There shouldn't be any BFGS history failures for `damp` or `fire`, I'd think?

src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -482,6 +482,52 @@ def handle_vcrelax_converged_except_final_scf(self, calculation):
...
+    def handle_relax_recoverable_ionic_convergence_bfgs_history_error(self, calculation):
+        """Handle failure of the ionic minimization algorithm (BFGS).
+
+        When BFGS history fails, this can mean two things: the structure is close to the global minimum,
+        but the moves the algorithm wants to do are smaller than `trust_radius_min`, or the structure is
+        close to a local minimum (hard to detect). For the first, we restart with lowered trust_radius_min.
+        For the first case, one can lower the trust radius; for the second one, one can exploit a different
+        algorithm, e.g. `damp` (and `damp-w` for vc-relax).
+        """
+        trust_radius_min = self.ctx.inputs.parameters.setdefault('IONS', {}).setdefault('trust_radius_min', 1.0e-3)
```
mbercx, line 501: Do you want to `setdefault` here? Or are you just trying the get the user-specified value _or_ the default?

It would probably be good to specify QE defaults in the typical `qe_defaults` dict:

https://github.com/aiidateam/aiida-quantumespresso/blob/b9c7517744e645a93d4afc9b1999881fc39a0e46/src/aiida_quantumespresso/workflows/pw/base.py#L29C4-L29C4

Althought it does make me wonder why we pass this to the `defaults` `AttributeDict` specified as a class attribute on the `PwBaseWorkChain`... Seems like an unnecessary step? But ok, won't introduce scope creep here.

src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -482,6 +482,52 @@ def handle_vcrelax_converged_except_final_scf(self, calculation):
...
+        calculation_type = self.ctx.inputs.parameters.setdefault('CONTROL', {}).setdefault('calculation', 'relax')
```
mbercx, line 502: Nitpick: The `CONTROL` top-level dictionary is already set to a default empty dictionary in the `setup` step:

https://github.com/aiidateam/aiida-quantumespresso/blob/b9c7517744e645a93d4afc9b1999881fc39a0e46/src/aiida_quantumespresso/workflows/pw/base.py#L247-L250

So it isn't necessary here.

Also, I think the default for `calculation` is `'scf'`?

https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm37

src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -482,6 +482,52 @@ def handle_vcrelax_converged_except_final_scf(self, calculation):
...
+        ion_dynamics = self.ctx.inputs.parameters.setdefault('IONS', {}).setdefault('ion_dynamics', 'bfgs')
```
mbercx, line 503: Similar comments on the `setdefault` and gathering the `qe_defaults`.

src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -482,6 +482,52 @@ def handle_vcrelax_converged_except_final_scf(self, calculation):
...
+        if calculation_type == 'vc-relax':
+            cell_dynamics = self.ctx.inputs.parameters.setdefault('CELL', {}).setdefault('cell_dynamics', 'bfgs')
+
+        if calculation_type == 'relax' and ion_dynamics == 'bfgs':
+            self.ctx.inputs.parameters.setdefault('IONS', {})['ion_dynamics'] = 'damp'
+            action = 'bfgs history (ionic only) failure: restarting with `damp` dynamics.'
+        elif calculation_type == 'relax' and ion_dynamics == 'damp':
```
mbercx, line 510: Will this ever happen? If you've already switched to `damp`, how can there be a BFGS failure?

mbercx: Maybe this logic should be moved to `handle_relax_recoverable_ionic_convergence_error`?

src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -482,6 +482,52 @@ def handle_vcrelax_converged_except_final_scf(self, calculation):
...
+            self.ctx.inputs.parameters.setdefault('IONS', {})['ion_dynamics'] = 'fire'
+            action = 'bfgs history (ionic only) failure: restarting with `fire` dynamics.'
+        elif trust_radius_min > 1.0e-4 and ion_dynamics == 'bfgs' and calculation_type == 'vc-relax':
```
mbercx, line 513: Nitpick: Just for consistency ^^

```suggestion
        calculation_type == 'vc-relax' and elif trust_radius_min > 1.0e-4 and ion_dynamics == 'bfgs':
```

src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -482,6 +482,52 @@ def handle_vcrelax_converged_except_final_scf(self, calculation):
...
+            self.ctx.inputs.parameters.setdefault('IONS', {})['trust_radius_ini'] = trust_radius_min  # start close
+            new_trust_radius_min = trust_radius_min / 10.0
```
mbercx, line 515: The canonical way would be to add this `10.0` as a default to our `defaults` class attribute:

https://github.com/aiidateam/aiida-quantumespresso/blob/b9c7517744e645a93d4afc9b1999881fc39a0e46/src/aiida_quantumespresso/workflows/pw/base.py#L28-L36

I know it might be a bit overkill and somewhat less readable, but I think at some point we want to move to making these configurable by the user as inputs, and then it'd be good to have them in one place.

mbercx (changes requested):
Thanks @bastonero! I've left some nitpicks/questions. ^^

src/aiida_quantumespresso/workflows/pw/base.py
bastonero: Yeah, just want to get the default, so I'll switch to `get`

src/aiida_quantumespresso/workflows/pw/base.py
bastonero: True, but here we are fixing a BFGS issue, so at least it should be a relax calculation

bastonero: Probably unnecessary anyways as it will be specified either way

src/aiida_quantumespresso/workflows/pw/base.py
bastonero: Fair enough ahah

bastonero: For the time being I just removed it. Probably there will be a specific `damp` related error that we will find out with time and add an error later on

src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -482,6 +483,53 @@ def handle_vcrelax_converged_except_final_scf(self, calculation):
...
+    def handle_relax_recoverable_ionic_convergence_bfgs_history_error(self, calculation):
+        """Handle failure of the ionic minimization algorithm (BFGS).
+
+        When BFGS history fails, this can mean two things: the structure is close to the global minimum,
+        but the moves the algorithm wants to do are smaller than `trust_radius_min`, or the structure is
+        close to a local minimum (hard to detect). For the first, we restart with lowered trust_radius_min.
+        For the first case, one can lower the trust radius; for the second one, one can exploit a different
+        algorithm, e.g. `damp` (and `damp-w` for vc-relax).
+        """
+        trust_radius_min = self.ctx.inputs.parameters['IONS'].get('trust_radius_min', qe_defaults.trust_radius_min)
```
mbercx, line 502: Don't hate me, but I think the `IONS` and `CELL` are _not_ set to empty dictionaries by default in the `setup` step of the outline. 😅 The `IONS` one is _required_ by Quantum ESPRESSO when `calculation == 'relax', 'md', 'vc-relax', or 'vc-md'`:

https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm923

`CELL` seems _optional_ when `calculation == 'vc-relax' or 'vc-md'`:

https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm1124

Could you add the `setdefault` lines to the `setup` step, but only in case the `calculation` is of the right type? I think this would make our lives easier and the code more readable since we can avoid the `setdefault` calls elsewhere in the code.

src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -482,6 +483,53 @@ def handle_vcrelax_converged_except_final_scf(self, calculation):
...
+            self.ctx.inputs.parameters['IONS']['ion_dynamics'] = 'damp'
+            action = 'bfgs history (ionic only) failure: restarting with `damp` dynamics.'
+
+        elif calculation_type == 'vc-relax' and ion_dynamics == 'bfgs' and trust_radius_min > 1.0e-4:
+            self.ctx.inputs.parameters['IONS']['trust_radius_ini'] = trust_radius_min  # start close
+            new_trust_radius_min = trust_radius_min * self.defaults.delta_factor_trast_radius_min
+            self.ctx.inputs.parameters['IONS']['trust_radius_min'] = new_trust_radius_min
+            action = f'bfgs history (vc-relax) failure: restarting with `trust_radius_min={new_trust_radius_min:.5f}`.'
+
+        elif calculation_type == 'vc-relax' and cell_dynamics == 'bfgs':
```
mbercx, line 519: Do we need ever need to look at the `cell_dynamics`, really? As I understand it, if you choose BFGS, you need to set both `ion_dynamics` and `cell_dynamics` to BFGS:

https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm942
https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm1127

So maybe we can just use `ion_dynamics` here and remove lines 506-507?

mbercx (changes requested):
Thanks @bastonero! I came up with some more comments. ^^

src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -33,6 +33,7 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         'delta_factor_max_seconds': 0.95,
         'delta_factor_nbnd': 0.05,
         'delta_minimum_nbnd': 4,
+        'delta_factor_trast_radius_min': 0.1,
```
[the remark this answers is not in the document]
bastonero: Annn the italian typos :D

src/aiida_quantumespresso/workflows/pw/base.py
bastonero: Sure!

src/aiida_quantumespresso/workflows/pw/base.py
bastonero: Good point, we can indeed just use `ion_dynamics`

bastonero: Following the same logic we can then skip entirely all the defaults, as it is related only to `bfgs` errors, hence we can drop the checks on `ion_dynamics`

mbercx (approved):
Thanks @bastonero! I think this one is good to go.

Issue #961, opened 2023-07-19
PwBaseCalculations number of electron_maxstep is fixed and repeat the same simulation


AndresOrtegaGuerrero:
When a calculation doesn't reach convergence within the number used in the workchain (80 steps) it does a similar restart.
Repeating the same parameters , this could be up to 4 times.
Perhaps at a second stage (restart) the workchain should just be kill to avoid repeating unsuccessful calculations.
On the other hand, should we consider if the system might need some extra steps? Should the restart try to include 50% more steps and if it doesn't success should it kill the job?

mbercx:
Thanks for raising the issue @AndresOrtegaGuerrero! Note that in case the electronic convergence fails, the `PwBaseWorkChain` reduces the `mixing_beta` :

https://github.com/aiidateam/aiida-quantumespresso/blob/5d4a7d9405b757e2ecf65a72c1ee92aa2fb36a39/src/aiida_quantumespresso/workflows/pw/base.py#L559-L576

However, we have already discussed in the past that:

1. The current factor for reducing the `mixing_beta` is too high:

https://github.com/aiidateam/aiida-quantumespresso/blob/5d4a7d9405b757e2ecf65a72c1ee92aa2fb36a39/src/aiida_quantumespresso/workflows/pw/base.py#L32

2. Perhaps it only makes sense to try this _once_.

Increasing the number of electronic steps `electron_maxstep` might also be sensible, but perhaps only in case we see that the SCF is converging?

AndresOrtegaGuerrero:
@mbercx I like the idea that it checks the output and it takes a decision upon the scf convergence and/change. But definitively to not do 4 times a restart that will end up in failure.

t-reents:
Hi together! @mbercx Just out of curiosity, have you also discussed to include `mixing_ndim` in the past? I included this parameter in my own workchains on top of the plugin. It will be also increased in steps of 4 (up to a given maximum) in case the `mixing_beta` is already quite small. I think that I remember some calculations where it seems that this actually helped. Of course, increasing `mixing_ndim` also increases the memory usage so this might be the reason why this is not included per default. If I remember correctly, increasing `mixing_ndim` and decreasing `mixing_beta` might increase the necessary steps to converge and for that reason I also increase `electron_maxstep` in case `mixing_ndim` is increased.

mbercx:
> But definitively to not do 4 times a restart that will end up in failure.

[...]

PR #986, opened 2023-11-23
`PhBaseWorkChain`: better handler for convergence

bastonero:
Plain lowering down the alpha mix could not work always. The `nmix_ph` is already an improvement, leading to the usage of more iterations into the mixing. The rate of the alpha mix change is also increased, and below a certain threshold a multi mix approach is tried. Afterwards, the handler finishes the strategies.

### DIFF
--- a/src/aiida_quantumespresso/workflows/ph/base.py
+++ b/src/aiida_quantumespresso/workflows/ph/base.py
@@ -24,8 +24,9 @@ class PhBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
 
     defaults = AttributeDict({
         'delta_factor_max_seconds': 0.95,
-        'delta_factor_alpha_mix': 0.90,
-        'alpha_mix': 0.70,
+        'delta_factor_alpha_mix': 0.5,
+        'nmix_ph': 4,
+        'alpha_mix': 0.7,
     })
 
     @classmethod
@@ -321,13 +322,33 @@ class PhBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
     @process_handler(priority=410, exit_codes=PhCalculation.exit_codes.ERROR_CONVERGENCE_NOT_REACHED)
     def handle_convergence_not_reached(self, node):
         """Handle `ERROR_CONVERGENCE_NOT_REACHED` exit code: decrease the mixing beta and restart."""
+        self.ctx.restart_calc = node
         factor = self.defaults.delta_factor_alpha_mix
+
+        nmix_ph = self.ctx.inputs.parameters.get('INPUTPH', {}).get('nmix_ph', self.defaults.nmix_ph)
+
+        if nmix_ph < 8:  # 8~20 is the recommended range on the Quantum ESPRESSO documentation
+            self.ctx.inputs.parameters.setdefault('INPUTPH', {})['nmix_ph'] = 8
+            action = f'increased number of mixing iteration from {nmix_ph} to 8 and restarting'
+            self.report_error_handled(node, action)
+
+            return ProcessHandlerReport(True)
+
+        # now try playing with the mixing parameters
         alpha_mix = self.ctx.inputs.parameters.get('INPUTPH', {}).get('alpha_mix(1)', self.defaults.alpha_mix)
         alpha_mix_new = alpha_mix * factor
 
-        self.ctx.restart_calc = node
-        self.ctx.inputs.parameters.setdefault('INPUTPH', {})['alpha_mix(1)'] = alpha_mix_new
+        if self.ctx.inputs.parameters.get('INPUTPH', {}).get('alpha_mix(20)', None) is not None:
+            action = 'no more efficient strategies for convergence.'
+            self.report_error_handled(node, action)
+            return ProcessHandlerReport(False)
 
-        action = f'reduced alpha_mix from {alpha_mix} to {alpha_mix_new} and restarting'
+        if alpha_mix_new >= 0.1:
+            self.ctx.inputs.parameters.setdefault('INPUTPH', {})['alpha_mix(1)'] = alpha_mix_new
+            action = f'reduced alpha_mix from {alpha_mix} to {alpha_mix_new} and restarting'
+        else:
+            self.ctx.inputs.parameters.setdefault('INPUTPH', {})['alpha_mix(20)'] = 0.4
+            action = f'introduing alpha_mix(20)={alpha_mix_new} and restarting'
+
         self.report_error_handled(node, action)
         return ProcessHandlerReport(True)
--- a/src/aiida_quantumespresso/workflows/protocols/ph/base.yaml
+++ b/src/aiida_quantumespresso/workflows/protocols/ph/base.yaml
@@ -12,6 +12,8 @@ default_inputs:
                 withmpi: True
         parameters:
             INPUTPH:
+                alpha_mix: 0.4
+                nmix_ph: 8
                 tr2_ph: 1.0e-18
 
 default_protocol: balanced
--- a/tests/workflows/ph/test_base.py
+++ b/tests/workflows/ph/test_base.py
@@ -122,15 +122,28 @@ def test_handle_convergence_not_reached(generate_workchain_ph):
     process.setup()
     process.validate_parameters()
 
+    result = process.handle_convergence_not_reached(process.ctx.children[-1])
+    assert isinstance(result, ProcessHandlerReport)
+    assert result.do_break
+    assert process.ctx.inputs.parameters['INPUTPH']['nmix_ph'] == 8
+
     alpha_new = PhBaseWorkChain.defaults.alpha_mix * PhBaseWorkChain.defaults.delta_factor_alpha_mix
+
+    for _ in range(2):  # this is dependent on PhBaseWorkChain.defaults.delta_factor_alpha_mix
+        result = process.handle_convergence_not_reached(process.ctx.children[-1])
+        assert isinstance(result, ProcessHandlerReport)
+        assert result.do_break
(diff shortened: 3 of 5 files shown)


src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -257,13 +258,39 @@ def handle_out_of_walltime(self, node):
...
+                nmix_ph_new = 20
+
+            self.ctx.inputs.parameters.setdefault('INPUTPH', {})['nmix_ph'] = nmix_ph_new
+
+            action = f'increased number of mixing iteration from {nmix_ph} to {nmix_ph_new} and restarting'
+            self.report_error_handled(node, action)
+
+            return ProcessHandlerReport(True)
+
+        # now try playing with the mixing parameters
```
mbercx, line 279: In my runs I had immediately set `alpha_mix` to 0.4 in the `overrides`. I didn't gather any statistics, but I did get a noticable improvement in convergence. Something to consider?

src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -22,8 +22,9 @@ class PhBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
 
     defaults = AttributeDict({
         'delta_factor_max_seconds': 0.95,
-        'delta_factor_alpha_mix': 0.90,
-        'alpha_mix': 0.70,
+        'delta_factor_alpha_mix': 0.65,
```
mbercx, line 25: Just out of curiosity: how did you decide on this value?

I recently tested reducing the `delta_factor_mixing_beta` for the `PwBaseWorkChain` on a set of failed SCF runs and didn't really help much (surprisingly, to be honest). There we do start from `mixing_beta` 0.4 (also see my other comment below).

src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -257,13 +258,39 @@ def handle_out_of_walltime(self, node):
...
+
+            return ProcessHandlerReport(True)
+
+        # now try playing with the mixing parameters
         alpha_mix = self.ctx.inputs.parameters.get('INPUTPH', {}).get('alpha_mix(1)', self.defaults.alpha_mix)
         alpha_mix_new = alpha_mix * factor
 
-        self.ctx.restart_calc = node
-        self.ctx.inputs.parameters.setdefault('INPUTPH', {})['alpha_mix(1)'] = alpha_mix_new
+        if self.ctx.inputs.parameters.get('INPUTPH', {}).get('alpha_mix(20)', -1) != -1:
```
mbercx, line 283: Nitpick: I would just use `None` here.

```suggestion
        if self.ctx.inputs.parameters.get('INPUTPH', {}).get('alpha_mix(20)', None) is not None:
```

src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -257,13 +258,39 @@ def handle_out_of_walltime(self, node):
...
+            action = 'no more efficient strategies for convergence.'
+            self.report_error_handled(node, action)
+            return ProcessHandlerReport(False)
+
+        if alpha_mix_new >= 0.1:
+            self.ctx.inputs.parameters.setdefault('INPUTPH', {})['alpha_mix(1)'] = alpha_mix_new
+            action = f'reduced alpha_mix from {alpha_mix} to {alpha_mix_new} and restarting'
+        else:
+            self.ctx.inputs.parameters.setdefault('INPUTPH', {})['alpha_mix(20)'] = 0.4
```
mbercx, line 292: Setting `alpha_mix(20)` to some value means that _from iteration 20_ the mixing factor will be adapted, right?

If in this case `alpha_mix(1)` is already e.g 0.1, it would help to suddenly increase the mixing at iteration 20?

src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -257,13 +258,39 @@ def handle_out_of_walltime(self, node):
...
+            action = f'introduing alpha_mix(20)={alpha_mix_new} and restarting'
```
mbercx, line 293: If the strategy above is indeed the desired one. ^^

```suggestion
            action = f'introducing alpha_mix(20)=0.4 and restarting'
```

src/aiida_quantumespresso/workflows/ph/base.py
mbercx: One additional note on the line below this one (which was already there ^^): The user can also set `alpha_mix` instead of `alpha_mix(1)` and this would work fine. If we only look for `alpha_mix(1)` and then set it, I think we override the `alpha_mix` specified by the user.

mbercx (changes requested):
Thanks @bastonero! Just left a few questions/comments.

src/aiida_quantumespresso/workflows/ph/base.py
bastonero: Yes, probably better to start from `alpha_mix(1) = 0.4` and work they way down. The delta factor I chose it in a way to go pretty fast below 0.1, so one can explore different mixings in the standard 5 max iterations.
Do you have any better idea?

src/aiida_quantumespresso/workflows/ph/base.py
bastonero: Yes, exactly. I remember seeing this being suggested in the QE forum for some tricky cases. The idea is to do have low mixing at the beginning, and then at the 20-th iteration one expect to have the mixed solution quite sabilized and then one can mix more the self-consistent solution (IDK if I explained myself).

src/aiida_quantumespresso/workflows/ph/base.py
bastonero: Absolutely! I left because I wanted to ask: can one make the handler to take some further _user input_? E.g. say th e user can adjust this logic to e.g. change the mix at some iterations `niterations` to some `alpha_mixes`? Could it be done? Or does it make sense?

mbercx: I think we can just make a top-level `Dict` input for `BaseRestartWorkChain`s where we add such inputs. I was already thinking the same for e.g. `delta_factor_mixing_beta` for the `PwBaseWorkChain`. But let's do that in another PR!

src/aiida_quantumespresso/workflows/ph/base.py
mbercx: Yes, I see how this makes sense. Curious to see it in action. ^^

I suppose we should gather some statistics similar as in https://github.com/aiidateam/aiida-quantumespresso/issues/961. I have a good number of `ph.x` runs, but would still have to parse the convergence info. Let's adapt the strategy for now and maybe revisit!

src/aiida_quantumespresso/workflows/ph/base.py
mbercx: Nothing based on hard facts. In https://github.com/aiidateam/aiida-quantumespresso/pull/987 my thinking was now to stick to three options by default: 0.4 -> 0.2 -> 0.1. So basically make the delta factor 0.5. If one of those three doesn't work, all hope is probably lost for this strategy?

src/aiida_quantumespresso/workflows/ph/base.py
```diff
@@ -22,8 +22,9 @@ class PhBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
 
     defaults = AttributeDict({
         'delta_factor_max_seconds': 0.95,
-        'delta_factor_alpha_mix': 0.90,
-        'alpha_mix': 0.70,
+        'delta_factor_alpha_mix': 0.5,
+        'nmix_ph': 4,
+        'alpha_mix': 0.4,
```
mbercx, line 27: I think we want to set this as the protocol default, no? So we should add this to the `protocols/ph/base.yaml` file instead.

mbercx (changes requested):

src/aiida_quantumespresso/workflows/ph/base.py
mbercx: I've always just used `alpha_mix`:

```
mabercx@uan04:/pfs/lustrep3/scratch/project_465000106/mabercx/88/a0/a0e0-3391-4e6b-b4da-fc98886eaa4a> grep alpha_mix aiida.in
  alpha_mix =   4.0000000000d-01
mabercx@uan04:/pfs/lustrep3/scratch/project_465000106/mabercx/88/a0/a0e0-3391-4e6b-b4da-fc98886eaa4a> grep alpha_mix aiida.out | head -n 5
      thresh= 1.000E-02 alpha_mix =  0.400 |ddv_scf|^2 =  1.410E-08
      thresh= 1.187E-05 alpha_mix =  0.400 |ddv_scf|^2 =  3.525E-09
      thresh= 5.937E-06 alpha_mix =  0.400 |ddv_scf|^2 =  7.637E-11
      thresh= 8.739E-07 alpha_mix =  0.400 |ddv_scf|^2 =  2.221E-13
      thresh= 4.713E-08 alpha_mix =  0.400 |ddv_scf|^2 =  3.612E-14
```

src/aiida_quantumespresso/workflows/ph/base.py
mbercx: You didn't like my nitpick? 😢

mbercx:
@bastonero integrated my nitpick. ^^ I'm happy to merge this now, unless you still have comments/updates since we started this PR 1.5 years go 🙈

mbercx:
Note: RTD was having trouble finding the GitHub references when cloning the repository and checking out the branch, seems closing and reopening the PR fixes this issue.

Issue #988, opened 2023-11-29
PdosWorkChain except when tot_magnetization is used


AndresOrtegaGuerrero:
The PdosWorkChain except when dealing with a calculation were `tot_magnetization` is considered.
The issue is from this line `self.ctx.nscf_fermi = workchain.outputs.output_parameters.dict.fermi_energy`

When using tot_magnetization there are two energies `fermi_energy_up` and `fermi_energy_down` instead of fermi_energy. 

This affects the rest of the logic of the workchain that only considers fermi_energy

PR #992, opened 2023-12-12
Fix Except PdosWorkChain when tot_magnetization is used

AndresOrtegaGuerrero:
This PR address #988

### DIFF
--- a/src/aiida_quantumespresso/workflows/pdos.py
+++ b/src/aiida_quantumespresso/workflows/pdos.py
@@ -458,7 +458,13 @@ class PdosWorkChain(ProtocolMixin, WorkChain):
         self.ctx.nscf_emin = workchain.outputs.output_band.get_array('bands').min()
         self.ctx.nscf_emax = workchain.outputs.output_band.get_array('bands').max()
         self.ctx.nscf_parent_folder = workchain.outputs.remote_folder
-        self.ctx.nscf_fermi = workchain.outputs.output_parameters.dict.fermi_energy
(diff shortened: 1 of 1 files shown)


src/aiida_quantumespresso/workflows/pdos.py
```diff
@@ -458,7 +458,13 @@ def inspect_nscf(self):
         self.ctx.nscf_emin = workchain.outputs.output_band.get_array('bands').min()
         self.ctx.nscf_emax = workchain.outputs.output_band.get_array('bands').max()
         self.ctx.nscf_parent_folder = workchain.outputs.remote_folder
-        self.ctx.nscf_fermi = workchain.outputs.output_parameters.dict.fermi_energy
+        if 'fermi_energy' in workchain.outputs.output_parameters.dict:
+            self.ctx.nscf_fermi = workchain.outputs.output_parameters.dict.fermi_energy
+        else:
+            self.ctx.nscf_fermi_energy_up = workchain.outputs.output_parameters.dict.fermi_energy_up
```
mbercx, line 464: Thanks @AndresOrtegaGuerrero! Quick question here: do we need to keep `nscf_fermi_energy_up` and `nscf_fermi_energy_down` in the context if we determine the Fermi level from the highest value in line 466?

AndresOrtegaGuerrero: Yes i will update this !

AndresOrtegaGuerrero:
I did another PR #993 so is in my personal account

PR #993, opened 2023-12-12
Fix/988/pdosworkchain

### LINKED ISSUE #988: PdosWorkChain except when tot_magnetization is used

AndresOrtegaGuerrero:
Fixes #988

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# pylint: disable=redefined-outer-name,too-many-statements
+# pylint: disable=redefined-outer-name,too-many-statements,unsubscriptable-object
 """Initialise a text database and profile for pytest."""
 from collections.abc import Mapping
 import io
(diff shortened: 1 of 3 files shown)


mbercx (approved):
Thanks @AndresOrtegaGuerrero!

tests/conftest.py
```diff
@@ -51,14 +51,17 @@ def fixture_code(fixture_localhost):
...
 
+        query = QueryBuilder().append(
+            InstalledCode,
+            filters={'label': label},
+        )
         try:
-            return load_code(label=label)
-        except exceptions.NotExistent:
+            return query.first()[0]
+        except TypeError:
```
sphuber, line 64: Why was this changed? @mbercx I don't understand this nor do I think it is correct. I would revert this

mbercx: I'm not 100% what was happening under the `pytest` hood, but when using `pytest.mark.parametrize` to test the two possible NSCF output parameters, it seems to somehow have created two instances of the `quantumespresso.pw` code, and the second test fails with:

```
aiida.common.exceptions.MultipleObjectsError: multiple Code entries found with LABEL<test.quantumespresso.pw>
```

My solution is a bit of a late-night hack, I'll admit. ^^ 

Looking into this further, it actually seems that VSCode running the test in parallel causes the issue. After reverting reverting the change [here](https://github.com/mbercx/aiida-quantumespresso/commit/b23e5eca84cd2a20b63c651d05955debd90f4063), the tests still pass fine. Running `pytest` manually locally also works without issue.

Happy to open a PR to revert, but maybe there is also a more elegant solution to have the tests run properly in parallel?

sphuber: Well, the tests are not designed to be run in parallel, especially not if they use the same storage backend. Even if you "fix" this case, there is bound to be a shitload of other problems lying in wait when you have multiple tests run against the same database. So I really think you just shouldn't run them in parallel.

mbercx: Opened https://github.com/aiidateam/aiida-quantumespresso/pull/995 to revert.

PR #994, opened 2023-12-12
Bump default SSSP version in protocol to v1.3

unkcpz:
This is required by https://github.com/aiidalab/aiidalab-qe/pull/578. It is not that easy to decouple reading protocol from `aiida-quatumespresso` in QeApp, if the override is not provided, it will always fallback to the default pseudo family which may not aligned between the qe plugin and the qe app. 

It would be great after this change, a new patch release can be made. Or it has to be a minor version bump for `aiida-quantumespresso`?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,7 +131,7 @@ jobs:
             run: verdi code create core.code.installed -n --config .github/config/code-ph.yaml --filepath-executable $(which ph.x)
 
         -   name: Setup SSSP
-            run: aiida-pseudo install sssp -v 1.2 -x PBEsol -p efficiency
+            run: aiida-pseudo install sssp -v 1.3 -x PBEsol -p efficiency
 
         -   name: Run test script
             run: python .github/scripts/run_documentation_scripts.py
--- a/docs/source/howto/calculations/pw.md
+++ b/docs/source/howto/calculations/pw.md
@@ -207,7 +207,7 @@ from ase.build import bulk
 from aiida.orm import StructureData, load_code, load_group
 
 # Load the pseudopotential family whose pseudos to use
-family = load_group('SSSP/1.1/PBE/effiency')
+family = load_group('SSSP/1.3/PBEsol/effiency')
 structure = StructureData(ase=bulk('GaAs', 'fcc', 5.4))
 
 builder = load_code('pw').get_builder()
@@ -226,7 +226,7 @@ from ase.build import bulk
 from aiida.orm import StructureData, load_code, load_group
 
 # Load the pseudopotential family whose pseudos to use
-family = load_group('SSSP/1.1/PBE/effiency')
+family = load_group('SSSP/1.3/PBEsol/effiency')
 structure = StructureData(ase=bulk('GaAs', 'fcc', 5.4))
 
 builder = load_code('pw').get_builder()
--- a/docs/source/installation/index.md
+++ b/docs/source/installation/index.md
@@ -212,10 +212,10 @@ $ pip install aiida-pseudo
 ```
 
 At a minimum, at least one pseudo potential family should be installed.
-We recommend using the [SSSP] with the PBEsol functional:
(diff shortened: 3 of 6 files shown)


mbercx:
Thanks @unkcpz! Also updated the documentation where I found references to the SSSP pseudos.

A patch release is a bit tedious since we have everything on `main`. But doing a minor release isn't so much work. How urgently do you need it? ^^

unkcpz:
Thanks for such a quick update! 

> How urgently do you need it?

Can I say very very urgent? 😉 My mistake was that I somehow missed this requirement for qeapp and it is what our big boss ask to include for the qeapp before release exactly.

mbercx:
> Can I say very very urgent? 😉

I'll start prepping a release branch then ^^

mbercx (approved):

mbercx:
Interesting. Re-requesting a review does not dismiss my old one. Was curious what would happen there. ^^

Anyways, let me also check the `aiida-pseudo` version where SSSP v1.3 was added, we should probably bump our dependency in case it's below that one.

unkcpz:
> let me also check the aiida-pseudo version where SSSP v1.3 was added, we should probably bump our dependency in case it's below that one.

I think `aiida-pseudo~=1.1` should work. I find it from changelog https://github.com/aiidateam/aiida-pseudo/blob/a4face268b9a911caacecedbcbc605db389dc68e/CHANGELOG.md?plain=1#L31

mbercx:
Seems [SSSP v1.3 support was only added in `aiida-pseudo==1.1.0`](https://github.com/aiidateam/aiida-pseudo/blob/main/CHANGELOG.md#110---2023-06-20).

mbercx:
> > let me also check the aiida-pseudo version where SSSP v1.3 was added, we should probably bump our dependency in case it's below that one.
> 
> I think `aiida-pseudo~=1.1` should work. I find it from changelog https://github.com/aiidateam/aiida-pseudo/blob/a4face268b9a911caacecedbcbc605db389dc68e/CHANGELOG.md?plain=1#L31

Hehe, beat me to it. ;)

mbercx (approved):
Now I only pray that I haven't missed something and @sphuber hits me with the spanking gifs in the morning. ^^

unkcpz:
Test locally on qeapp with these changes and works as expected.

sphuber:
> Now I only pray that I haven't missed something and @sphuber hits me with the spanking gifs in the morning. ^^

Nah, I already spent all my creative energy [on another post today](https://github.com/aiidateam/aiida-core/pull/6224#issuecomment-1853611007).

Changes look fine to me. It is just that people with existing profiles will now probably face an exception when updating. I think the `PwBaseWorkChain.get_builder_from_protocol` will now require the v1.3 PBEsol SSSP to be installed. The error will just provide a very generic message `Please use `aiida-pseudo install` to install it.` But users won't automatically now what the exact command is. We should improve this by either:
1) improve the error message to be explicit and give the full exact command
2) allow any installed SSSP version as a fallback and issue a warning if that is used

I think for most cases option 2) would probably be most user-friendly. Just for users that actually want a very specific version and forgot to install it (or mistyped it in the protocol) could be surprises, but not sure how common this will be.

unkcpz:
> I think the PwBaseWorkChain.get_builder_from_protocol will now require the v1.3 PBEsol SSSP to be installed. 

@sphuber True! this is the problem I want to avoid in QeApp by this change here, because in QeApp we install the SSSP and dojo by default, after bumping the version the `get_builder_from_protocol` all fails with not able to find the pseudo-family. 

But I think option 1 might be better. The user anyway needs to set up pseudo libraries the first time why bother to do it again, it also pushes users to use the new libraries as default. What do you think? @mbercx

mbercx:
I'm also leaning towards (1). Seems more straightforward to motivate the user to just install the new pseudos. But then it'd nice to have a more general command that converts a certain configuration into the corresponding CLI command to install it. And I suppose this would be more at home in the `aiida-pseudo` package. ^^

PR #995, opened 2023-12-13
Tests: Revert change to `fixture_code`

mbercx:
In commit a68e1e15c11f6ad4461921145c648d75c49ff26c the `fixture_code` fixture was adapted to query for the first `InstalledCode` with the test code label instead of using `load_code`, as the latter would fail when running the new parametrized PDOS tests in parallel.

However, running tests in parallel versus the same storage backend can potentially lead to all sorts of problems. So instead of providing a somewhat hacky fix for this case, we revert the changes made to `fixture_code` and avoid running the tests in parallel.


sphuber (approved):
Thanks @mbercx

PR #997, opened 2023-12-15
Protocols: Update k-point sampling protocols

mbercx:
Based on more extensive tests that also involved a broad range of smearings, a new set of protocols were devised:

| Name       | Smearing (Ry) | k-points distance (1/A) |
|------------|---------------|-------------------------|
| Precision  |        0.0125 |                    0.10 |
| Efficiency |        0.0200 |                    0.15 |
| Fast       |        0.0275 |                    0.30 |

Here we already update the values in the `PwBaseWorkChain` protocol file. In a future PR it would be nice to set up a system for protocol aliasing, so we can also change the names without breaking protocol methods downstream.

### DIFF
--- a/src/aiida_quantumespresso/workflows/protocols/pdos.yaml
+++ b/src/aiida_quantumespresso/workflows/protocols/pdos.yaml
@@ -35,11 +35,11 @@ default_inputs:
                     num_machines: 1
                 max_wallclock_seconds: 43200  # Twelve hours
                 withmpi: True
-default_protocol: moderate
+default_protocol: balanced
 protocols:
-    moderate:
+    balanced:
         description: 'Protocol to perform a projected density of states calculation at normal precision at moderate computational cost.'
-    precise:
+    stringent:
         description: 'Protocol to perform a projected density of states structure calculation at high precision at higher computational cost.'
         dos:
             parameters:
--- a/src/aiida_quantumespresso/workflows/protocols/ph/base.yaml
+++ b/src/aiida_quantumespresso/workflows/protocols/ph/base.yaml
@@ -14,11 +14,11 @@ default_inputs:
             INPUTPH:
                 tr2_ph: 1.0e-18
 
-default_protocol: moderate
+default_protocol: balanced
 protocols:
-    moderate:
+    balanced:
         description: 'Protocol to perform the computation at normal precision at moderate computational cost.'
-    precise:
+    stringent:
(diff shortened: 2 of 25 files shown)


demiranda-gabriel (approved):
Modifications in the standard protocols are in agreement with our latest extensive benchmark.

demiranda-gabriel (dismissed):

demiranda-gabriel:
Thank you @Marnik! I think your aliasing approach is super clean! We don’t expect to keep changing the names of the base protocols that much 😅 To be honest, if I want to run things with a custom protocol, I don’t have too much of a problem myself in keeping an extra YAML file that I load as a dict and pass as overrides. However, if we want to invest in user-defined protocols, do you have an idea what the simplest approach would be? Maybe passing a path to a YAML file with the custom protocol as an argument to `get_protocol_inputs()`? Perhaps we could still rely on the default protocols in case the user-defined one is missing any important parameters.

Just an important comment: from the last meetings we had with Nicola and Giovanni, it had been decided to call the protocols “stringent, balanced, and fast” instead of “precision, efficiency, and fast”, to not get it confused with the pseudo families. Sorry about that, I should have updated this issue when the decision was made, my fault. 

I see that you’re merging this PR from your own fork. Do you know if I would be able to make commits to this branch and have it automatically updated in this PR? So I can rename everything myself.

Screenshot 2025-03-26 at 10 45 19 AM

mbercx:
>  To be honest, if I want to run things with a custom protocol, I don’t have too much of a problem myself in keeping an extra YAML file that I load as a dict and pass as overrides. However, if we want to invest in user-defined protocols, do you have an idea what the simplest approach would be? Maybe passing a path to a YAML file with the custom protocol as an argument to get_protocol_inputs()? Perhaps we could still rely on the default protocols in case the user-defined one is missing any important parameters.

Yeah, using the `overrides` is also my preferred approach. I think it would make sense that you can pass a YAML file to the `protocol` input argument. Some initial thoughts:

1. This file should be able to define the complete protocol of a work chain, so it should pass the inputs of the sub-processes down to their `get_builder_from_protocol` methods.
2. This is currently done through the `overrides`. But then you still keep inputs from the original protocols and you can't remove them, see https://github.com/aiidateam/aiida-quantumespresso/issues/653.

Let's keep this discussion for another issue though. Thanks for updating me about the new protocol names, I'll integrate those and merge the PR!

PR #998, opened 2023-12-16
adapting BandsWorkChain for HubbardStructureData

AndresOrtegaGuerrero:
When using HubbardStructureData with PwBandsStructure the workchain doesnt use the hubbard information. 
Since it calls seekpath , the information is lost. 

Additionally the HubbardStructureData has a bug for StructureData with periodicity 1d and 2d. 

This PR addresses these two issues , #999


bastonero:
Thanks @AndresOrtegaGuerrero. I am wondering if then all the methods are working as expected. Could you please provide some tests and/or examples, with some structure? I think there should be already some fixtures for 2D. It is important that the functions work, and in the intended behaviour. Here, it is not clear if your fix is going to work out.

AndresOrtegaGuerrero:
Hi @bastonero , The 2D and 1D is already working at the PwBaseWorkChain (like the sampling of k-points) we did these with @mbercx. Regarding the PwBandsWorkChain, this branch manage to fix the. Another test is the PdosWorkChain that works without no problems with HubbardStructureData.  Is also tested in the AiiDAlab qe app (https://github.com/aiidalab/aiidalab-qe/pull/577)

AndresOrtegaGuerrero:
I think you meant to add test for the hubbard structure, and for bands workchain 😅

src/aiida_quantumespresso/data/hubbard_structure.py
```diff
@@ -109,7 +109,16 @@ def append_hubbard_parameter(
         :param hubbard_type: hubbard type (U, V, J, ...), defaults to 'Ueff'
             (see :class:`~aiida_quantumespresso.common.hubbard.Hubbard` for full allowed values)
         """
-        pymat = self.get_pymatgen_structure()
+        if self.pbc != (True, True, True):
+            #create a temporary StructureData to obtain periodic sites.
+            ase = self.get_ase()
+            ase.set_pbc((True, True, True))
+            temp_structure = StructureData(ase=ase)
```
unkcpz, line 116: I think it should be able to create a temp structure directly with StructureData (https://github.com/aiidateam/aiida-core/blob/06ea130df8854f621e25853af6ac723c37397ed0/src/aiida/orm/nodes/data/structure.py#L1592) instead of through ase structure. 

What will happened if the `pbc` is not 3D? It will raise an error or it will give the wrong results afterwards?

src/aiida_quantumespresso/workflows/pw/bands.py
```diff
@@ -209,14 +210,24 @@ def run_seekpath(self):
...
             }
         }
         result = seekpath_structure_analysis(self.ctx.current_structure, **inputs)
-        self.ctx.current_structure = result['primitive_structure']
+
+        # If the input structure was a HubbardStructureData
+        if isinstance(self.ctx.current_structure, HubbardStructureData):
+            self.ctx.hubbard_info = self.ctx.current_structure.hubbard
+            hubbard_structure = HubbardStructureData.from_structure(result['primitive_structure'])
+            hubbard_structure.hubbard = self.ctx.hubbard_info
```
unkcpz, line 226: I am a bit worried here, the `habbard_info` is bind with the original structure. But the `current_structure` will then be a primitive structure. 

I think it will be more proper that there will be a `get_primitive_structure` method for `HubbardStructureData` and `hubbard_info` is always attached with the corresponded structure.

unkcpz (changes requested):
@AndresOrtegaGuerrero thanks! I suggest to add a `get_primitive_structure` to `HubbardStructureData` and call it to extract the bind information from a single `HubbardStructureData` object. See the comment below. 

I am not super familiar what `habbard_info` contains, maybe it doesn't relate to whether the structure is primitive or not. So the change may not needed.

src/aiida_quantumespresso/workflows/pw/bands.py
bastonero: Just passing by: be very aware that getting a primitive structure would mean to also transfer the hubbard parameters CORRECTLY, e.g. in case of extended Hubbard parameters. This is still not implemented in e.g. `HubbardUtils`, for which there is a method that is only tested for supercells. And indeed, this way would most certainly set WRONG hubbard parameters for the primitive structure, where the order of atoms might change, as well as the number, and considering different lattice parameters this is a recepy for disaster. As such, is it REALLY necessary to have a primite structure? If not, then it would be wiser to just skip this if uncertain on what it is doing.

src/aiida_quantumespresso/data/hubbard_structure.py
AndresOrtegaGuerrero: if is different that 3D , it will try to use pymatgen, and then it will fail

src/aiida_quantumespresso/workflows/pw/bands.py
AndresOrtegaGuerrero: I wouldnt change the structure, since this function is just to pass the hubbard parameters, instead i proposed that 1D and 2D , are treated as 3D , just to pass the parameters to the HubbardStructureData,

src/aiida_quantumespresso/data/hubbard_structure.py
AndresOrtegaGuerrero: @unkcpz i think i cant use set_pbc since , is an StructureData that is saved

src/aiida_quantumespresso/workflows/pw/bands.py
AndresOrtegaGuerrero: @unkcpz Indeed when using seekpath and generating a `primitive_structure `, we CAN'T use hubbard_info. Is also quite tricky to generalize it since the HubbardStructureData can have hubbard_parameters for DFT+U, DFT+U+V. It seems it need a method to generalize the manifold to apply it on a` primitive_structure`

unkcpz: Then can you make it except when the scenario happened? If I understand correctly, with your implementation, if I run the band on the supercell with hubbard turned on, it will raise in some place that the calculation will fail. Can you try to add a validation and notify user about it.

AndresOrtegaGuerrero: @unkcpz I added a logic that generalize the hubbard_info , to be transformed to the `primitive_structure`. @bastonero could you give me your opinion , I will try to make a test with a conv unit cell , asap

src/aiida_quantumespresso/data/hubbard_structure.py
AndresOrtegaGuerrero: you mean like this `temp_structure = StructureData(ase=ase, pbc=[True, True, True])` ?

unkcpz: No I mean using as less structure type as possible. 
With your implementation, the structure type used are `HubbardStructureData` -> `asc structure` ->  `StructureData` -> pymatgen structure. However, the purpose here is to get the `sites` which are information that should include in the StructureData. 

I think we should try our best to not use ase/pymatgen, at least should try to not add more dependent on those. I want to see @mbercx's opinion on it.

mbercx: Thanks for the ping @unkcpz. 

I see your point that going through ASE to get a `StructureData` again to then get a `pymatgen` structure seems very roundabout, but I understand why @AndresOrtegaGuerrero did it. 

The issue is that `get_pymatgen()` would now work since it would return a pymatgen `Molecule`, and `get_pymatgen_structure()` fails since that only allows periodic boundary conditions... Ideally this _would_ actually be possible. I don't immediately understand why `StructureData` doesn't allow it, since a pymatgen `Structure` can also e.g. have `(True, False, False)` periodic boundary conditions. I'll open an issue on `aiida-core` to discuss this, but that'll take time.

Unfortunately, it's also not so easy to get an (unstored) copy of a `StructureData` directly. We could avoid using ASE in another way though. In the end the only thing the method needs is the sites, so the following should work for structures of any dimension:

```python
from pymatgen.core import PeriodicSite, Lattice

sites = [
    PeriodicSite(
        species=site.species,
        coords=site.coords,
        lattice=Lattice(self.cell),
        coords_are_cartesian=True
    ) for site in self.get_pymatgen().sites
]
```

And is probably cleaner than the if/else clause with the `StructureData` -> ASE -> `StructureData` -> `Structure` -> sites route for <3D structures.

As a final note, I know @unkcpz wants to avoid using ASE/pymatgen since these can be somewhat problematic dependencies (ASE because there are barely any releases, pymatgen since they often make breaking changes). But in the end I don't see a way to avoid using these tools, nor do I think it is desirable. That would involve reimplementing (and maintaining!) a lot of functionality ourselves.

src/aiida_quantumespresso/workflows/pw/bands.py
```diff
@@ -209,14 +210,46 @@ def run_seekpath(self):
...
         inputs = {
             'reference_distance': self.inputs.get('bands_kpoints_distance', None),
             'metadata': {
                 'call_link_label': 'seekpath'
             }
         }
         result = seekpath_structure_analysis(self.ctx.current_structure, **inputs)
-        self.ctx.current_structure = result['primitive_structure']
+
+        # If the input structure was a HubbardStructureData
```
mbercx, line 222: I haven't check the logic here, but instead of adding this to the `PwBandsWorkChain`, wouldn't it make more sense to move this into the `seekpath_structure_analysis`?

tests/data/test_hubbard_structure.py
```diff
@@ -61,6 +61,19 @@ def test_from_structure(generate_structure, generate_hubbard):
     assert len(hubbard_structure.kinds) == 2
 
 
+@pytest.mark.usefixtures('aiida_profile')
+@pytest.mark.parametrize('structure_name', ['2D-xy-arsenic', '1D-x-carbon', '1D-y-carbon', '1D-z-carbon'])
```
mbercx, line 65: This test doesn't seem to test the `append_hubbard_parameter` method, which is the one you are actually adapting?

mbercx:
Thanks @AndresOrtegaGuerrero! One more general note here is that you are making two quite different changes in one PR:

1. Provide support for low-dimensional structures to `HubbardStructureData`.
2. Provide support for `HubbardStructureData` to the `PwBandsWorkChain`.

No need to open another PR, but I'd split these changes up into two commits that describe each. Happy to do this once the PR is happy to merge, or discuss how to do this over Zoom.

src/aiida_quantumespresso/workflows/pw/bands.py
AndresOrtegaGuerrero: I tried the logic, it worked , should i move it to `seekpath_structure_analysis` calcfunction? (I am ok with :D)

AndresOrtegaGuerrero:
> Thanks @AndresOrtegaGuerrero! One more general note here is that you are making two quite different changes in one PR:
> 
> 1. Provide support for low-dimensional structures to `HubbardStructureData`.
> 2. Provide support for `HubbardStructureData` to the `PwBandsWorkChain`.
> 
> No need to open another PR, but I'd split these changes up into two commits that describe each. Happy to do this once the PR is happy to merge, or discuss how to do this over Zoom.

Thank you @mbercx sorry about the two changes, in my head they were quite small that since it was an aiidalab_qe issue i thought it was ok in one. It wont happen again :D

bastonero:
The pbc should be fixed by this PR https://github.com/aiidateam/aiida-core/pull/6281#issue-2125449409

AndresOrtegaGuerrero:
@bastonero , @mbercx  proposed a solution here without the need of using generate_structure

mbercx (changes requested):
Thanks @AndresOrtegaGuerrero. As @bastonero already mentioned, I think for intersite Hubbard parameters it's not so trivial to correctly preserve the parameters for the primitive structure provided by SeeK-path (see comments).

To still allow the user to find the standardized path in the intersite V case, we should see if we can't _fully_ preserve the structure and have SeeK-path return the corresponding path. You'd have to see if this is possible, maybe touch base with @giovannipizzi on the topic.

tests/calculations/functions/test_seekpath_analysis.py
```diff
@@ -0,0 +1,89 @@
...
+    structure = generate_hubbardstructure_conv()
+    conventional_parameters = structure.hubbard.parameters
+
+    result = seekpath_structure_analysis(structure)
+    primitive_parameters = result['primitive_structure'].hubbard.parameters
+
+    assert isinstance(
+        result['primitive_structure'], HubbardStructureData
+    ), 'Primitive structure should be a HubbardStructureData'
+    assert conventional_parameters != primitive_parameters, 'Primitive parameters should be different'
```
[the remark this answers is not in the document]
AndresOrtegaGuerrero: I guess we can check the logic, if the structure is different from the primitive seekpath is returning ?

src/aiida_quantumespresso/calculations/functions/seekpath_structure_analysis.py
```diff
@@ -28,4 +30,47 @@ def seekpath_structure_analysis(structure, **kwargs):
...
+    result = get_explicit_kpoints_path(structure, **unwrapped_kwargs)
+
+    # If the input structure was a HubbardStructureData, update the primitive structure with Hubbard parameters
+    if isinstance(structure, HubbardStructureData):
+        update_structure_with_hubbard(structure, result)
+
+    return result
+
+
+def update_structure_with_hubbard(structure, result):
```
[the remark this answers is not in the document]
AndresOrtegaGuerrero: But i am a lil puzzle from this example you give, you have different kinds to differentiate ,` V	Ni-3d	N-2p	2	45	1.1805 `, and `V	Ni-3d	N-2p	1	3	1.034` ??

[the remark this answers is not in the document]
AndresOrtegaGuerrero: I guess for V, we will need to check distancee between index of atoms , and them compare and find the distance of the Hubbard(inital) to find which atoms in the primitive have the same distance to assign the new  V, and in this case we shouldnt use initialize_intersites_hubbard, since the Hubbard you provide in your example considers a different V depending on the distance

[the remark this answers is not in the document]
mbercx: > I guess for V, we will need to check distancee between index of atoms , and them compare and find the distance of the Hubbard(inital) to find which atoms in the primitive have the same distance to assign the new V

Would the distance between sites be sufficient? Can't there be two pairs of atoms that have the same distance but a different intersite V? In that case we'd have to check that the primitivization respects each V pair, not just within the unit cell using the atom index, but also the neighbors via the translations. 

>  and in this case we shouldnt use initialize_intersites_hubbard, since the Hubbard you provide in your example considers a different V depending on the distance

Indeed, `append_hubbard_parameter` would allow you to specify the atom/site indices and translation.

mbercx:
@AndresOrtegaGuerrero @bastonero I'll move the seekpath-related changes to a different PR so we can focus on the <3D structures here.

I've added some validation on the site indices for `append_hubbard_parameter` in [f8d6933](https://github.com/aiidateam/aiida-quantumespresso/pull/998/commits/f8d6933f61dad601a5bde0accf6e6b4ae9603296).

I've also extended the tests of the method to check more cases and verify that adapting it for <3D structures doesn't break anything. ^^

tests/data/test_hubbard_structure/test_append_hubbard_parameters_parameters0_2D_xy_arsenic_.yml
```diff
@@ -0,0 +1,9 @@
+- - 0
```
AndresOrtegaGuerrero, line 1: Do we need all these files ? can it be in a single yml file ?

mbercx: These are automatically generated using the `pytest` `data_regression` fixture. The main reason I added these was to:

1. also test the case where `translation` is `None`.
2. not have the test rely on the `HubbardParameters.from_tuple` method, which `append_hubbard_parameter` uses internally.

AndresOrtegaGuerrero: ah ok , besides that it LGTM!

mbercx:
@bastonero do you still want to have a look?

bastonero:
> @bastonero do you still want to have a look?

Let's have a quick one

bastonero (changes requested):

bastonero:
Lol, I had a huge review with all the comments that nobody was seeing. That's why nobody was replying to those comments

mbercx:
> Lol, I had a huge review with all the comments that nobody was seeing. That's why nobody was replying to those comments

Haha, I read "let's have a quick one", then saw the massive review and almost fell out of my chair. 🤣

src/aiida_quantumespresso/data/hubbard_structure.py
```diff
@@ -109,8 +111,16 @@ def append_hubbard_parameter(
         :param hubbard_type: hubbard type (U, V, J, ...), defaults to 'Ueff'
             (see :class:`~aiida_quantumespresso.common.hubbard.Hubbard` for full allowed values)
         """
-        pymat = self.get_pymatgen_structure()
-        sites = pymat.sites
+        sites = [
+            PeriodicSite(
+                species=site.species, coords=site.coords, lattice=Lattice(self.cell), coords_are_cartesian=True
+            ) for site in self.get_pymatgen().sites
+        ]
```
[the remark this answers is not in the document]
bastonero: @mbercx just this comment

mbercx:
Now working on https://github.com/aiidateam/aiida-quantumespresso/pull/1007, will consider your comments there, but we already agreed that we'd stick to the onsite case for now.

bastonero (approved):
Thanks @mbercx , I do approve 🚀

bastonero:
@mbercx it seems you have to self approve yourself :D

mbercx:
> @mbercx it seems you have to self approve yourself :D

Oh right, I requested changes after all 😅

mbercx (approved):

mbercx:
Haha, seems we also need @unkcpz's approval now 😂

unkcpz (approved):
Looks cool, already out of my scope! Thanks everyone, approved.

PR #1005, opened 2024-02-11
Bug: `PhBaseWorkChain.set_qpoints` doesn't work as expected

### LINKED ISSUE #1004: `PhBaseWorkChain`: the new `set_qpoints` is untested and broken
A few lines in `PhBaseWorkChain` added recently are not covered by the test suite and have some typos.
The following:
https://github.com/aiidateam/aiida-quantumespresso/blob/b517686df8cd33765ef12cbf430620ea8e94f137/src/aiida_quantumespresso/workflows/ph/base.py#L186
should be:
`structure = self.ctx.inputs.parent_folder.creator.output.output_structure`
or 
`structure = self.inputs.ph.parent_folder.creator.output.output_structure`
and also
https://github.com/aiidateam/aiida-quantumespresso/blob/b517686df8cd33765ef12cbf430620ea8e94f137/src/aiida_quantumespresso/workflows/ph/base.py#L188
should be either
`structure = self.ctx.inputs.parent_folder.creator.inputs.structure`
or 
`structure = self.inputs.ph.parent_folder.creator.inputs.structure`

bastonero:
Fixes #1004 

The `set_qpoints` method was untested and contained many typos and bugs, which would break the usage of the workchain when a `qpoints_distance` is used and would set wrong `qpoints` in the context inputs.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -597,7 +597,11 @@ def generate_inputs_ph(
     """Generate default inputs for a `PhCalculation."""
 
     def _generate_inputs_ph(with_output_structure=False):
-        """Generate default inputs for a `PhCalculation."""
+        """Generate default inputs for a `PhCalculation.
+
+        :param with_output_structure: whether the PwCalculation has a StructureData in its outputs.
+            This is needed to test some PhBaseWorkChain logics.
+        """
         from aiida.common import LinkType
         from aiida.orm import Dict, RemoteData
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -603,7 +603,12 @@ def generate_inputs_ph(
 
         from aiida_quantumespresso.utils.resources import get_default_options
 
-        pw_node = generate_calc_job_node(inputs={'parameters': Dict(), 'structure': generate_structure()})
+        pw_node = generate_calc_job_node(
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -591,18 +591,37 @@ def _generate_inputs_q2r():
...
 
 @pytest.fixture
-def generate_inputs_ph(fixture_sandbox, fixture_localhost, fixture_code, generate_remote_data, generate_kpoints_mesh):
+def generate_inputs_ph(
+    generate_calc_job_node, generate_structure, fixture_localhost, fixture_code, generate_kpoints_mesh
+):
     """Generate default inputs for a `PhCalculation."""
 
-    def _generate_inputs_ph():
+    def _generate_inputs_ph(with_output_structure=False):
```
mbercx, line 599: Could you add a quick explanation of this argument to the docstring? It wasn't immediately clear to me that this mean the `parent_folder` creator was the one who'd get an output structure. ^^ I was also thinking of maybe renaming the argument to make this more clear, but could only come up with ridiculously long options.

mbercx (commented):
Thanks @bastonero! Changes look good, just left a small comment.

mbercx (approved):
Thanks @bastonero, let's merge this and do patch release. 🚀

PR #1006, opened 2024-02-12
`PhBaseWorkChain`: skip q-points validation if ports are excluded

mbercx:
Currently, the top-level `validate_inputs` validator for the `PhBaseWorkChain` will always check if either the `qpoints` or `qpoints_distance` inputs are provided. However, there are cases where a higher-level work chain that wraps the `PhBaseWorkChain` might provide the q-points input on the fly. Such a work chain would have to set the `PhBaseWorkChain` validator to `None` in its own `define` method.

However, as also discussed for a similar case for the `PwCalculation` in a389629, it's more elegant to check if the `qpoints_distance` and `qpoints` ports are present in the namespace. Typically a work chain that provides the q-points input on the fly will exclude these ports when exposing the `PhBaseWorkChain` inputs. By checking if the ports are present in the validator, we avoid having to set the validator to `None` as well in the higher-level work chain.

### DIFF
--- a/tests/workflows/ph/test_base.py
+++ b/tests/workflows/ph/test_base.py
@@ -3,7 +3,7 @@
 """Tests for the `PhBaseWorkChain` class."""
 from aiida import orm
 from aiida.common import AttributeDict, LinkType
-from aiida.engine import ProcessHandlerReport
+from aiida.engine import ProcessHandlerReport, WorkChain
 import pytest
 
 from aiida_quantumespresso.calculations.ph import PhCalculation
@@ -257,3 +257,22 @@ def test_merge_outputs(
     outputs = ph_base.get_outputs(node_2)
 
     data_regression.check(outputs['output_parameters'].get_dict())
+
+
+def test_validate_inputs_excluded_qpoints_distance(generate_inputs_ph):
+    """Test that q-points validation passes in case the ports are excluded in a parent work chain."""
+    from aiida.engine.utils import instantiate_process
+    from aiida.manage.manager import get_manager
+
+    class WrapPhBaseWorkChain(WorkChain):
+        """Minimal work chain that wraps a ``PhBaseWorkChain`` for testing excluding q-points inputs."""
+
+        @classmethod
+        def define(cls, spec):
+            super().define(spec)
+            spec.expose_inputs(PhBaseWorkChain, exclude=('qpoints', 'qpoints_distance'))
+
+    inputs = {'ph': generate_inputs_ph()}
+    inputs['ph'].pop('qpoints', None)
+    runner = get_manager().get_runner()
+    instantiate_process(runner, WrapPhBaseWorkChain, **inputs)
--- a/src/aiida_quantumespresso/workflows/ph/base.py
+++ b/src/aiida_quantumespresso/workflows/ph/base.py
@@ -69,9 +69,11 @@ class PhBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         # yapf: enable
 
     @classmethod
-    def validate_inputs(cls, inputs, ctx=None):  # pylint: disable=unused-argument
+    def validate_inputs(cls, value, port_namespace):  # pylint: disable=unused-argument
         """Validate the top level namespace."""
-        if 'qpoints_distance' not in inputs and 'qpoints' not in inputs:
+
+        if (('qpoints_distance' in port_namespace or 'qpoints' in port_namespace) and
+            'qpoints_distance' not in value and 'qpoints' not in value):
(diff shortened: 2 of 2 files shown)


bastonero (approved):
Thanks @mbercx , this looks good to me. Only a curiosity/question: can the logic of the namespace be tested?

mbercx:
Hehe, I was worried you might drop the word "tests". I'm also wondering what a good approach is for testing this change, maybe @sphuber can give a suggestion?

sphuber:
> Hehe, I was worried you might drop the word "tests". I'm also wondering what a good approach is for testing this change, maybe @sphuber can give a suggestion?

Something like this?

```python
from aiida.engine import WorkChain
from aiida_quantumespresso.workflows.ph.base import PhBaseWorkChain

class SubPhBaseWorkChain(WorkChain):

    @classmethod
    def define(cls, spec):
        super().define(spec)
        spec.expose_inputs(PhBaseWorkChain, exclude=('qpoints', 'qpoints_distance'))


def test_validate_inputs_excluded_qpoints_distance(generate_inputs_ph_base):
    from aiida.engine.utils import instantiate_process
    from aiida.manage.manager import get_manager

    inputs = generate_inputs_ph_base
    inputs.pop('qpoints', None)
    inputs.pop('qpoints_distance', None)
    runner = get_manager().get_runner()
    instantiate_process(runner, SubPhBaseWorkChain, **inputs)
```

tests/workflows/ph/test_base.py
```diff
@@ -257,3 +257,23 @@ def test_merge_outputs(
...
+        """Minimal work chain that wraps a ``PhBaseWorkChain`` for testing excluding q-points inputs."""
+
+        @classmethod
+        def define(cls, spec):
+            super().define(spec)
+            spec.expose_inputs(PhBaseWorkChain, exclude=('qpoints', 'qpoints_distance'))
+
+    inputs = {'ph': generate_inputs_ph()}
+    inputs['ph'].pop('qpoints', None)
+    inputs['ph'].pop('qpoints_distance', None)
```
bastonero, line 277: I don't think this is really needed right? Since it is an input of the PhBaseWorkChain

bastonero (commented):
Thanks @mbercx, very happy that a test has been added, and thanks @sphuber for the suggestion. Just a minor question, and it's good to go

tests/workflows/ph/test_base.py
bastonero: I mean the last line of the one highlighted.

mbercx: Oh fair, it's not in there anyways 👍

sphuber: > I don't think this is really needed right? Since it is an input of the PhBaseWorkChain

I don't follow. That is the whole point of this test. It is checking that, even though `qpoints` and `qpoints_distance` _are_ inputs to `PhBaseWorkChain` and so they are validated in the inputs namespace validator, when a wrapping workchain explicitly excludes them, the validator does not raise a warning. So the test needs to make sure that inputs are passed that do _not_ contain either `qpoints` or `qpoints_distance`. Right?

bastonero: Correct, but `generate_inputs_ph()` would generate inputs of `PhCalculation`, which doesn't have `qpoints_distance`, so it's one line more which is not needed. The two qpoints input for `PhBaseWorkChain` are moved into the "top level" namespace, and not nested in the `ph` namespace.

sphuber: Aaaah, I see, do you mean that since I used `generate_inputs_ph` which generates inputs for `PhCalculation` and not the workchain, there won't be a `qpoints_distance` to begin with. That makes sense. But then I think the real fix should be to replace `generate_inputs_ph` with a fixture that generates the inputs for `PhBaseWorkChain` and then pop the relevant inputs. This would be more future proof to changes in the input specs of the processes and the input generating fixtures.

mbercx: Yeah but I used the `generate_inputs_ph` fixture which wouldn't have the `qpoints_distance` (there is no `generate_inputs_ph_base` afaik).

bastonero: Fair enough, I do agree.

bastonero: I think one can use `generate_ph_workchain_node(return_inputs=True)` or similar (don't remember exactly the name of the fixutre)

sphuber: Ok, feel free to change as I suggested, but not necessary. Also fine to merge as is for me

bastonero (approved):
I think we can approve it this way for the moment being. Thanks @mbercx

PR #1011, opened 2024-02-19
First steps to make protocols scheduler agnostic

### LINKED ISSUE #1010: PwBandsWorkchain.get_builder_from_protocol() not working on SGE machines
The PwBandsWorkchain.get_builder_from_protocol() does not support SGE scheduler. This makes it cumbersome to do calculations SGE based machines. See error message below. This happens when providing a code that is run on a SGE computer.

```---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[20], [line 4](vscode-notebook-cell:?execution_count=20&line=4)
      [1](vscode-notebook-cell:?execution_count=20&line=1) for pk in scf_pk_calculated:
      [2](vscode-notebook-cell:?execution_count=20&line=2)     node = load_node(pk)
----> [4](vscode-notebook-cell:?execution_count=20&line=4)     PwBandsWorkChain.get_builder_from_protocol(code=code,structure=node.inputs.structure)

File [~/aiida-venv/lib/python3.10/site-packages/aiida_quantumespresso/workflows/pw/bands.py:139](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida_quantumespresso/workflows/pw/bands.py:139), in PwBandsWorkChain.get_builder_from_protocol(cls, code, structure, protocol, overrides, options, **kwargs)
    [136](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida_quantumespresso/workflows/pw/bands.py:136) inputs = cls.get_protocol_inputs(protocol, overrides)
    [138](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida_quantumespresso/workflows/pw/bands.py:138) args = (code, structure, protocol)
--> [139](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida_quantumespresso/workflows/pw/bands.py:139) relax = PwRelaxWorkChain.get_builder_from_protocol(
    [140](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida_quantumespresso/workflows/pw/bands.py:140)     *args, overrides=inputs.get('relax', None), options=options, **kwargs
    [141](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida_quantumespresso/workflows/pw/bands.py:141) )
    [142](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida_quantumespresso/workflows/pw/bands.py:142) scf = PwBaseWorkChain.get_builder_from_protocol(
    [143](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida_quantumespresso/workflows/pw/bands.py:143)     *args, overrides=inputs.get('scf', None), options=options, **kwargs
    [144](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida_quantumespresso/workflows/pw/bands.py:144) )
    [145](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida_quantumespresso/workflows/pw/bands.py:145) bands = PwBaseWorkChain.get_builder_from_protocol(
    [146](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida_quantumespresso/workflows/pw/bands.py:146)     *args, overrides=inputs.get('bands', None), options=options, **kwargs
    [147](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida_quantumespresso/workflows/pw/bands.py:147) )

File [~/aiida-venv/lib/python3.10/site-packages/aiida_quantumespresso/workflows/pw/relax.py:160](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida_quantumespresso/workflows/pw/relax.py:160), in PwRelaxWorkChain.get_builder_from_protocol(cls, code, structure, protocol, overrides, relax_type, options, **kwargs)
    [157](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida_quantumespresso/workflows/pw/relax.py:157)         raise ValueError(f'Structures with periodic boundary conditions `{structure.pbc}` are not supported.')
    [159](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida_quantumespresso/workflows/pw/relax.py:159) builder = cls.get_builder()
--> [160](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida_quantumespresso/workflows/pw/relax.py:160) builder.base = base
    [161](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida_quantumespresso/workflows/pw/relax.py:161) builder.base_final_scf = base_final_scf
...
    [126](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida/engine/processes/builder.py:126) # of a ``ProcessBuilderNamespace`` for the port of the attribute that is being set and than iteratively set
    [127](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida/engine/processes/builder.py:127) # all the values within the mapping that is being assigned to the attribute.
    [128](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/steven/~/aiida-venv/lib/python3.10/site-packages/aiida/engine/processes/builder.py:128) if isinstance(port, PortNamespace):

ValueError: invalid attribute value input `metadata.options.resources` is not valid for the `SgeScheduler` scheduler: `parallel_env` must be specified and must be a string```

### PR DESCRIPTION
**Pull request by t-reents**
Fixes #1010 

The current protocols implicitly rely on the `Slurm` scheduler, thus causing errors when users try to use ad different one, e.g. SGE as described in the issue. 
Here, the documentation of the `get_builder_from_protocol` method is extended to make the users aware of this behavior. Moreover, a warning is added in case no explicit `resources` are provided via the `options` argument, as this will cause errors in future versions.

@sphuber @mbercx Suggestions for a different phrasing or place where to put the additional documentation are very welcome. Moreover, I used a `UserWarning` because a `DeprecationWarning` somehow didn't appear. I'm not sure whether they are filtered at some point?

### DIFF
--- a/src/aiida_quantumespresso/workflows/pdos.py
+++ b/src/aiida_quantumespresso/workflows/pdos.py
@@ -381,6 +381,11 @@ class PdosWorkChain(ProtocolMixin, WorkChain):
             metadata_dos['options'] = recursive_merge(metadata_dos['options'], options)
             metadata_projwfc['options'] = recursive_merge(metadata_projwfc['options'], options)
 
+        metadata_dos['options'] = cls.set_default_resources(metadata_dos['options'], dos_code.computer.scheduler_type)
+        metadata_projwfc['options'] = cls.set_default_resources(
+            metadata_projwfc['options'], projwfc_code.computer.scheduler_type
+        )
+
         builder = cls.get_builder()
         builder.structure = structure
         builder.clean_workdir = orm.Bool(inputs['clean_workdir'])
--- a/src/aiida_quantumespresso/workflows/ph/base.py
+++ b/src/aiida_quantumespresso/workflows/ph/base.py
@@ -125,7 +125,9 @@ class PhBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         metadata = inputs['ph']['metadata']
 
         if options:
-            metadata['options'] = recursive_merge(inputs['ph']['metadata']['options'], options)
+            metadata['options'] = recursive_merge(metadata['options'], options)
+
+        metadata['options'] = cls.set_default_resources(metadata['options'], code.computer.scheduler_type)
 
         # pylint: disable=no-member
         builder = cls.get_builder()
--- a/src/aiida_quantumespresso/workflows/protocols/pdos.yaml
+++ b/src/aiida_quantumespresso/workflows/protocols/pdos.yaml
@@ -21,8 +21,6 @@ default_inputs:
                 DeltaE: 0.01
         metadata:
             options:
-                resources:
-                    num_machines: 1
                 max_wallclock_seconds: 43200  # Twelve hours
                 withmpi: True
     projwfc:
@@ -31,8 +29,6 @@ default_inputs:
                 DeltaE: 0.01
         metadata:
             options:
-                resources:
-                    num_machines: 1
                 max_wallclock_seconds: 43200  # Twelve hours
                 withmpi: True
 default_protocol: balanced
--- a/src/aiida_quantumespresso/workflows/protocols/ph/base.yaml
+++ b/src/aiida_quantumespresso/workflows/protocols/ph/base.yaml
@@ -6,8 +6,6 @@ default_inputs:
     ph:
         metadata:
             options:
-                resources:
-                    num_machines: 1
                 max_wallclock_seconds: 43200  # Twelve hours
                 withmpi: True
         parameters:
--- a/src/aiida_quantumespresso/workflows/protocols/pw/base.yaml
+++ b/src/aiida_quantumespresso/workflows/protocols/pw/base.yaml
@@ -10,8 +10,6 @@ default_inputs:
     pw:
         metadata:
             options:
-                resources:
-                    num_machines: 1
                 max_wallclock_seconds: 43200  # Twelve hours
                 withmpi: True
         parameters:
--- a/src/aiida_quantumespresso/workflows/protocols/utils.py
+++ b/src/aiida_quantumespresso/workflows/protocols/utils.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 """Utilities to manipulate the workflow input protocols."""
+import copy
 import pathlib
 from typing import Optional, Union
 import warnings
@@ -80,6 +81,26 @@ class ProtocolMixin:
             return recursive_merge(inputs, overrides)
 
         return inputs
+
+    @staticmethod
+    def set_default_resources(options: dict, scheduler_type: str) -> dict:
+        """Set default resources based on the scheduler type of a computer.
+
+        Temporary workaround to keep the default resources for the direct and Slurm schedulers until defaults can be
+        properly configured on computers, see https://github.com/aiidateam/aiida-quantumespresso/pull/1011
+
+        :param options: the options dictionary.
+        :param scheduler_type: the scheduler type, e.g. `core.slurm`.
+        """
+        new_options = copy.deepcopy(options)
+
+        if scheduler_type in ('core.direct', 'core.slurm', 'core.pbspro', 'core.torque'):
+            new_options.setdefault('resources', {}).setdefault('num_machines', 1)
+        if scheduler_type in ('core.sge',):
+            new_options.setdefault('resources', {}).setdefault('parallel_env', 'mpi')
+            new_options.setdefault('resources', {}).setdefault('tot_num_mpiprocs', 1)
+
+        return new_options
 
     @classmethod
     def _load_protocol_file(cls) -> dict:
--- a/src/aiida_quantumespresso/workflows/pw/base.py
+++ b/src/aiida_quantumespresso/workflows/pw/base.py
@@ -236,7 +236,9 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         metadata = inputs['pw']['metadata']
 
         if options:
-            metadata['options'] = recursive_merge(inputs['pw']['metadata']['options'], options)
+            metadata['options'] = recursive_merge(metadata['options'], options)
+
+        metadata['options'] = cls.set_default_resources(metadata['options'], code.computer.scheduler_type)
 
         # pylint: disable=no-member
         builder = cls.get_builder()
--- a/tests/workflows/protocols/pw/test_bands.py
+++ b/tests/workflows/protocols/pw/test_bands.py
@@ -6,6 +6,8 @@ import pytest
 from aiida_quantumespresso.common.types import ElectronicType, RelaxType, SpinType
 from aiida_quantumespresso.workflows.pw.bands import PwBandsWorkChain
 
+pytestmark = pytest.mark.usefixtures('pseudo_family')
+
 
 def test_get_available_protocols():
     """Test ``PwBandsWorkChain.get_available_protocols``."""
--- a/tests/workflows/protocols/pw/test_base.py
+++ b/tests/workflows/protocols/pw/test_base.py
@@ -355,3 +355,30 @@ def test_options(fixture_code, generate_structure):
 
     assert metadata['options']['queue_name'] == queue_name
     assert metadata['options']['withmpi'] == withmpi
+
+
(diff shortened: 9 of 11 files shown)

### CODE REVIEW

**Comment by sphuber** [comment]
> @sphuber @mbercx Suggestions for a different phrasing or place where to put the additional documentation are very welcome.

It would be great to also add this to the documentation in addition to the docstrings. I noticed that the `get_builder_from_protocol` functionality isn't really documented though. 

> Moreover, I used a `UserWarning` because a `DeprecationWarning` somehow didn't appear. I'm not sure whether they are filtered at some point?

Deprecation warnings are ignored by default in Python (https://docs.python.org/3/library/exceptions.html#DeprecationWarning). A user has to enable them manually to see them. I think a `UserWarning` is fine for now as an alternative.

**Comment by t-reents** [comment]
Hi @sphuber ! Sorry for picking this up just now.

I added a general description of the `get_builder_from_protocol` in the `How to - Workflows` section. To be honest, I wasn't 100% sure where to put it, but since it's a general remark about the workflows, I decided to put it there. Happy to discuss/adjust this.

Moreover, I realized that the issue of the protocols depending on SLURM also applies to the other workflows, not just the `pw` ones. Hence, I also added the warning to those.

**Comment by mbercx** [comment]
Thanks @t-reents, sorry for only looking at this PR now. A few preliminary comments, also after reading the corresponding issue:

1. If we don't remove the Slurm scheduler defaults from the protocol files, trying to use a code on an SGE-scheduler computer in e.g. the `PwRelaxWorkChain.get_builder_from_protocol` method will still fail outright, even if the correct SGE resources are specified through the options, because the `num_machines` input is still there.
2. Removing both these lines:

   https://github.com/aiidateam/aiida-quantumespresso/blob/435fc9008d012c9a0d81d8fad95624dde7b670cb/src/aiida_quantumespresso/workflows/protocols/pw/base.yaml#L13-L14

   will result in a failure in case the `resources` dictionary is not provided by the user. However, this should be a perfectly legitimate use case: for the `direct` scheduler specifying the resources is not really necessary. I think you can also set defaults for Slurm when setting up/configuring the computer (have to double-check, don't have Slurm locally).
3. I think a custom warning is not necessary. If we simply set the `resources` option to an empty dictionary in the protocol YAML files (removing the `num_machines` line), the scheduler inputs will still be validated before the calculation is run. _When_ this will happen depends if you are running the `PwBaseWorkChain` or a high-level work chain like the `PwRelaxWorkChain`. The former will only fail upon submission, but the latter will fail at this line:
   https://github.com/aiidateam/aiida-quantumespresso/blob/435fc9008d012c9a0d81d8fad95624dde7b670cb/src/aiida_quantumespresso/workflows/pw/relax.py#L160
   Because the validation does seem to happen at this point. Either way, a warning that checks the options doesn't seem necessary, because in case the user doesn't specify the correct resources the calculation will not run anyways?

**Comment by t-reents** [comment]
Thanks @mbercx for your comments!

I agree that simply removing the `num_machines` would basically fix it and the validation would do the rest of the job. I think that I also mentioned this as a second option in the issue. However, the idea of the warnings is just to have them for an intermediate period, to make the users aware that in one of the next releases, we'll expect explicitly defined resources (btw, I agree with you that it should be fine for the `local` scheduler). 

To be honest, I don't have a strong opinion on this, so if I understand your comments correctly, we can also directly adjust the protocol files and only add the information to the documentation, without introducing these warnings.

**Comment by mbercx** [comment]
> To be honest, I don't have a strong opinion on this, so if I understand your comments correctly, we can also directly adjust the protocol files and only add the information to the documentation, without introducing these warnings.

I think this should be the best way forward, considering that the warnings are sometimes unnecessary, and in case they are, the validation would stop the calculation from running anyways? Maybe I'm missing something though.

**Comment by mbercx** [comment]
@t-reents I see now we aren't removing the `num_machines` line from the protocol YAML files anymore. Unfortunately, that means that trying to use the protocol methods with e.g. the SGE scheduler will still fail with the following error even when you provide the correct resources in the options:

```python
ValueError: invalid attribute value input `metadata.options.resources` is not valid for the `SgeScheduler` scheduler: these parameters were not recognized: num_machines
```

So if we want to enable SGE users to actually use the protocol methods, I don't think we can get around removing the `num_machines:` line. Unfortunately, if we remove it, users will have to specify `num_machines` in the options for the direct and Slurm schedulers.

One option would be to follow [@sphuber's suggestion](https://github.com/aiidateam/aiida-quantumespresso/issues/1010#issuecomment-1947459351) and simply accept that users will have to specify the resources. I agree with him that implementing scheduler-specific logic in the protocol methods to provide some defaults is not really desirable. However, we'd be breaking backwards-compatibility in a sense, and make it a bit more difficult for users of these schedulers to run calculations. Especially for the direct scheduler this bugs me, why do I have to specify the `num_machines` if I'm just starting to use the plugin package on my own computer?

I think the real solution is that we should be able to specify scheduler defaults easily on the configured computer. But that will of course take quite a bit of time. As a stopgap, I'd be fine with:

1. Removing the `num_machines:` lines from the protocols, so we're not breaking the functionality for SGE users (and potentially others).
2. Adding some **temporary** scheduler-specific code for the direct and Slurm schedulers where necessary, e.g.:

```
        if code.computer.scheduler_type in ['core.slurm', 'core.direct']:
            inputs['pw']['metadata']['options']['resources'].setdefault('num_machines', 1)
```

Let me know if you agree. I'm happy to make the changes and add the necessary tests etc.

t-reents:
Hi @mbercx, yeah, I agree.

I'm very busy right now, so please feel free to continue, in case you have the time (If not, I'll do it at some point, but might take some time). I'm of course available to discuss

mbercx:
@t-reents I've updated the PR so default resources are added for most of the schedulers in `aiida-core`. Even though I don't think this workaround that adds scheduler-specific code to `aiida-quantumespresso` is the optimal solution, I think it's our best option in case we want to (1) allow for schedulers that don't use `NodeNumberJobResource` and (2) avoid requiring users to specify the default resources.

I've added the `add_option` `staticmethod` so all the logic regarding this workaround is in one place and doesn't need to be duplicated. Disadvantage is that someone looking at the code will have to check this method to see what is happening, and so it obscures the fact that default resources are set a bit. Maybe `set_options` is also a better name for the method? Let me know what you think, and I'll add some tests for the other schedulers if we agree on the approach and implementation.

mbercx:
@JPchico maybe give this branch a spin too if you'd like. Comments on the SGE defaults are welcome, since I've never used this scheduler myself.

PR #1013, opened 2024-02-27
XPS: Fix inputs for molecules

mhdzbert:
If is a molecular structure:
* Adds the 'mt' to `ch_scf` in order to treat charged molecular systems
* Performs the SCF calculations at Gamma point.


mhdzbert:
Many thanks @superstar54

superstar54 (approved):
Hi @mhdzbert, I set kpoints to [1, 1, 1] for the molecule case, otherwise, the default kpoints from protocol will conflict with the `gamma_only`.

mhdzbert:
Ok, thank you for correcting this.

PR #1016, opened 2024-03-19
add `should_run_scf` for PwBandsWorkChain

superstar54:
This PR sets the `scf` step as an option in the `PwBandsWorkChain`.

Use case:
- In aiidalab-qe, we want to run one `scf` in the main workflow, then pass the `remote_folder` to plugins, such as `bands` and `pdos`. This requires that the `scf` calculation is an option in the `PwBandsWorkChain`. 
 #https://github.com/aiidalab/aiidalab-qe/pull/631

### DIFF
--- a/src/aiida_quantumespresso/workflows/pw/bands.py
+++ b/src/aiida_quantumespresso/workflows/pw/bands.py
@@ -57,7 +57,11 @@ class PwBandsWorkChain(ProtocolMixin, WorkChain):
             'help': 'Inputs for the `PwRelaxWorkChain`, if not specified at all, the relaxation step is skipped.'})
         spec.expose_inputs(PwBaseWorkChain, namespace='scf',
             exclude=('clean_workdir', 'pw.structure'),
-            namespace_options={'help': 'Inputs for the `PwBaseWorkChain` for the SCF calculation.'})
+            namespace_options={
+                'help': 'Inputs for the `PwBaseWorkChain` for the SCF calculation.',
+                'required': False,
+                'populate_defaults': False,
+        })
         spec.expose_inputs(PwBaseWorkChain, namespace='bands',
             exclude=('clean_workdir', 'pw.structure', 'pw.kpoints', 'pw.kpoints_distance', 'pw.parent_folder'),
             namespace_options={'help': 'Inputs for the `PwBaseWorkChain` for the BANDS calculation.'})
@@ -81,8 +85,10 @@ class PwBandsWorkChain(ProtocolMixin, WorkChain):
             if_(cls.should_run_seekpath)(
                 cls.run_seekpath,
             ),
-            cls.run_scf,
-            cls.inspect_scf,
+            if_(cls.should_run_scf)(
+                cls.run_scf,
+                cls.inspect_scf,
+            ),
             cls.run_bands,
             cls.inspect_bands,
             cls.results,
@@ -104,7 +110,9 @@ class PwBandsWorkChain(ProtocolMixin, WorkChain):
             required=False,
             help='The parameters used in the SeeKpath call to normalize the input or relaxed structure.')
         spec.output('scf_parameters', valid_type=orm.Dict,
-            help='The output parameters of the SCF `PwBaseWorkChain`.')
+            required=False,
+            help='The output parameters of the SCF `PwBaseWorkChain`.'
+            )
         spec.output('band_parameters', valid_type=orm.Dict,
             help='The output parameters of the BANDS `PwBaseWorkChain`.')
         spec.output('band_structure', valid_type=orm.BandsData,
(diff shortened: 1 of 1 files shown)


src/aiida_quantumespresso/workflows/pw/bands.py
```diff
@@ -259,7 +271,9 @@ def run_bands(self):
         inputs.metadata.call_link_label = 'bands'
         inputs.kpoints = self.ctx.bands_kpoints
         inputs.pw.structure = self.ctx.current_structure
-        inputs.pw.parent_folder = self.ctx.current_folder
+        # override the parent_folder if the scf was run
+        if 'scf' in self.inputs:
+            inputs.pw.parent_folder = self.ctx.current_folder
```
sphuber, line 276: How would the `parent_folder` be defined though? It is excluded from the `bands` input namespace, and if scf is skipped this will therefore not be set?

sphuber (commented):
Thanks @superstar54 . Does this currently actually work? I believe if you skip the SCF step, then there won't be a `parent_folder` as input to the bands step, and the user also cannot specify it in the inputs since it is excluded. Besides that, as @mbercx mentioned during the meeting, if we strip out the SCF step, it is not much of a workflow anymore. If the user already has an SCF completed and wants the band structure, wouldn't they just be better off running a `PwBaseWorkChain` specifying the input parameters themselves? Using the `get_builder_from_protocol` they can get all inputs preset and then just switch the calculation to `bands`. The logic in `PwBandsWorkChain` specific to the bands is not very intricate anyway, it is just to define the total number of bands, which is just based on a top-level user input anyway.

src/aiida_quantumespresso/workflows/pw/bands.py
superstar54: Thanks for pointing out this. I didn't notice that the `parent_folder` is excluded. I run a simple example with Si by setting the `parent_folder`, it's strange that no error is reported, and the process finished successfully! Here is the process detail:
```console
$ verdi process show 28038
Property     Value
-----------  ------------------------------------
type         PwBandsWorkChain
state        Finished [0]
pk           28038
uuid         94b2d315-6a72-4f73-ae87-5a3b50d86d07
label
description
ctime        2024-03-21 15:34:17.077194+00:00
mtime        2024-03-21 15:34:24.276367+00:00

Inputs                  PK     Type
----------------------  -----  -------------
bands
    pw
        code            216    InstalledCode
        pseudos
            Si          19     UpfData
        parameters      28034  Dict
        parent_folder   28029  RemoteData
    max_iterations      28035  Int
bands_kpoints_distance  28036  Float
clean_workdir           28037  Bool
structure               28020  StructureData

Outputs                 PK  Type
-------------------  -----  -------------
band_parameters      28054  Dict
band_structure       28052  BandsData
primitive_structure  28042  StructureData
seekpath_parameters  28040  Dict

Called       PK  Type
--------  -----  ---------------------------
seekpath  28039  seekpath_structure_analysis
bands     28046  PwBaseWorkChain

Log messages
---------------------------------------------
There are 3 log messages for this calculation
Run 'verdi process report 28038' to see them
```
You see that the `parent_folder` is set. Here is the node graph
Screenshot from 2024-03-21 16-50-37

mbercx:
@superstar54 do you still think this is a value feature to have, considering @sphuber's comments above? Happy to discuss your use case, but if not I'll close this PR if that's alright.

mbercx:
@superstar54 I'm going to add this to the v5.0 milestone. I'm also considering removing the `PwRelaxWorkChain` from the `PwBandsWorkChain`. Idea being that I prefer to have more atomic work chains with a clear purpose, without too much overlap between them. These are then building blocks that can be tied together, e.g. using the `WorkGraph`.

While reworking the `PwBandsWorkChain`, I'll also add some documentation etc. Hopefully that will give me enough clarity to decide if the changes you made here provide a legitimate use case we want to integrate. :)

superstar54:
Hi @mbercx , thanks for looking at this again.

mbercx:
@superstar54 I've had some time to think about this, and I'm now leaning strongly towards accepting these changes, especially if we implement https://github.com/aiidateam/aiida-quantumespresso/issues/1146 (for which I'd love to hear your opinion btw ^^).

Although it's true you only run a single `bands` calculation in case you provide the SCF parent folder, the workflow does three other things to make your life easier:

1. Create a k-path with seekpath.
2. Set the number of bands based on the SCF.
3. Provide protocols for running the `bands` calculation.

superstar54:
Hi @mbercx, I see your point that if the user provides the `parent_folder` for SCF while also requiring to run' seekpath', there could be a mismatch in the structure. However, this can be avoided by validating the inputs.

My understanding is: this is an opt-in feature. The users still run the original workflow, but if they want to avoid duplicate scf (as in aiidalab-qe case), they can choose this feature, while the validation of the inputs also makes sure they are not running `seekpath`, and require providing the `bands_kpoints` at the same time.

I can add this validation if you think this is reasonable.

PR #1017, opened 2024-03-25
`PhCalculation`: add symmetry related exit codes

### LINKED ISSUE #1001: `PhBaseWorkChain`: incompatible FFT grid error doesn't have an exit code
The following error has not a dedicated exit code:
```
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
     Error in routine phq_setup (1):
     FFT grid incompatible with symmetry
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
```
It would be more informative to have an exit code specific to this error, although unrecoverable (FFT must be changed, i.e. the cutoff of a previous pw.x calculation).

bastonero:
Fixes #1001

Some symmetry errors that ph.x can stop with are added. They are related to how the PHonon code handles internally symmetry, which unfortunately is slightly different from the PW code, which is used as a necessary previous step to run ph.x. This is generally due to some interal hard-coded thresholds. The solution is generally to run pw.x by specifying the ibrav, instead of the general ibrav=0. This cannot be handled though, as it would require to re-run the pw.x code.


bastonero:
@mbercx very simple PR, but recently I was so annoyed by the fact I could not distinguish wether the PH code was stoppiing due to this error or due to some scheduler issues. 

As written in the PR, these errors don't have a straight solution if not rerunning pw.x with e.g. ibrav or more relaxed positions/cell. So I put them in the 300s range.

src/aiida_quantumespresso/calculations/ph.py
```diff
@@ -72,6 +72,11 @@ def define(cls, spec):
             message='The stdout output file was incomplete probably because the calculation got interrupted.')
         spec.exit_code(350, 'ERROR_UNEXPECTED_PARSER_EXCEPTION',
             message='The parser raised an unexpected exception: {exception}')
+        spec.exit_code(360, 'ERROR_INCOMPATIBLE_FFT_GRID',
+            message='The FFT grid is incompatible with the detected symmetries. Try using the correct ibrav.')
```
mbercx, line 76: Maybe we should explicitly state that the `ibrav` should be set in the parent `pw.x`. I also wasn't sure if we should use "correct" here:

```suggestion
            message='The FFT grid is incompatible with the detected symmetries. Try using the lattice-specific `ibrav` != 0 in the parent `pw.x` calculation.')
```

Do you know if these errors only happen when `ibrav == 0`?

Same for the other exit code message.

bastonero: I still didn't try, but reading on the QE documentation it seems that by specifying ibrav the error should solve (it is really due to how QE-ph handles symmetries)

mbercx: Alright, let's leave it like this for now. In case we ever see that the error pops up for `ibrav != 0` we'll adapt.

src/aiida_quantumespresso/calculations/ph.py
```diff
@@ -72,6 +72,12 @@ def define(cls, spec):
             message='The stdout output file was incomplete probably because the calculation got interrupted.')
         spec.exit_code(350, 'ERROR_UNEXPECTED_PARSER_EXCEPTION',
             message='The parser raised an unexpected exception: {exception}')
+        spec.exit_code(360, 'ERROR_INCOMPATIBLE_FFT_GRID',
+            message='The FFT grid is incompatible with the detected symmetries. Try using the lattice-specific '
+                    '`ibrav` != 0 in the parent `pw.x` calculation.')
+        spec.exit_code(361, 'ERROR_WRONG_REPRESENTATION',
+            message=('The representation found seems to be wrong according to the detected symmetries. '
+                     'Try using the correct ibrav.'))
```
mbercx, line 80: This one too please ^^

```suggestion
                     'Try using the lattice-specific `ibrav` != 0 in the parent `pw.x` calculation.'))
```

bastonero:
Ah sorry @mbercx , i thought you pushed the changes. coming with the commit to fix the message

mbercx (approved):
LGTM! 🚀 Thanks again, @bastonero!

PR #1021, opened 2024-03-27
Bands from protocol: fix `bands_kpoints` overrides

### LINKED ISSUE #1020: PwBandsWorkChain: the get_builder_from_protocol does not override `bands_kpoints`
If one specifies `bands_kpoints`, this is not corretly overridden.

bastonero:
Fixes #1020 

The `bands_kpoints` wouldn't be override if specified in the overrides and passed to the PwBandsWorkChain.get_builder_from_protocol. If specified, it is chosen instead of the `bands_kpoints_distance`.


sphuber (approved):
Thanks @bastonero

PR #1024, opened 2024-04-23
Release `v4.6.0`

mbercx:


sphuber (approved):
Thanks @mbercx . Guess this may have also just been a patch release since those improvements are mostly almost just fixes right? But anyway, doesn't really matter.

mbercx:
> Guess this may have also just been a patch release since those improvements are mostly almost just fixes right?

It's a fair question, I guess the line is a bit blurry sometimes. ^^ A minor release "felt" right.

PR #1026, opened 2024-04-29
`get_xspectra_structures`: Refactor and Improve Code

PNOGillespie:
# Overview
In this PR, we have re-factored various components of `get_xspectra_structures` as part of ongoing improvement work. These changes are the first of two improvements intended to enable users to set symmetry data manually instead of relying on automatic symmetry analysis - ultimately enabling the same feature for the `XspectraCrystalWorkChain` and any others which use the same structure preparation tools.

This PR will be followed by a second one in which the logic for the XSpectra `WorkChain`s will be added in order to properly exploit the changes in this PR. The overall changes have been separated into 2 PRs due to the number of changes made here as part of the re-factoring process.

# Changes
* Moves supercell creation, processes for molecules, and generation of `equivalent_sites_data` to separate functions.
* Adds functionality for users to manually provide symmetry data, thus enabling the `CalcFunction` to be used as a means to generate structures with user control over which exact sites to mark.
* Fixes a bug discovered in the case where non-hubbard structures with custom Kind names lost their Kind names when converting the ASE supercell to `StructureData` type.
* Removes unnecessary usage of `dict.keys()` iterables where simply iterating over the dictionary directly can be used instead.
* Fixes a small oversight in the tests where `spglib_settings` was mistakenly named `spglib_options`.


src/aiida_quantumespresso/workflows/functions/get_xspectra_structures.py
```diff
@@ -15,6 +15,160 @@
 from aiida_quantumespresso.utils.hubbard import HubbardStructureData, HubbardUtils
 
 
+def process_molecule_input(structure, **kwargs):  # pylint: disable=too-many-statements
+    """Generate symmetry data and supercell suitable for molecular systems using Pymatgen."""
+    from pymatgen.symmetry.analyzer import PointGroupAnalyzer
+
+    result = {'output_params': {}}
+    # If we are working with a molecule, check for pymatgen_settings
```
superstar54, line 23:
```suggestion

```

src/aiida_quantumespresso/workflows/functions/get_xspectra_structures.py
```diff
@@ -52,6 +206,14 @@ def get_xspectra_structures(structure, **kwargs):  # pylint: disable=too-many-st
                              from the parent structure. For instance, use_element_types = True would
                              consider sites for Kinds 'Si' and 'Si1' to be equivalent if both are sites
                              containing silicon. Defaults to True.
+        - parse_symmetry: a Bool object to indicate that the symmetry data should be generated using spglib
+                          for the input structure. If False, will instead use symmetry data provided
+                          via the `equivalent_sites_data` and `spacegroup_number` inputs and generate all
+                          the structures based on this information. Defaults to True.
+        - equivalent_sites_data: a Dict object taking the form of the `equivalent_sites_data` dict normally
```
superstar54, line 213: This is not clear to the user.

src/aiida_quantumespresso/workflows/functions/get_xspectra_structures.py
```diff
@@ -63,10 +225,14 @@ def get_xspectra_structures(structure, **kwargs):  # pylint: disable=too-many-st
...
     """
 
-    is_molecule_input = False
     elements_present = [kind.symbol for kind in structure.kinds]
     unwrapped_kwargs = {key: node for key, node in kwargs.items() if isinstance(node, orm.Data)}
-    if 'abs_atom_marker' in unwrapped_kwargs.keys():
+
+    is_molecule_input = False
+    if 'is_molecule_input' in unwrapped_kwargs:
+        is_molecule_input = kwargs['is_molecule_input'].value
```
superstar54, line 233:
```suggestion
    is_molecule_input = kwargs.get('is_molecule_input', False)
```
this should work because one can use AiiDA Bool data in the `if` condition directly.

superstar54 (changes requested):
Thanks, @PNOGillespie, for the work!

Could you also add a test when setting symmetry data manually.

PNOGillespie:
Hi @superstar54, Thanks for the feedback. I've added a simple test for manual symmetry data input and cleaned up a few things as you suggested doing in one of your comments.

I also added an example of how the `equivalent_sites_data` input needs to be formatted. Was that the part which was unclear in the docstring, or do you think there's more to do on that?

superstar54 (dismissed):
LGTM!

PNOGillespie:
Hi @superstar54. I fixed the issue with the docstring by formatting the example to be in-line. I also added a sentence to better explain what needs to be in the `equivalent_sites_data` dictionary.

Let me know if anything else is needed.

src/aiida_quantumespresso/workflows/functions/get_xspectra_structures.py
superstar54: Could you also fix this? The structure is a `molecule` inside this function.

PNOGillespie: Sorry about that. I thought I'd removed the comment in the first set of changes that I committed. It's been removed in the [latest commit](https://github.com/aiidateam/aiida-quantumespresso/pull/1026/commits/fbba65160a15df4399c1d74e025eb0a046201c6c)

superstar54 (approved):
LGTM!

PR #1033, opened 2024-06-15
Basic implementation of `bands.x`

bastonero:
An implementation that wraps the binary `bands.x` was missing. This program is used to find the symmetries of wavefunctions, to re-order the bands, and to perform basic post-processings, such as the calculation of the momentum operator.

The current parser only performs basic parsing, with no specialized outputs for k-point wavefunction symmetries, nor momentum operator, nor bands. This can be, for instance, implemented over time depending on user request.


src/aiida_quantumespresso/calculations/bands.py
```diff
@@ -0,0 +1,39 @@
+# -*- coding: utf-8 -*-
+"""`CalcJob` implementation for the q2r.x code of Quantum ESPRESSO."""
```
sphuber, line 2:
```suggestion
"""`CalcJob` implementation for the bands.x code of Quantum ESPRESSO."""
```

src/aiida_quantumespresso/calculations/bands.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+
+    _MOMENTUM_OPERATOR_NAME = 'momentum_operator.dat'
+    _BANDS_NAME = 'bands.dat'
+
+    _default_namelists = ['BANDS']
+    _blocked_keywords = [
+        ('BANDS', 'outdir', NamelistsCalculation._OUTPUT_SUBFOLDER),  # pylint: disable=protected-access
+        ('BANDS', 'prefix', NamelistsCalculation._PREFIX),  # pylint: disable=protected-access
+        ('BANDS', 'filband', _BANDS_NAME),  # pylint: disable=protected-access
```
sphuber, line 25: I think this rule is not used, right?
```suggestion
        ('BANDS', 'filband', _BANDS_NAME),
```

tests/parsers/fixtures/bands/default/aiida.out
sphuber, line 1: Can you please cut this output to the bare minimum? Just whatever is needed for the parsing to work

sphuber (changes requested):
Thanks @bastonero all looks good. Just some minor requests.

As a side note, not sure if you are already aware of `aiida-shell` but for it makes it really easy to run these small kind of utilities without having to write dedicated plugins. You can see [an example of QE the with use of `bands.x`](https://aiida-shell.readthedocs.io/en/latest/examples/qe.html#id6) in the online docs.

src/aiida_quantumespresso/calculations/bands.py
bastonero: Right, thanks!

tests/parsers/fixtures/bands/default/aiida.out
bastonero: Sure

bastonero:
Thanks a lot @sphuber for the review.

I was aware of the plugin, but thanks for the reminder, I wasn't aware of the very nice docs. Here I was interested to use the CalcJob/WorkChain for future composite WorkChains. But good to keep in mind about aiida-shell.

bastonero:
PS: any idea why the docs is always failing lately?

mbercx:
> PS: any idea why the docs is always failing lately?


Hmm, let me have a look 👀

sphuber (approved):

sphuber:
Fixed the RTD build so merging this. Thanks @bastonero

PR #1039, opened 2024-07-10
CLI: Fix bug in `aiida-quantumespresso workflow launch pw-base`

sphuber:
The command would not specify the `CONTROL` key for the input parameters resulting in an exception in the upload step. This was not discovered by tests because the workflow is not actually run and the exception is only hit when the calcjob is actually executed.


sphuber:
@mbercx the exception in question is hit here https://github.com/aiidateam/aiida-quantumespresso/blob/91c3e1d35939491663a697d201dcccdf90c076c6/src/aiida_quantumespresso/calculations/__init__.py#L720
It essentially means that the `CONTROL.calculation` key is required for the `PwCalculation`, even though it is made required by the `BasePwCpInputGenerator` class. I wonder if we therefore should not add this check to the validator of the `PwCalculation` class itself instead of letting it except in the base class when the upload task is executed.

GeigerJ2 (approved):
LGTM!

PR #1049, opened 2024-11-21
NebCalculation: Add intermediate image structures

ionuorah:
Intermediate structures that might be need to guide the NEB path are enabled and added as trajectory data.

### DIFF
--- a/src/aiida_quantumespresso/calculations/neb.py
+++ b/src/aiida_quantumespresso/calculations/neb.py
@@ -62,6 +62,8 @@ class NebCalculation(CalcJob):
         spec.input('metadata.options.parser_name', valid_type=str, default='quantumespresso.neb')
         spec.input('first_structure', valid_type=orm.StructureData, help='Initial structure')
         spec.input('last_structure', valid_type=orm.StructureData, help='Final structure')
+        spec.input('intermediate_structures', valid_type=orm.TrajectoryData, required=False, 
+            help='intermediate structure(s) only')
         spec.input('parameters', valid_type=orm.Dict, help='NEB-specific input parameters')
         spec.input('settings', valid_type=orm.Dict, required=False,
             help='Optional parameters to affect the way the calculation job and the parsing are performed.')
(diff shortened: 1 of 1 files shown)


rikigigi:
I had in my plans to implement this too 🙂

Only a question: why do we have to provide also
first_structure and last_structure as input when intermediate_structures trajectory is provided? In this way I have to split my input trajectory in three nodes (suppose for example that I'm modifying one that was calculated by this plugin).

I would allow the user to use or first_structure and last_structure, or intermediate structure, not both, and maybe deprecate the old input and in a next major release delete it

What do you think? Did you have other specific usecases?

Thank you

ionuorah:
> I had in my plans to implement this too 🙂
> 
> Only a question: why do we have to provide also first_structure and last_structure as input when intermediate_structures trajectory is provided? In this way I have to split my input trajectory in three nodes (suppose for example that I'm modifying one that was calculated by this plugin).
> 
> I would allow the user to use or first_structure and last_structure, or intermediate structure, not both, and maybe deprecate the old input and in a next major release delete it
> 
> What do you think? Did you have other specific usecases?
> 
> Thank you

I have  no particular usecase, I only want to be able to include the intermediate structures. Maybe  it is better to allow the user to provide either first/last structures  or a trajectory data of all (first, intermediate(s) , last)  structures.

bastonero:
I agree to just use one of the two, and in a next major release to switch to something like `trajectory`, which might contain only start and last, or the former two plus intermediates. I would also call it `trajectory` or `path` or similar, as `intermediate_structures` is a bit too long and it is also less general. Cheers

rikigigi:
in our timeline probably we are going to start to work on this around mid-august

mbercx:
Thanks all for the discussion and contributions! If I understand correctly, this PR is superseded by https://github.com/aiidateam/aiida-quantumespresso/pull/1106, so I'll close it. Feel free to reopen in case I misunderstood.

PR #1050, opened 2024-11-25
Adding support for aiida-atomistic `StructureData`

mikibonacci:
This PR is still in draft request, not ready to be reviewed.

### DIFF
--- a/src/aiida_quantumespresso/utils/magnetic.py
+++ b/src/aiida_quantumespresso/utils/magnetic.py
@@ -1,0 +1,90 @@
+# -*- coding: utf-8 -*-
+"""Utility class for handling the :class:`aiida_quantumespresso.data.hubbard_structure.HubbardStructureData`."""
+# pylint: disable=no-name-in-module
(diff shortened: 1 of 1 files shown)


mikibonacci:
Hi @mbercx, The PR is in draft state but you can already have a look at the modifications, so I to know if I am going in wrong directions (before touching the whole code base). 

Thanks!

mbercx:
@mikibonacci sorry for never getting around this... It's been a while since I was able to dedicate time to `aiida-quantumespresso`.

I suppose this work is outdated with the latest changes in `aiida-atomistic`? Once I get the `aiida-quantumespresso` plugin back into fighting shape and release v5.0, I can dedicate some proper time to helping out here.

For now: I'm planning to change the formatter, so that will probably cause horrible merge conflicts. Is that ok?

mbercx:
Thanks for getting back to me @mikibonacci! Let's first get our docs here in a less shameful state, and then we can push the new `StructureData` to 100%.

PR #1052, opened 2024-12-17
Updated the protocols to accomodate the suggestions proposed on issue #959 in aiidalab_qe.

Minotakm:
Hey all,

There is an ongoing [discussion](https://github.com/aiidalab/aiidalab-qe/issues/959) regarding why the DOS sometimes misses peaks at low energies and why PDOS and DOS results can differ. The discussion concluded with the following changes:

- Decreasing the `DeltaE` value to 0.01 in the DOS protocol.
- Changing the occupations setting from `tetrahedra` to `tetrahedra_opt`.

Following that I did some tests, where here we can see that the low peak now is tracked. 
image

Additionally, the DOS calculated with dos.x, projwfc.x, and the sum of the projections from pDOS are now consistent.
image

Furthermore, I updated the `PdosWorkChain` to check for the `tetrahedra` parameters.

Closes https://github.com/aiidalab/aiidalab-qe/issues/959

PR #1063, opened 2025-01-22
🚀 Release `v4.8.0`

mbercx:


mbercx:
Hmm, let me see why it's still trying those `tests (3.8` which aren't running...

mbercx:
Was since the `tests (3.8)` were still one of the tests that had to pass according to the branch protections.

sphuber (approved):

mbercx:
@AndresOrtegaGuerrero I will most likely not have time to do so this week, and [your comment here](https://github.com/aiidateam/aiida-quantumespresso/pull/1052#issuecomment-2589638471) regarding the issues with parsing of the `projwfc.x` means getting that fixed properly might still take some time, especially if we want to try to merge https://github.com/aiidateam/aiida-quantumespresso/pull/749, which technically should fix that problem but is quite a beast.

Doing a release is quite quick, really. So I'd rather get this out now instead of blocking it for https://github.com/aiidateam/aiida-quantumespresso/pull/1052.

AndresOrtegaGuerrero:
> @AndresOrtegaGuerrero I will most likely not have time to do so this week, and [your comment here](https://github.com/aiidateam/aiida-quantumespresso/pull/1052#issuecomment-2589638471) regarding the issues with parsing of the `projwfc.x` means getting that fixed properly might still take some time, especially if we want to try to merge #749, which technically should fix that problem but is quite a beast.
> 
> Doing a release is quite quick, really. So I'd rather get this out now instead of blocking it for #1052.

Sure

PR #1066, opened 2025-02-04
Deprecate `ADDITIONAL_RETRIEVE_LIST` in `settings`

### LINKED ISSUE #939: 🗑 Deprecate `ADDITIONAL_RETRIEVE_LIST` setting in favor of option
Now that `aiida-core` has support for specifying the `additional_retrieve_list` setting as an option a `CalcJob`:

https://github.com/aiidateam/aiida-core/commit/17b77181d87abed9211fdaf1f432e00e276c1c11

We should deprecate the use of this setting and remove it in the next major release.

t-reents:
Fixes #939.

Deprecate the `ADDITIONAL_RETRIEVE_LIST` key in the `settings` input. Instead, one should use the `CalcJob.metadata.options.additional_retrieve_list` option to specify files that should be retrieved in addition to the default ones.


src/aiida_quantumespresso/calculations/__init__.py
```diff
@@ -183,6 +183,13 @@ def validate_inputs(cls, value, port_namespace):
         if 'settings' in value:
             settings = _uppercase_dict(value['settings'].get_dict(), dict_name='settings')
 
+            if 'ADDITIONAL_RETRIEVE_LIST' in settings:
+                warnings.warn(
+                    'The key `ADDITIONAL_RETRIEVE_LIST` in the settings input is deprecated and will be removed in '
+                    'the future. Use the `CalcJob.metadata.options.additional_retrieve_list` input instead.',
+                    FutureWarning
```
sphuber, line 190: Think these warnings are not shown by default in Python. This is why we use a separate class `aiida.common.warnings.AiidaDeprecationWarning` (https://github.com/aiidateam/aiida-core/blob/f43a51010b3c84b1e9526d167401fc9ac07c556a/src/aiida/common/warnings.py#L15)

I am not sure if we were already using this approach in plugins, but may be a nice way to ensure users see the warning (and that it can be turned off through `verdi config warnings.showdeprecations False`).

sphuber (approved):
Changes look good to me, but just suggest to look into changing the warning class to ensure users see it

src/aiida_quantumespresso/calculations/__init__.py
t-reents: Thanks @sphuber for the review and reminder! You are right, those wouldn't be shown. I'll adapt it

t-reents: Just for context: I actually remember now why I used `FutureWarning` instead of `DeprecationWarning`, because `DeprecationWarnings` are silenced per default but `FutureWarnings` shouldn't be. Nonetheless, I think it makes sense to use the AiiDA specific subclass for the other reason outlined above.

mbercx:
@t-reents if you're ok with `AiidaDeprecationWarning`, I suppose we can adapt this and get the PR merged?

t-reents:
Thanks for the ping @mbercx ! Yes, I am, just need to pick it up again. I'll do it later today and let you know once it's done.

t-reents:
@mbercx Please have a final look

mbercx (approved):

PR #1074, opened 2025-03-27
📚 Add basic `PwBaseWorkChain` documentation

mbercx:
Mainly testing if the docs do build properly still for branches on the `aiidateam` repository.


mbercx:
Note: closing the PR and reopening it fixed the RTD build issue. I'm assuming this forced Github to update the references again, which can then be properly picked up by RTD.

mbercx:
Now it's failing at the `checkout` stage...

```
git clone --depth 1 https://github.com/aiidateam/aiida-quantumespresso.git . 0s
git fetch origin --force --prune --prune-tags --depth 50 pull/1074/head:external-1074 0s
git checkout --force cadfc7fd310c20e4a34a51526dd5c0f22605508d
```

With the error:

```
fatal: reference is not a tree: cadfc7fd310c20e4a34a51526dd5c0f22605508d
```

mbercx:
It seems that the RTD build starts _too fast_ somehow. I can reproduce the error locally, and I see that the latest commit is missing from the log. If I wait a bit longer, the checkout command does work. At that point I can rebuild the documentation on RTD and all is well.

mbercx:
Tried to integrate a change in the `.readthedocs.yaml` file to make the build wait a bit, but that didn't work. I suppose we could add it to the CI, but maybe this is a temporary issue. For now I guess we can just wait a bit and retrigger the RTD build.

bastonero (changes requested):
Thanks a lot @mbercx !

This is really needed, and thanks for pushing it. I think it would be great to have at least two or three how-tos:
1. One showing how to get the builder and "fill it up" (or even just creating an `inputs` dictionary with the minimal structure
2. The second showing how to easily get it using the `get_builder_from_protocol`
3. The third showing how to override the default inputs

Then we should also add an extra how-to page to just explain in general the usage of the `get_builder_from_protocol` and how to get other info about protocols.

mbercx:
Haha, thanks for the comments @bastonero! This was really just a _super_ initial draft to test why the RTD build was failing, but yes I did want to get started on this, documentation should be a priority after we merge as many of the open PRs as possible I think.

I'll try to have a more solid crack at adding these first parts later this week!

Minotakm:
Hey @mbercx, we all agreed in yesterdays meeting that it is a nice consise `how-to`. The only issue with that is the `run` vs `submit` method. On my point of view, we just need to make sure that the use understands the difference between these two methods. That in the first case the cell is blocked until the calculation stops, you do it to show to them that `something` is calculated, also  because the calculation is relatively quick. In real word,  most of the cases you use `submit`.

mbercx:
Superseded by https://github.com/aiidateam/aiida-quantumespresso/pull/1129

PR #1075, opened 2025-04-11
add new qe schema 240411 for 7.4.1

pietrodelugas:
this is a slighly update scheme, the only change with qes_241015 is a new optional element describing the sawtooth potential

### DIFF
--- a/src/aiida_quantumespresso/parsers/parse_xml/schemas/qes_241104.xsd
+++ b/src/aiida_quantumespresso/parsers/parse_xml/schemas/qes_241104.xsd
@@ -1,0 +1,1405 @@
+<?xml version="1.0"?>
+<!--
+Copyright (c), 2021-2024, Quantum Espresso Foundation. All rights reserved.
+This file is distributed under the terms of the MIT License. See the
+file 'LICENSE' in the root directory of the present distribution, or
+http://opensource.org/licenses/MIT.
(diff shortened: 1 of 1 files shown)


Minotakm:
Hey @pietrodelugas  thank you very much! There is something wrong probably with my git or how it renders the differences and I cannot see them. Is it possible to update the correct schema on QE-schemas and we can have at look ? Thank you again!

gsavva:
Thanks Pietro!
I think there is a bit of a confusion with the filenames.
The new schema should have been `qes_241104.xsd` which is actually missing, whereas the edit was done on another file with similar name `qes_240411.xsd`

pietrodelugas:
@Minotakm @gsavva, it's not the diff that is having issues; it's someone else here around. 
I forgot to add the file before committing :-( 
Now, it should be fine. I have also added it to the qeschemas repository.  
Thanks!!

Minotakm:
Hahaha thank you so much! I will give it a try :D

Minotakm:
Hey @pietrodelugas . Thanks again for the schema! I tested locally and it works fine. To approve it we need to update the tests accordinly. Shall I commit these changes on this PR?

mbercx:
Superseded by https://github.com/aiidateam/aiida-quantumespresso/pull/1087

PR #1077, opened 2025-04-14
Fix a docstring typo

npaulish:


Minotakm (approved):
LGTM 🚀

mbercx:
@npaulish thanks for the fix! I wanted to fix the `pre-commit`, but couldn't force push to the source branch from the PR, probably because it's your `main` which has similar protection rules as the `aiidateam` one. So I opened https://github.com/aiidateam/aiida-quantumespresso/pull/1078 to reintroduce the fix and merge that one.

PR #1079, opened 2025-04-17
Fix in PdosWorkChain to allow retrieval of nbands params from finished SCF calc when no SCF is to be launched

ValkScripter:
In the case where no SCF workchain is explicitly launched, `workchain_scf` is not in `self.ctx` so an error occurs when trying to get the `output_parameters` of the SCF workchain and retrieve the number of bands used when launching Nscf WorkChain.

### Workflow functionality enhancement:
* Updated the `run_nscf` method to handle the scenario where an SCF workchain is not run by adding `workchain_scf` to `self.ctx` when 'scf' is not in `self.inputs`. The `workchain_scf` is the caller of the creator of the `parent_folder` specified in 'nscf' inputs.

### DIFF
--- a/src/aiida_quantumespresso/workflows/pdos.py
+++ b/src/aiida_quantumespresso/workflows/pdos.py
@@ -46,12 +46,11 @@ Related Resources:
     (see `this post <https://lists.quantum-espresso.org/pipermail/users/2017-November/039656.html>`_).
 
 """
+import jsonschema
 from aiida import orm, plugins
 from aiida.common import AttributeDict
 from aiida.engine import ToContext, WorkChain, if_
 from aiida.orm.nodes.data.base import to_aiida_type
-import jsonschema
-
 from aiida_quantumespresso.utils.mapping import prepare_process_inputs
 
 from .protocols.utils import ProtocolMixin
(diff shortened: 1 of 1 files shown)


ValkScripter:
Hi @t-reents,

Thanks for the reply and I think you are absolutely right for the no-wrapping case.
Then it is maybe better not to get the - eventually non-existing - `workchain_scf` from the `parent_folder` but just to set the value of `self.ctx.scf_parent_folder` to the input `parent_folder` and modify the parts of the code which rely on `self.ctx.workchain_scf` to make them use the creator of the `parent_folder`.

I will do some test calculations to see where other errors appear.

ValkScripter:
Hi again,

I modified my fix so that the `nbnd` is retrieved directly from `nscf.pw.parent_folder.creator`.
There is only one other occurrence in the code which used the `self.ctx.workchain_scf` without checking its existence and it's when `serial_clean` is True in `inspect_nscf` : `clean_workchain_calcs(self.ctx.workchain_scf)` is called.
So without changing this part of the code I added input validation instead to make sure there is no error during execution of the workflow but rather at the beginning. Plus I don't think using `serial_clean` with an external SCF calc/workchain is a good use case. 

I performed tests with this new fix using the `remote_folder` of a bare `PwCalculation` as `parent_folder` and it worked fine for me.

PR #1082, opened 2025-05-19
Update CI config

danielhollas:
~~Two tests started failing after the recent click 8.2.0. However, the tests were actually testing the `aiida-pseudo` CLI so  we can remove them here! :grin:~~

~~I've added them to the aiida-pseudo repo in https://github.com/aiidateam/aiida-pseudo/pull/186~~

I am doing a bunch of CI cleanup here. In particular, I am using `uv` for a much faster installation (up to 50s shaved off).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,9 @@ name: ci
 
 on: [push, pull_request]
 
+env:
+    FORCE_COLOR: 1
+
 jobs:
 
     pre-commit:
@@ -11,21 +14,19 @@ jobs:
         steps:
         -   uses: actions/checkout@v4
 
-        -   name: Cache Python dependencies
-            uses: actions/cache@v4
-            with:
-                path: ~/.cache/pip
-                key: pip-pre-commit-${{ hashFiles('**/setup.json') }}
-                restore-keys:
-                    pip-pre-commit-
-
         -   name: Set up Python
             uses: actions/setup-python@v5
             with:
                 python-version: '3.9'
 
+        - name: Setup uv
+          uses: astral-sh/setup-uv@v6
+          with:
+              activate-environment: true
+              python-version: '3.9'
+
         -   name: Install Python dependencies
-            run: pip install -e .[pre-commit,tests]
+            run: uv pip install -e .[pre-commit,tests]
 
         -   name: Run pre-commit
             run: pre-commit run --all-files || ( git status --short ; git diff ; exit 1 )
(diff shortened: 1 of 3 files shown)


.github/workflows/nightly.yml
```diff
@@ -33,33 +35,30 @@ jobs:
...
             run: |
-                pip install -e .[tests]
-                pip install git+https://github.com/aiidateam/aiida-core@main#egg=aiida-core[atomic_tools]
+                uv pip install -e .[tests] aiida-core[atomic_tools]@git+https://github.com/aiidateam/aiida-core@main
 
         -   name: Run pytest
             id: tests
             env:
                 AIIDA_WARN_v3: 1
-            run: pytest -sv tests
```
danielhollas, line 59: The verbose output is excessive and it's hard to find the actual errors.

.github/workflows/ci.yml
```diff
@@ -49,26 +50,24 @@ jobs:
...
+
         -   name: Install Python dependencies
-            run: pip install -e .[tests]
+            run: uv pip install -e .[tests]
 
         -   name: Run pytest
             env:
                 AIIDA_WARN_v3: 1
-            run: pytest -sv tests
+            run: pytest tests
```
danielhollas, line 70: The verbose output is excessive and it's hard to find the actual errors. We made a similar change in aiida-core recently.

.github/workflows/nightly.yml
```diff
@@ -10,17 +10,21 @@ on:
...
 
-        if: github.repository == 'aiidateam/aiida-quantumespresso'  # Prevent running the builds on forks as well
+        # Don't run scheduled tests on forks, but allow them on PRs
+        if: github.repository == 'aiidateam/aiida-quantumespresso' || github.event_name != 'schedule'
         runs-on: ubuntu-latest
 
         strategy:
+            fail-fast: true
             matrix:
-                python-version: ['3.9', '3.10', '3.11', '3.12']
```
danielhollas, line 23: Not sure if you need to test all these in the nightly tests?

tests/cli/test_commands.py
```diff
@@ -1,46 +0,0 @@
-# -*- coding: utf-8 -*-
-"""Tests for CLI commands."""
-from __future__ import annotations
-
-import subprocess
-
-from aiida_pseudo.cli import cmd_root
```
sphuber, line 7: This was indeed a mistake, testing the `aiida_pseudo` cli. But instead of removing this entirely, I would simply keep it and actually test the CLI of this package, as intended:
```
from aiida_quantumespresso.cli import cmd_root
```

sphuber (changes requested):

tests/cli/test_commands.py
danielhollas: Done, (I didn't realize you have CLI even here :-D)

.github/workflows/nightly.yml
danielhollas: Hmm, after recent experience I am not sure about this, added `-v` back

danielhollas:
Hmm, I don't know where all these new failures come from, they seem unrelated to the changes here (and tests were passing yesterday!)

sphuber:
> Hmm, I don't know where all these new failures come from, they seem unrelated to the changes here (and tests were passing yesterday!)

All failures seem related to there no log messages having been captured by the `PwParser` tests. May have something to do with your removing of the `-sv` flag from `pytest`? Try reinstating those and see if that changes anything, just a guess for now

.github/workflows/nightly.yml
mbercx: I think just testing the lower/upper limit is fine indeed. 👍

mbercx:
> I don't have time to dig into this more, this has already been a distraction to me. :-D Hopefully somebody else can pick up these PRs (though I think both can be merged since the test failures seem unrelated).

Thanks @danielhollas! Appreciate the contribution, and apologies for leaving it open so long. I'm also a bit confused why the tests are failing, but let me investigate. 🔍

mbercx:
Removed the change in https://github.com/aiidateam/aiida-quantumespresso/pull/1083 here, and the xspectra code was still using `numpy.product`, which was removed in `numpy` v2, and installed by the nightly build.

mbercx (approved):
LGTM! Thanks again @danielhollas 🙏

PR #1083, opened 2025-05-26
Test aiida_quantumespresso.cli instead of aiida_pseudo.cli

danielhollas:
Split from #1082

### DIFF
--- a/tests/cli/test_commands.py
+++ b/tests/cli/test_commands.py
@@ -1,38 +1,43 @@
 # -*- coding: utf-8 -*-
+# pylint: disable=line-too-long
 """Tests for CLI commands."""
+
 from __future__ import annotations
 
 import subprocess
 
-import click
 import pytest
 
-from aiida_quantumespresso.cli import cmd_root
 
-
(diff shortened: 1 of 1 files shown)


mbercx:
Quite impressive how this correction in the `test_commands.py` module somehow breaks so many parser tests. 😅  I'm looking into it.

danielhollas:
> Quite impressive how this correction in the test_commands.py module somehow breaks so many parser tests. 😅 I'm looking into it.

I'll take that as a compliment. :smile:

mbercx:
@danielhollas this is quite a puzzle. I've managed to isolate the issue quite well, but I'm still not sure what's happening. In the end though, I already spent too much time trying to fix an issue with a fancy feature for running the basic `--help` tests for the CLI recursively. So I've just replaced the recursion function with an explicit list.

I've learned quite a bit about `pytest` on the way, and discovered potential issues with some of the AiiDA profile test fixtures (opening an issue soon). So all in all a morning well spent. 😂

I'll go ahead and merge this. That said, I left a little easter egg for future developers to play with. Feel free to have a crack at it and open a PR if you can solve it. 😉

PR #1087, opened 2025-07-01
PwParser: Add the XML schema for Quantum ESPRESSO v7.4.1

Minotakm:
This PR adds the xml schema for the Quantum Espresso v7.4.1 as well as  the tests for it. It uses the schema from #1075.


README.md
```diff
@@ -12,7 +12,7 @@ The matrix below assumes the user always install the latest patch release of the
 
 | Plugin | AiiDA | Python | Quantum ESPRESSO |
 |-|-|-|-|
-| `v4.8 < v5.0` | ![Compatibility for v4.0][AiiDA v4.0-pydantic2] |  [![PyPI pyversions][Python v3.9-v3.12]](https://pypi.org/project/aiida-quantumespresso/4.8.0/) | ![Quantum ESPRESSO compatibility][QE v6.6-7.4] |
```
mbercx, line 15: Unfortunately the badges aren't that simple, see

https://github.com/Minotakm/aiida-quantumespresso/tree/Add_support_QE_v7.4.1

You also have to update them at the end of the Markdown file. ^^

Also, this change would only be released in `aiida-quantumespresso` v4.11, so I think we need to add another line to the compatibility matrix, no?

mbercx: It will get bigger and bigger though... Maybe we should move a part to the documentation somewhere, and only leave the 4+ versions. I'll have a think.

mbercx (changes requested):
Thanks @Minotakm! Just a quick comment on the badge, and it seems the `precommit` is failing. Do you know how to run this locally? Let me know if you want to have a crack at fixing it or want me to take over.

Minotakm:
Thanks @mbercx I will fix them no worries.

Minotakm:
@mbercx I am getting a variety of errors with yapf not related to my PR. Maybe you can check as well.

mbercx:
I think one problem is that we're still using `yapf` 😅 . I'll have a look, no worries!

README.md
mbercx: Yup, that's also what I was thinking. ^^

bastonero:
Hey guys, good if we could merge this one and do a `4.11` release ASAP, as due to the old aiida-pseudo constraints we are blocking tests and packages deployment (e.g. aiida-hubbard). Cheers

README.md
mbercx: Opened https://github.com/aiidateam/aiida-quantumespresso/issues/1092 to track this, don't want to slow down this PR any further.

mbercx (approved):

mbercx:
Thanks again @Minotakm @pietrodelugas for the work1

PR #1096, opened 2025-07-15
Updated raw parser with newer format for Hubbard U standard output for pw.x calculation (QE> 7.2)

alberto-carta:


mbercx:
Thanks @alberto-carta! @bastonero as our resident Hubbard expert, can you have a look at this?

@alberto-carta I'd also like to add some tests for the new parser functionality. The setup is a bit tricky, so I will take care of it, but it would be very helpful to have a minimal example, i.e. a run for a small structure with few k-points etc, so we don't have to add a file with 100k lines. ^^

alberto-carta:
So for instance, this now returns to output a big dictionary where each entry looks like this

"2": {
        "magnetic_moment": 0.58571,
        "occupations": {
            "down": 3.93578,
            "total": 8.45726,
            "up": 4.52148
        },
        "spin_data": {
            "down": {
                "eigenvalues": [0.035, 0.968, 0.968, 0.982, 0.982],
                "eigenvectors": [
                    [-0.0, 0.976, -0.216, 0.004, -0.001],
                    [-0.577, -0.002, -0.003, 0.492, 0.651],
                    [-0.577, -0.001, 0.003, 0.318, -0.752],
                    [-0.0, 0.216, 0.976, 0.001, 0.004],
                    [0.577, -0.003, 0.001, 0.81, -0.101]
                ],
                "occupation_matrix": [
                    [0.968, 0.0, 0.0, -0.0, 0.0],
                    [0.0, 0.667, -0.316, 0.0, 0.316],
                    [0.0, -0.316, 0.667, -0.0, 0.316],
                    [-0.0, 0.0, -0.0, 0.968, 0.0],
                    [0.0, 0.316, 0.316, 0.0, 0.667]
                ]
            },
            "up": {
                "eigenvalues": [0.78, 0.78, 0.985, 0.985, 0.991],
                "eigenvectors": [
                    [0.963, -0.233, -0.133, 0.003, -0.0],
                    [0.075, 0.079, 0.386, -0.711, 0.577],
                    [0.031, -0.104, 0.423, 0.69, 0.577],
                    [0.233, 0.963, 0.003, 0.133, 0.0],
                    [0.105, -0.026, 0.809, -0.021, -0.577]
                ],
                "occupation_matrix": [
                    [0.784, -0.011, -0.011, -0.0, -0.022],
                    [-0.011, 0.985, 0.003, -0.019, -0.003],
                    [-0.011, 0.003, 0.985, 0.019, -0.003],
                    [-0.0, -0.019, 0.019, 0.784, -0.0],
                    [-0.022, -0.003, -0.003, -0.0, 0.985]
                ]
            }
        }
        
        
        This mimics the QE structure of the std.out, where "2" at the beginning refers to "ATOM 2"
        
        Can you specify how you want it? so if the atom is a nickel something like this?
        
        {
    "Ni2": {
        "magnetic_moment": 0.58571,
        "occupations": {
            "down": 3.93578,
            "total": 8.45726,
            "up": 4.52148
        },
        "occupation_matrices": {
            "down": {
                "occupation_matrix": [
                    [0.968, 0.0, 0.0, -0.0, 0.0],
                    [0.0, 0.667, -0.316, 0.0, 0.316],
                    [0.0, -0.316, 0.667, -0.0, 0.316],
                    [-0.0, 0.0, -0.0, 0.968, 0.0],
                    [0.0, 0.316, 0.316, 0.0, 0.667]
                ]
            },
            "up": {
                "occupation_matrix": [
                    [0.784, -0.011, -0.011, -0.0, -0.022],
                    [-0.011, 0.985, 0.003, -0.019, -0.003],
                    [-0.011, 0.003, 0.985, 0.019, -0.003],
                    [-0.0, -0.019, 0.019, 0.784, -0.0],
                    [-0.022, -0.003, -0.003, -0.0, 0.985]
                ]
            }
        }
    }
}


In the case of SOC you will also have off diagonal down-up part

bastonero:
Many thanks for the comments!

I agree that the functionality is broken and we should totally restore it. The problem with extending it is that the format of the outputs is a `Dict` object. This will store information directly in the databae, and will pollute and slow down queries in AiiDA. Therefore, my suggestion would be to keep it simple as it was before, avoiding to add extra information and think it through. For sure, we should totally not use the stdout, as major effort was done also on QE side to store the occupations in the XML file.

In passing, indices should be referred to python indices (i.e. related to StructureData), not fortran indices. I don't know if the old implementation was keeping fortran indices, but that would be bad as we should keep consistency with python and AiiDA, not with QE (see e.g. all the Hubbard implementation). 

> However, as a user, this approach for accessing fundamental outputs doesn't seem simple at all to me.

Could you be more specific? Maybe you mean "intuitive"? Length wise is as long as accessing a usual output. Ideally, this will be more powerfull and useful than returning a dictionary that than one needs to understand how to use with respect to a simple function (or more) that already does the job.

Of course, this should be accompanied with suitable documentation.

On the one hand, I understand your point on having the outputs ready where one should expect them. On the other hand, I think many other information that QE spits out are not "properly" stored and showed in the `.outputs`. So, I think we should balance whether it is worth it to duplicate the data.
For instance, I understand why one should also (re)store the trajectory, comprising energy, forces, ect, as you may want to _query_ for it. So probably the good question is: do we want to query for this piece of data?

Probably yes, but i would still avoid the way it is extended at the moment in this PR and devise e.g. a new `Data` class where we store the data as json file and dump it in the repository. This will also allow us to get access to occupations in a much better way. Like:
```python
pw_calc_node.outputs.occupations.asdict()
pw_calc_node.outputs.occupations.get_occupations(kind_name='Fe')
pw_calc_node.outputs.occupations.get_eigenvalues(kind_name='Fe')
pw_calc_node.outputs.occupations.get_eigenvectors(kind_name='Fe')
...
```

mbercx:
@bastonero @alberto-carta what's the status on this? Any way I can help get this merged?

mbercx:
Quick note: I'm close to starting to move parsing efforts to `qe-tools`. Ok if I move this functionality there?

PR #1097, opened 2025-07-21
Fixing parse_xml if no pseudo-wfcs are there in the pseudo file.

mikibonacci:
Previously, we always assume we have the `num_of_atomic_wfc` defined. However, if pseudo files do not contain info on these pseudo wavefunctions (like in the `sg15` case), this will give an error in the parsing. 

The error shows up if we use an enough recent version of Quantum ESPRESSO (> 7, but I have not tested all versions), because we use `parse_xml_post_6_2`.


t-reents (approved):
Thanks @mikibonacci! LGTM

mbercx (commented):
Thanks @mikibonacci! Just a quick question on the change.

mikibonacci:
Hi both, sorry for the late reply. Indeed, no wfcs available in the pp file. I can change to None

Inviato da Outlook per Android<https://aka.ms/AAb9ysg>
________________________________

PR #1102, opened 2025-07-30
🐛 `get_builder_from_protocol`: Fix bugs when overriding pseudopotentials

mbercx:
As mentioned in [this comment](https://github.com/aiidateam/aiida-quantumespresso/pull/1100?notification_referrer_id=NT_kwDOAYa_GLQxNzgzMTQ4NTEwNzoyNTYwNzk2MA#issuecomment-3130219768), https://github.com/aiidateam/aiida-quantumespresso/pull/1100 highlighted two issues with the current logic for using custom pseudo potentials through the `overrides` input of the `get_builder_from_protocol()` method:

1. If the user overrides the `pseudos` input, the `starting_magnetization` etc is still constructed taking the `z_valence` from the `pseudo_family`.
1. If the user overrides the `pseudos` input but doesn't override the cutoffs, these are still (silently) taken from the `pseudo_family`.

Here we try to fix these two issues. Moreover, if the user completely specifies the `pseudos` and cutoffs through the overrides, it should be fine in case `pseudo_family` is `None` or `False`. This is important in the context of adding custom pseudo potential functionality in the QEapp: https://github.com/aiidalab/aiidalab-qe/pull/1348.

### DIFF
--- a/src/aiida_quantumespresso/workflows/pw/base.py
+++ b/src/aiida_quantumespresso/workflows/pw/base.py
@@ -156,31 +156,47 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
 
         natoms = len(structure.sites)
 
-        try:
-            pseudo_set = (PseudoDojoFamily, SsspFamily, CutoffsPseudoPotentialFamily)
-            pseudo_family = orm.QueryBuilder().append(pseudo_set, filters={'label': pseudo_family}).one()[0]
-        except exceptions.NotExistent as exception:
-            raise ValueError(
-                f'required pseudo family `{pseudo_family}` is not installed. Please use `aiida-pseudo install` to'
-                'install it.'
-            ) from exception
-
-        try:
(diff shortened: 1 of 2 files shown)


mbercx:
@edan-bainglass I added specific tests to make sure the `pseudo_family` can still be set to `None`, as I'm assuming you do in the QEapp. Give it a spin and let me know in case you run into issues!

edan-bainglass:
@mbercx is there a reason why this needed a separate PR from #1100? As far as I can tell, it's ~90% the same changes 😅 + tests (which we discussed you would cover). I was under the impression you were just going to adjust #1100 to cover the extra bits.

mbercx:
Mainly to keep your branch + commits in place, along with the test runs. Happy to force push to your branch, but is there a reason why you'd prefer that?

t-reents (approved):

mbercx:
Thanks @t-reents for the review, and @edan-bainglass for the OG fix in https://github.com/aiidateam/aiida-quantumespresso/pull/1100! Will rebase + merge, close that PR and open a release PR momentarily!

PR #1108, opened 2025-08-22
Add num_unrecoverable_restart variable to allow extra run attempt

AndresOrtegaGuerrero:
Introduces a num_unrecoverable_restart variable that enables one additional
attempt to rerun in cases where a run would otherwise be marked as
unrecoverable. This helps to avoid problems with workflows failing due to sporadic node failure in the computer.
Needed for the QEApp , for long workflows.

### DIFF
--- a/tests/workflows/pw/test_base.py
+++ b/tests/workflows/pw/test_base.py
@@ -3,7 +3,7 @@
 """Tests for the `PwBaseWorkChain` class."""
 from aiida.common import AttributeDict
 from aiida.engine import ExitCode, ProcessHandlerReport
-from aiida.orm import Dict, Int
+from aiida.orm import Dict
 import pytest
 
 from aiida_quantumespresso.calculations.pw import PwCalculation


mbercx:
Interesting, looking at the code, I see your point:

https://github.com/aiidateam/aiida-quantumespresso/blob/c17831e1a2806057d03d22e2a3d5daf1d31f3f68/src/aiida_quantumespresso/workflows/pw/base.py#L414-L428

(I suppose you meant the `handle_unrecoverable_failure` process handler is triggered, not `handle_known_unrecoverable_failure`?)

This was added a long time ago, before the `BaseRestartWorkChain` was migrated to `aiida-core`: cab1a22d5c47028be35c3c9b16aa9c3f546df908.

I think we should remove this error handler, and deprecate the exit code. Then the functionality of the `BaseRestartWorkChain` should be triggered, i.e. in case there is an unhandled error (e.g. node failure), it will restart, but in case there are two consecutive unhandled failures the work chain will stop. Would that be sufficient for your purposes @AndresOrtegaGuerrero?

The `handle_known_unrecoverable_failure` should also be checked, that it never triggers at the moment. This one makes sense: if a know failure mode is triggered which we can't recover from, that we _know_ can't be fixed by a simple restart, we don't want to trigger the unhandled restart feature from the `BaseRestartWorkChain`.

AndresOrtegaGuerrero:
@mbercx  ,Yes, I meant 'unrecoverable handler.' In principle, what we'd like for the workflow is that in the case of a node failure, the BaseRestartWorkChain tries at least once. Since that behavior isn't present, I agree that it's part of aiida-core, so should I modify the PR and try to remove that handler?

AndresOrtegaGuerrero:
> > Since that behavior isn't present, I agree that it's part of aiida-core, so should I modify the PR and try to remove that handler?
> 
> Yes, I think that would be a cleaner solution. If you have the time, it would be great! We should keep the exit code, but "deprecate" it somehow, e.g. just add `[deprecated]` to the help? Open to suggestions here.

I did a new PR #1110 , let me know I can then close this PR and we continue discussing on the new one.

PR #1110, opened 2025-08-28
Remove unrecoverable handler

AndresOrtegaGuerrero:
Following the conversation with @mbercx , I've decided to create a fresh pull request. This new PR will remove the unrecoverable handler, allowing the BaseRestartWorkChain to attempt a single resubmission in the event of a node failure or an 'unrecoverable' error.

### DIFF
--- a/src/aiida_quantumespresso/workflows/bands/base.py
+++ b/src/aiida_quantumespresso/workflows/bands/base.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 """Workchain to run a Quantum ESPRESSO bands.x calculation with automated error handling and restarts."""
 from aiida.common import AttributeDict
-from aiida.engine import BaseRestartWorkChain, ProcessHandlerReport, process_handler, while_
+from aiida.engine import BaseRestartWorkChain, while_
 from aiida.plugins import CalculationFactory
 
 BandsCalculation = CalculationFactory('quantumespresso.bands')
--- a/src/aiida_quantumespresso/workflows/bands/base.py
+++ b/src/aiida_quantumespresso/workflows/bands/base.py
@@ -28,7 +28,7 @@ class BandsBaseWorkChain(BaseRestartWorkChain):
             cls.results,
         )
         spec.exit_code(300, 'ERROR_UNRECOVERABLE_FAILURE',
-            message='The calculation failed with an unrecoverable error.')
+            message='The calculation failed with an unrecoverable error. [deprecated]')
         # yapf: enable
 
     def setup(self):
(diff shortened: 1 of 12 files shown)


src/aiida_quantumespresso/workflows/bands/base.py
```diff
@@ -28,7 +28,7 @@ def define(cls, spec):
             cls.results,
         )
         spec.exit_code(300, 'ERROR_UNRECOVERABLE_FAILURE',
-            message='The calculation failed with an unrecoverable error.')
+            message='The calculation failed with an unrecoverable error. [deprecated]')
```
mbercx, line 31: I'd put the `[deprecated]` at the front of the message, so it's more clear which exit codes are deprecated when a user runs `verdi plugin list`.

src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -81,10 +81,10 @@ def define(cls, spec):
         spec.exit_code(211, 'ERROR_INVALID_INPUT_AUTOMATIC_PARALLELIZATION_UNRECOGNIZED_KEY',
             message='Unrecognized keys were specified for `automatic_parallelization`.'
                     'This exit status has been deprecated as the automatic parallellization feature was removed.')
-        spec.exit_code(300, 'ERROR_UNRECOVERABLE_FAILURE',
+        spec.exit_code(300, 'ERROR_UNRECOVERABLE_FAILURE [deprecated]',
```
mbercx, line 84: This should be in the message.

src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -81,10 +81,10 @@ def define(cls, spec):
...
             message='The calculation failed with an unidentified unrecoverable error.')
         spec.exit_code(310, 'ERROR_KNOWN_UNRECOVERABLE_FAILURE',
-            message='The calculation failed with a known unrecoverable error.')
+            message='The calculation failed with a known unrecoverable error. [deprecated]')
```
mbercx, line 87: This exit code is not deprecated, is it?

mbercx (commented):
Thanks @AndresOrtegaGuerrero! I fixed the pre-commit, but still have a few comments.

Some of the work chains now have no error handlers, but I think this is fine. The `BaseRestartWorkChain` will still fix some issues. I might repurpose the `ERROR_UNRECOVERABLE_FAILURE` exit code in the future to abort for failures where we _know_ a restart won't succeed, but for now I think the best way forward is to remove this handler.

src/aiida_quantumespresso/workflows/pw/base.py
AndresOrtegaGuerrero: That should be fine , known unrecoverable , means (at least to me) , there is a change or modification applied to avoid the error and those will still be active no ?

AndresOrtegaGuerrero: Yeah, you are right (There were too many of them :D )

AndresOrtegaGuerrero:
> Thanks @AndresOrtegaGuerrero! I fixed the pre-commit, but still have a few comments.
> 
> Some of the work chains now have no error handlers, but I think this is fine. The `BaseRestartWorkChain` will still fix some issues. I might repurpose the `ERROR_UNRECOVERABLE_FAILURE` exit code in the future to abort for failures where we _know_ a restart won't succeed, but for now I think the best way forward is to remove this handler.

A bit odd since I ran it locally, there must be something that doesnt match locally and the CI

mbercx (dismissed):
LGTM! 🚀 Thanks again @AndresOrtegaGuerrero 🙏

mbercx:
Oh, wait. I just noticed the `PhBaseWorkChain` _does_ still use the `ERROR_UNRECOVERABLE_FAILURE`, in case it runs out of solutions for the cholesky errors (`ERROR_COMPUTING_CHOLESKY`). Let me fix that.

PR #1114, opened 2025-09-01
Feature/projwfc : Add support for LDOS parsing with .tdosinboxes. and total PDOS with new Pdos node

ValkScripter:
This pull request introduces significant improvements to the `ProjwfcCalculation` and its parser, enhancing support for different types of projections and improving output handling. The changes add new output nodes for LDOS and PDOS, implement parsing logic for projections onto boxes, and introduce new exit codes for error handling. The codebase also receives formatting and clarity improvements.

### Enhanced output handling and support for LDOS/PDOS projections

* Added new output nodes to `ProjwfcCalculation` for LDOS (`Ldos`) and PDOS (`Pdos`), including conditions for their generation and improved help descriptions in the process specification.
* Updated `_retrieve_temporary_list` to include the `ldos_boxes` file, ensuring it is available for parsing when box projections are used.
* Implemented parsing logic in the parser to handle both atomic orbital and box projections, including reading from `ldos_boxes` and setting the appropriate output nodes (`Dos`, `Pdos`, `Ldos`).
* Leaves untouched the previously coded section to parse projections on atomic orbitals, except adding a new node for the total PDOS (`Pdos`) which is the sum of all projections (either on orbitals or on boxes).

### Improved error handling

* Added new exit codes for missing or unreadable `pdos_tot` and `ldos_boxes` files, and for unknown projection types in the output, improving robustness and clarity of error reporting.

### Code formatting and clarity improvements

* Standardized string quoting throughout the codebase for consistency and readability.
* Refactored and clarified regular expressions and key assignments in orbital and band parsing logic, improving maintainability and correctness.

### Parser interface and method improvements

* Refactored parser methods for clarity, including argument formatting, static method declarations, and error handling returns.

---

Most important changes:  
- Enhanced output handling and support for LDOS/PDOS projections
- Improved error handling with new exit codes
- Code formatting and clarity improvements 
- Parser interface and method improvements


mbercx:
Thanks a lot @ValkScripter, this looks great! 🙏 

Since there are quite a few changes, I'm going to need a bit more time to review. Will try to find some time next week to do so.

mbercx:
Alright, coming back to this one now! Before I start a complete review, I have two questions:

1. Would it be a lot of work to regenerate the testing fixtures using runs with fewer kpoints? We want to keep the repository as light as possible, and don't need this many kpoints to test the parsing functionality.
2. I see there are quite a few changes related to formatting, mostly changing quotes and reducing line length. Note that we have pre-commit set up to deal with such formatting issues and keep consistency in the code. The changes here actually cause the precommit to fail (see [here](https://github.com/aiidateam/aiida-quantumespresso/actions/runs/17384414324/job/49655271100?pr=1114)). Also note that [we use 120 for our max line length](https://github.com/aiidateam/aiida-quantumespresso/blob/cd0190ef685f0056628a5e705f3c223a34628700/pyproject.toml#L157).

   Not a hard and fast rule, but in general, I try to avoid mixing logic and formatting changes too much in a single PR. Formatting changes can lead to a lot of lines to review, and then I have to double check for every change if something changed in the logic as well.

   I've fixed the precommit and squashed all the commits into [a single one in my fork](https://github.com/mbercx/aiida-quantumespresso/commit/2ccaec6144822cf0044a2aaaaf9b11abfb4f30cf). Would it be ok if I update your PR? We can still see how to best split up the commits after review.

ValkScripter:
Hi @mbercx ! I indeed forgot about the pre-commit tool which is pretty useful in this case. Sorry for the amount of lines to review.

1/ I can regenerate the fixtures with fewer kpoints and energy steps to keep them light.

2/ Sure, go ahead! Thanks a lot for fixing the precommit and cleaning things up. We can definitely revisit the commit structure after review.

mbercx (changes requested):
Thanks a lot again @ValkScripter, I'm very happy with most of the changes, great to expand the functionality of the `ProjwfParser`!

> Sorry for the amount of lines to review.

No worries at all, I appreciate the contribution! 

I've left some questions and small fixes. Will try to add a commit myself that validates that the user only tries to run with `tdosinboxes` for `nspin` = 1 or 2.

I managed to add the validator just fine, but realised another issue, see https://github.com/aiidateam/aiida-quantumespresso/issues/1141. So let's merge this _without_ the validation, and come back to that another time. Here's the validator I was adding for future reference:

```python
    def validate_inputs(cls, value, _):
        from aiida_quantumespresso.utils.defaults.calculation import pw as pw_defaults
        from aiida_quantumespresso.utils.defaults.calculation import projwfc as projwfc_defaults
        parent_folder_creator = value['parent_folder'].creator

        if parent_folder_creator is not None:
            nspin = parent_folder_creator.inputs.parameters.get('SYSTEM', {}).get('nspin', pw_defaults.nspin)
            tdosinboxes = value.get('parameters', {}).get('PROJWFC', {}).get('tdosinboxes', projwfc_defaults.tdosinboxes)

            if nspin == 4 and tdosinboxes is True:
                return f'Computing the local DOS '
```

with the following addition to `aiida_quantumespresso.utils.defaults.calculation.__init__` (which should also be moved to `qe-tools`.... So much work to do 😓 ).

```python
projwfc = AttributeDict({
    'tdosinboxes': False,
})
```

mbercx:
Good to know @AndresOrtegaGuerrero! I'll prioritise this then, I should be able to generate some new texts fixtures myself, @ValkScripter has done his part. :)

ValkScripter:
> @ValkScripter do you have time to work on this? I can definitely integrate my comments, but having smaller fixture files would be very helpful. 🙏
> 
> I'm planning to update our devops stack soon, which will also mean (_finally_) switching to Ruff. That might introduce quite a few merge conflicts in this PR.

Hi @mbercx ! I see there are a lot of corrections / modifications to perform. 
I sadly do not have much time at the moment :cry: 
That being said I just pushed the new fixtures and resulting .yml files from the tests. 
I hope they are better and light enough.
As soon as I have more time to get back onto it I will.

ValkScripter:
I really don't mind, please feel free !

mbercx:
@ValkScripter I've integrated my suggestions. Please have a final look and let me know if I can merge!

ValkScripter:
Hi @mbercx, all changes look very good to me.

Just to clarify something here:
https://github.com/aiidateam/aiida-quantumespresso/pull/1114#discussion_r2356898310
> Just to make sure I understand: `projwfc.x` can do three types of projections:
> 
> 1. orbitals (default)
> 2. volumes (what you are adding here)
> 3. k-resolved DOS (still unsupported)
> 
> And so the only way this error can happen is in case [3]? But that option is blocked by our (still rather user-unfriendly) `ProjwfcCalculation._blocked_keywords`, right? So can this exit code ever happen?

Actually there are only 2 types of projections (as far as I know):

1. on orbitals (default behavior)
2. onto boxes (the volumes)

The error was added in case there is another projection type I am not aware of.

The `kresolveddos` option makes the DOS resolved in terms of k-points instead of energy points (k-point weights all being the same).
This allows to plot the contribution of the DOS, PDOS, and projections inside the Brillouin Zone for example instead of a standard DOS vs. Energy graph.
It can be done along k-paths also allowing to plot bandstructures "projected" onto orbitals (common thing) and maybe even onto boxes (just thought about it but never tried).

I actually have already implemented that functionality long time ago along with `tdosinboxes`.
It was before the removal of the "floating" dictionary to get all parsing parameters, so I have to adapt my code and make it compatible with the new file architecture to allow this functionality (one which I really want to be able to use with the latest features since then).

mbercx:
Thanks a lot for the explanation @ValkScripter! I think the best approach is to raise a warning with `warnings.warn`. I understand your initial inclination to create an exit code, but if e.g. the stdout is incomplete, the returned exit code would indicate the projection type is unknown, whereas actually the stdout simply doesn't contain the string required to detect it.

I'll go ahead and merge this PR with that change. Thanks again for the work, looking forward to reviewing a PR for `kresolveddos`. :)

mbercx (approved):

PR #1125, opened 2025-09-12
Add how-to documentation for `PwRelaxWorkChain`

Minotakm:
This PR adds documentation for the `PwRelaxWorkChain`.

### DIFF
--- a/docs/source/howto/workflows/pw/relax.md
+++ b/docs/source/howto/workflows/pw/relax.md
@@ -37,7 +37,7 @@ builder.base_relax.pw.parameters['SYSTEM']['ecutwfc'] = 60.0
 builder.base_relax.pw.parameters['SYSTEM']['ecutrho'] = 300.0
 
 # Submit the work chain
-node = submit(builder)
+workchain_node = submit(builder)
 print(f"Launched {workchain_node.process_label} with PK = {workchain_node.pk}")
 ```
 
--- a/docs/source/howto/workflows/pw/relax.md
+++ b/docs/source/howto/workflows/pw/relax.md
@@ -38,7 +38,7 @@ builder.base_relax.pw.parameters['SYSTEM']['ecutrho'] = 300.0
 
 # Submit the work chain
 node = submit(builder)
-print(f"Launched {node.process_label}<{node.pk}>")
+print(f"Launched {workchain_node.process_label} with PK = {workchain_node.pk}")
 ```
 
 ---
--- a/docs/source/howto/workflows/pw/relax.md
+++ b/docs/source/howto/workflows/pw/relax.md
@@ -28,17 +28,13 @@ builder = PwRelaxWorkChain.get_builder_from_protocol(
     structure=structure,
     protocol="fast",   # choose from: fast, moderate, precise
     relax_type=RelaxType.POSITIONS_CELL,  # relax both positions and cell
+    options={"resources": {"num_machines": 2},
+             "max_wallclock_seconds": 7200}
 )
 
 # Optionally adjust calculation parameters
 builder.base_relax.pw.parameters['SYSTEM']['ecutwfc'] = 60.0
 builder.base_relax.pw.parameters['SYSTEM']['ecutrho'] = 300.0
-
-# Optionally adjust resources
-builder.base_relax.pw.metadata.options = {
-    "resources": {"num_machines": 2},
-    "max_wallclock_seconds": 7200,
-}
 
 # Submit the work chain
 node = submit(builder)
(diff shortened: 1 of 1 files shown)


mbercx:
Thanks @Minotakm! Note that I already had an open PR for the `PwRelaxWorkChain` how-to: https://github.com/aiidateam/aiida-quantumespresso/pull/934 😅

We can already see a bit of difference in style. I agree though that for the how-tos can start with a single code snippet that submits a calculation. I would still perhaps add the section on the `RelaxType`, so users know what they can select here?

docs/source/howto/workflows/pw/relax.md
```diff
@@ -2,4 +2,73 @@
...
+
+## Inspecting results
+
+After the work chain finishes, the main outputs are:
+
+* `output_structure`: the relaxed structure.
+* `output_parameters`: a dictionary with key results from Quantum ESPRESSO (total energy, Fermi energy, etc.)
+
+```python
+from aiida.orm import load_node
```
mbercx, line 54: You've already imported `orm` above. Why not just do `orm.load_node`?

docs/source/howto/workflows/pw/relax.md
```diff
@@ -2,4 +2,73 @@
...
+    options={"resources": {"num_machines": 2},
+             "max_wallclock_seconds": 7200}
+)
+
+# Optionally adjust calculation parameters
+builder.base_relax.pw.parameters['SYSTEM']['ecutwfc'] = 60.0
+builder.base_relax.pw.parameters['SYSTEM']['ecutrho'] = 300.0
+
+# Submit the work chain
+node = submit(builder)
```
mbercx, line 40: I'd adapt the variable name to make it clear which node it is.

```suggestion
workchain_node = submit(builder)
```

docs/source/howto/workflows/pw/relax.md
```diff
@@ -2,4 +2,73 @@
...
+print(f"Launched {node.process_label}<{node.pk}>")
```
mbercx, line 41: Maybe add `PK` to the `print` so it's crystal clear what the number is.

```suggestion
print(f"Launched {workchain_node.process_label} with PK = {workchain_node.pk}")
```

docs/source/howto/workflows/pw/relax.md
```diff
@@ -2,4 +2,73 @@
...
+* `output_parameters`: a dictionary with key results from Quantum ESPRESSO (total energy, Fermi energy, etc.)
+
+```python
+from aiida.orm import load_node
+
+wc = load_node(node.pk)
+print("Work chain finished ok:", wc.is_finished_ok)
+
+# Get the relaxed structure
+if "output_structure" in wc.outputs:
```
mbercx, line 60: Why have the `if` statement here but not for the `output_parameters`? I understand you want to avoid having an error when the work chain is not `finished_ok`, but then this cell will still raise when you try to get the `outputs_parameters`.

This is of course why I initially used `run` in the `PwBaseWorkChain`, and suggested `wait = True` in [my comment here](https://github.com/aiidateam/aiida-quantumespresso/pull/1074#issuecomment-3277523544). But I think for the How-to we can expect the reader to be competent enough to understand that in case the workflow is not completed successfully, the outputs might be non-existent. So I'd vote to remove the `if`.

docs/source/howto/workflows/pw/relax.md
```diff
@@ -2,4 +2,73 @@
...
+
+After the work chain finishes, the main outputs are:
+
+* `output_structure`: the relaxed structure.
+* `output_parameters`: a dictionary with key results from Quantum ESPRESSO (total energy, Fermi energy, etc.)
+
+```python
+from aiida.orm import load_node
+
+wc = load_node(node.pk)
```
mbercx, line 56: This one looks a little strange to me. I understand you want to:

1. Show the user they can load the node from the PK
2. Avoid that they have to copy the PK when running these code snippets e.g. in a notebook.

But in case the `node` variable is still active, the line is not necessary. In case it's not, it will still fail. I think we can assume that a reader of the how-to has enough AiiDA competency to know how to load a node, and we nicely print the PK above. So I would just directly use `workchain_node`?

mbercx (changes requested):
Thanks again @Minotakm! I left some nitpicks, but I agree we shouldn't be bike-shedding too much. Feel free to comment/adapt how you see fit, then I'll merge and we can get a few example documentation pieces in place to see what we think is best.

docs/source/howto/workflows/pw/relax.md
Minotakm: Sure, I assumed you first run the first snippet which prints the pk of the node. Then you load it and do the rest of the how-to. But sure I will modify it and assume that everything is done in one go.

mbercx (approved):
LGTM! 🚀 Let's get this merged, add a few more pages, and then get a feel for if we want to still make changes.

Thanks again @Minotakm!

Issue #1122, opened 2025-09-07
‼ Remove calculation namespace in `BaseRestartWorkChain`s


mbercx:
All of the `BaseRestartWorkChain`s we have implemented expose the inputs of the process class they wrap into a new namespace, e.g.:

https://github.com/aiidateam/aiida-quantumespresso/blob/830aef35445e8c6d596e54651c9a891ce6b4b77e/src/aiida_quantumespresso/workflows/pw/base.py#L39-L44

However, this adds an extra layer of depth to the already often complicated inputs of our workflows. I vote to expose the inputs of these work chains at the top level.

To me, the `PwBaseWorkChain` should be as close as possible to a drop-in replacement for the `PwCalculation`, with some additional benefits:

* Automatic k-points mesh generation via `kpoints_distance`.
* Error handling
* Protocols (although we may choose to implement https://github.com/aiidateam/aiida-quantumespresso/issues/947 instead).

---

Some other notes (may be updated):

* We (thankfully) already expose the outputs directly on the top level, so why not for the inputs? This would also make our API more consistent in that regard.

mbercx:
Opinions welcome @t-reents @Minotakm @bastonero @mikibonacci @npaulish! Obviously a _very_ breaking change, but I think it's worth it.

bastonero:
This is going to be _really_ painfull, but indeed maybe worth it?

mbercx:
Good suggestion @bastonero! I think we should do so indeed, just ignoring inputs that aren't used is terrible UX. I'll see if I can find some time to work on this ASAP, before pushing the breaking change here.

mbercx:
@dbidoggia made [a good remark here](https://github.com/aiidateam/aiida-quantumespresso/pull/1106#issuecomment-3279009233) that just directly exposing the input namespace would create a conflict for the `metadata` namespace. I've implemented a workaround in https://github.com/aiidateam/aiida-quantumespresso/pull/1126, which can serve as an example implementation for removing the `pw` namespace in the `PwBaseWorkChain`.

[...]

PR #1126, opened 2025-09-14
‼️ Remove `pw` namespace

mbercx:


mbercx:
The tests here will most assuredly fail. The PR is still a work in progress, mainly a demonstrator for https://github.com/aiidateam/aiida-quantumespresso/issues/1122.

mbercx:
As noted [here](https://github.com/aiidateam/aiida-quantumespresso/issues/1122#issuecomment-3290843855), I've passed the `metadata.disable_cache` input down to the `PwCalculation`. I would argue that in case a user sets `disable_cache` to `True` on the `PwBaseWorkChain`, they don't want to use caching in the `PwCalculation`.

PR #1133, opened 2025-09-15
Add NebCiWorkChain for climbing image NEB calculations

dbidoggia:
Dear all, I have created a `NebCiWorkChain` to first perform a no-CI NEB calculation followed by a refinement with climbing image using either automatic or manual method.


dbidoggia:
Dear @mbercx, yes, absolutly. I tried everything, but could not understand why those commits still apear there

mbercx:
I think you must have branched off before pulling the latest changes in `main`. But thankfully a rebase on the `aiidateam` `main` branch seems to work like a charm:

```console
❯ git rebase origin/main
warning: skipped previously applied commit 7b505a2
warning: skipped previously applied commit 1b7a063
warning: skipped previously applied commit 3b6b7c4
hint: use --reapply-cherry-picks to include skipped commits
hint: Disable this message with "git config advice.skippedCherryPicks false"
Successfully rebased and updated refs/heads/pr/dbidoggia/1133.
```

Let me force push your `ci-neb` branch and see where we're at.

mbercx:
And all is well! (besides the RTD build, but I'll fix that later ^^) Thanks again for the contribution @dbidoggia. My main question before reviewing is: how stable is this workflow? As in: do you still expect to adapt it in the near future? I'm happy to add the workflow to `aiida-quantumespresso`, but that means once it's released we have to preserve backwards-compatibility, or deal with deprecations etc. Another approach we've been taking is to put more advanced workflows in separate packages (e.g. https://github.com/aiidateam/aiida-hubbard), which are still pre v1.0 and hence don't need to promise backwards compatibility between minor releases.

Something to consider, but again: also happy to review and add it to `aiida-quantumespresso`!

dbidoggia:
Thanks @mbercx, sooner or later I’ll learn how to avoid disasters on git…
This is a very basic WorkChain for now. Of course, it could be extended by adding checks on subworkchain outputs and refining the logic accordingly, but for our purposes, the current implementation is sufficient and I'm not planning to refine it deeper.

That said, I’ll leave the decision to you, since you probably know (or can better guess) the future perspectives coming from other contributions

PR #1135, opened 2025-09-16
📚 Update "getting started" process

### LINKED ISSUE #1132: 📚 Add installation + quickstart
Note: This is a piece of documentation written from @t-reents I split off from https://github.com/aiidateam/aiida-quantumespresso/pull/1011 that may serve as a starting point.

````
The AiiDA Quantum ESPRESSO plugin comes with a set of tested protocols. These protocols define default input parameters for different levels of precision. This concept enables the initialization of a pre-populated `builder` for most of the different `WorkChain`s, which can be obtained by calling the `get_builder_from_protocol` method.

The following example shows the working principle for the `PwBaseWorkChain`:

```
# code and structure are the mandatory input arguments

    code=pw_code, # load your pw-code
    structure=structure, # provide your input structure here
)

```
We refer to the documentation of the {{ get_builder_from_protocol }} method for further details about additional default input arguments.
````

Some other [notes from @bastonero](https://github.com/aiidateam/aiida-quantumespresso/issues/1107#issuecomment-3254156397) on the topic:

* using the builder from protocol with no modifications
* filling the builder manually
* using a dictionary inputs (i.e. no builder) and submitting it (like in workchain developments)
* using the builder from protocol with applied changes

and i would try to make clear (as @GeigerJ2 pointed out) how the nested inputs work. it is always not clear to users the concept of namespace and exposed inputs, so they don't get many times the structure of the inputs.

### PR DESCRIPTION
**Pull request by mbercx**
Fixes #1132 

Ok, this one is more or less ready. There are quite a few changes, so let's first see if we agree on the new structure before starting an in-depth review.

## Main Changes

 On the landing page, guide the users to the installation page and the (new) quick start.

### Installation

With `verdi presto`, things here can be much simplified. The basic installation now describes three things:

1. Install the package, no need to discuss install `aiida-core`. Then set up a profile with `verdi presto`, which also sets up the `localhost` computer.
2. Set up the `pw.x` code. Here I still want to make this simpler with https://github.com/aiidateam/aiida-quantumespresso/pull/1124
3. Installing the pseudopotentials with `aiida-pseudo`. 

I've moved the CLI tab-completion to an "Advanced", I think it should really not be a priority for most users. To keep users on point, I aded a box to guide them to the new Quick start.

### Quick start

The first place where we want users to start learning how to use the package.

1. Instead of starting by using the CLI, we show them how to run a `pw.x` calculation with the protocols via the Python API. I decided on this, because this is how I _want_ beginners to use the package. Here I still use `run`, to keep things simple and robust.
2. I think one of the first things users will want to do is adapt the inputs. I show them here in order of (my) preference: overrides, adapting builder, empty builder, passing dictionary to `run`. For the last two: I would frankly move them elsewhere. Showing 4 ways of setting inputs is unnecessarily overwhelming, even if I tried to keep it succinct.
3. The last thing I show is how to `submit` the process instead of run it. For _anything_ real they want to run, they have to understand this. By using `submit`, it also becomes natural to discuss the daemon, and how to inspect processes.

---

That's it! Then we can start pointing them to other tutorials/how tos etc. We can still recover how to run with the CLI, and an explanation of setting the inputs steps by step. But I want the quick start to avoid too much complexity and show people how to run workflows with the protocols for a structure.

### CODE REVIEW

File: docs/source/get_started/quick_start.md [hunk 1]
```diff
@@ -0,0 +1,272 @@
...
+builder = PwBaseWorkChain.get_builder_from_protocol(
+    code=code,
+    structure=structure,
+    protocol='fast',
+    overrides=overrides
+)
+builder
+```
+
+:::{warning}
```
mbercx, line 96: This warning can be removed after we merge https://github.com/aiidateam/aiida-quantumespresso/pull/1136

docs/source/get_started/quick_start.md
```diff
@@ -0,0 +1,272 @@
...
+
+Alternatively, you can also adapt the inputs on the `builder` afterwards:
+
+```python
+builder.pw.parameters['SYSTEM']['nbnd'] = 15
+builder
+
+```
+
+::::{note}
```
mbercx, line 111: As mentioned in the OP, I would move this information elsewhere. We can write a more dedicate tutorial/how to on "setting inputs".

docs/source/get_started/quick_start.md
```diff
@@ -0,0 +1,272 @@
...
+Let's see if it's running:
+
+:::{margin}
+**Note**: Lines starting with `!` aren’t Python.
+This is an [IPython/Jupyter shell escape](https://ipython.readthedocs.io/en/stable/interactive/reference.html#system-shell-access) — the command after `!` is run in your system shell (e.g. `bash` or `zsh`), not in Python itself.
+Alternatively, you can open a terminal and run these commands there (minus the `!`).
+:::
+
+```
+!verdi daemon status
```
mbercx, line 207: The idea is that this tutorial can be downloaded as a Jupyter notebook, in `.ipynb` format. Unfortunately, our somewhat outdated documentation setup doesn't offer that option. I'll make sure to fix this after we merge our first round of documentation PRs.

docs/source/get_started/installation.md
```diff
@@ -0,0 +1,155 @@
...
+
+```console
+$ verdi presto
+```
+
+(installation-setup-code)=
+
+## Basic setup
+
+### `pw.x` code
```
mbercx, line 29: With @t-reents's new sexy command in https://github.com/aiidateam/aiida-quantumespresso/pull/1124, this can be simplified. I would also _only_ show the CLI here, and move the Python API to the "advanced" section below.

---

## Draft new section

To run a Quantum ESPRESSO code, it should first be setup in AiiDA.
This can be done easily with the `aiida-quantumespresso` command line interface (CLI).
Here we will setup the `pw.x` code that is installed on the computer where AiiDA is running (`localhost`):

```console
$ aiida-quantumespresso setup codes localhost pw.x
```

:::{important}
The command will look for the first `pw.x` binary your `PATH`.
You can find out the absolute path to this binary using the `which` UNIX command:

```console
which pw.x
```

If this returns nothing or a Quantum ESPRESSO version you don't want to run, pass the directory of the correct absolute path as the filepath executable.

```console
$ aiida-quantumespresso setup codes --directory /path/to/bin localhost pw.x
```
:::

---

I'd also point to our own custom "setup" section (once we write it ^^). Finally, I'm wondering if we should set up more than one binary here now that we can. Perhaps all binaries we need for the `pw.x` workflows (`pw.x`, `dos.x`, `projwfc.x`)? Then they can see the command sets up multiple codes, and they'll have them for the `PdosWorkChain`.

mbercx:
> Thanks, Marnik! @afiorent noticed a typo in the ph.x pag

Thanks! Fixed in https://github.com/aiidateam/aiida-quantumespresso/pull/1150. Any comments on the changes in this PR?

docs/source/get_started/quick_start.md
```diff
@@ -0,0 +1,272 @@
...
+workchain_node.is_finished_ok
+```
+
+:::{tip}
+Restarted your kernel and lost your variables?
+Don’t worry — your data is still in the AiiDA database.
+You can load a node back into a variable with its UUID or PK:
+
+```python
+workchain_node = orm.load_node(<PK>)  # Replace <PK> here!
```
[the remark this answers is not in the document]
t-reents: But I still think that it's good to show it once in such a quick start

PR #1137, opened 2025-09-17
🔧 `.gitignore`: Add `.jupytext-sync-ipynb`

mbercx:
To write the documentation, it would be ideal to be able to just write a Jupyter notebook, so you can just run the cells locally and test the content. However, the `.ipynb` format is JSON, which is very cumbersome and leads to large and hard to read diffs.

Jupytext sync is a tool for writing notebooks in `.ipynb` format, and syncing them to a `.md` file, see:

https://marketplace.visualstudio.com/items?itemName=caenrigen.jupytext-sync

The Markdown format is also useable by MyST, our preferred documentation tool. Here we add the directory that Jupytext sync creates to the `.gitignore` to avoid committing the `.ipynb` files it generates.


mbercx:
This is only a minor change to the `.gitignore`, and quite essential for my current approach to writing documentation, so I'm going to merge this with admin powers. ^^

Once the checks pass, of course. I'm not an animal.

PR #1149, opened 2025-09-19
`PwRelaxWorkChain`: Fix meta-convergence

t-reents:
@mbercx Just copying the commit message here to explain the problems and solutions:

This PR fixes two issues related to the meta-convergence of the `PwRelaxWorkChain`.

* In the current version, we check whether the maximum number of iterations has been reached in the `should_run_relax` method. If yes, we return an exit code. However, we don't check whether the last iteration was actually successful and converged with respect to the meta-convergence. To also cover those cases, the `self.ctx.converged` variable is (re)introduced.

* Related to the previous aspect, we currently return an exit code within the `should_run_relax` method in case the maximum number of meta-convergence iterations was reached. However, this method is used as the condition of the `While_` loop and the `plumpy` implementation only supports boolean return values
(https://github.com/aiidateam/plumpy/blob/7df19a1bc5e59c8b9e67616ac9309145966ae867/src/plumpy/workchains.py#L401-L414). Hence, the current version could get stuck in infinite loops (e.g., the k-mesh keeps changing), as the returned exit code is not respected. This was fixed by changing the position of this statement and moving it to the `results` method

### DIFF
--- a/src/aiida_quantumespresso/workflows/pw/relax.py
+++ b/src/aiida_quantumespresso/workflows/pw/relax.py
@@ -200,6 +200,8 @@ class PwRelaxWorkChain(ProtocolMixin, WorkChain):
                 'No change in volume possible for the provided base input parameters. Meta convergence is turned off.'
             )
             self.ctx.meta_convergence = False
+        # Already converged if volume can not change
+        self.ctx.converged = volume_cannot_change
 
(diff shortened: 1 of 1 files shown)


mbercx (approved):
Thanks a lot @t-reents! These changes look good to me, good to have this fixed. 🚀

PR #1151, opened 2025-09-22
Revisited the logic for the number of bands determination, in the case of spin-orbit calculations.

Minotakm:
In band-structure calculations with spin–orbit coupling, each band can hold only one electron, so the current logic is insufficient. This PR revises the band-counting logic.

### DIFF
--- a/src/aiida_quantumespresso/workflows/pw/bands.py
+++ b/src/aiida_quantumespresso/workflows/pw/bands.py
@@ -267,7 +267,6 @@ class PwBandsWorkChain(ProtocolMixin, WorkChain):
         inputs.pw.parameters.setdefault('SYSTEM', {})
         inputs.pw.parameters.setdefault('ELECTRONS', {})
 
-        is_soc = inputs.pw.parameters['SYSTEM'].get('lspinorb', False)
         electrons_per_band = 1 if inputs.pw.parameters['SYSTEM'].get('noncolin', False) else 2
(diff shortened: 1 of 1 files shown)


src/aiida_quantumespresso/workflows/pw/bands.py
```diff
@@ -267,13 +267,15 @@ def run_bands(self):
         inputs.pw.parameters.setdefault('SYSTEM', {})
         inputs.pw.parameters.setdefault('ELECTRONS', {})
 
+        is_soc = inputs.pw.parameters['SYSTEM'].get('lspinorb', False)
+        electrons_per_band = 1 if is_soc else 2
```
mbercx, line 271:
```suggestion
        electrons_per_band = 1 if inputs.pw.parameters['SYSTEM'].get('noncolin', False) else 2
```

src/aiida_quantumespresso/workflows/pw/bands.py
```diff
@@ -267,13 +267,15 @@ def run_bands(self):
         inputs.pw.parameters.setdefault('SYSTEM', {})
         inputs.pw.parameters.setdefault('ELECTRONS', {})
 
+        is_soc = inputs.pw.parameters['SYSTEM'].get('lspinorb', False)
```
mbercx, line 270:
```suggestion
```

mbercx (changes requested):

mbercx (approved):

PR #1155, opened 2025-09-23
Feature/constraints

cpignedoli:
This is the first draft allowing for the constraint "distance"
It forces the key ion_dynamics 'fire' in the 'relax' calculation.
Still, I think we should warn/enforce no_symm.
Following this line, I could add other constraints (angle and dihedral); @mbercx what do you think?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,9 @@ name: ci
 
 on: [push, pull_request]
 
+env:
+    FORCE_COLOR: 1
+
 jobs:
 
     pre-commit:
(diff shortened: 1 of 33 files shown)


src/aiida_quantumespresso/calculations/__init__.py
```diff
@@ -709,6 +717,17 @@ def _generate_PWCPinputdata(cls, parameters, settings, pseudos, structure, kpoin
             kpoints_card = ''.join(kpoints_card_list)
             del kpoints_card_list
 
+        # ============ I prepare the CONSTRAINTS card =============
+        constraints = settings.pop('CONSTRAINTS', None)
+        if constraints is not None:
+            input_params['IONS']['ion_dynamics'] = 'fire'
```
mbercx, line 723: Is this necessary? If so, I would validate this in `validate_inputs`. We can also set this there if it's more user-friendly, and raise a warning.

mbercx:
> I think some changes and check should be done in PwRelaxWorkChain (src/aiida_quantumespresso/workflows/pw/relax.py), since in principle the CONSTRAINTS are applied to the relaxation workflow, depending on the RelaxType (src/aiida_quantumespresso/common/types.py)

Can you elaborate on the changes/checks you think are required? I'm not very familiar with `CONSTRAINTS`, so I'm trying to understand.

PR #1162, opened 2025-09-26
👌 `setup_codes`: warn when executable does not exist with `directory`

mbercx:
When using the `directory` input of the `setup_codes` tool, the command will raise a `FileNotFoundError` when the constructed `<directory>/<executable>` path does not exist. This means that in case the user relies on another approach for making the executable accessible than the prepend text, they are unable to set up their codes with `setup_codes` and the corresponding new CLI command.

Here we adapt the `get_executable_paths` function to raise a warning instead of an error when the `<directory>/<executable>` path does not exist.


t-reents (approved):
LGTM 🚀 Thanks @mbercx

t-reents (approved):
Still LGTM! Thanks for taking care of both. The computer prepend text makes indeed a lot of sense.

Enjoy your Feierabend! :beers:

src/aiida_quantumespresso/tools/setup.py
```diff
@@ -118,7 +119,7 @@ def get_executable_paths(
                 exec_path = (directory / executable).as_posix()
 
                 if not transport.path_exists(exec_path):
-                    raise FileNotFoundError(f'Executable<{exec_path}> does not exist on computer<{computer.label}>')
+                    warnings.warn(f'Could not find Executable<{exec_path}> on computer<{computer.label}>')
```
t-reents, line 122:
```suggestion
                    warnings.warn(f'Could not find executable<{exec_path}> on computer<{computer.label}>')
```

t-reents (approved):

mbercx:
> I think we made it work 😄

Post-Feierabend-on-train coding is the best. 🍻

PR #1163, opened 2025-09-27
🐛 Fix units for forces/stresses

mbercx:


Minotakm (commented):
Looks good

t-reents (dismissed):

mbercx:
Ok, thanks guys! Added a test for the default XML files where we have forces and stresses calculated.

PR #1164, opened 2025-09-27
🚀 Support release v4.15.0

mbercx:
Besides actual features/improvement/bug fixes, I also added devops changes like the reformatting with Ruff. Main reason is to keep the support branch as close to `main` as possible, so cherry-picking commits stays relatively easy (conflict-free).


Minotakm (approved):
LGTM 🚀

PR #1165, opened 2025-09-27
✨`pw.x`: parse electric field

mbercx:


src/aiida_quantumespresso/parsers/parse_xml/parse.py
```diff
@@ -103,7 +103,7 @@ def parse_xml_post_6_2(xml):
         [x * CONSTANTS.bohr_to_ang for x in outputs['atomic_structure']['cell']['a3']],
     ]
 
-    has_electric_field = inputs.get('electric_field', {}).get('electric_potential', None) == 'sawtooth_potential'
+    has_electric_field = 'electric_field' in outputs
```
mbercx, line 106: @rikigigi I've pinged you since you last touched this line. Is this logic still fine for your use case?

mbercx: Ah, right! Actually, it seems the `sawtooth_potential` logic was added way back in 660b86a2745faf0fa7de64a14fa1e4804bc29403. I'm not sure _why_ we should only set `has_electric_field` to true in this specific case, so I'm fine just adapting it.

PR #1171, opened 2025-09-29
📚 Developer: add release instructions

mbercx:


docs/source/developer.md
```diff
@@ -28,3 +28,100 @@ Originally discussed in [this issue](https://github.com/aiidateam/aiida-quantume
...
+    git pull origin
+
+Next, tag the final release commit:
+
+    git tag -a v3.3.0 -m 'Release `v3.3.0`'
+
+**IMPORTANT**: once you push the tag to GitHub, a workflow will start that automatically publishes a release on PyPI.
+Double check that the tag is correct, and that the `main` branch looks in good shape.
+
+If you accidentally tagged the wrong commit, you can delete the local tag using the following command:
```
Minotakm, line 359: @mbercx Maybe we discuss also if we already pushed the wrong tag?

mbercx: Hmm, maybe? It never really happened. ^^ We have a script that runs in the CD that validated the release tag:

https://github.com/aiidateam/aiida-quantumespresso/blob/main/.github/workflows/validate_release_tag.py

So probably the CD would fail, and nothing would be pushed to the PyPI. At that point you can delete the remote and local tag, and then fix the proper tag.

I'd say: if you mess this up, contact me. 😉

Minotakm (approved):
Thanks @mbercx !

PR #1175, opened 2025-10-02
🔧 Diversify and document our devops tools

mbercx:
This PR makes a bunch of related changes that makes it possible for contributors to work with the tools they prefer:

1. Installing all "extras" in their development environment: easy, straightforward.
2. Using `uv` environments via the `dev` dependency group. This one can also be used for [1] via `--group` for `pip>=25.1`.
3. Hatch environments and scripts.

In [4e2662a](https://github.com/aiidateam/aiida-quantumespresso/pull/1175/commits/4e2662a0ea8d0016f3be0c5a45d201f5fbb097c2) we mainly update `pytest` and fix some incorrect testing with regard to warnings. This was believed necessary to use `hatch test`, but in hindsight that is probably not the case. always good to update packages though!

In https://github.com/aiidateam/aiida-quantumespresso/pull/1175/commits/0a98d60b23987ed570b7853a4aec4d826f09090a we switch to using the Ruff commit hook directly. This means developers won't need to install Hatch.

Finally, in [cda3518](https://github.com/aiidateam/aiida-quantumespresso/pull/1175/commits/cda3518f965b3712004c54af6d9945272cf66a52) we reintroduce the original extras and the new dependency group, and adapt the Hatch environments.

The various options are now also documented, see

https://aiida-quantumespresso--1175.org.readthedocs.build/en/1175/developer.html

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -74,7 +74,7 @@ jobs:
         -   name: Run pytest
             env:
                 AIIDA_WARN_v3: 1
-            run: hatch run pytest -sv tests
+            run: hatch test -v -- -s
 
 
     publish:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,4 +55,4 @@ jobs:
         -   name: Run pytest
             env:
                 AIIDA_WARN_v3: 1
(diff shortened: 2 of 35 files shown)


Minotakm (approved):
Thanks @mbercx nice docs!

PR #1195, opened 2025-10-13
📚 `README.md`: trim down and refer to documentation

mbercx:
The current README file shows the entire compatibility matrix, and still refers to tools for development that we no longer support (e.g. `tox`). Instead of updating the content, we clean up the README and refer to the documentation.

We also correct a small issue with regard to the Python versions in the compatibility matrix in the documentation.


mbercx:
Alright, README adapted! [Have another look](https://github.com/mbercx/aiida-quantumespresso/tree/docs/readme?tab=readme-ov-file#aiida-quantumespresso) @bastonero. I decided to put the emoji at the start of each section title, and that each section deserved an emoji, else they'd be jealous of the Installation and Getting started one.

Now I just hope @sphuber doesn't come hunt me down in Aus because of all the emojis I'm stuffing into `aiida-quantumespresso`. 🙈

PR #1197, opened 2025-10-14
💥 Parsers: remove legacy code

mbercx:
Still cleaning up, so in draft state for now.

@rikigigi I had to remove the "default" `cp.x` parser test here, because that was still in the old `data-file.xml` format. Are the remaining tests sufficient coverage, or should we generate a new one with a more recent QE version?


rikigigi:
I think that the "new" fixtures cover more than the old one, the old are safe to delete. Anyway, those fixtures are 5 years old...

PR #1202, opened 2025-10-23
❌ `PwBandsWorkChain`: deprecate `relax` namespace

mbercx:
The `PwBandsWorkChain` was originally designed to handle both structure optimization and band structure calculations in a single workflow. However, this design conflates two conceptually distinct tasks and creates unnecessary complexity. Users who only want band structures are forced to understand relaxation parameters, while those who need optimization must navigate a nested workflow structure.

From v5.0 onwards, the `PwBandsWorkChain` will focus exclusively on band structure calculations. Users should optimize their structures using the `PwRelaxWorkChain` separately, then pass the relaxed structure to the `PwBandsWorkChain`. This separation provides clearer workflow semantics and better modularity.

The `relax` namespace remains functional during the deprecation period but issues a warning when used. The `get_builder_from_protocol` method no longer sets up relaxation inputs by default, reflecting the future v5.0 behavior.

### DIFF
--- a/src/aiida_quantumespresso/workflows/pw/bands.py
+++ b/src/aiida_quantumespresso/workflows/pw/bands.py
@@ -1,8 +1,10 @@
 """Workchain to compute a band structure for a given structure using Quantum ESPRESSO pw.x."""
 
+import warnings
 from aiida import orm
 from aiida.common import AttributeDict
 from aiida.engine import ToContext, WorkChain, if_
+from aiida.common.warnings import AiidaDeprecationWarning
 
 from aiida_quantumespresso.calculations.functions.seekpath_structure_analysis import seekpath_structure_analysis
 from aiida_quantumespresso.utils.mapping import prepare_process_inputs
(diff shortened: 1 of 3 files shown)


mbercx:
@t-reents will fix the tests if we agree on the deprecation. Currently I also adapt the `get_builder_from_protocol()` method so it doesn't add the `relax` namespace. It makes sense to do this because else the method returns a deprecated set of inputs, but technically it's backwards-incompatible.

What do you think?

t-reents (dismissed):
Changes look good, so I approve, thanks!

Regarding the get_builder_from_protocol, I agree with you. While it is True that this is backwards incompatible, I think that it indeed makes sense to adjust it accordingly and to be honest, it is also fairly easy to be adjusted/fixed by the user to just pass a relaxed structure. 
Moreover, I'm obviously not saying that one should generally ignore it, but in cases like the one here, I tend to be in favor of actually doing the changes instead of always taking care of features that we don't want anymore etc., just because of backwards compatibility (as I said, it's obviously a trade-off and should be decided case by case)

mbercx:
Alright, the tests have been fixed! This should be reviewed together with https://github.com/aiidateam/aiida-quantumespresso/pull/1203. If you can have a final look, @t-reents, I'd much appreciate it. If you sign off on that PR or this one, I'll go ahead and merge both. I'm planning to do a v4.16 release early next week with all the deprecations.

PR #1203, opened 2025-10-24
💥 `PwBandsWorkChain`: remove relaxation

mbercx:
The `PwBandsWorkChain` was originally designed to handle both structure optimization and band structure calculations in a single workflow. However, this design conflates two conceptually distinct tasks and creates unnecessary complexity. Users who only want band structures are forced to understand relaxation parameters, while those who need optimization must navigate a nested workflow structure. From v5.0 onwards, the `PwBandsWorkChain` will focus exclusively on band structure calculations. Users should optimize their structures using the `PwRelaxWorkChain` separately, then pass the relaxed structure to the `PwBandsWorkChain`.

Here we remove the relaxation from the `PwBandsWorkChain`, along with all related logic.


mbercx:
> And apologies for the delay in the review process, but the CSCS proposal wanted my attention as well...

No worries at all @t-reents! I know all about CSCS proposals 😉. Also I was focussing my attention on the project here in Aus.

PR #1204, opened 2025-10-24
💥 `parameters`: enforce consistent casing

mbercx:
Throughout the codebase, we always face the issue that while FORTRAN namelists and
the parameters specified therein are not case-sensitive, the dictionaries we store the
parameters in are. This means we always have to consider the case when we e.g.:

- Write calculation input files
- Perform input validation
- Execute any logic based on the parameters

This makes our code much more verbose, and error prone (see below). Moreover, allowing
freedom in the case of the input parameters makes it much more difficult to write robust
queries. Simply put: allowing for case-insensitive inputs in the parameters is not a
sustainable approach.

Here, we add validation to the `parameters` input port of the various calculation
classes that makes sure they adhere to the following conventions:

1. Namelist keys should be UPPERCASE, and their values must be a dictionary of
   parameters.
2. Parameter keys should be lowercase.

The new `validate_parameters` validator auto-corrects the inputs and emits warnings in
case the node is not yet stored, and rejects stored nodes with error message.

By adding this validation, we can remove any logic that guards against differences in
case, typically via the `_uppercase_dict`/`_lowercase_dict` functions in the
`prepare_for_submission` methods.

Note that some other logic in the codebase has to be adapted to adhere to these new 
conventions. An example here is the `la2F` parameter of the `INPUT `namelist of `q2r.x`.
Here we have to adapt all of the logic to use `la2f` instead. This points to an issue
that was previously not considered: in case the user would have used a different casing,
all of this logic would fail.

Similarly, we have to update the protocol/validation/logic for the `PDosWorkChain` for:

- `DeltaE` -> `deltae`: note that here the QE input manual is inconsistent between codes. It uses
  `DeltaE` for `dos.x`, but `deltaE` for `matdyn.x`. To be fair, the units are also
  different.
- `Emin`/`Emax` → `emin`/`emax`

In all cases, we stick to one convention: lowercase for parameters.


tests/calculations/test_q2r.py
```diff
@@ -44,11 +44,11 @@ def test_q2r_default(fixture_sandbox, generate_calc_job, generate_inputs_q2r, fi
 @pytest.mark.parametrize(
     ('key', 'value'),
     [
-        ('la2F', True),
+        ('la2f', True),
         ('zasr', 'simple'),
     ],
 )
-def test_q2r_parameters(fixture_sandbox, generate_calc_job, generate_inputs_q2r, file_regression, key, value):
+def test_parameters(fixture_sandbox, generate_calc_job, generate_inputs_q2r, file_regression, key, value):
```
mbercx: This change is to avoid issues with updating the data regression file for `la2f`. `pytest`'s data regression _is_ case sensitive when it looks for the YAML file to compare with, but `git` is not always case sensitive. So by renaming the test and regenerating the data regression files we avoid nasty issues where git doesn't properly track the changes.

mbercx:
Thanks @t-reents for the review! I'll go ahead and merge. I was also working on adding a deprecation path, but realised this is not so simple. 

I was thinking of only raising a deprecation warning and _not_ fixing the case. But then we should still adapt the case in the protocols for  `DeltaE`, `Emin` and `Emax`, else our protocols raise warnings.  But then we should also fix the validation/logic in e.g. the `PdosWorkChain`, else that will break. But then using the old `DeltaE` will break, so we're making breaking changes. We can add logic to be lenient and accept both cases here. But we'll be making the code more messy, which is what we're trying to avoid.

I've opened https://github.com/aiidateam/aiida-quantumespresso/pull/1210 with my current version of the deprecation. We can also decide not to merge it in the end: I think by updating on the fly with a warning we are helping to avoid any issues for _most_ users. Also: the deprecation may be more annoying because it _doesn't_ fix things on the fly. 😅 Have a look and let me know what you think on the PR!

PR #1210, opened 2025-11-12
❌ `parameters`: deprecate inconsistent casing

mbercx:
Deprecation PR of https://github.com/aiidateam/aiida-quantumespresso/pull/1204

>[!NOTE] 
> Reposting [this comment](https://github.com/aiidateam/aiida-quantumespresso/pull/1204#issuecomment-3520235796) on the deprecation pathway below

@t-reents I was also working on adding a deprecation path, but realised this is not so simple. I was thinking of only raising a deprecation warning and not fixing the case. But then we should still adapt the case in the protocols for DeltaE, Emin and Emax, else our protocols raise warnings. But then we should also fix the validation/logic in e.g. the PdosWorkChain, else that will break. But then using the old DeltaE will break, so we're making breaking changes. We can add logic to be lenient and accept both cases here. But we'll be making the code more messy, which is what we're trying to avoid.

We can also decide not to merge this PR in the end: I think by updating on the fly with a warning we are helping to avoid any issues for most users. Also: the deprecation may be more annoying because it doesn't fix things on the fly. 😅 Have a look and let me know what you think!

### DIFF
--- a/src/aiida_quantumespresso/calculations/__init__.py
+++ b/src/aiida_quantumespresso/calculations/__init__.py
@@ -18,6 +18,7 @@ from qe_tools.converters import get_parameters_from_cell
 from aiida_quantumespresso.data.hubbard_structure import HubbardStructureData
 from aiida_quantumespresso.utils.convert import convert_input_to_namelist_entry
 from aiida_quantumespresso.utils.hubbard import HubbardUtils
+from aiida_quantumespresso.utils.validation.parameters import validate_parameters
 
 from .base import CalcJob
 from .helpers import QEInputValidationError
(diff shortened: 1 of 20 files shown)


t-reents:
I pretty much resonate with all your comments. Initially, I'd also say, yes, let's add warnings and don't auto-fix yet. However, we then need to do all the changes that you outlined. While I'd still say that it might be OK to adjust the protocols accordingly etc., I'm (also) not sure whether it's really worth it to also adjust the validation logic of the PDOS WorkChain etc. and adding the complexity to the code

In summary, I would almost tend to your last comment
> We can also decide not to merge this PR in the end: I think by updating on the fly with a warning we are helping to avoid any issues for most users. Also: the deprecation may be more annoying because it doesn't fix things on the fly. 😅

Since all those warnings could indeed be more annoying than just fixing it.

Sorry for not having much to add, but as I said, I'd personally tend to not add this "complexity"

Issue #1156, opened 2025-09-24
👌 Tracking possible improvements for setup codes command


mbercx:
### Dealing with slow connections

I ran

```
(mc3d-relax) aiida@thanos:~/project/mc3d-relax/git/aiida-quantumespresso$ aiida-quantumespresso setup codes -l sirius-{} eiger pw.x
```

And the command seemed like it got stuck. Probably my connection is slow, and also initially I had a quite large "safe interval". We should probably add some status updates and notes that these commands can take a while.

mbercx:
On a more positive note, running

```
aiida-quantumespresso setup codes eiger pw.x dos.x projwfc.x -i
```

Copying in the `module load statements`, and seeing

```console
Success: Code<pw> for pw.x created with pk<1> on Computer<eiger>
Success: Code<dos> for dos.x created with pk<2> on Computer<eiger>
Success: Code<projwfc> for projwfc.x created with pk<3> on Computer<eiger>
```

Is so satisfying! 😍

giovannipizzi:
Just to clarify one of the points: what is put as prepend text in the setup command is added also as a prepend text of the computer, or not? I think it should

PR #1222, opened 2025-12-05
Automatic code setup: validation and `--all` option

t-reents:
Addressing parts of #1156

* adding validation
* `--all` option. To make this more convenient, I also added the `--ignore-missing` and `--skip-existing` options. I agree with you @mbercx, that I'd try to keep this as minimal as possible. I figured that `--all` (and the related ones) make sense. However, I don't really see the need for a dry-run.

Since we have the `--all` option now, the `executables` argument is optional now.

Even without the new `--all` option, adding a `--ignore-missing` makes sense. I just realised that the command would fail when specifying three executables of which one is missing, which was a bit annoying.

### DIFF
--- a/src/aiida_quantumespresso/tools/setup.py
+++ b/src/aiida_quantumespresso/tools/setup.py
@@ -8,9 +8,9 @@ from typing import Union
 from aiida import orm
 
 VALID_QE_EXECUTABLES = [
+    'bands.x',
     'cp.x',
     'dos.x',
-    'epw.x',
     'matdyn.x',
     'neb.x',
     'ph.x',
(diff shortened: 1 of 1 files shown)


t-reents:
>Instead of adding an --ignore-missing option, should raise a warning instead of an error by default?
Now any issues with missing executables are entirely silences with ignore_missing = True. Is that a good idea?

Indeed, probably not such a good idea. Thinking about it, I actually like the first option, e.g. using a warning. Nonetheless, one advantage of raising by default is that it is very explicit, i.e. users will immediately realize that something went wrong. However, one will probably also realize quickly that the code is missing and look at the warning at that point

PR #1224, opened 2025-12-05
Fixing Hubbard formulation

### LINKED ISSUE #1220: 'liechtenstein' Hubbard formulation
When trying to set the Hubbard formulation to 'liechtenstein' in an instance of the `Hubbard` class, I'm getting 

```bash
File ".../aiida_quantumespresso/utils/hubbard.py", line 69, in get_hubbard_card
    line = f'{pre}\t{atom_i}-{man_i} \t{value}'  # noqa: F821
              ^^^
UnboundLocalError: cannot access local variable 'pre' where it is not associated with a value
```

Version 4.16.0

t-reents:
Addresses #1220 

@bastonero I applied the fix as discussed in the issue. Moreover, I also added "ALPHA" to the supported values. Please have a look whether my understanding/logic for the dudarev formulation was correct.

One additional aspect:

I was wondering, do we actually need this check here:
https://github.com/aiidateam/aiida-quantumespresso/blob/0f4318b8ae408aa89bc2123b04ddb8ab81c01bd8/src/aiida_quantumespresso/utils/hubbard.py#L65-L66

Isn't this automatically validated in:
https://github.com/aiidateam/aiida-quantumespresso/blob/0f4318b8ae408aa89bc2123b04ddb8ab81c01bd8/src/aiida_quantumespresso/common/hubbard.py#L151-L152


bastonero (changes requested):
thanks a lot @t-reents. would you be a lamb and add a little test when actually using the liechtenstein "keyword"?

t-reents:
Ciao @bastonero !

Apologies, it obviously took my half a year to look into it again..

I should have addressed all your comments in ea518edb2813fa9423923cac52f8203501fdb582.
Could I kindly ask you for a review?

P.S.: In general, since I'm now trying to take care of the maintenance (since @mbercx left 😢), are you still available for Hubbard related questions/reviews, or are you no longer working on any of these tools? Just to know :)

src/aiida_quantumespresso/utils/hubbard.py
```diff
@@ -62,26 +62,23 @@ def get_hubbard_card(self) -> str:
...
 
             # This variable is to meet QE implementation. If intersite interactions
             # (+V) are present, onsite parameters might not be relabelled by the ``hp.x``
             # code, causing a subsequent ``pw.x`` calculation to crash. That is,
             # we need to avoid writing "U Co-3d 5.0", but instead "V Co-3d Co-3d 1 1 5.0".
             is_intersite = is_intersite_hubbard(hubbard=hubbard)
             if hubbard.formulation == 'dudarev':
-                if param.hubbard_type == 'J':
-                    pre = 'J'
+                if param.hubbard_type in ['J', 'J0', 'ALPHA']:
```
bastonero, line 74: I am not entirely sure what's more consistent with the Dudarev formulation, since I have never effectively used J (but only Ueff = U-J). I suppose we can leave this "open" to the user, and worst case QE will throw some error. Somehow as with pw.x, we do not want to do input checks on our side (at least, just the basics). So i guess for the time being this is fine.

bastonero (approved):
Thanks a lot @t-reents ! I think all in good shape. Just a comment, but not nothing really needed. I think we can merge this as is.

Re maintainance/hubbard related things: I now left Bremen and my new position is not particularly devoted to the maintainance of the plugin, although i am very happy to keep it going with some reviews, especially for the Hubbard side. Just ping me, and I'll let you know if in that specific moment i have time or not to do the review.

src/aiida_quantumespresso/utils/hubbard.py
t-reents: Thanks. I wasn't a 100% myself, as I'm also not a heavy user of the Hubbard part, but converged to this solution with Claude. I will indeed leave it as is and we can always come back. Thanks!

t-reents:
>Re maintainance/hubbard related things: I now left Bremen and my new position is not particularly devoted to the maintainance of the plugin, although i am very happy to keep it going with some reviews, especially for the Hubbard side. Just ping me, and I'll let you know if in that specific moment i have time or not to do the review.

OK, great – thank you! Indeed, my main idea was to potentially ping you for the Hubbard related aspects, since you have a lot of expertise there, and potentially ask you for a review. Other than that, I think that the maintenance work should be fine on our side.

Btw, all the best of course for the next chapter!!

PR #1250, opened 2026-04-10
fix developer guide url

Bud-Macaulay:
URL to the developer guide in the readme is broken,

not sure how changing URLs are handled for the docs.
If you would rather fix this with a redirect elsewhere feel free to close.


mbercx (approved):
Right, I moved that in 1b809205560dbcefc3dbe49c451872c96a73c0ee, nice catch, thanks for the fix @Bud-Macaulay!

mbercx:
Docs build fails because of this warning, I think. Will fix in a bit:

```
WARNING: failed to reach any of the inventories with the following issues:
intersphinx inventory 'https://pymatgen.org/objects.inv' not fetchable due to <class 'requests.exceptions.HTTPError'>: 404 Client Error: Not Found for url: https://pymatgen.org/objects.inv
```

mbercx:
> If you would rather fix this with a redirect elsewhere feel free to close.

We don't do fancy stuff like that 'round here 😉

t-reents:
Thanks @Bud-Macaulay !
CI (including the docs) pass now, I'll merge in a moment

PR #1252, opened 2026-04-10
fix double encode of hubbardJSON

### LINKED ISSUE #1251: "Hubbard json file" in repository is not valid json object
Currently I believe that the saved Hubbard file is not a valid json object due to it being double encoded by two seperate methods (json.dumps and model_dump_json())

``` python
    @hubbard.setter
    def hubbard(self, hubbard: Hubbard):
        """Set the full Hubbard information."""
        if not isinstance(hubbard, Hubbard):
            raise ValueError('the input is not of type `Hubbard`')

        serialized = json.dumps(hubbard.model_dump_json()) 
        self.base.repository.put_object_from_bytes(serialized.encode('utf-8'), self._hubbard_filename)
```
This means that althought the RESTapi is correctly serving a download endpoint of these files as json the browser does not correctly parse and format it. 

This is confusing for users using aiida-explorer or the rest api programmatically as they would require a manual reformatting step. 

A fix would likely require also changing the getter method to be compatible with normal json and this double wrapped json? I have seen in other issues that perhaps that there are a few other discussions regarding the setter (ex #1219), so if there is some to rewrite this module in the future this can wait.

Bud-Macaulay:
fixes #1251 

Removed double encoding of hubbard json file and extended getter to handle the both cases.


t-reents:
Thanks @Bud-Macaulay !

The change looks good to me in general. Could you maybe add 2 tests?

1) Check that after your fix the stored file is just a JSON object as expected and not a double encoded/escaped string.
2) Check that the `getter` is indeed backwards compatible and works with the old double encoded files

src/aiida_quantumespresso/data/hubbard_structure.py
```diff
@@ -59,16 +59,29 @@ def hubbard(self) -> Hubbard:
...
+            # Try normal case first
+            return Hubbard.model_validate_json(content)
+        except Exception:
+            # Fallback: old double-encoded format
+            try:
+                decoded = json.loads(content)  # unwrap string
+                return Hubbard.model_validate_json(decoded)
+            except Exception as exc:
+                raise ValueError(f'Invalid Hubbard JSON format: {exc}')
+
```
t-reents:
```suggestion
        except pydantic.ValidationError:
            # Fallback: old double-encoded format
            try:
                decoded = json.loads(content)  # unwrap string
                return Hubbard.model_validate_json(decoded)
            except (json.JSONDecodeError, pydantic.ValidationError) as exc:
                raise ValueError(f'Invalid Hubbard JSON format: {exc}') from exc

```

src/aiida_quantumespresso/data/hubbard_structure.py
t-reents, line 6:
```suggestion
import numpy as np
import pydantic
```

t-reents (changes requested):
Thanks @Bud-Macaulay !

Something I was already thinking before: the `except Exception` is probably quite broad, considering that this should only capture the legacy double encoding, shouldn't it?

I added a suggestion to adapt it (the ValidationError was my part, I further iterated with Claude to also make the inner except block more specific). Do you agree with this? If yes, I also added a suggestion to add the corresponding tests that explicitly test those exceptions. If you agree, you can just commit them.

Apologies for the additional work 😅

tests/data/test_hubbard_structure.py
```diff
@@ -171,3 +173,29 @@ def test_get_symbol_indices(generate_hubbard_structure):
...
+    original = hubbard_structure.hubbard
+
+    # Simulate the old double-encoded format: json.dumps(hubbard.model_dump_json())
+    double_encoded = json.dumps(original.model_dump_json())
+    hubbard_structure.base.repository.put_object_from_bytes(
+        double_encoded.encode('utf-8'), 'hubbard.json'
+    )
+
+    restored = hubbard_structure.hubbard
+    assert restored == original
```
t-reents, line 201:
```suggestion
    assert restored == original


def test_hubbard_getter_invalid_json_raises_value_error(generate_hubbard_structure):
    """Test that unparsable content raises a `ValueError`, exercising the `json.JSONDecodeError` fallback."""
    hubbard_structure = generate_hubbard_structure()
    hubbard_structure.base.repository.put_object_from_bytes(b'not even json', 'hubbard.json')

    with pytest.raises(ValueError, match='Invalid Hubbard JSON format'):
        hubbard_structure.hubbard  # pylint: disable=pointless-statement


def test_hubbard_getter_propagates_unexpected_errors(generate_hubbard_structure, monkeypatch):
    """Test that errors unrelated to (legacy) JSON parsing are not silently swallowed."""
    from aiida_quantumespresso.common.hubbard import Hubbard

    hubbard_structure = generate_hubbard_structure()

    def _raise(*args, **kwargs):
        raise RuntimeError('unexpected failure')

    monkeypatch.setattr(Hubbard, 'model_validate_json', _raise)

    with pytest.raises(RuntimeError, match='unexpected failure'):
        hubbard_structure.hubbard  # pylint: disable=pointless-statement

```

Bud-Macaulay:
Tbh, i can barely remember opening this PR ---
Have mashed enter - If you're happy i'm happy.

t-reents (approved):

t-reents:
Haha, perfect 😄 

I think that I'm happy now, thanks again!

PR #1253, opened 2026-04-13
Remove silent failure for periodic boundary conditions

elinscott:
Previously, `structure.pbc = (False, True, True)` would be silently ignored and treated as a fully-periodic system. This minor PR updates the code so that this instead an error is raised in such cases.

### DIFF
--- a/src/aiida_quantumespresso/workflows/pw/base.py
+++ b/src/aiida_quantumespresso/workflows/pw/base.py
@@ -1,4 +1,5 @@
 """Workchain to run a Quantum ESPRESSO pw.x calculation with automated error handling and restarts."""
+import warnings
 
 from aiida import orm
 from aiida.common import AttributeDict, exceptions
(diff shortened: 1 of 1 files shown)


elinscott:
My fix was targeting 2D systems but clearly 0D and 1D systems have been caught up in the changes. What is the desired behaviour here?

- For 0D we can either use Martyna-Tuckerman or Makov-Payne (cubic cells only)
- For 1D there is no built-in solution as far as I can tell
- Should we even be setting `assume_isolated` at all? (Maybe only for charged systems?)

Regardless of the answers to the above questions, changes to the codebase are necessary because currently the behaviour is inconsistent: image corrections are applied for 2D but not for 0D or 1D.

mbercx:
Thanks @elinscott! My MC**3D** bias is showing again here. ^^ In fact, the only reason the 2D version of `assume_isolated` because it was added by @AndresOrtegaGuerrero in https://github.com/aiidateam/aiida-quantumespresso/commit/ef21642ba8c715b384e726628f866f7a3df63890. I will let him comment on your third point.

Other than that, I can only comment that since the protocols were testsed for 3D systems, perhaps we should in any case raise a big fat "here be dragons" warning in case the structure is not fully periodic? Perhaps we can rope in some other non-3D experts: @ccigna @Crivella (feel free to ask others).

elinscott:
Indeed, long-term the builder should probably support different periodicity, but until then there should definitely be a "here be dragons" warning, as a lot of this builder's logic will give nonsensical values for non-3D systems (e.g. > 1 **k**-point in aperiodic directions).

Perhaps I could rewrite my modifications so that the logic is more general and can serve as a location where — in the future — 0D, 1D, and 2D-specific logic for multiple keywords (not just `assume_isolated`) could be added. Is that a good solution for now?

AndresOrtegaGuerrero:
> My fix was targeting 2D systems but clearly 0D and 1D systems have been caught up in the changes. What is the desired behaviour here?
> 
> * For 0D we can either use Martyna-Tuckerman or Makov-Payne (cubic cells only)
> * For 1D there is no built-in solution as far as I can tell
> * Should we even be setting `assume_isolated` at all? (Maybe only for charged systems?)
> 
> Regardless of the answers to the above questions, changes to the codebase are necessary because currently the behaviour is inconsistent: image corrections are applied for 2D but not for 0D or 1D.

For 0D systems (i.e., isolated molecules), both Makov–Payne (MP) and Martyna–Tuckerman (MT) charge corrections can in principle be considered. However, additional checks are required to ensure that the simulation cell is sufficiently large for these corrections to be valid. It is therefore unclear whether such constraints should be explicitly enforced.

For 2D systems, we use the implemented correction as-is, since it is currently the only one available for structures periodic in the x–y plane. This is also the convention adopted in the AiiDAlab-QE App.

For 1D systems, the choice is left to the user: they may either apply charge corrections or instead increase the cell size along the non-periodic directions (e.g., the z-axis). In any case, using 1D systems should not raise an error, as we still aim to sample the Brillouin zone with a grid of the form (n, 1, 1).

Crivella:
I'm on the side of giving a warning to the user but not an error, so they can set the relevant parameters themselves.

Setting this parameters in reduced dimensionality systems will in general require additional tests as mentioned by @AndresOrtegaGuerrero

elinscott:
> I'm on the side of giving a warning to the user but not an error, so they can set the relevant parameters themselves.
> 
> Setting this parameters in reduced dimensionality systems will in general require additional tests as mentioned by @AndresOrtegaGuerrero

I'm not suggesting we should return an error for 0D/1D, the question is whether or not there is anything sensible we can automate rather than leaving everything to the user

ccigna:
Dear all,

if the question is whether to use the "assume isolated" flag for 1D system, since there is not the correct 1D correction public available, I'd say don't use it: use the "2D" one is less correct that not using at all. 
Raising a warning to correctly converge the vacuum space is a good solution. The size of this vacuum depends on the scope of the calculations (if you aim to energy only, of forces like phonons etc.) therefore I'd let the choice to the user.

What can be automated for 1D (and 2D) system is the flag "cell_dofree" is you are doing a vc-relax, and maybe the "mixing_mode = local-TF" if the system does not reach convergence.

Best,
Chiara

PR #1255, opened 2026-04-21
adding a monitor for accuracy stuck

mikibonacci:
In this PR I add a specific monitor function to the plugin. This monitor checks the SCF accuracy during a `PwCalculation` and it kills the job if the accuracy is the same after 5 iterations. This is a sort of a bug in [QE](https://gitlab.com/QEF/q-e/-/work_items?sort=created_date&state=opened&author_username=mikibonacci&first_page_size=20&show=eyJpaWQiOiI4MTQiLCJmdWxsX3BhdGgiOiJRRUYvcS1lIiwiaWQiOjE3MzI2NDQ4OX0%3D), and up to now there is not solution. 

This check is particularly useful in the case of large supercell calculations (like in muon spectroscopy).

This monitor can be attached to the calculation simply via:

```python
from aiida_quantumespresso.tools.monitors import monitor as accuracy_stuck_monitor

builder.monitors = {"accuracy_stuck": Dict({"entry_point": "quantumespresso.accuracy_stuck"})}
```

as also explained in the corresponding doc page (part of this PR). 

For now, the same monitor (with a less polished implementation) is [here](https://github.com/mikibonacci/aiida-monitor.git), but as it is specific to QE might be useful to have it in this package.

After this PR, I will open another one to implement the related error handler.


mbercx:
Thanks Miki! Very curious to try it. Maybe we can look at this together next week? When are leaving again for the land of pasta and primitivo?

mikibonacci:
I'll be surrounded by Pasta starting from friday :( .  We can have a call next week, happy to discuss! also related to the following PR I would like to submit, we need for sure to discuss on that

t-reents (approved):
Thanks @mikibonacci !

Fixed the tests and made the number of consecutive iterations before terminating configurable.

PR #1256, opened 2026-04-24
Added support for parsing MD calculations

tsthakur:
- Added MD-specific parsing logic using "Entering Dynamics:" keyword
- Implemented velocity calculation from positions using velocity-Verlet algorithm used by Quantum ESPRESSO
- Supports both force-based and simple velocity calculations
- Now handles incomplete stdout for MD calculations allowing structure extraction from a trajectory in cases where restarting from wavefunction is not possible
- Added MD parameters extraction (dt, iprint) to trajectory metadata
- Added proper unit conversions for velocities and simulation time
- Now marks problematic trajectories with 'never_concatenate' attribute to avoid issues when combining multiple MD runs
- Backward compatibility is still preserved for non-MD calculations
- Added suggestions for future improvements in comments to preserve compatibility with non-MD use-cases

### DIFF
--- a/src/aiida_quantumespresso/parsers/parse_raw/pw.py
+++ b/src/aiida_quantumespresso/parsers/parse_raw/pw.py
@@ -531,7 +531,22 @@ def parse_stdout(stdout, input_parameters, parser_options=None, parsed_xml=None,
     # (cell, initial positions, kpoints, ...) and I skip them.
     # In case, parse for them before this point.
     # Put everything in a trajectory_data dictionary
-    relax_steps = stdout.split('Self-consistent Calculation')[1:]
+    
+    # For MD calculations, I now skip the data calculated for the 0th step i.e. for the
+    # input structure. And nothing is calculated for the final structure, so in the trajectory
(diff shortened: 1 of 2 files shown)


mbercx:
Thanks @tsthakur! First order of business: tests

1. The tests are failing. Seems a key isn't there that your code expects it to be for non-MD calculations. Look at the tests, figure out the root cause. Being able to run the tests locally will help (see [the developer guide](https://aiida-quantumespresso.readthedocs.io/en/stable/developer.html), stick with default).
2. We need tests for your new parsing functionality. I'll help you set up the tests on Friday, but what I need from you is the files: i.e. the output files of an MD run we can use for testing the parsing. Here it is _very_ important that these files are small; else our repo will bloat over time and working on it becomes a pain. So the files should be minimal but complete: as small as possible while still testing all the parsing functionality you introduce here. This will probably require a run that makes no sense (e.g. with 2 k-points, only 10 steps, ...), but that's fine for the parsing: the results don't matter here.

Does that make sense? Let me know if you have any questions!

PR #1257, opened 2026-04-28
🐛 Parsers: pass `pbc` to `TrajectoryData.set_trajectory`

mbercx:
`aiida-core` v2.8.0 deprecates omitting `pbc` when `cells` is provided and will raise in v3.0.0. Pass the input structure's `pbc` from the `pw`, `cp`, and `neb` parsers to silence the warning and stay forward-compatible.

The `aiida-core` lower bound is bumped to `~=2.8` since `pbc` is not accepted by `set_trajectory` in earlier versions.


mbercx:
Note: we'll have to adapt compatibility matrix

PR #1261, opened 2026-05-13
Remove ppcg diagonalization method

### LINKED ISSUE #1091: QE v7.5 drops support for PPCG diagonalization
`PwBaseWorkChain` implements an error handling that switches to different diagonalization algorithms, among which the `ppcg` one. QE v7.5 drops support for it, so when we will support v7.5 we will also need to e.g. make a check of the QE version an not allow the `ppcg` flag in the error handling.

Reference: https://gitlab.com/QEF/q-e/-/blob/develop/Doc/release-notes?ref_type=heads

stillfast:
Remove ppcg diagonalization method as QE v7.5 no longer supports it, improving error handling for users.#1091

### DIFF
--- a/src/aiida_quantumespresso/workflows/neb/base.py
+++ b/src/aiida_quantumespresso/workflows/neb/base.py
@@ -243,7 +243,7 @@ class NebBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
 
         if 'diagonalizations' not in self.ctx:
             # Initialize a list to track diagonalisations that haven't been tried in reverse order or preference
-            self.ctx.diagonalizations = [value for value in ['cg', 'paro', 'ppcg', 'david'] if value != current.lower()]
+            self.ctx.diagonalizations = [value for value in ['cg', 'paro', 'david'] if value != current.lower()]
 
         try:
             new = self.ctx.diagonalizations.pop()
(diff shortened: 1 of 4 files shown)


mbercx:
Thanks @stillfast! I ran the CI, but you'll probably still have to update the corresponding tests. Want to have a crack at it? I can also take care of this.

mbercx:
The `NebBaseWorkChain` also seems to still switch to `ppcg`. I assume support for `ppcg` has been removed from `neb.x` as well.  @dbidoggia: Do you know?

mbercx:
Suggested commit message for squash and merge

```
🐛 `Pw/NebBaseWorkChain`: drop `ppcg` from fallback list

Quantum ESPRESSO v7.5 removes support for the `ppcg` diagonalization algorithm,
so `handle_diagonalization_errors` can no longer switch to it as a recovery
step. When it does, the restarted calculation fails on unsupported input rather
than recovering, defeating the purpose of the handler.

Remove `'ppcg'` from the candidate list in `handle_diagonalization_errors`,
leaving `['cg', 'paro', 'david']` as the fallback cycle. No version check is
added: supporting `ppcg` only on pre-7.5 versions would complicate the handler
for an algorithm that has probably been removed from Quantum ESPRESSO for a good
reason.
```

stillfast:
> ppcg

Thank you for the reminder. I will revise neb.x immediately and  I'll take a crack at it.

mbercx:
Cheers! Let's indeed also go ahead and remove `ppcg` from the `NebBaseWorkChain`. If it _is_ still supported for whatever reason, adding it again is trivial.

mbercx (approved):
All checks green! Thanks for taking care of this, @stillfast. 🙏

stillfast:
You're welcome! I really appreciate the guidance. @mbercx

PR #1262, opened 2026-05-13
Remove silent failure for periodic boundary conditions

elinscott:
Previously, `structure.pbc = (False, True, True)` would be silently ignored and treated as a fully-periodic system. This minor PR updates the code so that this instead an error is raised in such cases.

Replaces #1253 after migrating the corresponding branch to my fork; see that PR for preceeding discussion.


elinscott:
@t-reents I think this is ready for review. The one design decision that needs to be made is if `pbc = (True, False, True)` should emit an error or only warn. See prior discussion in the old version of this PR.

t-reents:
Thanks @elinscott !

Personally, I'd be in favour of an error, i.e., keeping your current implementation, as this is the "safest" option. Moreover, to some extent it's also consistent with the logic in the PwRelaxWorkChain for the different relaxation modes:

https://github.com/aiidateam/aiida-quantumespresso/blob/633c24cdce00aa6dc2eb6e526d819f0d4653a56f/src/aiida_quantumespresso/workflows/pw/relax.py#L179-L191

I assume, since you implemented it in this way, that you agree with the error instead of warning approach? If yes, I will approve and merge it – just let me know

elinscott:
Indeed, what I was doing was to be consistent with that logic. However, note that this new restriction is more strict — that earlier catch is relevant only to vcrelex calculations, whereas my `ValueError` raises for _any_ pw calculation. For such calculations, as far as I know `True True False` is not inherently safer than `False True True` (unless we have a charged system in which case `assume_isolated` becomes important). For this reason, I actually think that probably a warning is more appropriate, and I will downgrade to a warning (unless you object).

elinscott:
... hmmm, now I've convinced myself that it is distatesful that if we have a 2D system we will/won't set `assume_isolated` based on how the Cartesian coordinates are arranged. (But at the same time, we don't set `assume_isolated` for 0D or 1D systems at all!) Probably worthwhile referring back to the discussion of the old PR (see description above).

t-reents:
I was just replying :D So yeah, in principle I'd be fine with the warning, but in that case, I wouldn't set `assume_isolated`, exactly for the reason that the orientation shouldn't change which parameters we set. 

Moreover, regarding consistency, I was considering that this is quite strict for all pw calculations, but I was in principle OK with it (not an argument for or against, but in practice, almost any workflow involves a relaxation at some point, so the orientation should probably follow the (True, True, False) anyway, so it wouldn't be a too strict limitation – I guess).

Regarding the other discussion in #1253, I somehow thought/got that for 2D it would be a valid/reasonable choice (also consistent with AiiDAlab), whereas for 1D and 0D there isn't an obvious default to set. 

In the end, the question is essentially do we want to set `assume_isolated` in case of 2D? If yes, I'd say we should raise an error, and I generally like the idea of having consistent behaviour between AiiDA-QE and AiiDAlab. Otherwise, if we go for the warning, I wouldn't set `assume_isolated` for the reason we discussed before.

elinscott:
OK. great. in that case I think we're in agreement. Let's keep `assume_isolated` for 2D but not for 1D and 0D, which means that raising an error for x-z-periodic and y-z-periodic systems is the rational thing to do. Happy for you to merge!

t-reents (approved):
Thanks again @elinscott ! 🚀

PR #1271, opened 2026-07-06
🐛 `PwBaseWorkChain`: use FR pseudo family for `SPIN_ORBIT` without overrides

Minotakm:
The fully-relativistic PseudoDojo default for `spin_type == SpinType.SPIN_ORBIT` was only applied when `overrides` was not `None`. A builder created without any overrides therefore silently received the SSSP pseudopotentials. The default is now applied whenever no explicit `pseudo_family` is provided, and a test is added.


t-reents (approved):

t-reents:
Thanks @Minotakm

PR #1272, opened 2026-07-08
Add `hyperqueue` to the schedulers receiving default resources

elinscott:
## Problem
`ProtocolMixin.set_default_resources` fills in `resources = {'num_machines': 1}` for various schedulers as a  temporary workaround until defaults can be configured on computers (#1011). Computers using the [aiida-hyperqueue](https://github.com/aiidateam/aiida-hyperqueue) scheduler get no such default, so every `get_builder_from_protocol` call fails at builder validation with `invalid attribute value required value was not provided for 'resources'` unless the user manually threads `metadata.options.resources` through every override namespace.

## Change
Add `hyperqueue` to the scheduler tuple. `HyperQueueJobResource` accepts the `num_machines`/`num_mpiprocs_per_machine` backward-compatibility shape and derives `num_cpus` from it, so the same conservative single-machine default the other schedulers receive is well-defined here too.

## Notes
This is the first non-`core.*` entry in the tuple, but the coupling is by string only (no import) and I _think_ we can trust the aiida-hyperqueue maintainer 😉

## Testing
Tested end-to-end by a downstream Koopmans workflow driving `PwBaseWorkChain`/`PwBandsWorkChain` protocols on a HyperQueue localhost computer: before the change every protocol-built calcjob failed validation; after it, jobs submit and run.


t-reents (approved):

t-reents:
Thanks @elinscott, will merge it now.

Issue #1273, opened 2026-07-16
Make protocol overrides a TypedDict


elinscott:
## The problem

One of the most painful things about writing an aiida workflow that depends on `PwBaseWorkChain`, `PwBandsWorkChain` (or indeed other workchains outside of `aiida-quautumespresso`) is working out the appropriate keys for `overrides` in `get_builder_from_protocol`. `overrides` is a schemaless nested dict that mirrors the workchain's input namespace tree. If you get it wrong, nothing validates its shape, and a misplaced/misspelled key is silently ignored.

## Proposed solution
Make `overrides` `TypedDict`s with `total=False` that mirror each protocol builder's expected `overrides` tree, e.g. for `PwBaseWorkChain`:

```python
class PwParametersOverrides(TypedDict, total=False):
    CONTROL: dict[str, Any]
    SYSTEM: dict[str, Any]
    ELECTRONS: dict[str, Any]
    # ...

class PwCalculationOverrides(TypedDict, total=False):
    parameters: PwParametersOverrides
    pseudos: dict[str, Any]
    metadata: dict[str, Any]
    # ...

class PwBaseOverrides(TypedDict, total=False):
    pw: PwCalculationOverrides
    kpoints_distance: float
    pseudo_family: str
    clean_workdir: bool
    # ...
```

and annotate `get_builder_from_protocol(..., overrides: PwBaseOverrides | None = None)`.

Runtime behavior is unchanged — `TypedDict` is a plain `dict` at runtime and `total=False` keeps every key optional, so existing untyped callers keep working and no validation cost is added. But now mypy/pyright will flag misspelled and misplaced keys in downstream code, and IDEs can autocomplete.

### Avoiding a second source of truth
These mirrors of the input namespace would be a duplicate of the process spec and would drift as ports are added or renamed. Two ways to keep the spec authoritative:

- **Drift-guard tests**: one test per overrides TypedDict asserting its keys are a subset of the corresponding workchain's `spec().inputs` port names plus builder-consumed protocol keys
- **Generation from the spec**: derive the TypedDicts from `WorkChain.spec()` — either a small script committed alongside the types and run in CI in `--check` mode (regenerate-and-diff, the same contract as formatters), or a build-time codegen step. The spec stays the single source of truth.

## Route forward
If maintainers are open to this, I'd be glad to contribute the PR.

elinscott:
@t-reents @Minotakm @mbercx @giovannipizzi interested in your thoughts on this

[...]

PR #1274, opened 2026-07-16
✨ Typed `overrides` for `get_builder_from_protocol`

elinscott:
This PR adds `TypedDict` definitions describing the nested `overrides` mapping accepted by `get_builder_from_protocol`, so that callers get editor autocompletion and static type-checking for what is otherwise an untyped, deeply-nested `dict`. This PR introduces the scheme for `PwBaseWorkChain` and `PwBandsWorkChain`; the remaining work chains can follow the same pattern in later PRs.

Partially addresses #1273

## Changes
A new module `src/aiida_quantumespresso/workflows/protocols/overrides.py` defines `total=False` `TypedDict`s that mirror the overrides tree.

The two `get_builder_from_protocol` signatures are annotated `overrides: PwBaseOverrides | None = None` and `overrides: PwBandsOverrides | None = None` respectively.

The `TypedDict`s carry no runtime behaviour — they are ordinary `dict`s at runtime and the builder logic is untouched. Existing `overrides={...}` call sites keep working exactly as before.

## Drift guards

Because these types duplicate information that already lives in each work chain's `spec()`, they can rot. `tests/workflows/protocols/test_overrides.py` asserts, for each `TypedDict`, that its keys are **equal** to the corresponding `spec().inputs` port names, **plus** a small, explicitly-declared set of "builder-consumed" keys — protocol keys such as `meta_parameters` and `pseudo_family` that `get_builder_from_protocol` reads directly (`inputs.pop(...)`) without exposing them as input ports — **minus** the few ports the builder always sets from its own `code`/`structure` arguments. Each nested namespace is checked against its own sub-spec: `pw`, the top-level `scf`/`bands`, and the `scf.pw`/`bands.pw` sub-namespaces.

Because `PwBandsWorkChain` excludes `clean_workdir` from `scf`/`bands` and `pw.parent_folder` from `bands`, those namespaces use narrowed `TypedDict`s rather than the full `PwBaseOverrides`, and the sub-spec guards lock that narrowing in.

The `pw.parameters` namelist keys are guarded only against a second hand-maintained copy — there is no authoritative in-repo list of every `pw.x` namelist — with the namelists QE dispatches by calculation type additionally checked against the plugin's `_automatic_namelists`. If a future spec change removes or renames a port, the matching test fails until the `TypedDict` is updated.

## Notes for maintainers

On the single-source-of-truth question raised in #1273 this implementation keeps things hand-written rather than generating types from the `spec` at runtime (which wouldn't give static checkers anything to see). Happy to revisit the generation approach if preferred.

### DIFF
--- a/src/aiida_quantumespresso/workflows/protocols/overrides.py
+++ b/src/aiida_quantumespresso/workflows/protocols/overrides.py
@@ -1,0 +1,100 @@
+"""Typed dictionaries describing the ``overrides`` accepted by ``get_builder_from_protocol``.
+
+These ``TypedDict`` definitions mirror the nested ``overrides`` mapping that each work chain's
+``get_builder_from_protocol`` recursively merges onto the protocol defaults. They exist purely to
+provide editor autocompletion and static type checking for callers; they carry no runtime behaviour
+and the builder logic remains unchanged whether or not they are used.
+
+To keep them from silently drifting away from the work chain input specs, the keys of each
+``TypedDict`` are asserted to be *equal* to the corresponding ``spec().inputs`` port names (plus a
+small set of "builder-consumed" keys that the protocol logic reads directly without exposing them as
+input ports, minus an explicit set of ports the builder always overwrites from its own arguments) in
+``tests/workflows/protocols/test_overrides.py``. Equality (rather than a one-sided subset) means the
+guard fires in both directions: an orphaned typed key *and* a newly added upstream port that the
+``TypedDict`` is missing both break the test.
+"""
+
+from __future__ import annotations
+
+from typing import Any, TypedDict
+
+
+class PwParametersOverrides(TypedDict, total=False):
+    """Overrides for the ``pw.parameters`` Quantum ESPRESSO namelists.
+
+    The keys are the QE ``pw.x`` namelist names; the values are free-form mappings of namelist
+    keyword to value. This enumerates the full ``pw.x`` namelist set so that any namelist a caller
+    may legitimately merge in through the ``overrides`` is autocompleted.
+    """
+
+    CONTROL: dict[str, Any]
+    SYSTEM: dict[str, Any]
+    ELECTRONS: dict[str, Any]
+    IONS: dict[str, Any]
+    CELL: dict[str, Any]
+    FCP: dict[str, Any]
+    RISM: dict[str, Any]
+
+
+class PwCalculationOverrides(TypedDict, total=False):
+    """Overrides for the ``pw`` namespace of ``PwBaseWorkChain`` (the ``PwCalculation`` inputs).
+
+    ``code`` and ``structure`` are deliberately omitted: the builder always sets them from its own
+    ``code``/``structure`` arguments, so an override there would be ignored (see the drift-guard's
+    ``INTENTIONALLY_UNTYPED`` set).
+    """
+
+    parameters: PwParametersOverrides
+    pseudos: dict[str, Any]
+    metadata: dict[str, Any]
+    settings: dict[str, Any]
+    parallelization: dict[str, Any]
+    monitors: dict[str, Any]
+    hubbard_file: Any
+    parent_folder: Any
+    remote_folder: Any
+    vdw_table: Any
+
+
+class PwMetaParameters(TypedDict, total=False):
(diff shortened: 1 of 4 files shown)


t-reents:
I checked the changes and also tested them and don't have any further requests. 

While I'm in principle in favor of the generation approach to avoid the manual implementation (which is not even a problem in the era of Claude :D), I definitely see the point of static type checkers, which I would also consider as more important. Hence, I'm fine with the implementation as is.

If there is nothing else from your side, I'd approve and merge. As you said, we can easily extend it to other `WorkChains` in follow-up PRs.

elinscott:
Thanks @t-reents! I just had a second look at this and make a few superficial tweaks:

- Rename to delinate more clearly from the Calculation overrides:
  - `PwBaseOverrides` → `PwBaseProtocolOverrides`
  - `PwBandsOverrides` → `PwBandsProtocolOverrides`
- Fold the eight drift guard tests into one parametrized test

Ready for you to review/approve/merge when you are.

t-reents:
I'm fine with your latest changes and agree with the naming etc. 

However, I actually noticed an issue now – which is a general problem but related. You correctly added all the ports to the overrides, such as `monitors` or `on_unhandled_failure` etc. The problem is that the `get_builder_from_protocol` only explicitly sets specific inputs, so those overrides never get passed to the actual builder:

https://github.com/aiidateam/aiida-quantumespresso/blob/d890a83e1305c773b8816a78ffc230ce7421cf19/src/aiida_quantumespresso/workflows/pw/base.py#L315-L331

Hence, the IDE will suggest those overrides as valid overrides but they are never passed to the actual builder. 

1. This behavior should be fixed, probably in a separate PR first, i.e., correctly supporting also those keys (btw, the same applies to the `vdw_table`, the `hubbard_file` and potentially more). 
2. I was thinking whether it would make sense to have another test that actual checks that the overrides specified in your `TypedDict`s are correctly forwarded to the actual builder, i.e., to avoid such cases in the future.

elinscott:
> However, I actually noticed an issue now – which is a general problem but related. You correctly added all the ports to the overrides, such as `monitors` or `on_unhandled_failure` etc. The problem is that the `get_builder_from_protocol` only explicitly sets specific inputs, so those overrides never get passed to the actual builder:
> 
> https://github.com/aiidateam/aiida-quantumespresso/blob/d890a83e1305c773b8816a78ffc230ce7421cf19/src/aiida_quantumespresso/workflows/pw/base.py#L315-L331
> 
> Hence, the IDE will suggest those overrides as valid overrides but they are never passed to the actual builder.
> 
> 1. This behavior should be fixed, probably in a separate PR first, i.e., correctly supporting also those keys (btw, the same applies to the `vdw_table`, the `hubbard_file` and potentially more).
> 2. I was thinking whether it would make sense to have another test that actual checks that the overrides specified in your `TypedDict`s are correctly forwarded to the actual builder, i.e., to avoid such cases in the future.

Hmmm, having overrides silently ignored is very bad. Naive question: wky does the builder need to handle each keyword by hand? Can't we somehow recursively iterate over `inputs` without having to enumerate each key?

elinscott:
Would definitely be amenable to adding a series of tests that checks that every possible ovverides key gets set appropriately.

t-reents:
>Naive question: wky does the builder need to handle each keyword by hand? Can't we somehow recursively iterate over inputs without having to enumerate each key?

I was wondering the same, and the original implementation was done much earlier, so I also don't know the initial motivation (or potential limitations back then). One way to argue could be that the motivation of the `get_builder_from_protocol` is just a convenient wrapper to get a pre-populated builder. Afterwards, the user can still modify and set every input that is available in the builder. Also, when reading the docstring of the `get_builder_from_protocol`:

https://github.com/aiidateam/aiida-quantumespresso/blob/d890a83e1305c773b8816a78ffc230ce7421cf19/src/aiida_quantumespresso/workflows/pw/base.py#L177

One could literally interpret it that it's only intended to override the keys exposed/pre-populated by the protocol. However, this is not even true in the current implementation, as the method also sets and overrides the `settings`, even though that key is not specified in the protocol.
As I said, this was just an attempt to understand why it was potentially done in this way.

In any case, I agree that one would ideally just iterate over all the inputs and populate the underlying builder – so happy to follow that approach.

Pinging @mbercx as he might still have some comments

PR #1276, opened 2026-07-23
🐛 `PdosWorkChain`: thread `settings` overrides to dos/projwfc

elinscott:
## Problem
`PdosWorkChain.get_builder_from_protocol` never populated the `settings` of the `dos` and `projwfc` namespaces. As a result any `settings` override for those namespaces — for example `overrides={'projwfc': {'settings': {'CMDLINE': ['-npool', '4']}}}` — was silently dropped and never reached the builder, so it had no effect on the launched `DosCalculation` / `ProjwfcCalculation`.

## Changes
Both namespaces had the identical gap, since `DosCalculation` and `ProjwfcCalculation` both derive from `NamelistsCalculation`, which exposes the `settings` port. The fix sets `builder.dos.settings` and `builder.projwfc.settings` from the merged protocol inputs whenever a `settings` override is provided, matching the way the sibling `code`/`parameters`/`metadata` inputs are already threaded.

## Tests
Added `test_settings_overrides` to `tests/workflows/protocols/test_pdos.py`, which passes `settings` overrides for both the `dos` and `projwfc` namespaces and asserts they land on the builder. The test fails on the pre-fix code (the `settings` input is absent) and passes with the fix.

### DIFF
--- a/src/aiida_quantumespresso/workflows/pdos.py
+++ b/src/aiida_quantumespresso/workflows/pdos.py
@@ -383,9 +383,13 @@ class PdosWorkChain(ProtocolMixin, WorkChain):
         builder.dos.code = dos_code
         builder.dos.parameters = orm.Dict(inputs.get('dos', {}).get('parameters'))
         builder.dos.metadata = metadata_dos
+        if inputs.get('dos', {}).get('settings'):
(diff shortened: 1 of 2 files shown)


t-reents (approved):
Thanks @elinscott! Changes look fine

PR #1278, opened 2026-08-04
🐛 Protocols: fix spurious warning for dynamic namespace overrides

t-reents:
The validation of overrides emits warnings for dynamic namespaces with undeclared children, such as `monitors` or `pseudos` in case of the `PwBaseWorkChain`.

This PR fixes this behavior but still emits warnings for dynamic namespaces with declared children.

Script to reproduce the issue (on main one will still see the warnings but this branch should fix it):

```python

from aiida import orm
from aiida_quantumespresso.workflows.pw.base import PwBaseWorkChain

pseudo_fam = orm.load_group('SSSP/1.3/PBEsol/efficiency')

code = orm.load_code('YOUR CODE')

structure = orm.StructureData(cell=[[2.7, 2.7, 0], [0, 2.7, 2.7], [2.7, 0, 2.7]])
structure.append_atom(position=(0, 0, 0), symbols='Si')
structure.append_atom(position=(1.35, 1.35, 1.35), symbols='Si')

pseudos = pseudo_fam.get_pseudos(structure=structure)

overrides = {
    'pw': {
        'monitors': {'monitor1': {'entry_point': 'quantumespresso.accuracy_stuck'}},
        'pseudos': pseudos,
        'parameters': {
            'SYSTEM': {
                'ecutwfc': 30,
                'ecutrho': 240,
            },
        }
    },
}

builder = PwBaseWorkChain.get_builder_from_protocol(code, structure, overrides=overrides)

```

PR #1279, opened 2026-08-05
🐛 Protocols: fix silently dropped `overrides` in `get_builder_from_protocol`

t-reents:
`get_builder_from_protocol()` built each work chain's inputs by manually copying a hardcoded whitelist of fields onto the builder, wrapping each in its AiiDA type by hand. Any `overrides` key not in that whitelist (e.g. `monitors`) was silently ignored instead of being applied or raising an error.

- Replace the manual per-field copying with `builder._update(inputs)` in `PwBaseWorkChain`, `PhBaseWorkChain`, `PdosWorkChain`, `PwBandsWorkChain`, `PwRelaxWorkChain` and `NebBaseWorkChain`. This existing `ProcessBuilderNamespace` method already does everything the manual code was doing by hand: it merges a plain dict onto the builder, applies the default `to_aiida_type` serializer per port, and raises `AttributeError` for unknown keys at a non-dynamic level.
- A typo'd or misplaced key at a work chain's root namespace now raises instead of being dropped. Keys nested inside a dynamic namespace (e.g. `pw`, `base_relax`) still can only be caught with the existing `UserWarning` from `_validate_override_keys`, since dynamic namespaces accept arbitrary keys by design — this soft check is unchanged and keeps working exactly as introduced in #1136.
- While this changes the behavior, i.e., failing for top level keys, I'd still say that it's fine. I also had a version just emitting a warning and ignoring those keys, but I found it unnecessarily complicated. In the end, an error is only raised for top level keys and if a user tries to manually set them on the builder, it will anyway fail. Also, in #1102 we introduced a new raising behavior that was included in a minor release, so I think that we could accept this change and properly document it in the CHANGELOG.

If there are other opinions, especially in favour of checking those top level keys and dropping them with a warning, instead of raising, we can of course adjust it

### DIFF
--- a/src/aiida_quantumespresso/workflows/neb/base.py
+++ b/src/aiida_quantumespresso/workflows/neb/base.py
@@ -130,19 +130,28 @@ class NebBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
             options=options,
             **kargs,
         )
-        builder = cls.get_builder()
-        builder.neb.code = code
-        builder.neb.images = images
-        builder.neb.pw.pseudos = pw_base.pw.pseudos
-        builder.neb.pw.parameters = pw_base.pw.parameters
-        builder.neb.metadata.options = pw_base.pw.metadata.options
+        inputs = {
+            'neb': {
+                'code': code,
+                'images': images,
+                'pw': {
+                    'pseudos': pw_base.pw.pseudos,
+                    'parameters': pw_base.pw.parameters,
+                },
+                'metadata': {'options': pw_base.pw.metadata.options},
+            },
+            'kpoints_force_parity': pw_base['kpoints_force_parity'],
+            'max_iterations': pw_base['max_iterations'],
+        }
 
         if 'kpoints' in pw_base:
-            builder.kpoints = pw_base['kpoints']
+            inputs['kpoints'] = pw_base['kpoints']
         else:
-            builder.kpoints_distance = orm.Float(pw_base['kpoints_distance'])
-        builder.kpoints_force_parity = orm.Bool(pw_base['kpoints_force_parity'])
-        builder.max_iterations = orm.Int(pw_base['max_iterations'])
+            inputs['kpoints_distance'] = pw_base['kpoints_distance']
+
+        builder = cls.get_builder()
+        builder._update(inputs)
+
         return builder
 
     def setup(self):
--- a/src/aiida_quantumespresso/workflows/neb/base.py
+++ b/src/aiida_quantumespresso/workflows/neb/base.py
@@ -150,7 +150,7 @@ class NebBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
             inputs['kpoints_distance'] = pw_base['kpoints_distance']
 
         builder = cls.get_builder()
-        builder._update(inputs)
+        builder._merge(inputs)
 
         return builder
 
--- a/src/aiida_quantumespresso/workflows/pw/base.py
+++ b/src/aiida_quantumespresso/workflows/pw/base.py
@@ -312,23 +312,20 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
 
         metadata['options'] = cls.set_default_resources(metadata['options'], code.computer.scheduler_type)
 
-        builder = cls.get_builder()
-        builder.pw['code'] = code
-        builder.pw['pseudos'] = pseudos
-        builder.pw['structure'] = structure
-        builder.pw['parameters'] = orm.Dict(parameters)
-        builder.pw['metadata'] = metadata
-        if 'settings' in inputs['pw']:
-            builder.pw['settings'] = orm.Dict(inputs['pw']['settings'])
-        if 'parallelization' in inputs['pw']:
-            builder.pw['parallelization'] = orm.Dict(inputs['pw']['parallelization'])
-        builder.clean_workdir = orm.Bool(inputs['clean_workdir'])
+        inputs['pw']['code'] = code
+        inputs['pw']['structure'] = structure
+        inputs['pw']['pseudos'] = pseudos
+        inputs['pw']['parameters'] = parameters
+        inputs['pw']['metadata'] = metadata
+
+        # `kpoints` and `kpoints_distance` are mutually exclusive: only pass on the one that should be used.
         if 'kpoints' in inputs:
-            builder.kpoints = inputs['kpoints']
+            inputs.pop('kpoints_distance', None)
         else:
-            builder.kpoints_distance = orm.Float(inputs['kpoints_distance'])
-        builder.kpoints_force_parity = orm.Bool(inputs['kpoints_force_parity'])
-        builder.max_iterations = orm.Int(inputs['max_iterations'])
+            inputs.pop('kpoints', None)
+
+        builder = cls.get_builder()
+        builder._update(inputs)
 
         return builder
 
--- a/src/aiida_quantumespresso/workflows/pw/base.py
+++ b/src/aiida_quantumespresso/workflows/pw/base.py
@@ -325,7 +325,7 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
             inputs.pop('kpoints', None)
 
         builder = cls.get_builder()
-        builder._update(inputs)
+        builder._merge(inputs)
 
         return builder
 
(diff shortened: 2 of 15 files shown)


t-reents:
@elinscott I'd suggest that we first agree on this and get it merged and afterwards, we can add the tests to your PR #1274 that validate that all possible overrides according to your `TypedDicts`s actually get merged into the builder.

src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -312,23 +312,20 @@ def get_builder_from_protocol(
...
-            builder.kpoints = inputs['kpoints']
+            inputs.pop('kpoints_distance', None)
         else:
-            builder.kpoints_distance = orm.Float(inputs['kpoints_distance'])
-        builder.kpoints_force_parity = orm.Bool(inputs['kpoints_force_parity'])
-        builder.max_iterations = orm.Int(inputs['max_iterations'])
+            inputs.pop('kpoints', None)
+
+        builder = cls.get_builder()
+        builder._update(inputs)
```
elinscott: **`_update` raises `KeyError` for any dict-valued key at the root namespace, including valid ones.**

`ProcessBuilderNamespace._update` does `self[key].update(value)` when the value is a `Mapping`. But `self[key]` is `self._data[key]` which is only pre-populated for *namespace* ports, which for e.g. `PwBaseWorkChain` is just `metadata` and `pw`, so `handler_overrides` (a genuine `Dict` input port) hits `KeyError` instead of landing:

```python
PwBaseWorkChain.get_builder_from_protocol(
    code, structure,
    overrides={'handler_overrides': {'handle_known_unrecoverable_failure': {'enabled': False}}},
)
# main: builds, value silently dropped
# here: KeyError: 'handler_overrides'
```

This is also an issue one level down, where the key check finds nothing wrong and so the user gets no warning at all, only the bare `KeyError`:

```python
PwBandsWorkChain.get_builder_from_protocol(
    code, structure,
    overrides={'scf': {'handler_overrides': {'handle_known_unrecoverable_failure': {'enabled': False}}}},
)
```

The same mechanism means an *invalid* dict-valued root key gives `KeyError: 'relax'` rather than the `AttributeError: Unknown builder parameter: relax` the PR description promises.

This was not caught because the tests only cover scalar-valued root keys (`clean_wokdir: True`, `pseudo_family: '...'`), which is the half that works.

`builder._merge(inputs)` appears to be a drop-in fix: it routes everything through `__setattr__`, so unknown keys give the intended `AttributeError` regardless of value type, and `handler_overrides` serialises to a `Dict` as it should. Claude tested the `_update` → `_merge` swap and everything seems to work as intended.

tests/workflows/protocols/pw/test_base.py
```diff
@@ -288,11 +288,10 @@ def test_parallelization_overrides(fixture_code, generate_structure):
         # CORRECT overrides for `Dict` node with incorrect keys
         # The key check should _not_ validate the inputs, that is the job of the port validator
         ({'pw': {'parameters': {'NON-EXISTENT': {'param': 1}}}}, None),
-        # WRONG overrides with typo
-        ({'clean_wokdir': True}, UserWarning),
-        # WRONG overrides with process input at incorrect level
+        # WRONG overrides with process input at incorrect level, nested inside the dynamic `pw` namespace: the
+        # namespace itself cannot reject unknown keys, so this is only caught with a warning.
         ({'pw': {'options': {}}}, UserWarning),
```
elinscott: **Nested keys in a dynamic namespace land raw.**

The description says these are "still only caught with the existing `UserWarning`", but since `pw` is dynamic, `_update` now stores them, and the builder stops validating:

```python
PwBaseWorkChain.get_builder_from_protocol(
    code, structure, overrides={'pw': {'options': {'max_wallclock_seconds': 60}}}, # should be pw.metadata.options.max_wallclock_seconds
)
# main: warning, key absent, spec().inputs.validate() passes
# here: warning, pw.options present, validate() -> "Invalid type <class 'int'> for dynamic port
#       value: expected <class 'aiida.orm.nodes.data.data.Data'>"
```

This is an ugly way to crash — make the loud failure here is informative.

src/aiida_quantumespresso/workflows/neb/base.py
```diff
@@ -130,19 +130,28 @@ def get_builder_from_protocol(
...
-            builder.kpoints = pw_base['kpoints']
+            inputs['kpoints'] = pw_base['kpoints']
         else:
-            builder.kpoints_distance = orm.Float(pw_base['kpoints_distance'])
-        builder.kpoints_force_parity = orm.Bool(pw_base['kpoints_force_parity'])
-        builder.max_iterations = orm.Int(pw_base['max_iterations'])
+            inputs['kpoints_distance'] = pw_base['kpoints_distance']
+
+        builder = cls.get_builder()
+        builder._update(inputs)
```
elinscott: **Test coverage only tracks two of the six conversions.**

None of `pdos.py`, `ph/base.py`, `pw/bands.py` nor `neb/base.py` are pinned by the tests. The NEB rewrite is the largest of the six and has no protocol test at all, which seems worth at least a smoke-level builder test given it also changes how `neb.pw` and `neb.metadata` are assigned.

elinscott (changes requested):
Thanks, looks good by and large, and the tests are discriminating — I reverted `pw/base.py` and `pw/relax.py` against them and they fail.

One blocking issue and some smaller things listed in comments below.

src/aiida_quantumespresso/workflows/pw/base.py
t-reents: Ahh, I explicitly pointed Claude to `_update` and `_merge` as it wasn't aware of them, but it apparently convinced me choosing the wrong one, as I only used the non-nested cases for testing. Thanks! I will revert to `_merge`

tests/workflows/protocols/pw/test_base.py
t-reents: Hm, but this would require changing the behavior of the overrides validation introduced in #1136 in general, right? Because there it was discussed that `_validate_override_keys` only emits the warning and the validation will occur when submitted/via the port validator. 
So just to understand correctly, what you're saying is, that `_validate_override_keys` should generally raise instead of emitting warnings when detecting unrecognised keys?

elinscott: Yes.

As per #1136 `_validate_override_keys` warns whenever an override key matches no declared port  e.g.
```
overrides={'pw': {'options': {'max_wallclock_seconds': 60}}}
```
where the real port is `pw.metadata.options`, so `options` triggers
```
Found unrecognised key in overrides
```

The question is what happens to such a key after the warning:
- **On main** it was silently dropped — not by design, but because the old item-by-item copying (`builder.pw.code = ...`, `builder.pw.parameters = ...` over a fixed list of names) never assigned anything else onto the builder
- **With this PR** the whole `inputs` dict is merged onto the builder, so the unrecognised key lands inside the dynamic `pw` namespace and survives to submission, where it fails with `Invalid type <class 'int'> for dynamic port value: expected <class 'aiida.orm.nodes.data.data.Data'>`.

This PR already half-hardens things anyway:
- an unrecognised root-level key fails immediately (builders reject unknown names: `_merge` gives `Unknown builder parameter: ...`)
- a dynamic subspace such as `pw` accepts any key name at assignment, and its extras are only type-checked at submission.
 
So whether the same typo raises immediately or festers to a cryptic submit-time error depends purely on whether the enclosing namespace happens to be dynamic.

t-reents: Agree. I think the warning + builder validation approach would only be helpful in case one would get reasonable error messages, which is not the case. 

I will adapt it. Moreover, since the behavior was always a bit strange, i.e., overrides got silently dropped, I think it's also justified to include this in a `5.1` release.

src/aiida_quantumespresso/workflows/neb/base.py
t-reents: I just learned that I already had some tests locally that we're not committed 😅

t-reents:
@elinscott should be ready for another review.

The latest commits should have addressed all review comments. `_update` to `_merge`, changed warnings to strict validation in `utils.py` and added tests. The tests are also discriminating, i.e., failing on main but pass here. 

I also added a test for the NEB `WorkChain`. However, this is not very informative, since the override handling in that workchain is a bit strange. All the overrides are just passed to the `PwBaseWorkChain` and then only specific keywords are merged into the actual `neb` builder. This being said, it's not possible to override actual inputs of the neb namespace directly. Nonetheless, restructuring that part is beyond the scope of this PR.
There is already an open PR related to NEB #1213, so we could request to adapt the behavior there or open a separate one.

The strict validation actually catched that the CLI specified some overrides at the wrong level.

t-reents:
Just as a comment, since we move from warnings to errors, this PR partially reverts the changes introduced in #1278.

Issue #931, opened 2023-05-09
Protocols: Add `aiida-pseudo` install command to `ValueError`


mbercx:
When a user tries to use a pseudopotential family that isn't installed, e.g.:

```python
from aiida import orm, load_profile
from aiida_quantumespresso.workflows.pw.base import PwBaseWorkChain

load_profile('dev')

structure = orm.load_node(15151) # Si
code = orm.load_code('qe-7.2-pw@localhost')

builder = PwBaseWorkChain.get_builder_from_protocol(
    structure=structure,
    code=code,
    overrides={'pseudo_family': 'SSSP/1.1/PBEsol/precision'}
)
```

The `get_builder_from_protocol()` command returns a `ValueError`:

```python
ValueError                                Traceback (most recent call last)
Cell In[15], line 9
      6 structure = orm.load_node(15151) # Si
      7 code = orm.load_code('qe-7.2-pw@localhost')
----> 9 builder = PwBaseWorkChain.get_builder_from_protocol(
     10     structure=structure,
     11     code=code,
     12     overrides={'pseudo_family': 'SSSP[/1.1/PBEsol/precision](https://vscode-remote+ssh-002dremote-002btheos-002dpc.vscode-resource.vscode-cdn.net/1.1/PBEsol/precision)'}
     13 )
...
    166     ) from exception
    168 try:
    169     cutoff_wfc, cutoff_rho = pseudo_family.get_recommended_cutoffs(structure=structure, unit='Ry')

ValueError: required pseudo family `SSSP/1.1/PBEsol/precision` is not installed. Please use `aiida-pseudo install` toinstall it.
```

Besides the absence of a space between `toinstall`, we could add specific instructions to install the family the user requested. I.e. in this case:

```
aiida-pseudo install sssp -v 1.1 -x PBEsol -p precision 
```

To do this properly, we'd need a utility method that converts the pseudo configuration (specified via the label) into the corresponding `aiida-pseudo install` command (perhaps also considering the defaults). This should probably be implemented in `aiida-pseudo` and used here.

The utility command can then also immediately do a check to see if the configuration exists at all. I was inclined to have the error message indicate that it's not a valid `aiida-pseudo` configuration, but of course users can also install custom families with recommended cutoffs. So it should probably say something along the lines of:

```
ValueError: required pseudo family `best-pseudos-ever-yo` is not installed.
Since it is not an "established" configuration, use `aiida-pseudo install family` to install it from a suitably formatted archive.
```

PR #1281, opened 2026-08-07
👌 Protocols: support pseudo families without recommended cutoffs

elinscott:
### Summary

This PR allows non-SSSP/PseudoDojo pseudo families to be used with the builders.

`aiida-pseudo install family <folder> <label>` installs a family the builders then refuse:

```
ValueError: required pseudo family `MyPseudos/local` is not installed.
Please use `aiida-pseudo install` toinstall it.
```

It is installed. `PwBaseWorkChain.get_builder_from_protocol` looks only for the family classes that can recommend cutoffs, and asks for those cutoffs before fetching the pseudos, so a family without them cannot be used at all — even when the `overrides` already give both (which take precedence anyway). A `CutoffsPseudoPotentialFamily` whose stringency is never set fails the same way, with `no default stringency has been defined`.

### Changes
**First commit.** The recommended cutoffs are requested only when the `overrides` do not already define both; the pseudos are fetched either way. The error still fires when nothing provides them, now naming the alternative.

**Second commit.** The query now names the base class rather than the three cutoff-recommending subclasses, and the capability is checked where the cutoffs are needed (`isinstance(..., RecommendedCutoffMixin)`):

```
ValueError: pseudo family `PseudoPotentialFamily<MyPseudos/local>` cannot recommend cutoffs.
Provide both `ecutwfc` and `ecutrho` in the `overrides`, or use a family that recommends them.
```

The old query's line was not "families whose cutoffs track the protocol": `get_recommended_cutoffs` is called with no stringency, and only `stringent` varies the family — by naming a different one, which a cutoffs family follows no better than a plain one.

### A question for the maintainers
Is the intent that `get_builder_from_protocol` always returns a builder with cutoffs set — so a plain family plus override cutoffs is legitimate, as implemented — or that only cutoff-recommending families are supported at all, in which case the right fix is just a clearer error and no widening? I have taken the first; the second commit can be dropped on its own.

### Notes
- this adds the first direct `aiida_pseudo` import to `src/` (entry points cover group classes, not mixins)
- this drops the now-unused module-level `SsspFamily` / `PseudoDojoFamily` / `CutoffsPseudoPotentialFamily` names from `workflows/pw/base.py`. Say if you would rather keep them.
- #931 proposes improving the neighbouring "not installed" error. This does not reword it, but it does stop it firing for families that are installed.

### DIFF
--- a/src/aiida_quantumespresso/workflows/pw/base.py
+++ b/src/aiida_quantumespresso/workflows/pw/base.py
@@ -216,15 +216,16 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
         # Update the parameters based on the protocol inputs
         parameters = inputs['pw']['parameters']
 
+        system_overrides = (overrides or {}).get('pw', {}).get('parameters', {}).get('SYSTEM', {})
+        cutoffs_in_overrides = all(key in system_overrides for key in ('ecutwfc', 'ecutrho'))
+
         if overrides and 'pseudos' in overrides.get('pw', {}):
             pseudos = overrides['pw']['pseudos']
 
             if sorted(pseudos.keys()) != sorted(structure.get_kind_names()):
                 raise ValueError(f'`pseudos` override needs one value for each of the {len(structure.kinds)} kinds.')
 
-            system_overrides = overrides['pw'].get('parameters', {}).get('SYSTEM', {})
-
-            if not all(key in system_overrides for key in ('ecutwfc', 'ecutrho')):
+            if not cutoffs_in_overrides:
                 raise ValueError(
                     'When overriding the pseudo potentials, both `ecutwfc` and `ecutrho` cutoffs should be '
                     f'provided in the `overrides`: {overrides}'
@@ -244,15 +245,20 @@ class PwBaseWorkChain(ProtocolMixin, BaseRestartWorkChain):
                     'install it.'
                 ) from exception
 
-            try:
-                parameters['SYSTEM']['ecutwfc'], parameters['SYSTEM']['ecutrho'] = (
-                    pseudo_family.get_recommended_cutoffs(structure=structure, unit='Ry')
-                )
-                pseudos = pseudo_family.get_pseudos(structure=structure)
-            except ValueError as exception:
-                raise ValueError(
-                    f'failed to obtain recommended cutoffs for pseudo family `{pseudo_family}`: {exception}'
-                ) from exception
+            # Families that do not define recommended cutoffs can still be used, as long as the `overrides` provide
+            # both cutoffs themselves, since these are applied further down and take precedence anyway.
+            if not cutoffs_in_overrides:
+                try:
+                    parameters['SYSTEM']['ecutwfc'], parameters['SYSTEM']['ecutrho'] = (
+                        pseudo_family.get_recommended_cutoffs(structure=structure, unit='Ry')
+                    )
+                except ValueError as exception:
+                    raise ValueError(
+                        f'failed to obtain recommended cutoffs for pseudo family `{pseudo_family}`: {exception} '
+                        'If the family does not define any, specify both `ecutwfc` and `ecutrho` in the `overrides`.'
+                    ) from exception
+
+            pseudos = pseudo_family.get_pseudos(structure=structure)
 
         parameters['CONTROL']['etot_conv_thr'] = natoms * meta_parameters['etot_conv_thr_per_atom']
         parameters['ELECTRONS']['conv_thr'] = natoms * meta_parameters['conv_thr_per_atom']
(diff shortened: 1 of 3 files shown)


t-reents:
I'll have a closer look later today. 

Regarding your question:

>Is the intent that get_builder_from_protocol always returns a builder with cutoffs set — so a plain family plus override cutoffs is legitimate, as implemented — or that only cutoff-recommending families are supported at all, in which case the right fix is just a clearer error and no widening?

Based on the historical implementation, I think the latter is/was true (not sure about the intention). Nonetheless, I think that the most important aspect is that the builder should always be populated when it gets returned. Hence, I don't see a reason why one should limit it to cutoff recommending families. 

I think there are/should be 3 valid options:

1. A cutoff recommending family is used: nothing else needed
2. A PP family without recommended cutoffs: Check that cutoffs are provided in the overrides. If not, raise. So the important part is indeed that this should not work without the overrides.
3. Custom pseudos via overrides: As done in #1102, ensure that pseudos are provided for all elements via the overrides and that the cutoffs are also manually specified.

As long as those 3 cases are covered, I don't see a problem. Based on the approach that you selected, I guess that you agree with my interpretation, so no need to change anything.

As I said, I will have a look at the actual implementation later, but if there are no further issues, I'm happy to continue with your "design".

src/aiida_quantumespresso/workflows/pw/base.py
```diff
@@ -216,43 +215,52 @@ def get_builder_from_protocol(
...
                 )
 
         else:
             try:
-                pseudo_set = (
-                    PseudoDojoFamily,
-                    SsspFamily,
-                    CutoffsPseudoPotentialFamily,
+                pseudo_family = (
+                    orm.QueryBuilder().append(PseudoPotentialFamily, filters={'label': pseudo_family}).one()[0]
```
t-reents: This was not introduced by this PR, but would you mind adding another exception for `MultipleObjectsError`, in case multiple families with the same label are found?

Other than that, I reviewed the implementation and all looks good.

elinscott:
Thanks @t-reents !

> I think there are/should be 3 valid options:
> 
> 1. A cutoff recommending family is used: nothing else needed
> 2. A PP family without recommended cutoffs: Check that cutoffs are provided in the overrides. If not, raise. So the important part is indeed that this should not work without the overrides.
> 3. Custom pseudos via overrides: As done in [🐛 `get_builder_from_protocol`: Fix bugs when overriding pseudopotentials #1102](https://github.com/aiidateam/aiida-quantumespresso/pull/1102), ensure that pseudos are provided for all elements via the overrides and that the cutoffs are also manually specified.

Agreed. Cases 1 and 2 are what the PR implements; I checked case 3 against the current code rather than assuming it, and it behaves as you describe. The `pseudos`-in-`overrides` branch compares `sorted(pseudos.keys())` with the structure's kind names, so pseudos for only some kinds are refused, as are keys naming something the structure does not contain, and both `ecutwfc` and `ecutrho` have to be in the `overrides`. On a water molecule I get:

```
# pseudos for O only, both cutoffs given
ValueError: `pseudos` override needs one value for each of the 2 kinds.

# pseudos for both kinds, ecutwfc only
ValueError: When overriding the pseudo potentials, both `ecutwfc` and `ecutrho` cutoffs should be provided in the `overrides`: ...
```

t-reents (approved):

t-reents:
Thanks for adding the test and also confirming the expected behavior! 
I'll merge once the CI passes (which was already the case before I merged main).

src/aiida_quantumespresso/workflows/pw/base.py
elinscott: Added in the latest commit. To be fair, it's not actually entirely unrelated: group labels are unique per type string, so a plain `PseudoPotentialFamily` and an `SsspFamily`/`CutoffsPseudoPotentialFamily` can share one. so this PR makes this clash **more** likely.

Now gives an error such as

```
    ValueError: the label `custom/shared-label` matches more than one installed pseudo family:
    `PseudoPotentialFamily<custom/shared-label>`, `CutoffsPseudoPotentialFamily<custom/shared-label>`.
    Please delete or relabel all but one, or pass the `pseudos` in the `overrides` instead.
```

elinscott:
> Thanks for adding the test and also confirming the expected behavior! I'll merge once the CI passes (which was already the case before I merged main).

No problem — thanks! (In case you missed it, see my in-thread reply I just posted)

t-reents:
Ah yes, just saw the comment. Indeed, this is also why I raised it. While it was already possible before, the changes made it a bit more likely to actually occur.

elinscott:
> Ah yes, just saw the comment. Indeed, this is also why I raised it. While it was already possible before, the changes made it a bit more likely to actually occur.

Turns out this is an edge case but it _can_ happen if you're sadistic: aiidateam/aiida-pseudo#206

t-reents:
Haha, yes, I also played with it yesterday before asking for the test. I will have a look at `aiida-pseudo` next week, but I also think that this should be handled differently

PR #1284, opened 2026-09-17
✨ PwParser: Add XML schema for Quantum ESPRESSO v7.6

tsthakur:
I have tested this for all pw calculations, including r2scan vc-relax calculations on Alps-Daint.


Minotakm (commented):
Thanks @tsthakur for taking the time to do it! 

One thing on the compatibility tables: with `> 5.0` on row 1 and `4.16 <= 5.0` on row 2, the released
v5.0.0 lands in row 2, which is badged `AiiDA >=2.5.0,<3.0.0`. But v5.0.0 pins `aiida_core~=2.8`, so
the table would be telling v5.0.0 users that aiida-core 2.5–2.7 is fine.

Could we add a new row instead of moving the existing boundaries? That's what the earlier schema PRs
did (75808a9 for QE 7.4 and 3c2ba30 for 7.5) . What do you think @t-reents ?

t-reents:
@Minotakm yes, we should definitely do this. The current table of this PR wouldn't be correct.

@tsthakur Would be great if you could adjust it in the way that Michail described (i.e., a new row with the current status and aiida-qe `5.0 < 5.1` and then the first row should become `5.1 < 6.0`. Happy to merge it afterwards :)

tsthakur:
Thank you for the feedback @Minotakm ! I changed the rows as suggested by @t-reents , I guess squashing the 2 commits would be best before merging.

Minotakm (approved):
LGTM! Thanks @tsthakur

tsthakur:
@Minotakm, not sure if you were waiting for me to merge the PR, because it says merging is blocked as I am not authorised to push this branch. Could you have a look?

t-reents:
@tsthakur No, all good, I will do it. I just haven't done it yet, as I was planning to do a release afterwards and haven't looked into that yet.

t-reents:
@tsthakur Done 😄

